PR #32, opened 2021-05-05
check if protocol is allowed before execution

metal-messiah:

---

### Overview
Please describe the changes present in the pull request and if applicable, describe why the changes are needed.

This PR addresses a bug where the loader object is empty when opening in a file:// protocol containing the agent code. This empty object causes nested property assignment down the chain to error out.  The solution proposed here is to prevent the instrument files from executing if the protocol is invalid (file://).

### Related Github Issue
Please include a link to the related GitHub issue, if applicable

[JIRA](https://newrelic.atlassian.net/browse/JS-7890)

### Testing
The browser agent includes a suite of tests which should be used to
verify that your changes don't break existing functionality. These tests will run through Github Actions when:

 1) A pull request is made _and_ 
 2) The pull request is marked as ready for testing by the internal team. 

##### Steps to test:
1. Directly open from your file directory (file://) an html file containing the newly built agent code.
2. Ensure no runtime errors display in the browser console.

---   

### DIFF
--- a/feature/stn/instrument/index.js
+++ b/feature/stn/instrument/index.js
@@ -3,9 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-var protocolAllowed = require('../../../loader/protocol-allowed')
-if (!protocolAllowed(window.location)) return
-
 if (!(window.performance &&
   window.performance.timing &&
   window.performance.getEntriesByType
@@ -30,8 +27,8 @@ var PUSH_STATE = 'pushState'
 
 // Turn on feature harvesting
 var loader = require('loader')
-loader = Object.assign({}, loader, {features: Object.assign(loader.features || {}, {stn: true})})
-console.log('stn loader')
+if (loader.disabled) return
+
 loader.features.stn = true
 
 // wrap history ap
--- a/feature/stn/instrument/index.js
+++ b/feature/stn/instrument/index.js
@@ -3,6 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+var protocolAllowed = require('../../../loader/protocol-allowed')
+if (!protocolAllowed(window.location)) return
+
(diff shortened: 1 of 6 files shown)


feature/stn/instrument/index.js
```diff
@@ -27,6 +30,8 @@ var PUSH_STATE = 'pushState'
 
 // Turn on feature harvesting
 var loader = require('loader')
+loader = Object.assign({}, loader, {features: Object.assign(loader.features || {}, {stn: true})})
```
martinkuba, line 33: We initialize the `features` property here
https://github.com/newrelic/newrelic-browser-agent/blob/main/loader/index.js#L53

But I see that the early return in loader/index.js changes what is exported. Maybe we should still export something - maybe just a flag that says the agent is disabled. That might be better than running the same protocol check in multiple files.

feature/stn/instrument/index.js
```diff
@@ -27,6 +30,8 @@ var PUSH_STATE = 'pushState'
...
+console.log('stn loader')
```
martinkuba, line 34: remove console log

martinkuba (commented):
It would be good to have a regression test for this if not too difficult.

feature/stn/instrument/index.js
metal-messiah: Good call, this is the info i was looking for.

feature/stn/instrument/index.js
metal-messiah: whoops forgot to remove both of these lines before commiting.

feature/stn/instrument/index.js
metal-messiah: I also just removed that and moved where the early return happens so that the object is no longer empty.

martinkuba (approved):
I tested it locally and did not see the original error.

PR #43, opened 2021-05-10
JIL Cleanup

metal-messiah:
### Overview
This PR does a clean up of some basic functionality for JIL.  Its adds:
- Simplified interface commands for running tests
    - All tests can be ran using `npm run test`
    - `npm run test` can accept the same params as the cli script
        - NOTE: As per the NPM documentation, you can pass args to npm scripts, but you must separate the cli params from the npm script params using empty double dashes (`--`). example: `npm run test -- -f merged` . If this is too weird we can pivot to something else, this just utilizes what is already in place (`npm run test`)
    - A script command for spinning up the JIL-Server `npm run test-server`
- Updates the documentation
    - To match the changes
    - Clear up formatting
- Fortifies the NPM Install command, which now first installs the root package dependencies, and then runs a postscript which recursively traverses the children directories looking for package.json files, and subsequently installing them.

### Related Issue
[JIRA](https://newrelic.atlassian.net/browse/JS-7724)

### Testing
- Please clear your node_modules and try a `npm install`
    - Please check that node_modules exist in root and in `tools/jil`
- Please read the updated `README.md` and try some of the test commands using `npm run test` as the base command.
    - Make sure the readme makes sense, and that the tests execute as expected.
    - Please try the test server command
    - Please try running the tests directly from the cli file as before and ensure that still works as well.


metal-messiah:
It would be helpful if you review the recursive NPM install and ensure this is a better solution.  I was told the problem is that pre-post NPM installs were only sporadically successful with JIL.  This may help eliminate that problem.

post-installer.js
```diff
@@ -0,0 +1,43 @@
...
+const root = process.cwd()
+
+// Since this script is intended to be run as a "postinstall" command,
+// it will do `npm install` automatically inside the root folder before any of the subfolders.
+print('Installed Root Package')
+
+npm_install_recursive(root)
+
+function print(msg){
+    console.log(`\n================================\n${msg}\n================================\n`)
```
martinkuba, line 14: We use 2 spaces for indents, I would recommend changing to be consistent.

martinkuba:
This all looks good to me. The post-install script being recursive is a little unnecessary, but I am not opposed to it.

martinkuba (changes requested):

Issue #38, opened 2021-05-10
Full Fetch Support


rr0214:
**The Problem**

Customers are not seeing all of their Fetch calls.

**Why Now**

We are focusing on efforts on health of the agent and robust instrumentation. There have been increased support tickets and FRs around missing Fetch calls.

**Appetite**

2 engineer x 2.5 weeks = 5 engineering weeks

**What will we do**

SPA captures XHR and Fetch as events but only if they are part of route changes. We will now support all Fetch calls, sending them as metrics<br/>

##### Related Operational Links

- [ **Koan Key Result (** ](https://newrelic.koan.co/team/c30807d5-c686-4bb7-a146-fb8cc5441b35/goals/f/goals/045fb87c-9df7-466b-bf16-6b35f3296106?view=nested&modal=&status=active&columns=leads%2CdueDate%2Cparent%2Cmetrics%2Cconfidence%2CcloseOutDetails&rosterModal=false) [ **FY22** ](https://newrelic.koan.co/team/c30807d5-c686-4bb7-a146-fb8cc5441b35/goals/f/goals/045fb87c-9df7-466b-bf16-6b35f3296106?view=nested&modal=&status=active&columns=leads%2CdueDate%2Cparent%2Cmetrics%2Cconfidence%2CcloseOutDetails&rosterModal=false) [ **June** ](https://newrelic.koan.co/team/c30807d5-c686-4bb7-a146-fb8cc5441b35/goals/f/goals/045fb87c-9df7-466b-bf16-6b35f3296106?view=nested&modal=&status=active&columns=leads%2CdueDate%2Cparent%2Cmetrics%2Cconfidence%2CcloseOutDetails&rosterModal=false) [ **)** ](https://newrelic.koan.co/team/c30807d5-c686-4bb7-a146-fb8cc5441b35/goals/f/goals/045fb87c-9df7-466b-bf16-6b35f3296106?view=nested&modal=&status=active&columns=leads%2CdueDate%2Cparent%2Cmetrics%2Cconfidence%2CcloseOutDetails&rosterModal=false)
[...]

PR #64, opened 2021-05-28
Record fetch calls as metrics

martinkuba:
### Overview
Fetch calls are currently recorded only as AjaxRequest events with SPA browser interactions. This change records fetch calls as AJAX metrics, which will make them visible in the AJAX UI charts.

### Related Github Issue
https://github.com/newrelic/newrelic-browser-agent/issues/38


metal-messiah (commented):
Looks great, as we discussed, a test for validating failing fetch calls would be appropriate for this.

metal-messiah (approved):

PR #68, opened 2021-06-04
Allow Cumulative Layout Shift values of 0 to be reported

aubreymasten:
### Overview
From internal notes:
> 0 is the perfect CLS score and needs to be included in our calculations (e.g. 75th percentile). Currently the agent only records a CLS value when the browser provides one (when content actually shifts). When there is no shift, we do not report anything, and the cumulativeLayoutShift attribute is not added to the PageViewTiming events.

To address this, I've added a check for a user agent's CLS support using the [PerformanceObserver.supportedEntryTypes property](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/supportedEntryTypes).

By default `clsSupported` will be false, unless the user agent's supportedEntryTypes includes `layout-shift`.

Optionally, I added Chrome 77 to the list of browsers which we test. This was to test the case where `PerformanceObserver.supportedEntryTypes` returned false when checking for `layout-shift` because all other browsers which didn't support CLS in our list did not support `supportedEntryTypes` and were false because the try/catch loop swallows the null reference.

### Testing
I've added two tests: one to verify that CLS values of 0 are reported, and another to validate that browsers which do not support CLS will not record CLS values.


tests/functional/timings.test.js
```diff
@@ -391,6 +392,76 @@ function runClsTests(loader) {
...
+      })
+      .catch(fail)
+
+    function fail (err) {
+      t.error(err)
+      t.end()
+    }
+  })
+
+  testDriver.test(`${loader} agent does not collect cls attribute on unsupported browser`, unsupportedCls.and(reliableFinalHarvest), function (t, browser, router) {
```
martinkuba, line 429: I think this test should match only browsers that do not support CLS (supportedCls.inverse()).

martinkuba: Never mind, you got it :)

martinkuba (commented):
Looks good to me - will approve once all tests are passing.

martinkuba (approved):

PR #82, opened 2021-07-06
Handle max payloads

metal-messiah:

---

##### The problem:
* If an app has thousands of AJAX requests in the harvest interval, the body could be too big and lots of data could be dropped.  There is a defined limit that can be placed on the agent.  We can check against that and break up the requests into multiple pieces if needed.

##### The approach:
* Payload is evaluated in the AJAX serialization for size
* Payload is split into array items if exceeds the size limit on the loader
* The harvester loops through all array items in the harvest and sends to NR1 for each.

### Related Github Issue
[GITHUB Card](https://github.com/newrelic/newrelic-browser-agent/projects/1#card-64211355)

##### Automated testing is simply ensuring that:
if I add 2,000 requests to the harvest queue, that it breaks it into multiple items in the array, and that each is under the size limit

##### To test this manually, you can:
* Open the mock-bam-server
* paste in these file contents 
```<html>
<head>
  <script type="text/javascript">
    window.NREUM||(NREUM={});

    // agent loaded from a local directory
    // reporting to local fake consumer
    NREUM.info={
      agent:"localhost:8181/agent/dev/spa",
      beacon:"staging-bam.nr-data.net",
      errorBeacon: "staging-bam.nr-data.net",
      licenseKey:"2fec6ab188",
      applicationID:"34824659",
      sa: 1
    };

  </script>
  <script type="text/javascript" src="/loader/dev/spa"></script>
</head>
<body>
  hello
  <script>
      const loaded = [];
      const getRandomNumber = () => Math.floor(Math.random() * 10000)
      let count = 0;
      const sendHello = () => {
        let xhr = new XMLHttpRequest()
            // const num = getRandomNumber();
            xhr.open('GET', `/hello`)

            xhr.setRequestHeader('Content-Type', 'text/plain')
            xhr.onload = function (e) {
                if (count < 3000) {
                    count++
                    sendHello();
                }
            }

            xhr.send()
      }
      setTimeout(() => {
            sendHello()
        }, 250)
    
  </script>
</body>
</html>
```
* run against the file, which calls many ajax requests, and see that request is broken into multiple items and each is sent

### DIFF
--- a/feature/xhr/aggregate/index.js
+++ b/feature/xhr/aggregate/index.js
@@ -21,6 +21,8 @@ var ajaxEvents = []
 var spaAjaxEvents = {}
 var sentAjaxEvents = []
 
+var MAX_PAYLOAD_SIZE = loader.maxPayloadSize || 1000000
+
 // bail if not instrumented
 if (!loader.features.xhr) return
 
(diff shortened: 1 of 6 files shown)


martinkuba:
@metal-messiah Can you please rebase on main, so that it does not include the old commits?

feature/xhr/aggregate/index.js
```diff
@@ -168,3 +178,8 @@ function onEventsHarvestFinished(result) {
     sentAjaxEvents = []
   }
 }
+
+function exceedsSizeLimit(payload) {
+  return new Blob([payload]).size > loader.maxBytes
```
martinkuba, line 183: Blob is not supported on all browsers - we should have a feature detection around this. There are not many browsers that do not support it, but we do not want to cause errors anywhere. Also, I think it's ok drop the data for those old browsers when the payload size is too large (if it is not easy to do something else).

feature/xhr/aggregate/index.js
```diff
@@ -146,19 +151,24 @@ function getPayload (events) {
...
-      payload += ';' + attrParts.join(';')
+      insert += ';' + attrParts.join(';')
     }
 
-    if ((i + 1) < events.length) payload += ';'
-  }
+    if ((i + 1) < events.length) insert += ';'
 
+    // check if the current payload string is too big, if so then increment to a new bucket idx
+    if (exceedsSizeLimit(payload[payloadIdx] + insert)) payload[++payloadIdx] = ver
```
martinkuba, line 169: I think this will lose the string table for subsequent payloads - strings are included in the payload the first time they appear, and subsequent occurrences are referenced only. You will need to restart the string table for each payload, and definitely should have a test for this.

martinkuba: How expensive is it to call `exceedsSizeLimit` for every event? Is it better than checking the final size, and if too big starting from scratch with multiple smaller buckets?

metal-messiah: i think with this, we are already iterating over the string concat so at minimum it would be O(n) no matter what we do.  Not sure how expensive it is to create a new blob n number of times vs iterating multiple times.  Your above comment however makes me think it's better to just create smaller buckets and start over from scratch because of the referenced string payloads.

metal-messiah: Also, if pure speed is the name of the game we could also look into using array joins instead of the iterative string concat, but i think at this scale its not going to make a huge difference

feature/xhr/aggregate/index.js
```diff
@@ -168,3 +194,17 @@ function onEventsHarvestFinished(result) {
...
   }
 }
+
+function exceedsSizeLimit(payload, maxPayloadSize) {
+  maxPayloadSize = maxPayloadSize || MAX_PAYLOAD_SIZE
+  return window.Blob ? new Blob([payload]).size > maxPayloadSize : false
+}
+
+function chunk(arr, chunkSize) {
+  if (chunkSize <= 0) throw new Error('Invalid chunk size')
```
martinkuba, line 204: Can this happen? If yes, we should handle the error, so that the error does not show up as uncaught error in the customer app (which the customer cannot address).

feature/xhr/aggregate/index.js
```diff
@@ -168,3 +194,17 @@ function onEventsHarvestFinished(result) {
     sentAjaxEvents = []
   }
 }
+
+function exceedsSizeLimit(payload, maxPayloadSize) {
+  maxPayloadSize = maxPayloadSize || MAX_PAYLOAD_SIZE
+  return window.Blob ? new Blob([payload]).size > maxPayloadSize : false
```
martinkuba, line 200: What is the advantage of using Blob instead of just `payload.length`?

feature/xhr/aggregate/index.js
```diff
@@ -104,62 +105,87 @@ baseEE.on('interactionDiscarded', function (interaction) {
...
+  var payload = []
+  var chunkSize = events.length / chunks
+  var eventChunks = chunk(events, chunkSize)
+
+  for (var i = 0; i < eventChunks.length; i++) {
+    adders.push(getAddStringContext())
+    payload.push(ver)
+    var currentChunk = eventChunks[i]
+
+    for (var j = 0; j < currentChunk.length; j++) {
```
martinkuba, line 143: This is a detail, but it might be easier to extract serialization of a single chunk into a separate function. It would make this more understandable - no need to have an array of adders and nested for loops.

feature/xhr/aggregate/index.js
```diff
@@ -24,7 +24,8 @@ var sentAjaxEvents = []
 // bail if not instrumented
 if (!loader.features.xhr) return
 
-var harvestTimeSeconds = config.getConfiguration('ajax.harvestTimeSeconds') || 60
+var harvestTimeSeconds = config.getConfiguration('ajax.harvestTimeSeconds') || 10
+var MAX_PAYLOAD_SIZE = config.getConfiguration('ajax.maxPayloadSize') || 500
```
martinkuba, line 28: Should the default be 1MB?

tests/functional/xhr/ajax-events.test.js
```diff
@@ -33,6 +33,45 @@ testDriver.test('capturing XHR ajax events', xhrBrowsers, function (t, browser,
...
+    if (ajaxEvents.length) ajaxBundles++
+    console.log('AJAX RESPONSE!', ajaxBundles)
+    if (ajaxBundles < 2) router.expectAjaxEvents().then(r => parseAjaxPromise(r))
+    else {
+      t.ok(ajaxBundles > 1, 'AJAX load is split into multiple payloads and sent')
+      t.end()
+    }
+  }
+
+  Promise.all([ajaxPromise, loadPromise, rumPromise])
```
martinkuba, line 64: This will wait only for one ajax harvest; should it wait for multiple?

feature/xhr/aggregate/index.js
metal-messiah: yes thanks, forgot to set this back after i was manually testing it

tests/functional/xhr/ajax-events.test.js
metal-messiah: it will wait for multiple i believe, in the `parseAjaxPromise` method

feature/xhr/aggregate/index.js
metal-messiah: Are there ever going to be cases where characters will be encoded in different ways?

Converting to a blob removes all the unknowns of encoding, special characters, or variable byte sizes from the string, but if we are never going to face those we could just assume the characters are 2 bytes per character i suppose

feature/xhr/aggregate/index.js
metal-messiah: fair, i'll remove this as I don't think this will be possible the way it is written.  This was mostly a developer safety rail so Its not needed and I don't want errors to propagate to NR1

martinkuba (approved):

PR #83, opened 2021-07-09
Ajax deny list

martinkuba:
### Overview
This PR adds filtering of ajax events from a deny list configuration.

### DIFF
--- a/feature/xhr/aggregate/deny-list.js
+++ b/feature/xhr/aggregate/deny-list.js
@@ -16,7 +16,8 @@ function shouldCollectEvent(params) {
     if (parsed.hostname === '*') {
       return false
     }
-    if (compareDomain(parsed.hostname, params.hostname) && comparePath(parsed.pathname, params.pathname)) {
+    if (compareDomain(parsed.hostname, params.hostname) &&
+        comparePath(parsed.pathname, params.pathname)) {
       return false
     }
   }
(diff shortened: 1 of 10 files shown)


feature/xhr/aggregate/deny-list.js
```diff
@@ -0,0 +1,80 @@
...
+
+function setDenyList(denyListConfig) {
+  denyList = []
+  if (!denyListConfig || !denyListConfig.length) {
+    return
+  }
+  for (var i = 0; i < denyListConfig.length; i++) {
+    var url = denyListConfig[i]
+    if (url.indexOf('http://') === 0) {
+      url = url.substring(7)
```
metal-messiah, line 35: Just to be aware, there are some web applications that use a URL pattern that includes two or more cases of http(s):// in the string

If the url starts with `https://` but includes `http://` in the rest of the url, it could in theory cause an unexpected substring from the `url.substring(7)`

[example link](https://web.archive.org/web/20101224003257/http://webmasters.stackexchange.com/)

martinkuba: I think it would still be ok since the substring call is removing just the protocol at the very beginning of the URL. If there was a second protocol string, it would be in the path (as in the example). I have added a test that covers this. Let me know if there is something I am missing.

metal-messiah (approved):

PR #84, opened 2021-07-23
add way to build a subpath into output

metal-messiah:
### Overview
This adds a env var to build outputs with the agent linked to a nested path.
ie. `https://js-agent.newrelic.com/<SUBPATH>/nr-loader-spa.min.js`. this was used for custom testing with the extension and loaded manually into S3


martinkuba:
I would change the `subPath` env var to upper case (by convention). Otherwise 👍

PR #85, opened 2021-07-23
Final harvest for ajax events

martinkuba:
### Overview
When page is unloading, send any buffered ajax events that have not been harvested yet.

### DIFF
--- a/tests/functional/final-harvest.test.js
+++ b/tests/functional/final-harvest.test.js
@@ -239,7 +239,11 @@ testDriver.test('final harvest sends timings data', reliableFinalHarvest, functi
 // It does not check all of them, just errors and resources.  This is sufficient for the
 // test.  Sending more than that makes the test very fragile on some platforms.
 testDriver.test('final harvest sends multiple', reliableResourcesHarvest.and(stnSupported), function (t, browser, router) {
-  let url = router.assetURL('final-harvest-timings.html')
+  let url = router.assetURL('final-harvest-timings.html', {
+    init: {
+      ajax: { deny_list: ['*'] }
+    }
+  })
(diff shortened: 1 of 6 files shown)


tests/functional/final-harvest.test.js
```diff
@@ -239,7 +242,11 @@ testDriver.test('final harvest sends timings data', reliableFinalHarvest, functi
 // It does not check all of them, just errors and resources.  This is sufficient for the
 // test.  Sending more than that makes the test very fragile on some platforms.
 testDriver.test('final harvest sends multiple', reliableResourcesHarvest.and(stnSupported), function (t, browser, router) {
-  let url = router.assetURL('final-harvest-timings.html')
+  let url = router.assetURL('final-harvest-timings.html', {
+    init: {
+      ajax: { deny_list: ['*'] }
```
metal-messiah, line 247: I'm not sure this [test](https://github.com/newrelic/newrelic-browser-agent/blob/4f897e220c6c9ce2ee9ccea31e0a21b13a3fd4b2/tests/functional/final-harvest.test.js#L278) gives a definitive differentiation between timings and AJAX objs (could be either).  If not, we should probably be explicit and make sure there are no AJAX events in that payload since the `*` is used here.

martinkuba: I have added the deny_list to basically keep the test working the same as before. I think I will just remove checking the events count since the test is good enough to check just resources and errors (the comment above the test even says that).

martinkuba:
There is a test for FCP that started failing and is not related to this PR. It looks like it is because of a new release of iOS on Saucelabs. I will update the browser matrix as part of this PR, so that it can pass all tests.

metal-messiah (approved):

PR #86, opened 2021-07-23
handle endTime better

metal-messiah:
### Overview
This PR attempts to address the issue that was observed on Nike.com.  EndTime is sometimes undefined. This made `timeToLoad Event Start` invalid on NR1, and also broke `qp.decode()` on the extension because that prop had NaN instead of a number.

This solution adds a check in the `end` emission, and also adds a check where the event object is made, using timings from the metrics object.  It was a little unclear still where the property was lost, but in all cases the timings still remained on the metric obj, so it should be safe to use as a check condition. Let me know if there's a better way to handle this.

### Testing
You can manually test using the [extension](https://github.com/newrelic-experimental/nr-ba-injector-extension) and the dev build on S3 [here](https://js-agent.newrelic.com/dev/nr-loader-spa.min.js) and going to nike.com, then checking that events decode properly in the console and that AjaxRequest event objects in NR1 have timeToLoad values.

### DIFF
--- a/feature/xhr/aggregate/index.js
+++ b/feature/xhr/aggregate/index.js
@@ -75,6 +75,8 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     return
   }
 
+  var useMetrics = startTime === undefined || startTime === null || startTime < 0 || !endTime || endTime < startTime
+
   var event = {
     method: params.method,
     status: params.status,
@@ -83,8 +85,8 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     requestSize: metrics.txSize,
     responseSize: metrics.rxSize,
     type: type,
-    startTime: startTime && endTime ? startTime : metrics.time,
-    endTime: startTime && endTime ? endTime : metrics.time + metrics.duration,
+    startTime: !useMetrics ? startTime : metrics.time,
+    endTime: !useMetrics ? endTime : metrics.time + metrics.duration,
     callbackDuration: metrics.cbTime
   }
 
--- a/feature/xhr/aggregate/index.js
+++ b/feature/xhr/aggregate/index.js
@@ -83,8 +83,8 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     requestSize: metrics.txSize,
     responseSize: metrics.rxSize,
     type: type,
-    startTime: startTime,
-    endTime: endTime,
+    startTime: startTime && endTime ? startTime : metrics.time,
+    endTime: startTime && endTime ? endTime : metrics.time + metrics.duration,
     callbackDuration: metrics.cbTime
   }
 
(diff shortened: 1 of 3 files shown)


feature/xhr/instrument/index.js
```diff
@@ -298,6 +298,7 @@ ee.on('fetch-done', function (err, res) {
 
 // Create report for XHR request that has finished
 function end (xhr) {
+  this.endTime = this.endTime || loader.now()
```
metal-messiah, line 301: This was added to mimic the functionality seen in `fetch-done` seen (here)[https://github.com/newrelic/newrelic-browser-agent/blob/2c22571e5e260512000335f4926c417f295413f8/feature/xhr/instrument/index.js#L278]

feature/xhr/aggregate/index.js
```diff
@@ -83,8 +85,8 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     requestSize: metrics.txSize,
     responseSize: metrics.rxSize,
     type: type,
-    startTime: startTime,
-    endTime: endTime,
+    startTime: !useMetrics ? startTime : metrics.time,
+    endTime: !useMetrics ? endTime : metrics.time + metrics.duration,
```
metal-messiah, line 89: If for some reason after all of the event handlers have finished and  it still doesn't have endTime for building the event, the `storeXHR` method does have access to the metrics, which is what is used to build the start and endTime props anyway.  This can serve as a useful fallback.

tests/browser/xhr/ajax-events.browser.js
```diff
@@ -224,6 +224,59 @@ test('prepareHarvest correctly serializes an AjaxRequest events payload', functi
   t.end()
 })
 
+test('prepareHarvest correctly serializes an AjaxRequest events that is lacking an endTime', function (t) {
+  const context = { spaNode: undefined }
+  const expected = {
```
martinkuba, line 229: This expected object is not used for any assertions - should it be? Or, is the test intended to only test that the event is serialized without checking values?

feature/xhr/aggregate/index.js
```diff
@@ -75,6 +75,8 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     return
   }
 
+  var useMetrics = startTime === undefined || startTime === null || startTime < 0 || !endTime || endTime < startTime
```
martinkuba, line 78: I think this is redundant because having duration in metrics implies that we got the end time. The places where we set metrics.duration are the same places where we collect end time.

metal-messiah: For whatever reason, This was not true in the case that the loader and agg were mismatched. The metric data existed, the endTime passed into the method did not

metal-messiah: Hence why it stated working on Nike after this branch was pushed up to s3

PR #97, opened 2021-08-19
add flag for disabling AJAX feature

metal-messiah:
### Overview
This PR adds a flag to enable/disable `All AJAX` feature set.

### DIFF
--- a/feature/xhr/aggregate/index.js
+++ b/feature/xhr/aggregate/index.js
@@ -40,10 +40,12 @@ baseEE.on('feat-err', function() {
     return { body: agg.take([ 'xhr' ]) }
   })
 
-  scheduler = new HarvestScheduler(loader, 'events', { onFinished: onEventsHarvestFinished, getPayload: prepareHarvest })
-  scheduler.startTimer(harvestTimeSeconds)
+  if (allAjaxIsEnabled()) {
+    scheduler = new HarvestScheduler(loader, 'events', { onFinished: onEventsHarvestFinished, getPayload: prepareHarvest })
+    scheduler.startTimer(harvestTimeSeconds)
 
-  subscribeToUnload(finalHarvest)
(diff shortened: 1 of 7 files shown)


feature/xhr/aggregate/index.js
```diff
@@ -98,7 +98,7 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     spaAjaxEvents[interactionId] = spaAjaxEvents[interactionId] || []
     spaAjaxEvents[interactionId].push(event)
   } else {
-    ajaxEvents.push(event)
+    if (allAjaxIsEnabled()) ajaxEvents.push(event)
```
martinkuba, line 101: I would recommend moving the `allAjaxIsEnabled` call to line 78. That way events will not be created at all.

feature/xhr/aggregate/index.js
```diff
@@ -121,7 +121,7 @@ baseEE.on('interactionDiscarded', function (interaction) {
 function prepareHarvest(options) {
   options = options || {}
 
-  if (ajaxEvents.length === 0) {
+  if (ajaxEvents.length === 0 || !allAjaxIsEnabled()) {
```
martinkuba, line 124: Instead of checking here, I would recommend not starting the harvesting scheduler at all, which is on line 43 - 44.

feature/xhr/aggregate/index.js
```diff
@@ -233,3 +233,7 @@ function Chunk (events) {
     return this.payload.length * 2 > maxPayloadSize
   }
 }
+
+function allAjaxIsEnabled() {
+  return (!!config.getConfiguration('ajax.enabled') || false)
```
martinkuba, line 238: I don't think you need the `|| false` the way it is written.

martinkuba (approved):

PR #98, opened 2021-08-24
Add Span information to non-SPA AjaxRequest events

aubreymasten:
### Overview
This PR adds the attributes required to generate a Span event from an AjaxRequest event to the payloads of non-SPA AjaxRequest events.

### Testing
Functional tests added to validate span attributes were created correctly 
https://github.com/newrelic/newrelic-browser-agent/blob/7ba7567e79691697f53e8d110f5269b00e0b0571/tests/functional/xhr/ajax-events.test.js

### DIFF
--- a/feature/xhr/aggregate/index.js
+++ b/feature/xhr/aggregate/index.js
@@ -95,10 +95,9 @@ function storeXhr(params, metrics, startTime, endTime, type) {
   }
 
   if (xhrContext.dt) {
-    event.dt = {}
-    event.dt.spanId = xhrContext.dt.spanId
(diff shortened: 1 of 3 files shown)


feature/xhr/aggregate/index.js
```diff
@@ -92,6 +94,13 @@ function storeXhr(params, metrics, startTime, endTime, type) {
     callbackDuration: metrics.cbTime
   }
 
+  if (xhrContext.dt) {
+    event.dt = {}
```
martinkuba, line 98: We could put these values directly on the event object, so that we don't have to create another object.

Issue #105, opened 2021-09-08
CLS - Update Agent Logic


metal-messiah:
## MMF
#89 

## Summary

Update the agent to implement the approach outlined in [Issue 104](https://github.com/newrelic/newrelic-browser-agent/issues/104)

## Desired Behavior

- [ ] Agent should capture CLS according to new specs as decided on in [MMF Doc](https://docs.google.com/document/d/1T514Q6NisxehFtNzm9AOWujwvD7iZoTQfpKKzODmYnw/edit#) and [Issue 104](https://github.com/newrelic/newrelic-browser-agent/issues/104)
- [ ] Agent should have new unit and functional tests to validate new capture method
- [ ] Engineer should validate previous/existing tests still work

PR #108, opened 2021-09-10
update CLS to to max in window of time

metal-messiah:
### Overview
This PR modifies the way CLS is calculated to include looking at windows of time, and taking the maximum. CLS used to be cumulative for the whole page, however now we need to split it to look at a new CLS measurement after 1s between shifts or 5s total, and only keep the biggest one.

### Related Github Issue
[Issue](https://github.com/newrelic/newrelic-browser-agent/issues/105)

### DIFF
--- a/agent/timings.js
+++ b/agent/timings.js
@@ -79,8 +79,8 @@ function recordLcp() {
       'size': lcpEntry.size,
       'eid': lcpEntry.id
     }
-    // LCP is currently implemented in all browsers which support CLS, so this case should always be true
-    if (clsSupported) {
+    // collect 0 only when CLS is supported, since 0 is a valid score
+    if (cls > 0 || clsSupported) {
       attrs['cls'] = cls
     }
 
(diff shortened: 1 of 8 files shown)


agent/timings.js
```diff
@@ -90,17 +90,28 @@ function recordLcp() {
 
 function updateLatestLcp(lcpEntry) {
   if (lcp) {
-    var previous = lcp[0]
+    var previous = lcp
```
martinkuba, line 93: I think this logic with LCP should be kept as before. This is because we want to report the CLS value that was at the time the LCP occurred. LCP is not reported immediately - it may or may not change by the time it is collected.

Not great that our tests did not catch this.

agent/timings.js
```diff
@@ -90,17 +90,28 @@ function recordLcp() {
...
     }
   }
-  lcp = [lcpEntry, cls]
+  lcp = lcpEntry
 }
 
 function updateClsScore(clsEntry) {
   if (clsSupported) {
-    cls += clsEntry.value
+    var n = now()
```
martinkuba, line 103: clsEntry.startTime might be better to use here, since it is guaranteed to be the exact time the shift occurred

tests/functional/timings.test.js
```diff
@@ -561,6 +561,80 @@ function runClsTests(loader) {
...
+        let timingsPromise = router.expectTimings()
+        let domPromise = browser.get(router.assetURL('/'))
+        return Promise.all([timingsPromise, domPromise, loadPromise])
+      })
+      .then(([timingsResult, domResult, loadResult]) => {
+        const {body, query} = timingsResult
+        const timings = querypack.decode(body && body.length ? body : query.e)
+        const timing = timings.find(t => t.name === 'lcp')
+        const cls = timing.attributes.find(a => a.key === 'cls')
+        console.log('CLS COULD BE ANY OF:', loadResult)
```
martinkuba, line 583: Remove console logs?

tests/functional/timings.test.js
```diff
@@ -561,6 +561,80 @@ function runClsTests(loader) {
...
+    Promise.all([rumPromise, loadPromise])
+      .then(() => {
+        let timingsPromise = router.expectTimings()
+        let domPromise = browser.get(router.assetURL('/'))
+        return Promise.all([timingsPromise, domPromise, loadPromise])
+      })
+      .then(([timingsResult, domResult, loadResult]) => {
+        const {body, query} = timingsResult
+        const timings = querypack.decode(body && body.length ? body : query.e)
+        const load = timings.find(t => t.name === 'load')
```
martinkuba, line 616: I am wondering how this is passing since the HTML page is waiting for the window load event to add shifting content (which would not show up on the load timing event). Ultimately, I think it would make sense to inspect the windowUnload timing event.

agent/timings.js
metal-messiah: this can be changed because it now references CLS directly instead of a passthrough which was out of sync previously (and broke tests).  See [line 75](https://github.com/newrelic/newrelic-browser-agent/pull/108/files#diff-895b7b82ab7cdcd68bb5afe6ba6ad5c1ba82e3d5a628afc88fe078f53ba85d45R75), it used to reference the CLS value as a clone of the value which was passed through like `[lcpEntry, cls]`, and subsequently became out of sync.  Now it directly references the global `var cls` which will always be in sync.

tests/functional/timings.test.js
metal-messiah: `loadPromise` is doubly async here, which is why it is in the `Promise.all()` for a second time on `611`, and that is what is returning the `loadResult` array of CLS timings.  working with `wd` is certainly a little cumbersome to retrieve some of these basic properties, but this was just one way I found to pass that data along before the `browser.get('/')` was executed.  Feel free to change it for easier understanding!

metal-messiah: also this `lcpCls` test will need to be changed to match the conclusion we came to in our conversation, as it will no longer be the `largest` cls, you may have to find a different way to ensure it is the correct value.

martinkuba (approved):

aubreymasten (approved):

Issue #90, opened 2021-08-10
Add passive flag for touchstart EventListener


zebau:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )
## Summary
Currently, the `touchstart` EventListener is not using the passive flag therefore the Lighthouse passive event listener audit fails. This was already brought up in the [NewRelic Forum](https://discuss.newrelic.com/t/feature-idea-browser-agent-make-touch-listener-passive/122661) in November 2020 but the issues still seem to persist.

[NOTE]: # ( Tell us how the new feature should work. Be specific. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

## Possible Solution
Add the `{passive: true}` flag to the `touchstart` EventListener 

## Additional context
Google Lighthouse flags event listeners that may delay page scrolling: [web.dev documentation](https://web.dev/uses-passive-event-listeners/)

rr0214:
check to see if it impacts pageview timing event for first interaction.

TheRealJAG:
Thank you New Relic for looking into the issue!

jaGarcia:
+1 vote from me!

fahu:
Hey @metal-messiah I'm super happy that you fixed this issue so fast! Can you give an estimate when this will be released?

sunilit42:
Hey any update on release

PR #113, opened 2021-09-22
add passive flag to touchstart event

metal-messiah:
### Overview
This PR adds a passive flag to `touchstart` event listeners, which was impacting scores from lighthouse.

### Related Github Issue
[Issue 90](https://github.com/newrelic/newrelic-browser-agent/issues/90)

### DIFF
--- a/tools/jil/driver/browser.js
+++ b/tools/jil/driver/browser.js
@@ -21,7 +21,7 @@ function browserTest (name, spec, test) {
 }
 
 window.addEventListener
-  ? window.addEventListener('load', loaded)
+  ? window.addEventListener('load', loaded, false)
   : window.attachEvent('onload', loaded)
 
 function loaded () {


martinkuba:
Should we add the passive flag to all of the events since we are not preventing the default behavior with this listener?

The [Lighthouse documentation](https://web.dev/uses-passive-event-listeners/#how-to-make-event-listeners-passive-to-improve-scrolling-performance) says that there should be a feature detection for older browsers. Is this something we should worry about?

metal-messiah:
> Should we add the passive flag to all of the events since we are not preventing the default behavior with this listener?
> 
> 
> 
> The [Lighthouse documentation](https://web.dev/uses-passive-event-listeners/#how-to-make-event-listeners-passive-to-improve-scrolling-performance) says that there should be a feature detection for older browsers. Is this something we should worry about?

It was my understanding that passive flags are used to optimize scrolling events. So not sure what sort of impact applying passive will do to non-scrolling events.

martinkuba:
The Lighthouse warning is specifically about scrolling performance. The passive flag applies to any event. I was just thinking that there is no reason to split our listeners to passive/non-passive since we never call `preventDefault()`.

martinkuba:
I think we also should add feature detection for this, see [this explainer](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection).

martinkuba (approved):

Issue #122, opened 2021-09-30
Supportability Metrics - Agent


martinkuba:
## Summary

Project: https://github.com/newrelic/newrelic-browser-agent/projects/3

Implement mechanism for collecting Supportability Metrics in the agent.

This should include
* internal API for recording a metric (from anywhere in the agent code)
* harvesting of supportability metrics

PR #126, opened 2021-10-14
added api calls and methods for incrememnting custom metrics and supp…

metal-messiah:
…ortability metrics

---

### Overview
This PR introduces an API method for incrementing custom metrics as well as supportability metrics.  These are generated in the body of the jserrors endpoint.  They are implemented using `customMetrics.incrementCounter([tag], [type])` and `customMetrics.incrementValue([tag], [value | obj], [type])`. One other thing to note is that custom metrics (cm), used to be delivered in the query of the jserror packet.  That has been changed to being supplied in the body, and is where the supportability metric data (sm) will be sent as well.

### Related Github Issue
[GitHub Issue 122](https://github.com/newrelic/newrelic-browser-agent/issues/122)


loader/custom-metrics.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Increments the value of a named property, which defaults to "stats" if no object is supplied
+ * @param {string} name - Name of the metric
+ * @param {object | number} value - The named property to increment, defaults to {stats: number} if no object is supplied
+ * @param {boolean} [isSupportability] - determines which bucket ('cm' or 'sm') to store the metric
+ * @returns void
+ */
+function incrementValue (name, value, isSupportability) {
```
martinkuba, line 29: `recordValue` might be a slightly better name, since it's a new value to be aggregated (not incremented)

loader/api.js
```diff
@@ -70,12 +71,16 @@ mapOwn('actionText,setName,setAttribute,save,ignore,onEnd,getContext,end,get'.sp
 
 function apiCall (name, notSpa, bufferGroup) {
   return function () {
+    var cleanedName = name.replace('api-', '')
+    console.log('an API was called, increment counter for', cleanedName)
```
martinkuba: remove console log

loader/api.js
```diff
@@ -70,12 +71,16 @@ mapOwn('actionText,setName,setAttribute,save,ignore,onEnd,getContext,end,get'.sp
...
+    customMetrics.incrementCounter('API/' + cleanedName + '/called', true)
```
martinkuba, line 75: It might be a little better to have a convenience method for recording supportability metrics, something like what the Node agent has with the [`recordSupportability` function](https://github.com/newrelic/node-newrelic/blob/68da8b2343e5a349f1718ab09c675fd2ab08737e/lib/agent.js#L842)

loader/custom-metrics.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Increments the count of a named metric by 1
+ * @param {string} name - Name of the metric
+ * @param {boolean} [isSupportability] - determines which bucket ('cm' or 'sm') to store the metric
+ * @returns void
+ */
+function incrementCounter (name, isSupportability) {
+  handle('api-storeCustomMetric', [
+    loader.now(),
```
martinkuba, line 15: I think passing the timestamp through may not be necessary, but I understand that it's being consistent what's already there.

tools/jil/util/browser-matcher.js
```diff
@@ -270,6 +270,7 @@ features.addEventListenerOptions = new BrowserMatcher()
 
 features.pushstate = new BrowserMatcher()
   .exclude('ie', '<10')
+  .exclude('ios', '>11.2')
```
martinkuba, line 273: Why is ios excluded from the pushstate feature?

martinkuba (commented):
Over all it looks good, with a few comments
* we probably should have a functional test for recording a supportability metric with a value
* it might be good to have unit tests for the two functions in the new custom-metrics module
* it might make sense to call the `custom-metrics` modules just `metrics` and export functions like `recordSupportability` and `recordCustom`. But this is just a suggestion, will not hold up the PR for this.

PR #154, opened 2021-11-18
add pvt metrics to stn

metal-messiah:
### Overview
This PR adds PVT metrics to the STN payload. This also calculates and adds FID as an event relative to the rest of the timeline.

### DIFF
--- a/feature/stn/aggregate/index.js
+++ b/feature/stn/aggregate/index.js
@@ -18,7 +18,7 @@ var config = require('config')
 if (!harvest.xhrUsable || !loader.xhrWrappable) return
 
 var ptid = ''
-var ignoredEvents = {mouseup: true, mousedown: true, load: true}
+var ignoredEvents = {mouseup: true, mousedown: true}
 var toAggregate = {
   typing: [1000, 2000],
   scrolling: [100, 1000],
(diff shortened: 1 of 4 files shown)


feature/stn/aggregate/index.js
```diff
@@ -115,9 +115,18 @@ ee.on('feat-stn', function () {
   registerHandler('bstXhrAgg', storeXhrAgg)
   registerHandler('bstApi', storeSTN)
   registerHandler('errorAgg', storeErrorAgg)
+  registerHandler('pvtAdded', processPVT)
 })
 
-function storeTiming (_t) {
+function processPVT (name, value, attrs) {
+  var t = {}
+  var owner = name === 'load' ? 'window' : 'document'
```
martinkuba, line 123: I don't recall if the owner value is significant somehow for the backend. We might want to check with the visualization team. If yes, and the distinction between window and document is important, then we should probably also use window for `unload`. I am not sure about the others (fp, fcp, fi, lcp, pageHide).

feature/stn/aggregate/index.js
```diff
@@ -115,9 +115,18 @@ ee.on('feat-stn', function () {
...
+  t[name] = value
+  storeTiming(t, true, owner)
+  if (hasFID(name, attrs)) storeEvent({type: 'fid', target: owner}, owner, value, value + attrs.fid)
```
martinkuba, line 126: Do we want to store FID as an event? The event is really the FI (first input), and FID is its attribute. I can't think of a different way without adding a field in the payload, but wanted to double check.

feature/stn/aggregate/index.js
metal-messiah: this has been deemed insignificant by dempla

martinkuba (approved):

PR #168, opened 2022-01-04
Update changelog for 1214

aubreymasten:


martinkuba:
We need to increase the version number as well, and I think it would make sense for it to be in this PR.

martinkuba (approved):

PR #171, opened 2022-01-13
add detection of popular frameworks, with tests to validate

metal-messiah:
### Overview
This PR introduces framework detection for supportability metrics.  This will help the browser agent determine the ratio of different builds and help align future work priorities.  This is a simple detection method, by inspecting the DOM and the window object for commonly appended tags known to be introduced by popular frontend frameworks.  This is not foolproof, and some cases may be missed, but this methodology will have a much smaller performance hit than something which checks the DOM tree more deeply.

### Testing
This PR includes a new test file, and has introduced a built React app and a built Angular app into the test assets.  The test will simply check whether the agent is able to effectively identify whether the DOM was built with React, Angular, or no framework, and check that supportability metrics were generated from that detection.

#### Metrics are generated under the following pattern:
- Framework/***frameworkName***/Detected

#### Framework names:
`['React', 'Angular', 'AngularJS', 'Backbone', 'Ember', 'Vue', 'Meteor', 'Zepto', 'Jquery']`


loader/framework-detection.js
```diff
@@ -0,0 +1,53 @@
...
+    // not supported?
+  }
+}
+
+function detectReact() {
+  if (!!window.React || !!window.ReactDOM || !!window.ReactRedux) return true
+  if (document.querySelector('[data-reactroot], [data-reactid]')) return true
+  var divs = document.querySelectorAll('body > div')
+  for (var i = 0; i < divs.length; i++) {
+    if (Object.keys(divs[i]).includes('_reactRootContainer')) return true
```
aubreymasten, line 36: [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#browser_compatibility) may not be compatible with older browsers. Could consider replacing this with `Array.indexOf() !== -1`

aubreymasten (approved):

darrinmn9:
Hi @metal-messiah, apologies if I am saying something you already know... but in the case of `Vue`, the most complete detection implementation can be found by looking at the Vue devtools https://github.com/vuejs/devtools/blob/main/packages/shell-chrome/src/detector.js.

As your PR description mentions, maybe you didnt go this route due to the potential performance implications. I just wanted to mention this though. For example, my company's app uses Vue2 + NR agent, but the only way to know Vue is on the webpage is to look for the `__vue__` property on the specific DOM node that our application mounts to (which is impossible for NR to know ahead of time). In that scenario, the vue devtools has to call `document.querySelectorAll('*')` and look through all nodes (which may be too big of a perf hit for NR implementation).

Issue #55, opened 2021-05-25
Run tests on PRs in headless browsers


rr0214:
Currently, all browser-based tests in the agent are run using SauceLabs. This requires a private authentication key and means PRs from external contributors require a manual review by a member of the team to ensure the code is safe to run in an environment where these SauceLabs authentication keys could be accessed.

Instead of running cross-browser tests using SauceLabs, we could instead host a Selenium WebDriver in the same way to run these tests against headless versions of the browsers within the GitHub workflow itself.

This may require some change to the protocol we use to communicate between the test and the WebDriver. Some options for hosting the selenium server are below.

Use WebDriver IO: https://developers.google.com/web/updates/2017/04/headless-chrome#drivers

Host Selenium Hub as a service in the GitHub Action (example here)

Value: currently we have to add a label to run tests, this would allow tests to run automatically on the PR. This would have an additional advantage of contributor PRs auto-running tests

Medium Effort

PR #176, opened 2022-01-31
Workflow for headless browser testing

martinkuba:
### Overview
This adds a new workflow, which runs tests in a headless Chrome browser when a PR is open. The main benefit is that tests are run automatically when PR is open. They should also be faster and more reliable since there is no network latency (unlike with Saucelabs).

The workflow runs both unit and functional tests, which takes about 7 - 8 minutes. Some functional tests had to be excluded (e.g. related to paint, which seems unreliable in headless context). Unit tests, however, work without any issues and take only about 2 minutes. If you decided to run unit tests only, just pass the `--unit-only` flag to the test command in the workflow file. 

### Related Github Issue
Addresses https://github.com/newrelic/newrelic-browser-agent/issues/55

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
     "packages/*",
     "tools/test-builds/build-time-mfe/*"
   ],
-  "version": "0.0.9-beta.121"
+  "version": "0.0.9"
 }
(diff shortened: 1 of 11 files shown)


martinkuba:
@metal-messiah @aubreymasten Have you been able to review this? Is there anything I can do or should I close?

metal-messiah:
> @metal-messiah @aubreymasten Have you been able to review this? Is there anything I can do or should I close?

I will bring this back up in stand-up this morning!  We reviewed it shortly after you posted it and then it has sat dormant with the other work we have had, but this is worth incorporating!

PR #185, opened 2022-02-15
Trace IDs should be 32 bytes of random hex chars

jeffalder:

---

### Overview
The function as-is will always generate a trace ID that ends in 0. The reason is that the array passed to crypto.getRandomValues() only has 31 elements. That's fine when you're looking at the GUID function, which only has 31 random characters.

In the current code, the array is created with 31 elements, indexed 0-30. The last character is pulled using `randomVals[31]` which returns undefined, which gets translated to 0 (`undefined & 15 === 0`).

### Related Github Issue
none

### Testing
Just made a quick test in Firefox.

---   

### DIFF
--- a/loader/unique-id.js
+++ b/loader/unique-id.js
@@ -62,7 +62,7 @@ function generateRandomHexString(length) {
   var rvIndex = 0
   var crypto = window.crypto || window.msCrypto
   if (crypto && crypto.getRandomValues && Uint8Array) {
-    randomVals = crypto.getRandomValues(new Uint8Array(31))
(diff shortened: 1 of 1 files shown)


coreyarnold:
Great catch @jeffalder! We'll take a look at this change and see about getting this PR merged and ready for the next release. :connected: :passionate:

jeffalder:
Thanks @coreyarnold ! I followed up and changed it slightly so that it would only allocate `length` bytes if that's all it needs.

jeffalder:
🍠

PR #231, opened 2022-06-11
Fix: LCP metric no longer reported on initially hidden pages

### LINKED ISSUE #170: LCP should not be reported when page is opened and loaded in background
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )
The NewRelic agent reports LCP times for pages that are opened and loaded as a background tab (i.e. the `document.visibilityState` is initially `hidden` and remains as such until user focuses the tab). This is different from the web-vitals library which does not report an LCP when the page is hidden since navigation start.

## Steps to Reproduce

[NOTE]: # ( Please be as specific as possible. )
See video here for details: https://youtu.be/0gH6NeJqLZw

**tl;dr** Open page as background tab. Wait for approx 30 seconds (or however long). Make page visible. See inflated effects on LCP times reported by NR browser agent

## Expected Behavior

[NOTE]: # ( Tell us what you expected to happen. )
Pages that are initially opened and loaded in the background should not report their LCP

[NOTE]: # ( Please provide specifics of the local error logs, Browser Dev Tools console, etc. if appropriate and possible. )
See video for context

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )

* Chrome  96.0.4664.93

## Additional context

[TIP]:  # ( Add any other context about the problem here. )
https://github.com/GoogleChrome/web-vitals/blob/main/src/lib/getVisibilityWatcher.ts
note that `firstHiddenTime` is assigned a value at initialization to account for the scenario reported above.

cwli24:

---

### Overview
Previously, pages opened in the background and had a `visibilityState == 'hidden'` would continue to report the LCP metric, and an inflated one at that.

Now, in such cases the LCP will not be reported (or even buffered at all). 

### Related Github Issue
Fixes #170 

### Testing
- New html added to `tests/assets/`
- Test subcase appended to existing case in `tests/functional/timings.test.js`

In addition to manually pasting the Browser Agent, follow the test case in the related issue to add a basic performance observer and/or core-web-vitals' `getLCP` for comparison. Example:
image
It may be helpful to programmatically include this in the `lcpObserver` code right before the `handle("lcp", payload)` line:
`console.log("NR LCP Observer - handling lcp payload", payload);`
\
\
Now on your test server, test the workflow of opening a new in-focus tab and background tab.

- For the foreground tab, the regression test should show the LCP output in console, with the data reported to NR1:
image

- While the background tab should show the output from NR agent (and CWV) missing, with just the `performance observer LCP` output.

---   

### DIFF
--- a/cdn/webpack.config.js
+++ b/cdn/webpack.config.js
@@ -16,9 +16,9 @@ module.exports = {
   },
   output: {
     filename: '[name].js',
-    path: path.resolve(__dirname, './build'),
+    path: path.resolve(__dirname, '../build'),
     // publicPath: 'https://js-agent.newrelic.com/', // <- we need this to be set when we publish to the CDN
-    publicPath: '/cdn/build/', // <-- we need one property to be set when testing locally
+    publicPath: '/build/', // <-- we need one property to be set when testing locally
     library: {
       name: 'NRBA',
       type: 'umd'
--- a/packages/browser-agent-core/features/metrics/instrument/index.js
+++ b/packages/browser-agent-core/features/metrics/instrument/index.js
@@ -55,7 +55,7 @@ export class Instrument extends FeatureBase {
 
     singleChecks() {
         // note the browser agent version
-        this.recordSupportability(`Generic/Version/${VERSION}`)
+        this.recordSupportability(`Generic/Version/${VERSION}/Detected`)
 
         // frameworks on page
         getFrameworks().forEach(framework => {


metal-messiah (approved):
Great first PR.  Case appears to be covered.  Thanks for updating both code-sets.

PR #243, opened 2022-06-30
allow jil to skip .browser.js test files

metal-messiah:
### Overview
This PR updates JIL to allow `.browser.js` test files to be skipped based on spec rules using `browserMatcher`.  These are not able to be read from the file itself in browser tests, so spec files are now used to export the rules and supply to JIL when preprocessing.

Spec files exist in a 1:1 relationship to tests.  Not all tests need spec files.  These tests with no spec file will run in all spun-up browser instances.  Test files with a spec file will only run in browser instances that match the browserMatcher class exported in the spec.

Functional tests remain untouched.


tools/jil/runner/index.js
```diff
@@ -77,8 +77,14 @@ function loadDefaultFiles(cb) {
 function loadFiles(testFiles, cb) {
   for (let file of testFiles) {
     file = resolve(process.cwd(), file)
-    if (file.slice(-11) === '.browser.js' || (file.includes('/browser/') && file.endsWith('.test.js'))) {
-      loadBrowser(testDriver, file)
+    if (file.slice(-11) === '.browser.js') {
+      let spec
+      try {
+        spec = require(file.replace('.browser.', '.spec.'))
```
metal-messiah, line 83: This line in the runner now checks for a spec file to include in the `loadBrowser()` call that stages the browser tests. The spec file must be in the same directory as the test, and share the same name as the test (with spec.js suffix)

tests/browser/add-event-listener-options.browser.js
```diff
@@ -4,10 +4,8 @@
  */
 
 const jil = require('jil')
-let matcher = require('../../tools/jil/util/browser-matcher')
```
metal-messiah, line 7: You do not need to remove these lines from the test for it to work, as they are not evaluated anyway.  But these could be removed for clarity in all browser test files if desired.

tests/browser/add-event-listener-options.spec.js
```diff
@@ -0,0 +1,2 @@
+let matcher = require('../../tools/jil/util/browser-matcher')
+module.exports = matcher.withFeature('addEventListenerOptions')
```
metal-messiah, line 2: Example of what a `*.spec.js` file should export

cwli24:
+Noting that as a side effect, **all** existing `.test.js` files in browser tests folder are changed to `.browser.js`.

cwli24 (commented):
👍+Noting that as a side effect, **all** existing `.test.js` files in browser tests folder are changed to `.browser.js`.

PR #251, opened 2022-07-12
Test migration: All tests safari latest

metal-messiah:
### Overview 
**---Major changes PR---**

- Re-defines rudimentarily the browsers and versions cutoff that we support
- Updates to GH actions and JIL test suite
- Expand on last chrome@latest checkpoint to ensure **all** defined browser-version functional & unit test runs are passing
- Changes to some core source files and test files, especially SPA, to reflect working state

This is an all-passing checkpoint for the test migration project, barring cleanup and other post tasks.


packages/browser-agent-core/common/window/nreum.js
```diff
@@ -94,6 +94,26 @@ export function gosNREUMInitializedAgents(id, obj, target) {
   return nr
 }
 
+function gosNRDEBUG() {
```
aubreymasten, line 97: Previously this code was injected by JIL when tests were running. Is this temporary/will this be added to the production agent?

metal-messiah: After thinking, I think this should actually be temporary

cwli24: > After thinking, I think this should actually be temporary (and cleaned up)

Done in the cleanup branch; disregard this on here for now. Alternatively, can merge that in but not preferable as it'll bloat this one and make review harder for everyone to go through a bunch of commented-out removals.

cwli24:
Please don't squash this when merging. Would "lose" the last few weeks of granular history on new `dev`, plus `main` is now tangled into this history.

PR #252, opened 2022-07-13
Added sessionId and deprecated 'ck' param in query string

cwli24:
### Overview
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
Details of the change can be found in [this CDD](https://newrelic.atlassian.net/wiki/spaces/INST/pages/2522513791/JSESSIONID+Cookie+Change+Design+Document).
This commit:
- sunsets the `ck` param in harvest request URL to beacon
- creates and implement the new end-user generated `sessionId` in sessionStorage 

### Related Github Issue
N/A

### Testing
- `browser/harvest.browser.js` was updated to reflect this change
- `functional/harvest.test.js` was updated as well
 
**Other considerations:**
- (New) Unit testing deemed unnecessary at this time.
- Functionality of changes is closely tied to the functionality of sessionStorage itself. Hence, new automated functional tests are not added either.
-- Manually tested via navigation, e.g., duplicating tab, refreshing page, going back and forth.
-- Particularly, the case where navigating from "has-cookie.html" (`sessionId = 123`) to "no-cookie.html" (`sessionId = 0`) back to "has-cookie.html" will make `sessionId = 123` again.

### DIFF
--- a/packages/browser-agent-core/common/config/state/runtime.js
+++ b/packages/browser-agent-core/common/config/state/runtime.js
@@ -3,7 +3,7 @@ import { getLastTimestamp } from '../../timing/now'
 import * as userAgent from '../../util/user-agent'
 import { Configurable } from './configurable'
 import { gosNREUMInitializedAgents } from '../../window/nreum'
-import { getCurrentSessionId } from '../../window/session-storage'
+import { getCurrentSessionIdOrMakeNew } from '../../window/session-storage'
 import { getConfigurationValue } from '../config'
 
 var XHR = window.XMLHttpRequest
@@ -19,7 +19,7 @@ const model = agentId => { return {
   origin: '' + window.location,
   ptid: undefined,
   releaseIds: {},
-  sessionId: getConfigurationValue(agentId, 'privacy.cookies_enabled') === true ? getCurrentSessionId() : '0',
+  sessionId: getConfigurationValue(agentId, 'privacy.cookies_enabled') === true ? getCurrentSessionIdOrMakeNew() : '0',
   xhrWrappable: XHR && XHR_PROTO && XHR_PROTO['addEventListener'] && !/CriOS/.test(navigator.userAgent),
   userAgent
 }}
--- a/packages/browser-agent-core/common/config/state/runtime.js
+++ b/packages/browser-agent-core/common/config/state/runtime.js
@@ -3,23 +3,26 @@ import { getLastTimestamp } from '../../timing/now'
 import * as userAgent from '../../util/user-agent'
 import { Configurable } from './configurable'
 import { gosNREUMInitializedAgents } from '../../window/nreum'
+import { getCurrentSessionId } from '../../window/session-storage'
+import { getConfigurationValue } from '../config'
 
 var XHR = window.XMLHttpRequest
 var XHR_PROTO = XHR && XHR.prototype
 
-const model = {
-  origin: '' + window.location,
+const model = agentId => { return {
+  customTransaction: undefined,
+  disabled: false,
+  features: {},
   maxBytes: ieVersion === 6 ? 2000 : 30000,
   offset: getLastTimestamp(),
-  features: {},
-  customTransaction: undefined,
   onerror: undefined,
-  releaseIds: {},
-  xhrWrappable: XHR && XHR_PROTO && XHR_PROTO['addEventListener'] && !/CriOS/.test(navigator.userAgent),
-  disabled: false,
+  origin: '' + window.location,
   ptid: undefined,
+  releaseIds: {},
+  sessionId: getConfigurationValue(agentId, 'privacy.cookies_enabled') === true ? getCurrentSessionId() : '0',
+  xhrWrappable: XHR && XHR_PROTO && XHR_PROTO['addEventListener'] && !/CriOS/.test(navigator.userAgent),
   userAgent
-}
+}}
 
 const _cache = {}
 
@@ -31,6 +34,6 @@ export function getRuntime(id) {
 
 export function setRuntime(id, obj) {
   if (!id) throw new Error('All runtime objects require an agent identifier!')
-  _cache[id] = new Configurable(obj, model)
+  _cache[id] = new Configurable(obj, model(id))
   gosNREUMInitializedAgents(id, _cache[id], 'runtime')
 }
--- a/packages/browser-agent-core/common/window/session-storage.js
+++ b/packages/browser-agent-core/common/window/session-storage.js
@@ -7,7 +7,7 @@
  */
 import { generateRandomHexString } from '../ids/unique-id'
 
-export { getCurrentSessionId };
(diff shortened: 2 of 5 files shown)


packages/browser-agent-core/common/window/session-storage.js
```diff
@@ -0,0 +1,27 @@
...
+
+export { getCurrentSessionId };
+
+const ss = window.sessionStorage;
+const SESS_ID = Symbol("session id").toString(); // prevents potential key collisions in session storage
+
+/**
+ * @returns {string} This tab|window's session identifier, or a new ID if not found in storage
+ */
+function getCurrentSessionId() {
```
metal-messiah, line 18: I like this method, but can we change the name a bit to illustrate what it is doing more succinctly?

something like `getOrSetCurrentSessionId()` or whatever you like that illustrates that it is also setting state in certain circumstances.

tests/browser/harvest.browser.js
```diff
@@ -95,20 +95,22 @@ function validateUrl (t, actualUrl, expectedUrlTemplate, message) {
...
       ckValue = pair[1]
+    } else if (pair[0] === 's') {
+      sessionId = pair[1];
     }
   }
 
   // In addition to replacing timestamp, add in the ck parameter which goes after timestamp
-  let expectedUrl = expectedUrlTemplate.replace('{TIMESTAMP}', submissionTimestamp + '&ck=' + ckValue)
-
+  let expectedUrl = expectedUrlTemplate.replace('{TIMESTAMP}', `${submissionTimestamp}&ck=${ckValue}&s=${sessionId}`)
```
metal-messiah, line 112: Just want to confirm, Is `s=0` going to be our way of handling **both** cases of disabling? 
ie:
1.  if the end-user disables this behavior
3. if the customer disables this behavior?

packages/browser-agent-core/common/config/state/runtime.js
```diff
@@ -3,23 +3,26 @@ import { getLastTimestamp } from '../../timing/now'
...
+  features: {},
   maxBytes: ieVersion === 6 ? 2000 : 30000,
   offset: getLastTimestamp(),
-  features: {},
-  customTransaction: undefined,
   onerror: undefined,
+  origin: '' + window.location,
+  ptid: undefined,
   releaseIds: {},
+  sessionId: getConfigurationValue(agentId, 'privacy.cookies_enabled') === true ? getCurrentSessionId() : '0',
```
metal-messiah, line 22: Just to play devil's advocate, can you think of any cases where we would need to get this sessionID after load time?  Like if the end-user clears their cache in the middle of a page session, are they expecting future payloads to not have that value? Do we need to generate a new one?

This is likely a non issue, but would like to confirm. cc: @aubreymasten

tests/browser/harvest.browser.js
cwli24: That appears to be the case according to the doc, yes, and I don't see a need to distinguish them.
As it stands, from talking with @aubreymasten before, no.1 isn't even supported right now as in there's no mechanism to flip the switch during runtime right now while on the same page.

packages/browser-agent-core/common/config/state/runtime.js
cwli24: Can you specify what cache you mean? Like all the state files' internal `_cache` per agent? Or the sessionStorage?
Wrt sessionStorage:
- It's true eu's can manually clear their storage, but if we're making an analogous reference to cookies, they could've wiped cookies mid-session too.
- In the worst case where `getCurrentSessionId()` is called again, and the sessionStorage has been wiped, it would just generate a new value for all future payloads (and store that again), so the session got effectively "lost" (interrupted). I think that makes sense?
- More interestingly, I want to discuss current behavior. Making the `sessionId` part of the per-agent runtime properties rather than retrieving it from SS each use does alleviate that concern a bit. In that case, the `sessionId` would retain the old value even after SS has been wiped, _until the time when `setRuntime()` is called again_. This relates to the way the default `model` and `Configurable()` was done, and I should mention at that time, it'll also be dependent on the then-value of privacy.cookies_enabled.

packages/browser-agent-core/common/window/session-storage.js
cwli24: Revised

packages/browser-agent-core/common/config/state/runtime.js
metal-messiah: ```More interestingly, I want to discuss current behavior. Making the sessionId part of the per-agent runtime properties rather than retrieving it from SS each use does alleviate that concern a bit. In that case, the sessionId would retain the old value even after SS has been wiped, until the time when setRuntime() is called again. This relates to the way the default model and Configurable() was done, and I should mention at that time, it'll also be dependent on the then-value of privacy.cookies_enabled.```

Yep this is the part I was pointing to.  The behavior here would mean that if a client cleared their storage mid-session, we would still be sending "old" session data, which is probably fine.  Just wanted to confirm that that behavior is acceptable.

tests/browser/harvest.browser.js
metal-messiah: Right, but `#1` is on our board to implement for the current MMF so we should consult with John if that is still needed.

packages/browser-agent-core/common/config/state/runtime.js
cwli24: Oh gotcha. I think that behavior makes sense, since the conventional `session` is defined per lifetime of a tab or window. If anything, what's wrong in that scenario is that _something_ (not just the e.u. but I think fellow scripts under the same domain can also `.clear()` the storage) is going in and destroying that definition.

metal-messiah (approved):
Looks great!

packages/browser-agent-core/common/window/session-storage.js
```diff
@@ -0,0 +1,28 @@
...
+ * 
+ * @documentation https://newrelic.atlassian.net/wiki/spaces/INST/pages/2522513791/JSESSIONID+Cookie+Change+Design+Document
+ * @environment Browser script
+ */
+import { generateRandomHexString } from '../ids/unique-id'
+
+export { getCurrentSessionIdOrMakeNew };
+
+const ss = window.sessionStorage;
+const SESS_ID = Symbol("session id").toString(); // prevents potential key collisions in session storage
```
martinkuba, line 13: Symbol is not supported on all browsers

cwli24: Good call out. Since symbols would increase the polyfill from 62kb to 82kb, it's deemed unnecessary, and this is usage is removed.

tests/browser/harvest.browser.js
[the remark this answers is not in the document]
cwli24: Our Babel config includes this plugin. Tested and works on IE(11).

cwli24:
Has testing dependency on #251 .

PR #255, opened 2022-07-25
[WIP] Modularity Change Review

aubreymasten:

---

### Overview
Please describe the changes present in the pull request and if applicable, describe why the changes are needed.

### Related Github Issue
Please include a link to the related GitHub issue, if applicable

### Testing
The browser agent includes a suite of tests which should be used to
verify that your changes don't break existing functionality. These tests will run through Github Actions when:

 1) A pull request is made _and_ 
 2) The pull request is marked as ready for testing by the internal team. 

If applicable, please include details on how to test your changes.

---   


cdn/.env
```diff
@@ -0,0 +1,2 @@
+PATH=
```
aubreymasten, line 1: What does this file do?

packages/browser-agent/.env
```diff
@@ -0,0 +1 @@
+SUBPATH=
```
aubreymasten, line 1: Does this serve a purpose? Should we gitignore this?

cdn/.env
metal-messiah: I think nothing, remnant of early webpack inquiry, should be removed

tools/jil/server/asset-server.js
```diff
@@ -531,12 +590,12 @@ class AssetServer extends BaseServer {
       })
     } else {
       rsp.writeHead(404)
-      rsp.write(`Failed to locate asset at '${assetPath}'`)
+      // rsp.write(`Failed to locate asset at '${assetPath}'`)
```
aubreymasten, line 593: Should we uncomment this?

packages/browser-agent/.env
metal-messiah: Removing this may have an impact but only on the NPM package build process

tools/jil/server/asset-server.js
metal-messiah: I'll leave that up to you

packages/browser-agent-core/features/ajax/instrument/index.js
```diff
@@ -0,0 +1,391 @@
...
+    this.txSize = dataSize(opts.body) || 0
+  }
+
+  // we capture failed call as status 0, the actual error is ignored
+  // eslint-disable-next-line handle-callback-err
+  function onFetchDone(err, res) {
+    this.endTime = now()
+    if (!this.params) {
+      this.params = {}
+    }
```
aubreymasten, line 319: A check was removed for Data URLs that I think should still exist: https://github.com/newrelic/newrelic-browser-agent/blob/e0addb7c19e52429cc17b50a4e6f01892b3c1fc1/feature/xhr/instrument/index.js#L268-L269

packages/browser-agent-core/features/ajax/instrument/index.js
```diff
@@ -0,0 +1,391 @@
...
+    this.dt = new DT(this.agentIdentifier)
+
+    this.handler = (type, args, ctx, group) => handle(type, args, ctx, group, this.ee)
+    this.wrappedFetch = getWrappedFetch(this.ee)
+    wrapXhr(this.ee)
+    subscribeToEvents(this.agentIdentifier, this.ee, this.handler, this.dt)
+  }
+}
+
+// TODO update all of this to go into class and use this.ee for ee, handle, and register
```
aubreymasten, line 44: Do we want to take action on this before the release?

packages/browser-agent-core/features/ajax/instrument/index.js
```diff
@@ -0,0 +1,391 @@
...
+
+    // Don't instrument Chrome for iOS, it is buggy and acts like there are URL verification issues
+    if (!agentRuntime.xhrWrappable || agentRuntime.disabled) return
+
+    agentRuntime.features.xhr = true;   // declare that we are using xhr instrumentation
+
+    this.dt = new DT(this.agentIdentifier)
+
+    this.handler = (type, args, ctx, group) => handle(type, args, ctx, group, this.ee)
+    this.wrappedFetch = getWrappedFetch(this.ee)
```
aubreymasten, line 38: I don't see this used anywhere in the agent. Is this from the original modularity prototype? Can we delete it?

packages/browser-agent-core/features/ajax/instrument/index.js
aubreymasten, line 44: Should we create a ticket for this or update before release?

packages/browser-agent-core/features/ajax/instrument/index.js
aubreymasten, line 319: This is missing a supportability metric from the original code:
https://github.com/newrelic/newrelic-browser-agent/blob/e0addb7c19e52429cc17b50a4e6f01892b3c1fc1/feature/xhr/instrument/index.js#L287-L290

packages/browser-agent-core/features/ajax/instrument/index.js
```diff
@@ -0,0 +1,391 @@
...
+    var url
+    if (typeof target === 'string') {
+      url = target
+    } else if (typeof target === 'object' && target instanceof origRequest) {
+      url = target.url
+    } else if (window.URL && typeof target === 'object' && target instanceof URL) {
+      url = target.href
+    }
+    addUrl(this, url)
+
```
aubreymasten, line 305: Missing a check for Data URL:
https://github.com/newrelic/newrelic-browser-agent/blob/e0addb7c19e52429cc17b50a4e6f01892b3c1fc1/feature/xhr/instrument/index.js#L268-L269

packages/browser-agent-core/features/ajax/aggregate/index.js
```diff
@@ -0,0 +1,252 @@
...
+import { FeatureBase } from '../../../common/util/feature-base'
+
+export class Aggregate extends FeatureBase {
+  constructor(agentIdentifier, aggregator, externalFeatures = []) {
+    super(agentIdentifier, aggregator, externalFeatures)
+    let ajaxEvents = []
+    let spaAjaxEvents = {}
+    let sentAjaxEvents = []
+    let scheduler
+
```
aubreymasten, line 22: Missing runtime check for whether instrumentation was loaded:

https://github.com/newrelic/newrelic-browser-agent/blob/e0addb7c19e52429cc17b50a4e6f01892b3c1fc1/feature/xhr/aggregate/index.js#L26-L31

packages/browser-agent-core/features/ajax/aggregate/index.js
```diff
@@ -0,0 +1,252 @@
...
+    register('xhr', storeXhr, undefined, this.ee)
+
+    if (allAjaxIsEnabled()) {
+      scheduler = new HarvestScheduler('events', {
+        onFinished: onEventsHarvestFinished,
+        getPayload: prepareHarvest
+      }, this)
+
+      externalFeatures.forEach(feat => {
```
aubreymasten, line 60: This is the metrics harvest, should live outside of allAjaxIsEnabled check

packages/browser-agent-core/features/jserrors/instrument/index.js
```diff
@@ -0,0 +1,127 @@
...
+    })
+
+    state.ee.on('internal-error', (e) => {
+      handle('ierr', [e, now(), true], undefined, undefined, state.ee)
+    })
+
+    // Declare that we are using err instrumentation
+    // require('./debug')
+
+    const prevOnError = window.onerror
```
aubreymasten, line 60: This seems to have changed the behavior compared to the original code. Potentially causing the original onErrorHandler to be called twice.

packages/browser-agent-core/features/jserrors/instrument/index.js
```diff
@@ -0,0 +1,127 @@
...
+    window.onerror = (...args) => {
+      if (prevOnError) prevOnError(...args)
+      this.onerrorHandler(...args)
+      return false
+    }
+
+    try {
+      window.addEventListener('unhandledrejection', (e) => {
```
aubreymasten, line 68: What is this try/catch supposed to do and where did it come from? It looks like it was added in the Dev commit but I don't see it in the original code.

packages/browser-agent-core/features/jserrors/aggregate/index.js
```diff
@@ -0,0 +1,267 @@
...
+    super(agentIdentifier, aggregator)
+    this.stackReported = {}
+    this.pageviewReported = {}
+    this.errorCache = {}
+    this.currentBody
+
+    this.errorOnPage = false
+
+    // this will need to change to match whatever ee we use in the instrument
+    this.ee.on('interactionSaved',  (interaction) => {
```
aubreymasten, line 32: Could we make this a named class function and move it out of the constructor?

packages/browser-agent-core/features/page-view-timing/instrument/index.js
```diff
@@ -0,0 +1,149 @@
...
+    this.pageHiddenTime
+    this.performanceObserver
+    this.lcpPerformanceObserver
+    this.clsPerformanceObserver
+    this.fiRecorded = false
+
+
+    // // NREUM.debug("initialize page-view-timing instrument!", agentIdentifier)
+
+    if (this.isEnabled()) {
```
aubreymasten, line 25: This should be handled by getEnabledFeatures in the loader files:
https://github.com/newrelic/newrelic-browser-agent/blob/7922cefe29918f7a608e6f7c613b8d1c28de897f/cdn/agent-loader/lite.js#L23-L26

packages/browser-agent-core/features/ajax/instrument/index.js
metal-messiah: I think this is comment can be deleted, as `this.ee` is effectively passed into the dependent methods.

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/255#discussion_r931703594

packages/browser-agent-core/features/ajax/instrument/index.js
metal-messiah: Good catch. Needs to be added back.

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/projects/7#card-84572817

packages/browser-agent-core/features/ajax/instrument/index.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/projects/7#card-84572817

packages/browser-agent-core/features/ajax/aggregate/index.js
metal-messiah: The flag was needed before the modularity work, because features initialized as a side-effect of import.  Now we can be more direct when initializing the class instances.

packages/browser-agent-core/features/ajax/aggregate/index.js
metal-messiah: This is a nice catch.  This could be a new ticket for after the release, to extend the metrics feature to do more than just supportability and remove the dependency from AJAX feature.

packages/browser-agent-core/features/jserrors/instrument/index.js
metal-messiah: This is necessary for multiple instances of the agent on one page to have a page-level error handler. An HTML page starts with `window.onerror` as null by default, meaning it will only fire previous error handlers if they exist at the time of instrumenting that instance of the jserrors class (presumably by being set by one or more previous instances of the agent). 

This can be discussed further if necessary

packages/browser-agent-core/features/jserrors/instrument/index.js
metal-messiah: I **think** this was added in the original NPM work, to catch unhandled promise rejections.  This is likely unnecessary now that all the `wrap` files have been implemented, but it is worth checking.

packages/browser-agent-core/features/jserrors/aggregate/index.js
metal-messiah: agreed

packages/browser-agent-core/features/page-view-timing/instrument/index.js
metal-messiah: agreed, can be removed

packages/browser-agent-core/features/ajax/instrument/index.js
metal-messiah: up for discussion though

cwli24: I added this; you can delete it if you want; at this point, all of the instrument & aggregate could be re-evaluated for rework or styling changes.

packages/browser-agent-core/common/aggregate/aggregator.js
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SharedContext } from '../context/shared-context'
+import { mapOwn } from '../util/map-own'
+
+export class Aggregator extends SharedContext {
```
aubreymasten, line 9: Some clarity in the names of these functions has been lost because we've named the methods what they were named in the original export. Maybe we can review this file to rename and re-add some old comments:

https://github.com/newrelic/newrelic-browser-agent/blob/487a282f92dd02a32b2b94d41bd7b718f09ff30c/agent/aggregator.js#L11-L15

cwli24: Btw my 2 cents from working on tests: it did not make sense as reader for the code to rename a specialized function with a generalized version of its name. For example, don't expose "storeMetricUnderAgentEE" as "storeMetric". It's non-descript, vague, and potentially misleading. Expose methods as they are, or at least abbreviate them without losing context, e.g. "storeMetricAgtEE".

packages/browser-agent-core/common/window/load.js
```diff
@@ -0,0 +1,23 @@
...
+const doc = win.document
+const ADD_EVENT_LISTENER = 'addEventListener'
+const ATTACH_EVENT = 'attachEvent'
+
+function checkState (cb) {
+  if (doc.readyState === 'complete') cb()
+}
+
+export function onWindowLoad(cb) {
+  checkState(cb)
```
aubreymasten, line 13: This checkState will call the onWindowLoaded callback if the agent loads after the window.load and set the onload and load timings incorrectly.
https://github.com/newrelic/newrelic-browser-agent/blob/7e13aea02cc7ac928592bfaf71e5a6f0d18bac23/packages/browser-agent-core/features/page-view-event/instrument/index.js#L18-L28

aubreymasten: The old agent is clearer that `measureWindowLoaded()` (`windowLoaded()` in the old agent) should only ever be called more than once:

https://github.com/newrelic/newrelic-browser-agent/blob/7e13aea02cc7ac928592bfaf71e5a6f0d18bac23/loader/index.js#L77-L79

packages/browser-agent-core/common/window/load.js
```diff
@@ -0,0 +1,23 @@
...
+  if (doc[ADD_EVENT_LISTENER]) win[ADD_EVENT_LISTENER]('load', cb, eventListenerOpts(false))
+  else win[ATTACH_EVENT]('onload', cb)
+}
+
+export function onDOMContentLoaded(cb) {
+  checkState(cb)
```
aubreymasten, line 19: I think the comment above applies here too, `doc.readyState === 'complete'` could be any point after DOMContentLoaded. If we get to create a register before that event, we should call it, but if the event has already occurred we shouldn't call the callback immediately because it relies on `now()` and assumes we are at the moment of DOMContentLoaded.

packages/browser-agent-core/common/wrap/index.js
```diff
@@ -0,0 +1,46 @@
+
```
aubreymasten, line 1: What does this file do?

packages/browser-agent-core/common/wrap/wrap-events.js
```diff
@@ -0,0 +1,84 @@
...
+
+  function uniqueListener(args, obj) {
+    // Context for the listener is stored on itself.
+    return args[1]
+  }
+
+  return ee
+}
+
+export function scopedEE(sharedEE){
```
aubreymasten, line 81: I only see this `export function scopedEE` pattern used as an import for one test, and I'm not sure what it's doing.

I see scopedEE imported in this test, but I think the test needs to be changed to import and call wrapEvents

https://github.com/newrelic/newrelic-browser-agent/blob/7e13aea02cc7ac928592bfaf71e5a6f0d18bac23/tests/browser/add-event-listener-options.browser.js#L8-L10

Is there any other use for this extra export on all the wrap-* files?

tests/browser/add-event-listener-options.browser.js
```diff
@@ -4,11 +4,10 @@
...
 
 const jil = require('jil')
-let matcher = require('../../tools/jil/util/browser-matcher')
-let supported = matcher.withFeature('addEventListenerOptions')
 
-jil.browserTest('addEventListener options work when wrapped', supported, function (t) {
-  require('../../feature/wrap-events.js')
+jil.browserTest('addEventListener options work when wrapped', async function (t) {
+  const { scopedEE } = await import('../../packages/browser-agent-core/common/wrap/wrap-events');
+  scopedEE();
```
aubreymasten, line 10: I think this may want to be replaced with wrapEvents()

packages/browser-agent-core/common/wrap/wrap-history.js
```diff
@@ -0,0 +1,31 @@
...
+  var object = window.history
+  if (prototype && prototype.pushState && prototype.replaceState) {
+    object = prototype
+  }
+  // log('wrap history')
+  wrapFn.inPlace(object, [ 'pushState', 'replaceState' ], '-')
+
+  return ee
+}
+3
```
aubreymasten, line 28: Remove

packages/browser-agent-core/common/wrap/wrap-promise.js
```diff
@@ -0,0 +1,147 @@
...
+  
+        return returnVal
+      }
+    })
+  
+    OriginalPromise.prototype['catch'] = function wrappedCatch (fn) {
+      return this.then(null, fn)
+    }
+  
+    Object.assign(OriginalPromise.prototype, {constructor: {value: WrappedPromise}})
```
aubreymasten, line 66: Why was this changed?

packages/browser-agent-core/common/wrap/wrap-xhr.js
```diff
@@ -0,0 +1,192 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// wrap-events patches XMLHttpRequest.prototype.addEventListener for us.
+import './wrap-events'
```
aubreymasten, line 7: I don't think this will guarantee that events are wrapped. Other features call wrap-events, so for Lite, Pro, SPA this should work. But the Ajax feature on its own assumes that `import './wrap-events'` does the same thing as the previous `require('./wrap-events')` and auto-wraps.

To be safe we may want to call wrapEvents() around line 22.

This is a dependency that maybe we want to represent somewhere more unified as well.

packages/browser-agent-core/common/url/protocol.js
```diff
@@ -0,0 +1,20 @@
...
+export const protocol = {
+  isFileProtocol: isFileProtocol,
+  supportabilityMetricSent: false
+}
+
+function isFileProtocol () {
+  let win = getWindow();
+  let isFile = !!(win.location && win.location.protocol && win.location.protocol === 'file:');
+  if (isFile) {
+    //metrics.recordSupportability('Generic/FileProtocol/Detected') -- may be implemented later? Probably make sure it's once per window
```
aubreymasten, line 16: This should be re-implemented before release. We may also want to note this as a risk for multiple agent instances, like the comment says

packages/browser-agent-core/common/util/obfuscate.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+let recordedSupportability = false
+
+export class Obfuscator extends SharedContext {
+  constructor(parent) {
+    super(parent) // gets any allowed properties from the parent and stores them in `sharedContext`
+
+    /* Note: we may need these lines from legacy code -- (when passing in an optional metrics Instrument instance as argument)
+    if (shouldObfuscate()) metrics.recordSupportability('Generic/Obfuscate/Detected')
```
aubreymasten, line 17: These should be re-implemented before release

packages/browser-agent-core/common/util/obfuscate.js
```diff
@@ -0,0 +1,80 @@
...
+    for (var i = 0; i < rules.length; i++) {
+      var regex = rules[i].regex
+      var replacement = rules[i].replacement || '*'
+      obfuscated = obfuscated.replace(regex, replacement)
+    }
+    return obfuscated
+  }
+}
+
+// TO DO: this function should be inside the Obfuscator class since its context relates to agentID
```
aubreymasten, line 44: Just for consistency we should finish this task

packages/browser-agent-core/features/spa/instrument/index.js
```diff
@@ -0,0 +1,110 @@
...
+        const tracerEE = this.ee.get('tracer')
+        const jsonpEE = wrapJson(this.ee)
+        const promiseEE = wrapPromise(this.ee)
+        const eventsEE = wrapEvents(this.ee)
+        const timerEE = wrapTimer(this.ee)
+        const xhrEE = wrapXhr(this.ee)
+        const fetchEE = wrapFetch(this.ee)
+        const historyEE = wrapHistory(this.ee)
+        const mutationEE = wrapMutation(this.ee)
+
```
aubreymasten, line 48: Importing and running `xhr/instrument` was removed, which will cause problems if the Ajax feature is not required to be imported or named as an explicit dependency of SPA:

https://github.com/newrelic/newrelic-browser-agent/blob/e0addb7c19e52429cc17b50a4e6f01892b3c1fc1/feature/spa/instrument/index.js#L38

We should re-add this or create a dependency so we're at least warned if Ajax is ever disabled

packages/browser-agent-core/features/spa/aggregate/index.js
```diff
@@ -0,0 +1,708 @@
...
+        interaction.root.attrs.firstPaint = paintMetrics['first-paint']
+        interaction.root.attrs.firstContentfulPaint = paintMetrics['first-contentful-paint']
+      }
+      baseEE.emit('interactionSaved', [interaction])
+      state.interactionsToHarvest.push(interaction)
+      scheduler.scheduleHarvest(0)
+
+    }
+
+    function isEnabled() {
```
aubreymasten, line 693: Is this now handled by the CDN setup code?

packages/browser-agent-core/features/spa/aggregate/index.js
```diff
@@ -0,0 +1,708 @@
...
+        }, 0)
+      }, undefined, baseEE)
+
+      register(INTERACTION_API + 'onEnd', function (t, cb) {
+        this.ixn.handlers.push(cb)
+      }, undefined, baseEE)
+
+      register('api-routeName', function (t, currentRouteName) {
+        state.lastSeenRouteName = currentRouteName
+        if (state.currentNode) state.currentNode[INTERACTION].setNewRoute(currentRouteName)
```
aubreymasten, line 584: This looks like it was added post 1216, was this to address a bug found in testing?

cdn/agent-aggregator/util/featureDependencies.js
```diff
@@ -0,0 +1,10 @@
+export function getFeatureDependencyNames(feature) {
+    switch(feature){
+        case 'ajax':
+            return ['jserrors']
+        case 'session-trace':
+            return ['ajax']
```
aubreymasten, line 6: We should probably add a SPA depends on Ajax rule here

metal-messiah:
> Any plans drawn to optimize using `splitChunks` ? I imagine the chunks be: a. base loader + {metrics, page-view-event, page-view-timings} instrument b. **each** feature instrument individually c. base aggregator d. {m, pve, pvt} together + **each** feature aggregate individually

Plan would probably be an outcome of spiking on async loader impacts

packages/browser-agent-core/common/wrap/wrap-events.js
cwli24: Yeah you're right; good spotting. Same as the add-event-listener.

packages/browser-agent-core/common/window/load.js
metal-messiah: To be clear, the functions you linked, `onWindowLoad()` and `onDOMContentLoaded()`, are just event listeners that early resolve if the event has already happened.  The behavior you are referring to (setting timings, now(), mark, etc), should be handled by the callers (in the page-view-event module, for instance) and not in these functions, as these are generic.  What these functions could do however, is return a flag that lets the caller know that it resolved because it had already triggered, letting the callers know not to implement timings or whatnot.

metal-messiah: (its also used by the CDN to inject the aggregator files)

packages/browser-agent-core/common/window/load.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/255#discussion_r934991359

packages/browser-agent-core/common/wrap/index.js
metal-messiah: remnant of Martin's work, but serves as a single export file (imports a lot of different files, and exports them all, for ease of use)

packages/browser-agent-core/common/wrap/wrap-events.js
metal-messiah: only thing it does is shortcut accessing the namespaced ee without actually wrapping anything (and without needing to know which namespace belongs to which wrap module), (ie. ee.get('events'), ee.get('timings), etc).  Definitely not needed.

packages/browser-agent-core/common/wrap/wrap-promise.js
metal-messiah: I believe this was added by @cwli24 for the circular interaction ticket?

packages/browser-agent-core/common/wrap/wrap-xhr.js
metal-messiah: good catch

packages/browser-agent-core/common/url/protocol.js
metal-messiah: What is the risk?

metal-messiah: (this is implemented in the metrics feature)

packages/browser-agent-core/common/util/obfuscate.js
metal-messiah: This is implemented in the metrics feature

aubreymasten:
Integrated into: PR https://github.com/newrelic/newrelic-browser-agent/pull/259

Issue #73, opened 2021-06-04
Running agent tests does not work on Windows


rr0214:
Running `npm run test` on Windows errors with message:

'tools' is not recognized as an internal or external command.

Even when using `node` at the beginning of the command does not seem to run tests.

- small effort

cwli24:
> Even when using node at the beginning of the command does not seem to run tests.
@rr0214 added node to the beginning of the test command inside the base `package.json` file, like so:
<img width="517" alt="image" src="https://user-images.githubusercontent.com/28714891/180826623-6de52f27-482b-49f2-8bed-aeb9db7539df.png">
should make it functional. Please confirm this.

PR #257, opened 2022-07-28
Fix: potential interaction-node cycle

### LINKED ISSUE #132: InteractionNode object has a circular reference that can be problematic in some cases
## Description

The Interaction class contains an enumerable property with an instance of the InteractionNode class
https://github.com/newrelic/newrelic-browser-agent/blob/main/feature/spa/aggregate/Interaction.js#L30

And the InteractionNode class contains an interaction property that points back to the Interaction instance
https://github.com/newrelic/newrelic-browser-agent/blob/main/feature/spa/aggregate/InteractionNode.js#L13

This can have an unexpected impact when the application does a deep traversal of all properties of an object, and one of those properties contains an InteractionNode or Interaction instance. For example, the SPA instrumentation adds a property on all XHR instances.

## Steps to Reproduce

Create an XHR instance and write a code that recursively walks through all properties of the object.

## Additional context

A possible fix could be to use property getter that is not enumerable.

cwli24:
### Overview
Preventative fix to deep traversal of interaction & interaction-node codependency.

### Related Github Issue
Fixes #132 & #73.

### Testing
Locally ran functional `spa/*` tests on chrome & iOS, unit of all tests on all browsers-versions.

### Integration
Rebase & merge

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "author": "The New Relic Browser Team",
   "description": "Tests for the New Relic JavaScript agent",
   "scripts": {
-    "test": "tools/jil/bin/cli.js",
+    "test": "node tools/jil/bin/cli.js",
     "legacy:build": "rm -rf build && mkdir build && gulp build",
     "build:all": "npx lerna run build",
     "build-mac": "gulp && osascript -e 'display notification \"The agent has been built.\" with title \"Build Complete\"'",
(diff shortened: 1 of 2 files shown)


cwli24:
Trivial changes w/ no actual known case causing problems. Review not needed.

PR #258, opened 2022-07-31
[WIP] Build and ship

metal-messiah:
### Overview 

This PR will be for updating the GH actions to utilize the Webpack build strategies.


packages/browser-agent-core/common/harvest/harvest.js
```diff
@@ -155,7 +155,7 @@ export class Harvest extends SharedContext {
     return ([
       '?a=' + info.applicationID,
       encodeParam('sa', (info.sa ? '' + info.sa : '')),
-      encodeParam('v', VERSION),
+      encodeParam('v', `${VERSION}.${SUBVERSION}`),
```
cwli24, line 158: VERSION is being used in metrics instrument too; does that need to be changed? Can we rename this scheme to `{MAJOR}.{MINOR}.{PATCH}` per semantic versioning standard (naming)? Can you combine them to export `= {VERSION}` so we don't have to reconstruct them per caller?

metal-messiah: good catch.  Yes to both

PR #256, opened 2022-07-28
MMF: Npm migration cleanup

cwli24:
### Overview
Post #251 merge clean-up for test migration. We remove extraneous/unnecessary:

1. `console.log`
2. comments
3. debug on the NREUM object

### Testing
N/A -- not needed

### Integration
Regular 3-way merge

PR #259, opened 2022-08-02
Migration Cleanup

aubreymasten:
This PR includes final cleanup changes from comparing between v1216 and the current head of dev.

Merges feedback and changes from PR https://github.com/newrelic/newrelic-browser-agent/pull/255 & PR https://github.com/newrelic/newrelic-browser-agent/pull/256

### DIFF
--- a/feature/err/aggregate/index.js
+++ b/feature/err/aggregate/index.js
@@ -137,6 +137,7 @@ function canonicalizeStackURLs (stackInfo) {
 }
 
 function storeError (err, time, internal, customAttributes) {
+  // console.log('storeError', err)
   // are we in an interaction
   time = time || loader.now()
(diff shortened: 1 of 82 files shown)


feature/err/aggregate/index.js
```diff
@@ -137,6 +137,7 @@ function canonicalizeStackURLs (stackInfo) {
 }
 
 function storeError (err, time, internal, customAttributes) {
+  // console.log('storeError', err)
```
cwli24, line 140: This and the other console.log needs to be removed?

cwli24: Oops, nvm, this is old stuff.

aubreymasten: Good catch, removed from this file

PR #276, opened 2022-09-26
Web workers - Supportability metrics

cwli24:
### Overview
NEWRELIC-3702
Added reporting of supportability metrics for:
- Existence of required polyfill features [experimental]
- Usage of web workers API and per initialization of the combination of the following:
    - flavor: {dedicated, shared, service}
    - type: {classic, module}

### Testing

- Appended to functional metrics test
- Manually testable

### DIFF
--- a/cdn/agent-loader/spa.js
+++ b/cdn/agent-loader/spa.js
@@ -2,6 +2,8 @@
  * Copyright 2020 New Relic Corporation. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
+import { detectPolyfillFeatures } from './utils/feature-detection'
+const PolyfillFeatures = detectPolyfillFeatures();
 /* cdn specific utility files */
 import { stageAggregator } from './utils/importAggregator'
 import agentIdentifier from '../shared/agentIdentifier'
@@ -24,7 +26,7 @@ configure().then(() => {
     // lite features
     if (enabledFeatures['page_view_event']) new InstrumentPageViewEvent(agentIdentifier) // document load (page view event + metrics)
     if (enabledFeatures['page_view_timing']) new InstrumentPageViewTiming(agentIdentifier) // page view timings instrumentation (/loader/timings.js)
-    if (enabledFeatures.metrics) new InstrumentMetrics(agentIdentifier) // supportability & custom metrics
+    if (enabledFeatures.metrics) new InstrumentMetrics(agentIdentifier, PolyfillFeatures)   // supportability & custom metrics
     // pro features
     if (enabledFeatures.jserrors) new InstrumentErrors(agentIdentifier) // errors
     if (enabledFeatures.ajax) new InstrumentXhr(agentIdentifier) // ajax
--- a/packages/browser-agent-core/features/metrics/instrument/workers-helper.js
+++ b/packages/browser-agent-core/features/metrics/instrument/workers-helper.js
@@ -1,10 +1,14 @@
-export default {workersApiIsSupported, insertSupportMetrics};
+export { insertSupportMetrics };    // export list
 
 /**
- * True if Web Workers are supported in browser's execution context. Not all browser versions may support certain Workers or options however.
+ * True for each Worker type supported in browser's execution context. Not all browser versions may support certain Workers or options however.
  * - Warning: service workers are not available on unsecured HTTP sites
  */
-const workersApiIsSupported = Boolean(self.Worker && self.SharedWorker && self.ServiceWorker);
+const workersApiIsSupported = {
+    dedicated: Boolean(self.Worker),
+    shared: Boolean(self.SharedWorker),
+    service: Boolean(self.ServiceWorker)
+};
 
 let origWorker, origSharedWorker, origServiceWorkerCreate;
 /**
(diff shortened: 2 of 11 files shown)


cdn/agent-loader/spa.js
```diff
@@ -29,7 +31,7 @@ configure().then(() => {
     // lite features
     if (enabledFeatures['page_view_event']) new InstrumentPageViewEvent(agentIdentifier) // document load (page view event + metrics)
     if (enabledFeatures['page_view_timing']) new InstrumentPageViewTiming(agentIdentifier) // page view timings instrumentation (/loader/timings.js)
-    if (enabledFeatures.metrics) new InstrumentMetrics(agentIdentifier) // supportability & custom metrics
+    if (enabledFeatures.metrics) new InstrumentMetrics(agentIdentifier, PolyfillFeatures)   // supportability & custom metrics
```
metal-messiah, line 34: Just brain storming:: Would it make more sense to just emit a supportability metric event directly from the `detectPolyfillFeatures()` module instead of passing it around?

The metrics feature has this listener
`registerHandler('record-supportability', (...args) => this.recordSupportability(...args), undefined, this.ee)`

Inside the `detectPolyfillFeatures` method, it could do something like 
`handle('record-supportability', ['Ajax/Events/Excluded/Agent'], undefined, undefined, ee)`

Just an idea!

metal-messiah: On second thought, this won't be possible, since it would require the EE to run, and in your implementation you are purposeful about running as the first thing on the page.  Disregard the above!

metal-messiah: Heres the part that is difficult for me to comprehend, If we remove polyfills, and want to detect that polyfills don't exist, we likely wont be able to send these metrics, because the agent will break.

packages/browser-agent-core/features/metrics/instrument/workers-helper.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Injects code to report Web Workers supportability and usage as metrics, replacing the native API in global scope as a side effect.
+ * @param {func} report - handles reporting of supportability metrics to NR1
+ * @returns void
+ */
+function insertSupportMetrics(report) {
+    if (origWorker) return; // ensure we only insert the metrics once
+    if (!workersApiIsSupported) {
```
metal-messiah, line 29: I think this should be fine, there is a chance we would miss counting for envs that **do** have web workers but **don't** have service workers, but i think that is low risk.

cwli24: Yes, didn't see much merit in separating them instead of treating support as a whole, despite different flavors.

cwli24: Originally intended to just treat them together instead of separately, but now that I see Safari doesn't have SharedWorker support until the latest v16.. maybe they ought to be separated?

packages/browser-agent-core/features/metrics/instrument/workers-helper.js
```diff
@@ -0,0 +1,82 @@
+export default {workersApiIsSupported, insertSupportMetrics};
```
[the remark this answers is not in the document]
cwli24: Do you mean importing individual (list) rather than a default object?

cdn/agent-loader/spa.js
cwli24: 1. Your second thought + I try to keep things modularly contained; not a fan of the current "detached floating ghost" aka EE  that gets called here and there and makes tracking the data flow difficult. Like, helper files should just be helper files/fn's with no invisible/implicit side effects, and if EE is to be everywhere, we could've just made it part of the global NREUM in a tree structure (something similar to my redone EE poc actually).

2. You're right; I had done this with in mind that the current way of polyfilling would stay (so we determine -> polyfill -> report). But yeah if we're removing polyfills by default, we need to make sure that **at least** metrics instrument+aggregate runs, if nothing else. In general, that makes sense for any sm. 
- So is there a reason `configure()` has to be a Promise (what do you gain by making that step async, as I was under the assumption that most if not all of that work is done on 1st pass anyways)?
- We just have to make sure that's the first thing initialized and everything from entry to metrics is as ES5 like as possible. To that end, can we make the lite/pro/spa blocks IIAFEs instead?
- Ideally, each agent feature aggregates after they finished instrumenting, and failure in one feature only shuts that feature (and dependents) off rather than the entire agent. Do you agree, and how far are we from that vision?

packages/browser-agent-core/features/metrics/instrument/workers-helper.js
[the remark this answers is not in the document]
metal-messiah: when webpack sees `export default ...`  it will always have to include all the contents of the exports in each chunk that references that default export, because it is all imported even if not all the methods are used.  

Obviously not a huge deal here, but just wanted to say that was the pattern we were following.

It would be a syntax change from
`export default {workersApiIsSupported, insertSupportMetrics}`
to 
`export const workersApiIsSupported = ...`
`export function insertSupportMetrics() {...}`

packages/browser-agent-core/features/metrics/instrument/workers-helper.js
metal-messiah: Didnt realize the coverage was that late, probably should be separated.

cwli24: Done

cdn/agent-loader/spa.js
cwli24: Per discussion, leaving this as is. Acknowledged that, of the affected groups:
1. Clients with recent browsers not requiring polyfill - captured
2. Clients with old browsers requiring polyfills, and gets them - captured
3. Clients with old browsers requiring polyfills, but missing - NOT captured

This will skew the "Polyfill/X/NotSupported" counts lower.

metal-messiah (approved):

coreyarnold:
NEWRELIC-4370
NEWRELIC-4372

PR #277, opened 2022-09-26
Gh actions tests

metal-messiah:
### Overview
- Certain Sauce environments no longer support the syntax JIL uses to send and receive messages between JIL & Sauce (W3C).  This means newer android envs are now broken. (The old browser agent hadn't updated its test dependencies outside of Chrome@latest, firefox@latest, safari@latest, and edge@latest, so this was never detected as Android was running on a very old instance). 
- This will need to get fixed at some point.  The only way forward right now is to not test on Android VMs.
Will likely require a change to JIL in some way <or> a different way to test Chrome@mobile versions.
There is a very rigid structure in place right now for providing the browser specs to JIL (you can't just take the raw specs that sauce generates and plug them in to JIL... JIL breaks) -- This could (& should) be fixed in the future to simply accept the literal spec from Sauce
- The PR will have GH tests running with the following flow:
A call is made to Saucelabs to generate a list of "currently supported browsers" (Latest-10).  [This is saved here as a file](https://github.com/newrelic/newrelic-browser-agent/blob/gh-actions-tests/tools/jil/util/browsers-supported.json).  This could be referenced in the readme as evidence of what we tested a PR against, and also can be used for test debugging (plug and play from releases, branches, etc).
Separate GH jobs for each browser platform (Found an increase of testing speed by ~50% by doing this)
Overall, running tests on PRs used to take about 40 minutes.  These tests are appearing to run end to end in about 18-20 mins on average.
This could be dramatically sped up even more by caching NPM install dependencies.
Fixed the "Hanging Test" issue we see (where occasionally a test suite finishes successfully, but never closes, requiring a restart)
Added GitHub timeouts to the jobs, so that it doesn't wait 6 hours to kill the thread (now set at 25 minutes).
The Hanging Test issue itself is not fixed, however a final "cleanup" process happens after the tests run.  This process uses the insight API and queries our JIL NRQL dataset and checks for failures there.  
If failures are found, the GH test fails, and it spits out some info -- The failings tests + envs, and a JIL code snippet to run locally to reproduce the issue
If no failures are found in JIL, the GH test passes
	
- The PR also adds a polyfills test, which behaves the same way as above, except with older predefined static browsers.  This test would only run after a PR has been merged.  -- this will need further polishing, but will be used to inform us that the polyfill needs to be updated over time.


metal-messiah:
example --> https://github.com/newrelic/newrelic-browser-agent/actions/runs/3138453081

tools/jil/util/sauce-browsers.js
```diff
@@ -0,0 +1,161 @@
...
+const browserLength = name => {
+    // takes half of this in the end
+    switch (name) {
+        case "ios":
+        case "iphone":
+        case "ipad":
+        case "android":
+            return 5
+        default:
+            return 10
```
cwli24, line 54: Safari is 5 too, no?

cwli24 (commented):
👍

cwli24 (approved):

PR #284, opened 2022-10-13
Web workers - change global object scope and references NEWRELIC-4382

cwli24:
### Overview
Many parts of the code references `window` and `document` which do not exists in workers. This change expands the allowed and considered global scopes so that the agent runs in:
a. browser window (web) context w/o any behavior difference
b. WorkerGlobalScope on a best-effort basis, pending future tuning

### Related Issue
Part of NRELIC-2433 Epic

### Testing
Regression test via running the automated suite.

#### Other side effects
- Tests workflows now distinguishes attempts on a wf run when reporting to NRDB. This addresses an issue with the check-failures job.

### DIFF
--- a/.github/workflows/tests-polyfill.yml
+++ b/.github/workflows/tests-polyfill.yml
@@ -5,6 +5,9 @@ on:
     types: [labeled]
   workflow_dispatch:
 
+env:
+  BUILD_NUMBER: PR${{ github.event.number }}-job-${{ github.run_number }}-attempt-${{ github.run_attempt }}
+
 jobs:
   chrome-functional:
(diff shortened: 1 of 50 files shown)


.github/workflows/tests-polyfill.yml
```diff
@@ -5,6 +5,9 @@ on:
     types: [labeled]
   workflow_dispatch:
 
+env:
+  BUILD_NUMBER: PR${{ github.event.number }}-job-${{ github.run_number }}-attempt-${{ github.run_attempt }}
```
metal-messiah, line 9: 👍

metal-messiah (approved):

PR #286, opened 2022-10-21
Webworkers jserrors tests

metal-messiah:
### Overview

This PR adds tests for jserrors feature in webworkers.

### Testing 
Tests have been added that run against a Worker() (classic and module) for each applicable jserrors test that runs on the main page.


tests/functional/worker/notice-error.test.js
```diff
@@ -0,0 +1,118 @@
...
+      const { err } = JSON.parse(errResponse.body)
+      checkBasics(t, err)
+      t.deepEqual(err[0].custom, { hi: 'mom', worker: true }, 'Should have correct custom attributes')
+      t.end()
+    }).catch(fail(t));
+	})
+}
+
+
+function multipleMatchingErrorsTest(type, supportRegOrESMWorker) {
```
cwli24, line 88: Do need a test on not _exactly_ matching errors **don't** aggregate? Though that may tie into the existing bug of aggregation disregarding column number. Could you just note a to-do to come back and add that once that bucketing is fixed?

tools/jil/util/browser-matcher.js
```diff
@@ -413,12 +434,14 @@ features.es6 = new BrowserMatcher()
...
+  features.sharedWorkersFull = features.sharedWorkers.and(features.supportESMWorkers);
+
   features.serviceWorkers = new BrowserMatcher([
     new MatcherRule(TYPE_EXCLUDE, '*@*'),
     new MatcherRule(TYPE_INCLUDE, 'chrome@>60'),  // v60 has unexpected behavior of working on unsecured HTTP tests
     new MatcherRule(TYPE_INCLUDE, 'edge@>=17'),
     new MatcherRule(TYPE_INCLUDE, 'safari@>=11.1'),
     new MatcherRule(TYPE_INCLUDE, 'firefox@>=44'),
-    new MatcherRule(TYPE_INCLUDE, 'android@>=105'),
+    new MatcherRule(TYPE_INCLUDE, 'android@>=106'),
```
cwli24, line 445: I realized late that there's no android version 100+, and instead there's an obscured mapping between chromium version to android version. 
I see this from Sauce: https://changelog.saucelabs.com/en/update-to-google-chrome-version-100-on-android-emulators
So basically the default we've used up to now (JWP) goes up to version 9.0 and the new "GoogleAPI emulator" (W3C) goes up to 13.0 now.
Can you change this to `@>=6.0` while you're at it? And the one for ESMWorkers above (line 424) can be changed to probably `@>=9.0` which should be a safe equivalent of `chrome@>=80`.

tests/functional/worker/auto-instrumented-errors.test.js
```diff
@@ -0,0 +1,301 @@
...
+      t.ok(err[0].metrics.time.t > 0, 'Should have a valid timestamp')
+      t.equal(err[0].params.exceptionClass, 'RangeError', 'Should be RangeError class')
+      t.ok(!!err[0].params.message, 'Should have message')
+      t.ok(err[0].params.stack_trace, 'Should have a stack trace')
+      t.deepEqual(err[0].custom, { worker: true }, 'Should have correct custom attributes')
+      t.end()
+    }).catch(fail)
+
+    function fail(err) {
+      if (browser.match('firefox')) t.pass('This browser version does not throw errors in worker when max stack size is reached')
```
cwli24, line 232: You created a matcher for this too

cwli24 (changes requested):

cdn/agent-aggregator/aggregator.js
```diff
@@ -21,20 +22,19 @@ export function aggregator(build) {
...
   if (autorun) return initializeFeatures();
 
   async function initializeFeatures() {
-    const drain = getConfigurationValue(agentIdentifier, 'page_view_event.enabled') === false
     const features = await importFeatures(build)
     // gosNREUMInitializedAgents(agentIdentifier, features, 'features')
     // once ALL the enabled features all initialized,
     // add the activated features from the setToken call to the window for testing purposes
     // and activatedFeatures will drain all the buffers
     addToNREUM('activatedFeatures', activatedFeatures)
```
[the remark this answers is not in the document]
metal-messiah: not remove 36, but it could have a check for `isBrowser` before it as well

PR #288, opened 2022-10-26
Functionality fixes

metal-messiah:
### Overview
this PR addresses some issues uncovered in the agent core that were exposed during Web Worker testing.

### Fixes
* `configure()` call is now treated as a synchronous function, so that it does not push the execution to the next frame
* supportability metrics were not being set correctly when `ajax.deny_list` was being applied.


tools/jil/server/asset-server.js
```diff
@@ -242,7 +246,7 @@ class AgentInjectorTransform extends AssetTransform {
           let rspData = rawContent
             .split('{loader}').join(tagify(disableSsl + loaderContent))
```
cwli24, line 247: Noticed: it likely has no practical effect since nothing's supposed to go wrong in the config part, but customers don't get script tags on each line right? Noticed that as I was looking at final `rspData`

tools/jil/server/asset-server.js
```diff
@@ -112,6 +112,10 @@ class AgentInjectorTransform extends AssetTransform {
     return updatedConfig
   }
 
+  getAjaxDenyListString(){
+    return runnerArgs.denyListBam ? `window.NREUM||(NREUM={init:{}});NREUM.init.ajax=NREUM.init.ajax||{};NREUM.init.ajax.deny_list=NREUM.init.ajax.deny_list||[];NREUM.init.ajax.deny_list.push('bam-test-1.nr-local.net')` : ''
```
cwli24, line 116: Might want to put a `;` before and after the entire string for sanity. Idk if initContent necessarily has that

cdn/agent-loader/utils/configure.js
```diff
@@ -8,26 +8,20 @@ import { setConfiguration, setInfo, setLoaderConfig, setRuntime } from '@newreli
...
-            resolve(configured)
-            return
-        }
-        const nr = gosCDN()
-        
-        try {
+    if (configured) return
+    const nr = gosCDN()
+
+    try {
```
cwli24, line 14: Can we either move or create another try-catch around each of the lite/pro/spa entries? There's actually a low risk that something here goes wrong vs instrumenting & stageAggregator. Brought that up and this is prob a good time to put in such a trivial change that doesn't really require much testing. For the catch part, maybe attempt a sendBeacon to report that Agent ran into an error, then just silently quit. I think may be good idea to abort ee too.

cwli24 (changes requested):

cwli24 (approved):
Looks good to me

PR #290, opened 2022-10-27
Webworkers final harvest NR-58010

cwli24:
### Overview
Agent when running within workers will wrap the close method to simulate a web worker's equivalent of a window's unload event. Buffered data will be sent in a final harvest before thread closes. 

### Related Issue
Part of NEWRELIC-2433 epic

### Testing
New test ran on classic & module dedicated workers.

#### Other side effects
None


metal-messiah (dismissed):
👍

cwli24:
Will run test on dev after merge; expecting pass

PR #291, opened 2022-10-27
also run tests against shared-workers

metal-messiah:
### Overview

This PR adds `SharedWorker` to the array of workers that get tested against.

### DIFF
--- a/tests/assets/worker/js/shared-worker.js
+++ b/tests/assets/worker/js/shared-worker.js
@@ -1,0 +1,17 @@
+onconnect = function (event) {
+  const port = event.ports[0]
+
+  port.onmessage = function (e) {
+    if (e.data.type === 'startAgent') {
+      self.NREUM = e.data.payload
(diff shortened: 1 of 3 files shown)


cwli24 (approved):
Good stuff - should be a cinch of a pass on tests

cwli24:
If you look at the latest test from this PR, check-failure has an edge case where it succeeds despite an unhandledpromiserejection crashing the "check failed tests" stage. Could you just patch that too?

metal-messiah:
We have converted this PR to add infrastructure but not actually run the tests.  Tests can be run locally (only) using the infrastructure added.  Merging due to no run-time changes.

cwli24 (commented):
We decided to not include shared worker in automated testing due to the extra complication it adds to the test framework by nature while providing relatively low value in validation. (Auto testing dedicated classic & module workers + manual testing shared worker when applicable should suffice.)

PR #297, opened 2022-11-06
[WIP] One agent POC

metal-messiah:
### Overview 

Rough Draft at "one" agent APM, Copy/Paste + NPM.


packages/browser-agent-microfrontend/index.js
```diff
@@ -0,0 +1,75 @@
...
+                imports.forEach(({ Instrument, Aggregate }, i) => {
+                    const key = Object.keys(this.features)[i]
+                    this.features[key] = new (Instrument || Aggregate)(this.agentIdentifier, this.sharedAggregator)
+                    if (Instrument) chainedCompleted.push(this.features[key].completed)
+                })
+                Promise.all(chainedCompleted).then(() => {
+                    drain(this.agentIdentifier, 'api')
+                    gosNREUMInitializedAgents(this.agentIdentifier, this.features, 'features')
+                })
+            })
```
patrickhousley, line 67: Should either add a catch here or use async/await. Without this, an error will still generate an uncaught promise exception.

.github/workflows/publish-cdn.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: build:all
         run: npm run build:all
       - name: run tests
-        run: node ./tools/jil/bin/cli.js -f merged -b chrome@latest -s -t 85000 --concurrent=4
+        run: node ./tools/jil/bin/cli.js -f merged -b chrome@latest -s -t 85000 --concurrent=20
```
patrickhousley, line 25: Did you mean to leave this in here?

cdn/agent-loader/prebuilt.js
```diff
@@ -0,0 +1,7 @@
+import { BrowserAgent } from "@newrelic/browser-agent";
+
+const configs = {
+    init: { ajax: { enabled: false }, jserrors: {auto: false, enabled: true} },
+    info: { applicationID: '601303348', licenseKey: 'a60cc46d05', errorBeacon: 'bam.nr-data.net' }
+}
+window.agent = new BrowserAgent(configs)
```
patrickhousley, line 7: What does this do?

packages/browser-agent-microfrontend/index.js
metal-messiah: Actually, I changed the structure in the other loader to be agnostic to state of the promises, and i forgot to update this loader to match!

package.json
```diff
@@ -73,6 +75,7 @@
     "aws-sdk": "^2.894.0",
     "babel-loader": "^8.2.5",
     "babel-plugin-module-resolver": "^4.1.0",
+    "babel-plugin-transform-remove-strict-mode": "^0.0.2",
```
patrickhousley, line 78: This was probably added due to the test issues?? If so, we can remove this since we figured out what was causing those errors.

cdn/agent-loader/prebuilt.js
metal-messiah: This was present only for demonstration purposes, needs to be removed.

package.json
metal-messiah: 100%, yes thank you

packages/browser-agent-core/src/common/util/feature-base.js
```diff
@@ -1,10 +1,66 @@
...
+      const { Aggregate } = await import(`../../features/${this.featureName}/aggregate`)
+      new Aggregate(this.agentIdentifier, this.aggregator)
+      this.resolve()
+    } catch (err) {
+      this.reject(err)
+    }
+  }
+}
+
+export class AggregateBase extends FeatureBase {
```
patrickhousley, line 62: This is a nit for me but I prefer to have 1 class per file. If we don't want to expose the FeatureBase class, we don't have to include it in the package exports.

packages/browser-agent-core/src/common/util/obfuscate.js
```diff
@@ -55,17 +52,17 @@ export function validateRules(rules) {
...
-      if (console && console.warn) console.warn('An obfuscation replacement rule contains a "regex" value with an invalid type (must be a string or RegExp)')
+      if (console && console.warn) console.warn('New Relic: An obfuscation replacement rule contains a "regex" value with an invalid type (must be a string or RegExp)')
       invalidRegexDetected = true
     }
 
     var replacement = rules[i].replacement
     if (replacement) {
       if (typeof replacement !== 'string') {
-        if (console && console.warn) console.warn('An obfuscation replacement rule contains a "replacement" value with an invalid type (must be a string)')
+        if (console && console.warn) console.warn('New Relic: An obfuscation replacement rule contains a "replacement" value with an invalid type (must be a string)')
```
patrickhousley, line 65: Just a question on this. What happens if the obfuscation rule is invalid? Do we still capture the data and send it to NR1? We might want to consider throwing an error here and crashing the agent to prevent PII/PHI from possibly making it to NR1.

packages/browser-agent-custom/index.js
```diff
@@ -0,0 +1,57 @@
...
+import { getFeatureDependencyNames } from '@newrelic/browser-agent-core/src/loader/featureDependencies'
+import { featurePriority } from '@newrelic/browser-agent-core/src/loader/features'
+// common files
+import { Aggregator } from '@newrelic/browser-agent-core/src/common/aggregate/aggregator'
+import { gosNREUMInitializedAgents } from '@newrelic/browser-agent-core/src/common/window/nreum'
+import { generateRandomHexString } from '@newrelic/browser-agent-core/src/common/ids/unique-id'
+import { getConfiguration, getInfo, getLoaderConfig, getRuntime } from '@newrelic/browser-agent-core/src/common/config/config'
+
+
+export class BrowserAgent {
```
patrickhousley, line 13: Can we move this into a `src` directory to help keep a clean distinction between package configuration files and actual source code?

packages/browser-agent-microfrontend/index.js
```diff
@@ -0,0 +1,75 @@
...
+const nonAutoFeatures = [
+    'jserrors',
+    'page_action'
+]
+
+const autoFeatures = [
+    'metrics'
+]
+
+export class BrowserAgent {
```
patrickhousley, line 20: Same here on the `src` directory.

patrickhousley, line 20: Same here on the `src` directory.

packages/browser-agent/index.js
```diff
@@ -0,0 +1,81 @@
...
+const instruments = {
+    ajax: InstrumentXhr,
+    jserrors: InstrumentErrors,
+    page_view_event: InstrumentPageViewEvent,
+    page_view_timing: InstrumentPageViewTiming,
+    session_trace: InstrumentSessionTrace,
+    spa: InstrumentSpa
+}
+
+export class BrowserAgent {
```
patrickhousley, line 29: Same on the `src` directory here and this could replace the `src/index.ts`

.github/workflows/publish-cdn.yml
metal-messiah: Actually, i think we should leave this.  It could be helpful when its time to publish to CDN again to not have to wait 15 minutes for tests to run.

packages/browser-agent-core/src/common/util/obfuscate.js
metal-messiah: I think we took the approach of "its on them to configure it correctly" and we will give them messaging to aide in that but we didn't want to kill all data collection.  worth revisiting

PR #302, opened 2022-11-10
polish pr build and ship process

metal-messiah:
### Overview
This PR adds the ability to build and ship PR builds and dev builds directly to NR internal components

### DIFF
--- a/.github/workflows/dev-pr.yml
+++ b/.github/workflows/dev-pr.yml
@@ -3,13 +3,16 @@ name: Build and Deploy PR to NR Dev
 on:
   workflow_dispatch:
     inputs:
+      pr_number:
+        description: 'PR number'
+        required: true
       pr_build:
         description: 'PR build'
         required: true
         type: choice
         options:
-          - clear
           - build
+          - clear
   pull_request_target:
     types: [labeled, closed]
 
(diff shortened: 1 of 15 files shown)


metal-messiah:
Dempla does not currently parse the version in a way that is readable in the dashboard, but the process will not need to change and once dempla updates parsing the data will show up in the dashboards more clearly.

.github/workflows/dev-pr.yml
```diff
@@ -18,37 +21,48 @@ jobs:
...
       - name: clear S3
-        run: node tools/scripts/delete-s3-dir.js \
+        run: |
+          node ./tools/scripts/delete-s3-dir.js \
             --bucket ${{ secrets.AWS_BUCKET }} \
             --role ${{ secrets.AWS_ROLE_ARN }} \
-            --pr PR${{ github.event.pull_request.number }}
+            --pr PR-${{ github.event.pull_request.number || github.event.inputs.pr_number || 'workflow-dispatch' }}
       - name: build
         if: github.event.inputs.pr_build == 'build' || github.event.action != 'closed'
```
cwli24, line 41: Did you mean to use `&&` rather than `||` ?

metal-messiah: > Did you mean to use `&&` rather than `||` ?

Yes thanks nice catch

metal-messiah:
The internal builds will now use `setCustomAttribute()` to set a `buildID`. Dashboards should be able to show all data now regardless of a dempla fix. @cwli24 this is ready for a re-review

metal-messiah:
### you can target a page to use a specific loader by using one of the following:

#### query param:
`...?nrbaloader=PR-302`

!Some pages on NR1 override your query params on refresh, so an alternative is:!
#### localStorage:
`localStorage.setItem('nrbaloader', 'PR-302')`

tools/scripts/delete-s3-dir.js
```diff
@@ -0,0 +1,44 @@
...
+}
+if (!pr) {
+    console.log("PR field is empty!")
+    process.exit(1)
+}
+
+connectToS3(role, dry).then(() => {
+    emptyS3Directory(bucket, pr, dry)
+}).catch(err => {
+    console.log("err...", err)
```
patrickhousley, line 43: [question] Should we `process.exit(1)` here if we cannot delete whats in the s3 bucket?

cdn/package.json
```diff
@@ -6,12 +6,13 @@
   "main": "src/spa.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "webpack": "npx webpack --progress --config ./webpack.config.js",
+    "webpack": "rm -rf ../build && npx webpack --progress --config ./webpack.config.js",
```
patrickhousley, line 9: [suggestion] This breaks local watch builds. Can we remove the `rm -rf ../build && `? The build directory shouldn't exist in the scope of a GH workflow, right?

.github/workflows/publish-internal-prod.yml
```diff
@@ -0,0 +1,49 @@
...
+      current:
+        description: 'CURRENT "stable" version number (must be published to CDN root)'
+        required: true
+      next:
+        description: 'NEXT version number (must be published to CDN root)'
+        required: true
+
+jobs:
+  pr-dev-build:
+    if: contains(fromJson('["workflow_dispatch"]'), github.event_name)
```
patrickhousley, line 15: [question] If the wf `on` is set to only run on `workflow_dispatch`, is this really needed? I genuinely don't know on this one.

.github/workflows/publish-internal-prod.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    container: node:14
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    steps:
+      - uses: actions/checkout@v2
```
patrickhousley, line 22: [question] Should this change to `actions/checkout@v3`?

.github/workflows/dev-pr.yml
```diff
@@ -3,52 +3,51 @@ name: Build and Deploy PR to NR Dev
...
+      - name: purge fastly cache url
+        run: |
+          node ./tools/scripts/fastly.js \
+            --fastly-key ${{ secrets.FASTLY_PURGE_KEY }} \
+            --env dev \
+            --purge-internal
+      - name: Get current time
+        uses: josStorer/get-current-time@v2.0.2
+        id: current-time
+      - uses: mshick/add-pr-comment@v2
```
patrickhousley, line 49: [nit] We could use the github cli to do this instead of relying on a third-party workflow action.

https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows

tools/scripts/delete-s3-dir.js
metal-messiah: yes, thank you

.github/workflows/publish-internal-prod.yml
metal-messiah: nope this is a remnant that can be removed, TY

cdn/package.json
metal-messiah: Certain actions do trigger multiple builds for one reason or another (like building with and without sourcemaps).  Could probably change but thats why that was added

patrickhousley: All good....we will just have to accept that the watch does not work right now.

metal-messiah: for now will just move it out of the base `webpack` script

.github/actions/get-current-time/action.yml
```diff
@@ -0,0 +1,10 @@
+name: 'Get Current Time Stamp'
+description: 'Get the current time stamp'
+
+outputs:
+  timestamp:
+    description: "new Date().toString()"
+
+runs:
+  using: 'node14'
```
cwli24, line 9: Can you bump up **all** the node14 to node18 while you're at it? Across all the workflows.

metal-messiah:
testing from gh api

metal-messiah:
A new build of PR-302 has been shipped to NR Internal - https://staging.onenr.io/0M8jq5NPOjl

.github/actions/get-current-time/action.yml
metal-messiah: was seeing some issues after converting to run on node18, we should save this for a separate effort.

patrickhousley (approved):
Looks good.

cwli24 (approved):
+1

PR #303, opened 2022-11-14
fix(jserrors): add error type and message to aggregation check

patrickhousley:
### Overview

Updating the jserrors feature so it will include the error type and message when deciding if an error has already been seen and the error count increased.

### Related Github Issue

https://issues.newrelic.com/browse/NR-40043
https://issues.newrelic.com/browse/NR-33425

### Testing

A new functional test has been added to simulate and test a reproduction of the issue.


tests/functional/issues/nr-40043.test.js
```diff
@@ -0,0 +1,54 @@
...
+    metrics: {
+        enabled: false
+    },
+    jserrors: {
+        enabled: true,
+        harvestTimeSeconds: 5
+    }
+}
+
+testDriver.test('NR-40043: Multiple errors with noticeError', supported, function (t, browser, router) {
```
metal-messiah, line 23: I like the idea of tying this to the issue number

packages/browser-agent-core/features/jserrors/aggregate/index.js
```diff
@@ -153,7 +153,8 @@ export class Aggregate extends FeatureBase {
     var canonicalStack = this.buildCanonicalStackString(stackInfo)
 
     var params = {
-      stackHash: stringHashCode(canonicalStack),
+      stackHash: stringHashCode(`${stackInfo.name}_${stackInfo.message}_${canonicalStack}`),
+      // stackHash: stringHashCode(canonicalStack),
```
metal-messiah, line 157: Can we remove this hanging comment remnant?

metal-messiah (changes requested):
Everything looks good, nicely done.  Couple small requests:

- [ ] remove a remnant comment.  

Then I've been pondering the issue name in the filename
- [ ] Can we keep the tests lumped together (in the dir with the other errors tests) such that I can still easily tell that the new test is an errors test? If we decide to do an audit of errors tests for instance, and this test has been grouped separately there's a little disconnect there. 
- [ ] the test filename and the asset filename don't give us much info outside of the context of the ticket name. Maybe we keep the file names as a description of the test case, but make sure the issue ID in the body of the test still? Open to opinions
- [ ] finally can you add this change to the change log file 

Really appreciate the perspective of cataloging the bug in the asset.

metal-messiah (dismissed):
Nice work on this! looks great

metal-messiah (approved):

PR #304, opened 2022-11-15
Detect GH Workflow changes - NEWRELIC-5328

metal-messiah:
### Overview
This PR adds a GH action to detect if .yml files have changed

### Expected Behavior
* Action runs when PR is opened, synced, or re-opened
* Action detects changes to .yml files in PR
* Action warns via comment that a workflow has changed

### DIFF
--- a/.github/workflows/detect-yml-change.yml
+++ b/.github/workflows/detect-yml-change.yml
@@ -1,7 +1,6 @@
 name: Detect GH Actions Changes
 on:
   pull_request:
-    types: [synchronize, reopened]
 
 jobs:
   check:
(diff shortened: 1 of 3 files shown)


.github/workflows/detect-yml-change.yml
```diff
@@ -0,0 +1,28 @@
+name: Detect GH Actions Changes
+on:
+  pull_request:
+    types: [synchronize, reopened]
```
patrickhousley, line 4: Missing `opened` in the list for freshly opened PRs. Also, we could just leave this off since defaults is what I think we want.

```By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened```

metal-messiah: perfect thanks

patrickhousley:
Looks good.

patrickhousley (approved):

PR #305, opened 2022-11-17
Bug/63240 undefined hostname deny list NR-63240

bjfield:
### Overview
We've seen support tickets where the expected hostname doesn't exist on XHR request events. In our deny-list.js, we try to get the .length of the hostname, causing an exception and our agent to fail.

- Converted deny-list.js to TypeScript with an interface for XhrParams.
- Created a test for the bug condition using jest.
- Added a fix for errors on string comparison of undefined hostname.

### Related Jira Issue
[NR-63240](https://issues.newrelic.com/browse/NR-63240)

### Testing
Jest test lives alongside new deny-list.ts file in deny-list.spec.ts.


packages/browser-agent-core/src/common/deny-list/deny-list.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/** An array of filter objects {hostname, pathname} for identifying XHR events to be excluded from collection.
+ * @see {@link https://docs.newrelic.com/docs/browser/new-relic-browser/configuration/filter-ajax-request-events/ Filter AjaxRequest events}
+ * @type {Array.<{hostname: string, pathname: string}>}
+ */
+ let denyList: Array<{ hostname: string; pathname: string }> = []
+
+ /**
+  * Evaluates whether an XHR event should be included for collection based on the {@link denyList|AjaxRequest deny list}.
+  * @param {{method: string, hostname: string, port: string, protocol: string, host: string, pathname: string, status: number}} params - object with properties of the XHR event
```
patrickhousley, line 11: [nit] With tsdoc you should be able to reference the interface instead of typing out the param type

patrickhousley (changes requested):
[blocker] Please update the description with a link to the jira.

packages/browser-agent-core/src/common/deny-list/deny-list.ts
bjfield: Pushed change. Good catch!

dy-dx:
v1220 rolled out for us today and caused `data:` requests to throw errors. We had to disable the browser agent as a result.

Please expedite this if you can!

bjfield:
@dy-dx, thanks for taking the time to mention your issue and contributing to the conversation. If you wouldn't mind submitting a new issue (a bug report) here in the repo for your specific case, that would help expedite a fix. It's possible this PR would fix your issue, as data urls are one situation where hostnames are undefined, but either way we want to be sure we have good details about what's happening in your case.

bjfield:
This PR is closed and replaced with fresh [PR 306](https://github.com/newrelic/newrelic-browser-agent/pull/336).

PR #306, opened 2022-11-19
Fix js error bucketing logic

patrickhousley:
### Overview

Reverting the changes to the stack hashing algorithm in https://github.com/newrelic/newrelic-browser-agent/pull/303. The stack hash should match across browsers as much as possible and adding the error name and message would prevent that. The algorithm for creating the `stackHash` and `browserStackHash` values have be reset to what is in version 1220.

See https://docs.newrelic.com/attribute-dictionary/?event=JavaScriptError&attributeSearch=hash for more details on these data fields.

To fix the issue with bucketing, we are calculating a new hash value that is based on the error name, message, and original stack trace. This will ensure that identical errors from the same line but different columns and usage of noticeError in a central location, like a SPA error handler, will not be bucketed.

- Converted touched code to typescript.
- Fix test server arguments for port. It was defined as a boolean.
- Updates from sauce labs for supported browsers.
- Converted the unit tests in `tests/browser/err` to jest tests.
- Updated the tsconfig so the packages tsc build produces later ESM code that does not contain polyfills.
- Integrated latest version of husky and updated git hooks.
- Fixed issue with eslint not running correctly from previous PR.
- Updated the CDN webpack build to use the raw TS code instead of relying on TSC (nx build) being ran.

TODO: Write unit tests for the jserrors aggregate source file.

### Related Jira Issue

https://issues.newrelic.com/browse/NEWRELIC-3788
https://issues.newrelic.com/browse/NR-40043


patrickhousley:
Closing and breaking PR up

PR #308, opened 2022-11-23
BFCache: Pt.1 Redefining unload

cwli24:
### Overview
Enabled with a feature flag passed in the `init` configuration at start, the unload or end-of-life logic for the browser window context has been redefined. The new definition monitors the document visibility change as well as `pagehide` event, and removed the `unload` event that blocked back/forward cache eligibility.

### Related Issue
NEWRELIC-2441 epic
NEWRELIC-5080 feature
Public doc changes staged: https://github.com/newrelic/docs-website/compare/develop...cwli24:docs-website:browser-team-1222-rls

### Testing
- New additions to `functional/final-harvest.test.js`
- Modification & clean up of a few browser matchers
- Included back some tests for **iOS / Safari** versions, previously excluded due to driver problems
- None for new (temporary) on-page-restore custom metric emitted
- Manual validation on latest Chrome
- Sauce automated validation for modified test files for all iOS & Safari currently in support list

#### Other side effects
- An outdated file and its remaining usage removed
- A file was renamed
- Few variables were renamed to better convey lexicographic meaning
- `browser-supported.json` updated
- A couple trimming of redundant logic, particularly around eliminating `Harvest.sendFinal()` and reuse of `HarvestScheduler.runHarvest`


packages/browser-agent-core/src/common/event-listener/event-listener-opts.js
```diff
@@ -14,4 +14,13 @@ try {
...
   return supportsPassive ? {passive: true, capture: !!useCapture} : !!useCapture
+}
+
+/** Do not use this within the worker context. */
+export function windowAddEventListener(event, listener) {
+  window.addEventListener(event, listener, eventListenerOpts(false));
+}
+/** Do not use this within the worker context. */
+export function documentAddEventListener(event, listener) {
+  document.addEventListener(event, listener, eventListenerOpts(false));
```
patrickhousley, line 25: [question] Why are we creating functions that wrap `addEventListener`. These are pretty much just straight pass throughs and add bytes to the compiled code for seemingly no gain. Just wondering here.

packages/browser-agent-core/src/features/metrics/instrument/index.js
```diff
@@ -92,6 +94,15 @@ export class Instrument extends FeatureBase {
         this.recordSupportability(`Generic/Polyfill/ObjectAssign/${this.PfFeatStatusEnum.OBJECT_ASSIGN}`);
         this.recordSupportability(`Generic/Polyfill/ObjectEntries/${this.PfFeatStatusEnum.OBJECT_ENTRIES}`);
     }
+
+    eachSessionChecks() {
+        // [Temporary] Report restores from BFCache to NR1 while feature flag is in place in lieu of sending pageshow events.
+        windowAddEventListener('pageshow', (evt) => {
```
metal-messiah, line 100: Metrics feature is loaded in workers, this might break that

packages/browser-agent-core/src/common/event-listener/event-listener-opts.js
cwli24: Yeah, so there's many if not all places where the `addEventListener` is called with that third `eventListenerOpts(useCapture)` arg that's meant to set the `options.passive` flag if avail. -- I have no why that's needed, but I was attempting to refactor all of the `addEventListener(_, _, eventListenersOpts(false))` -> `addEventListener(_, _)` instead.
(I haven't replaced all of them. If this has become irrelevant and just a relic, we can make a cleanup task for all of these instead.)

packages/browser-agent-core/src/features/ajax/aggregate/index.js
```diff
@@ -58,14 +58,12 @@ export class Aggregate extends FeatureBase {
 
       externalFeatures.forEach(feat => {
         externalHarvestKeys.forEach(key => {
-          feat?.scheduler?.harvest?.on(key, () => {
+          feat?.scheduler?.harvest?.on(key, () => {   // this is some bad practice
```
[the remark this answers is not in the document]
cwli24: This is where the ajax->jserror (partial) dependency is happening. Essentially it's relying on the jserror feature to be initialized, then telling it to collect ajax ('xhr') stuff as part of its harvest payloads. There's no real reason ajax requires jserror aside from this afaik; if we want features to standalone, ajax should be self-responsible for reporting to both the /events endpt _and_ the /jserrors endpt. Better yet, have a central harvester. Alas, another issue for another time.

packages/browser-agent-core/src/features/metrics/instrument/index.js
cwli24: Gooood catch.

packages/browser-agent-core/src/features/ajax/aggregate/index.js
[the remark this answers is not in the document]
patrickhousley: Thanks. All good info to have.

tests/browser/harvest.browser.js
```diff
@@ -568,7 +570,7 @@ test('when sendBeacon returns false', function(t) {
     harvesterInst.on('jserrors', once(dummyPayload('jserrors')))
 
     setupFakeNr();
-    harvesterInst.sendFinal();
+    harvesterInst.sendX('jserrors', {unload: true});
```
patrickhousley, line 573: QQ: Should we just default all the tests to include `unload: true`?

patrickhousley: Nope that good.

metal-messiah (changes requested):

CHANGELOG.md
```diff
@@ -3,9 +3,11 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-
 ## v1222
 
+### Unblock instrumented pages from the back/forward cache (w/ feature flag)
+B/f cache eligibility was hampered by the agent's `unload` listener, which will be removed when a flag is on. With the `allow_bfcache` enabled in the `info` config, how the agent interprets the end of an user session will change and also allow an otherwise eligible page to persist.
```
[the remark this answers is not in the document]
cwli24: Corrected + reworded a bit, keeping in mind it only applies when the flag is on.

patrickhousley (approved):

metal-messiah (approved):

patrickhousley (approved):

jackkinsella:
I'm really excited to test-drive this new feature, but I couldn't figure out how to enable it via the Python agent. I wrote up my question on the forum here (https://discuss.newrelic.com/t/what-config-options-do-you-need-to-enable-to-the-bf-cache/192722), but I'll re-post here in case you can give your input.

jackkinsella:
Thanks for the clarification @cwli24 :)

I tried to get the copy & paste agent working (with the added confit item for bfcache), but I wasn't able to stop the auto-instrumentation of the default settings from clobbering things. I tried disabling auto-instrumentation by setting the following in the `newrelic.ini` file:

```
[newrelic]
browser_monitoring.auto_instrument = false
```

However, this had no effect and the default JS still showed.

jackkinsella:
Update: It wasn't possible to disable `browser_monitoring` within the `newrelic.ini` file when using the `newrelic-admin run-program X` CLI wrapper. What worked instead for removing this wrapper, then manually calling the Python New Relic agent early in my program's load order

```python
import newrelic.agent
# This must appear before all other imports
newrelic.agent.initialize('./newrelic.ini')
```

Then adding the manual JS within the HTML near the head.

Once I did this, BF Cache was working perfectly :))

nsunga:
@cwli24 Hello 👋  - Ive been reading the NR release notes

Via v1226 it looks like bfcache is enabled by default: https://docs.newrelic.com/docs/release-notes/new-relic-browser-release-notes/browser-agent-release-notes/4/#enable-backforward-cache

Does this mean we dont need to add the flag in the `NREUM.init` object?
> It's only available if you use the copy & paste agent and manually insert a ,allow_bfcache:true in the NREUM.init= line object.

Basically, we're on an older agent version, and it would be kind of silly if we upgraded to a version w/o bfcache compatibility

PR #312, opened 2022-11-29
RequireJS incompatibility bug fix (NEWRELIC-5435)

bjfield:
### Overview
This PR changes Webpack's configured output type (`output.library.type`) from `umd` to `self` due to conflicts with [RequireJS](https://requirejs.org/docs/errors.html). The UMD output type first checks for CommonJS and AMD module loading capabilities (which we do not need) before exposing the module globally on `self` by default. Because RequireJS is an AMD loader, in cases where it was present on the page, our Webpack loader was attempting to load the module using the available AMD `define` method—but without specifying a unique identifier, causing RequireJS to throw a "mismatched anonymous define" error and preventing the agent from loading.

### Related Ticket
[NEWRELIC-5435](https://issues.newrelic.com/browse/NEWRELIC-5435)

### Testing
Test coverage is provided by three new files:
- tests/assets/requirejs-compatibility.html
- tests/assets/js/vendor/require-2.3.6.js
- tests/functional/requirejs-compatibility.test.js


patrickhousley (approved):
Looks good!!

metal-messiah (approved):
nice work!

PR #314, opened 2022-11-30
Adding logic for route changes outside a current interaction

patrickhousley:
### Overview

In some SPA applications, our code creates an interaction for the click of a link but that interaction is prematurely closed before the url change occurs. This results in missing data in NR1 since no interaction for the route change is reported.

The change stores the current interaction in memory as `state.prevInteraction` any time the code attempts to set a new node (null or otherwise). If a route change then happens, and there is no current interaction (`state.currentNode === null`), we will check for the previous interaction. If it meets requirements of not being set to be ignored and not created via the interactions API, it will be re-used and updated with the url change.

The restored interaction is re-processed and transmitted to NR1.

### Related Github Issue

https://issues.newrelic.com/browse/NEWRELIC-3359

### Testing

To test locally, download the attached reproduction project.

1. Extract the project and open in your IDE
2. Create a new browser project in NR1-Staging
3. Use the copy/paste method and copy the generated script from NR1
4. In the reproduction project, open `pages/_document.tsx`
5. Paste in the below code into the `<Head>` tag, replacing `<COPY/PASTE CODE HERE>` with the copied code from NR1, making sure to remove the extra script tags from NR1
```
          <script
            type="text/javascript"
            dangerouslySetInnerHTML={{ __html: `
              <COPY/PASTE CODE HERE>
            `}}
          />
```
6. Remove the other script tag that is already present
7. Run `npm i && npm run dev`, a local web server should start at http://localhost:3000
8. Open the test page and click through links
9. In NR1 for your test browser app, open the meta data and copy the GUID
10. In NR1, query the data using the below nrql and you should get no results
    - `SELECT timestamp, actionText, previousUrl, targetUrl from BrowserInteraction where entityGuid = '<GUID>' since 5 minute ago`
    - Replace `<GUID>` with the quid for your test app, found in the app metadata
11. You should see some interactions reported, mainly the initial page load and interactions for pages that have already been lazy loaded, but you will most likely be missing interactions

To fix the issue using this branch:

1. Pull down this branch and run `npm i && npm run build:all`
2. Open the file `build/nr-loader-spa.js`
3. Search for `/build/` and replace it with `http://bam-test-1.nr-local.net:3333/build/`
4. Run `npm run test-server` to start the local agent server
5.  In the reproduction project, open `pages/_document.tsx` and remove the loader from the `dangerouslySetInnerHTML` string leaving only the init, config, and info pieces
6. Add the below script to `Head` below the existing script
```
          <script
            type="text/javascript"
[...]

PR #315, opened 2022-12-02
Fix angular lazy routes with api calls (NR-69140, NEWRELIC-4687)

patrickhousley:
### Overview

This PR expands upon #314 to add logic for recovering discarded interactions when fetch is called or when an xhr is created.

Additionally, this PR adds support for port numbers in the ajax deny list. Previously, entries with port numbers would never match. Now, if a port is provided, it will be used to match XHRs. If no port is provided, the port of the XHR will not be considered.

### Related Github Issue

https://issues.newrelic.com/browse/NR-69140
https://issues.newrelic.com/browse/NEWRELIC-4687

### Testing

To test locally, download the attached reproduction project.

1. Extract the project and open in your IDE
2. Create a new browser project in NR1-Staging
3. Use the copy/paste method and copy the generated script from NR1
4. In the reproduction project, open `src/index.html`
5. Paste in the code from NR1 into the head below the correct comment
6. Comment out the other script tags that pull in other versions of the loader
7. Run `npm i && npm start -- --configuration production --live-reload false  `, a local web server should start at http://localhost:4200
8. Open the test page and click through links
9. In NR1 for your test browser app, open the meta data and copy the GUID
10. In NR1, query the data using the below nrql and you should get no results
    - `SELECT timestamp, actionText, previousUrl, targetUrl from BrowserInteraction where entityGuid = '<GUID>' since 5 minute ago`
    - Replace `<GUID>` with the quid for your test app, found in the app metadata
11. You should see some interactions reported, mainly the initial page load and interactions for pages that have already been lazy loaded, but you will most likely be missing interactions and API calls

To fix the issue using this branch:

1. Pull down this branch and run `npm i && npm run build:all`
2. Open the file `build/nr-loader-spa.js`
3. Search for `/build/` and replace it with `http://bam-test-1.nr-local.net:3333/build/`
4. Run `npm run test-server` to start the local agent server
5.  In the reproduction project, open `src/index.html` and remove the loader from the script tag copied from NR1 leaving only the init, config, and info pieces
6. Un-comment the script tag for the local version of the loader
7. Open the test page and click through links
9. Rerun the query in NR1 and you should now see interactions reported for every link clicked along with all the API calls

**NOTE:** with this version of angular, you need to use node <= v16

[nr-lazy-loading.zip](https://github.com/newrelic/newrelic-browser-agent/files/10137179/nr-lazy-loading.zip)

You can also test this using the steps in #314 and #311


bjfield (changes requested):
I've reviewed the deny-list and port changes and left some suggestions. I tested your SPA changes with the Angular sample implementation a few weeks ago and confirmed the solution works. The code changes make sense and look right at a glance, but I can't comment reliably on potential unexpected impacts and implications without first spending quite a bit of time getting familiar with the SPA logic as a whole. In the interest of time I'll leave that fine-grained review to other.

metal-messiah (approved):

bjfield (approved):

PR #323, opened 2022-12-06
Differentiate babel targets for polyfill and non-polyfill builds (NEWRELIC-5374)

bjfield:
#### `cdn/webpack.config.js`
- Replaced parameter-driven splitting of entries with distinct composable config objects.
- Ended up with three composed objects (`standardConfig`, `polyfillConfig`, `webworkerConfig`) based on `commonConfig`.
- Explicitly specified `target` config property for each build (was a parameter before).
- Changed webpack target to `web` for modern browsers.
- For standard build, targeted modern browsers with babel-loader: last ten versions of Chrome, Firefox, Edge, and Safari.
- For polyfill build, targeted only IE 11 with babel-loader (not older versions of modern browsers also as story proposes).
- Differentiated filenames of chunk outputs for polyfill build to prevent overwriting dependency modules shared with other builds.
- I have not made an entry in the change log. **Please advise in review**.

#### `cdn/package.json`
Modified `browserslist` to include only IE11, as only polyfills are using the `browserslist` target.

#### `tools/jil/server/asset-server.js`
The code was attempting to load `this.polyfills` in all cases, even when no `-P` flag was passed. This would be a problem only in cases where we are not building the polyfills (in which case they cannot be loaded). This would be a rare occurrance but came up in development as I was isolating test cases.

### Related Ticket
[NEWRELIC-5417](https://issues.newrelic.com/browse/NEWRELIC-5417)

### Testing
- Care should be taken to ensure that all tests run, as transpilation targets have changed for all builds.
- In addition, the output of the build directory should be reviewed to ensure minor changes in filename format do not impact downstream systems.

### DIFF
--- a/cdn/webpack.config.js
+++ b/cdn/webpack.config.js
@@ -55,85 +55,113 @@ console.log("PUBLIC_PATH", PUBLIC_PATH)
 console.log("MAP_PATH", MAP_PATH)
 console.log("IS_LOCAL", IS_LOCAL)
 
-const config = (target) => {
-  const isWorker = target === 'webworker'
-  return {
+// The exported configs (standard, polyfill, webworker) compose from this common config.
+const commonConfig = {
+  devtool: false, // defer to SourceMapDevToolPlugin
+  mode: !IS_LOCAL ? 'production' : 'development',
+  optimization: {
+    minimize: true,
+    minimizer: [new TerserPlugin({
+      include: [/\.min\.js$/, /^(?:[0-9])/],
+      terserOptions: {
+        mangle: true
+      }
+    })],
+    flagIncludedChunks: true,
+    mergeDuplicateChunks: true
+  },
+  output: {
+    filename: `[name].js`,
+    chunkFilename: SUBVERSION === 'PROD' ? `[name].[hash:8]${PATH_VERSION}.js` : `[name]${PATH_VERSION}.js`,
+    path: path.resolve(__dirname, '../build'),
+    publicPath: PUBLIC_PATH, // CDN route vs local route (for linking chunked assets)
+    library: {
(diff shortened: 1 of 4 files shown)


cdn/webpack.config.js
```diff
@@ -163,8 +186,46 @@ const config = (target) => {
           }
         }
       ]
-    }
+    },
+    output: {
+      ...commonConfig.output,
```
patrickhousley, line 191: [suggestion] Use [webpack-merge](https://www.npmjs.com/package/webpack-merge) and you don't have to worry about doing stuff like this.

cdn/webpack.config.js
```diff
@@ -163,8 +186,46 @@ const config = (target) => {
...
+      ...{
+        // Because the ./agent-aggregator/aggregator.js dependency is async loaded, the output filename for that chunk will be the
+        // same across all bundles in non-production builds (where filenames aren't hashed) and will thus overwrite with either an
+        // ES5 or ES6 target. For differentiated transpilation of dynamically loaded dependencies in non-production builds, we can
+        // tag output filenames for chunks of the polyfills bundle with `-es5`.
+        chunkFilename: SUBVERSION === 'PROD' ? `[name].[hash:8]${PATH_VERSION}.js` : `[name]-es5${PATH_VERSION}.js`,
```
patrickhousley, line 197: [suggestion] It might actually be good to always have a tag like `-polyfill` on the polyfill files, even for production.

patrickhousley (changes requested):
Please use webpack-merge to merge multiple configs together.

cwli24 (commented):
Nice build/test clean ups & minor inconveniences scratched out here!

PR #328, opened 2022-12-07
Differentiate babel targets for polyfill and non-polyfill builds (NEWRELIC-5374)

bjfield:
#### `cdn/webpack.config.js`
- Replaced parameter-driven splitting of entries with distinct composable config objects.
- Ended up with three composed objects (`standardConfig`, `polyfillConfig`, `webworkerConfig`) based on `commonConfig`.
- Explicitly specified `target` config property for each build (was a parameter before).
- Changed webpack target to `web` for modern browsers.
- For standard build, targeted modern browsers with babel-loader: last ten versions of Chrome, Firefox, Edge, and Safari.
- For polyfill build, targeted only IE 11 with babel-loader (not older versions of modern browsers also as story proposes).
- Differentiated filenames of chunk outputs for polyfill build to prevent overwriting dependency modules shared with other builds.
- Modified `asset-server.js` logic to split browserify babel targets on -P flag.
- Updated `browsers-polyfill.json` to only IE11.
- I have not made an entry in the change log. **Please advise in review**.

#### `cdn/package.json`
Modified `browserslist` to include only IE11, as only polyfills are using the `browserslist` target.

#### `tools/jil/server/asset-server.js`
The code was attempting to load `this.polyfills` in all cases, even when no `-P` flag was passed. This would be a problem only in cases where we are not building the polyfills (in which case they cannot be loaded). This would be a rare occurrance but came up in development as I was isolating test cases.

### Related Ticket
[NEWRELIC-5374](https://issues.newrelic.com/browse/NEWRELIC-5374)

### Testing
- Care should be taken to ensure that all tests run, as transpilation targets have changed for all builds.
- In addition, the output of the build directory should be reviewed to ensure minor changes in filename format do not impact downstream systems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,5 @@ node_modules
 **/local-install
 tests/assets/test-builds/**/*
 packages/browser-agent-core/cjs
-**/webpack-analysis.html
+**/webpack-analysis*.html
 /cdn/build
--- a/cdn/package.json
+++ b/cdn/package.json
@@ -36,11 +36,6 @@
     "core-js": "^3.23.3"
   },
   "browserslist": [
-    "chrome >= 60",
-    "safari >= 11",
-    "firefox >= 56",
-    "ios >= 10.3",
-    "ie >= 11",
-    "edge >= 60"
+    "ie >= 11"
   ]
 }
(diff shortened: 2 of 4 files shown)


patrickhousley (approved):
Ran through multiple test cases locally and this looks good.

patrickhousley (approved):

metal-messiah (approved):

PR #330, opened 2022-12-07
Update PR Template

patrickhousley:
### Overview

Updating the PR template.


.github/workflows/tests-polyfill.yml
```diff
@@ -1,300 +1,21 @@
 name: Polyfill Tests
 
 on:
-  push:
-    branches:
-      - main
-      - dev 
+  pull_request_target:
+    types: [labeled]
```
metal-messiah, line 5: We went away from running polyfill tests running in PRs.  It causes some concurrency problems to do this in PRs with the other tests running, and is overkill.  I think polyfill tests can just be ran before we cut a release.

cwli24: For trunk-based + reduction to just IE11, you'd bring it back to per PR merge into trunk?

patrickhousley: I brnached from the wrong brnach guys. One sec

metal-messiah (approved):

cwli24:
What you think about a `Side effects` section?

patrickhousley:
> What you think about a `Side effects` section?

What would go in the section?

cwli24:
> > What you think about a `Side effects` section?
> 
> What would go in the section?

I'm thinking discrete or non obvious changes. The overview covers what the PR is mainly about but often there are small or subtle changes included (like we have a poor habit of adding other stuff either out of necessity (cleanup) or convenience). These are things that may or may not have stories and wouldn't need one if not.

patrickhousley:
> > > What you think about a `Side effects` section?
> > 
> > 
> > What would go in the section?
> 
> I'm thinking discrete or non obvious changes. The overview covers what the PR is mainly about but often there are small or subtle changes included (like we have a poor habit of adding other stuff either out of necessity (cleanup) or convenience). These are things that may or may not have stories and wouldn't need one if not.

IDK I think the overview should include details about everything that is being changed in the PR but I can add the section.

cwli24 (approved):

PR #335, opened 2022-12-08
Cleaning up repo files

patrickhousley:
### Overview

This PR cleans up duplicated source files and configurations that were left over from the 1220 project restructure.

### Related Github Issue

https://issues.newrelic.com/browse/NEWRELIC-5327

### Testing

All the tests should pass in GHA.


metal-messiah (approved):
Everything looks good. Still wondering what we should do with "zone-promise". Not sure if we want to keep that around for reference purposes. 

But as long as tests still pass on this I think it looks g2g

patrickhousley:
> Everything looks good. Still wondering what we should do with "zone-promise". Not sure if we want to keep that around for reference purposes.
> 
> But as long as tests still pass on this I think it looks g2g

It's in the repo history so we can always bring it back if we need to. However, since it was a POC and never really GAed, I deleted it.

metal-messiah (dismissed):
Looks good

cwli24 (changes requested):
Bottom of base `package.json` still has references to deleted stuff under the `"browser"` object; could you remove this too?

cwli24 (commented):
Also doesn't matter but `.nxignore` also is referring to deleted stuff. Fyi, if you want to strip them there.

.eslintignore
```diff
@@ -11,6 +11,5 @@ tools/test-builds/*
 tests/assets/test-builds/**/*.js
```
cwli24, line 11: There's a `post-installer.js` in here

tests/browser/spa/hashchange-popstate.browser.js
```diff
@@ -4,7 +4,7 @@
  */
 
 const jil = require('jil')
-let cleanUrl = require('../../../agent/clean-url')
+let cleanUrl = require('@newrelic/browser-agent-core/src/common/url/clean-url').cleanURL
```
cwli24, line 7: Just double checking that this doesn't get its own tests/lib version, since `cleanUrl` is used for both "actual" and "expected" output comparison.

.eslintignore
patrickhousley: Not sure what you mean or are expecting here.

cwli24 (approved):

cwli24:
Thanks for sweeping!

metal-messiah (approved):

PR #338, opened 2022-12-09
Breaking out the asset/collector server

patrickhousley:
### Overview

This PR creates a new test server package containing the asset and collector servers from the JIL test framework. This does not include any refactoring or other changes.

https://issues.newrelic.com/browse/NR-70850

### Testing

The npm command `test-server` and `test` should still function exactly as they do today.


bjfield (dismissed):
- General package relocating and reference updates look like a clean lift and the rationale for packaging (explained in the ticket) makes sense.
- I've built the branch and tested (for both browser and functional tests) and all commands work the same, as expected.
- Placing the `test-server` package one level deep inside `browser-agent-testing` makes sense for clarity should we choose to modularize other aspects of our test framework, but does not match the level of depth of the two other packages (e.g., `browser-agent/src`). Would like hearing the rationale, but this doesn't hold up approval.
- The directory name `test-server` matches the command line `npm run test-server` without being too long, and speaks to the fuller scope of what the component does than did `server`.
- There will be some conflict with the location of `asset-server.js` for pending [PR 328](https://github.com/newrelic/newrelic-browser-agent/pull/328/files#diff-58edc12c5ca482101fefd6ca617877004f65eb5f53c9b1ee6cb1dcfd612443fe). Might be good to get that PR in before merging this one. Otherwise, I don't see any problems with merging this POC when convenient.

patrickhousley:
> * Placing the `test-server` package one level deep inside `browser-agent-testing` makes sense for clarity should we choose to modularize other aspects of our test framework, but does not match the level of depth of the two other packages (e.g., `browser-agent/src`). Would like hearing the rationale, but this doesn't hold up approval.

My thinking for this is that we would eventually have more packages that are all around the idea of testing the browser agent and it would make sense to scope these inside a folder to keep the packages directory clean. These test packages are never meant to be published or used in any way outside of just testing the agent.

metal-messiah:
> > * Placing the `test-server` package one level deep inside `browser-agent-testing` makes sense for clarity should we choose to modularize other aspects of our test framework, but does not match the level of depth of the two other packages (e.g., `browser-agent/src`). Would like hearing the rationale, but this doesn't hold up approval.
> 
> 
> 
> My thinking for this is that we would eventually have more packages that are all around the idea of testing the browser agent and it would make sense to scope these inside a folder to keep the packages directory clean. These test packages are never meant to be published or used in any way outside of just testing the agent.
> 
> 
> 
> 

The current directory architecture of the repo is that "packages" are things that get published to npm (for example, cdn and jil are not included there even though they reference and use the "package" code for their various purposes)

Just for my understanding, what is the point of labeling and maintaining non-publishable directories as packages, instead of just keeping them as directories in the repo?

patrickhousley:
Closing this PR. Moving the server seems to have caused a lot of issues in the tests. Decoupling is going to take a lot more work than was first anticipated.

PR #341, opened 2022-12-12
update upload NR script to grab from CDN instead of build folder

metal-messiah:
### Overview
This PR updates the `publish-apm` flow to pull data from the CDN instead of building on the fly.  This is to ensure that changes in the trunk branch are not accidentally pulled in at execution time.  CDN assets need to be published to the CDN before running this script. The intended browser agent version needs to be included as an input to the workflow, ie "1221".

### DIFF
--- a/tools/scripts/upload-to-nr.js
+++ b/tools/scripts/upload-to-nr.js
@@ -158,7 +158,10 @@ async function run() {
     }
 
     console.log('Uploading loader ' + filename + ' to ' + environment + '...')
-    var options = util._extend(envOptions[environment], baseOptions)
+    var options = {
+      ...envOptions[environment],
+      ...baseOptions
+    }
 
     request(options, function (err, res, body) {
       if (err) return cb(err)
--- a/tools/scripts/upload-to-nr.js
+++ b/tools/scripts/upload-to-nr.js
@@ -25,6 +25,9 @@ var argv = yargs
   .string('eu-api-key')
   .describe('eu-api-key', 'API key to use for talking to EU RPM site to upload loaders')
 
+  .string('version')
+  .describe('version', 'Browser Agent version number')
+
   .boolean('skip-upload-failures')
   .describe('skip-upload-failures', "Don't bail out after the first failure, keep trying other requests")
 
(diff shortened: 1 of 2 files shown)


tools/scripts/upload-to-nr.js
```diff
@@ -33,137 +34,186 @@ var argv = yargs
...
+   * @returns {Promise<{[fileName]: string}[]>} Promise contains an array of objects {[filename]: body} --> {'nr-loader-spa-1221.min.js': ...scriptContents}
+   */
+  async function loaderFilenames() {
+    const loaderTypes = ['rum', 'full', 'spa']
+    const version = argv['version']
+    const fileNames = loaderTypes.map(type => [
+      `nr-loader-${type}-${version}.min.js`, 
+      `nr-loader-${type}-polyfills-${version}.min.js`,
+      `nr-loader-${type}-${version}.js`, 
+      `nr-loader-${type}-polyfills-${version}.js`,
```
patrickhousley, line 185: [question] Is there ever a chance that NR1 uses the non-minified loader? if not, does it make sense to not upload it?

metal-messiah: Not sure on that one.  I know there are a few flags in the connect service for "dev" builds, which *i think* signal it to return the unminified version.  Not sure if that is still used

tools/scripts/upload-to-nr.js
```diff
@@ -33,137 +34,186 @@ var argv = yargs
...
+          'X-Api-Key': argv['production-api-key']
+        }
       }
     }
-  }
 
-  console.log('Uploading loader ' + filename + ' to ' + environment + '...')
-  var options = util._extend(envOptions[environment], baseOptions)
+    console.log('Uploading loader ' + filename + ' to ' + environment + '...')
+    var options = util._extend(envOptions[environment], baseOptions)
```
patrickhousley, line 161: This could be replace with object spread syntax and we can remove reliance on this third-party.

```suggestion
    var options = {
      ...envOptions[environment],
      ...baseOptions
    }
```

tools/scripts/upload-to-nr.js
```diff
@@ -5,8 +5,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-var fs = require('fs')
-var path = require('path')
 var request = require('request')
```
patrickhousley, line 8: I am going to open a backlog tech debt ticket to remove our reliance on this third-party. The package has been deprecated for some time: see [here](https://github.com/request/request/issues/3142). Node provides a supported package, [Undici](https://github.com/nodejs/undici), that we should use instead.

patrickhousley: https://issues.newrelic.com/browse/NR-73098

metal-messiah: could also switch to using native fetch which is included in Node if we bump the node versions inside the GHA containers

patrickhousley (approved):
Looks good.

patrickhousley (approved):

cwli24 (approved):

PR #343, opened 2022-12-14
[chore] Converting unit tests for jserrors

patrickhousley:
### Overview

Converting unit tests for jserrors from jil to jest.

N/A

### Testing

Pull branch and run `npx nx run-many --target=test`


packages/browser-agent-core/src/features/jserrors/aggregate/compute-stack-trace.test.js
```diff
@@ -0,0 +1,292 @@
...
+    }))
+    expect(result.frames.length).toEqual(1);
+    expect(result.frames).toContainEqual(expect.objectContaining({
+      func: 'evaluated code'
+    }));
+  });
+
+  test("parser can handle stack with anonymous function", () => {
+    const mockError = browserErrors.utils.constructError({
+      ...browserErrors.data.chromeEdge98Errors[0],
```
metal-messiah, line 135: The way fixture data is used and is intended to be used as time goes on in this test is a little confusing. Are these values supposed to exist even as our testing window slides (at some point (soon) we won't support testing chrome 98, ie11, etc)

When do we update this list of fixture data and how?

Is there a way we can represent the fixture data in a more static way that stays evergreen?

patrickhousley: The data directory contains a README with details on how the test data was generated. We could take a subset of the data and remove the rest

metal-messiah (approved):

metal-messiah (approved):

cwli24 (approved):
Pog

PR #344, opened 2022-12-14
NR-73497: Removing phantomjs

patrickhousley:
### Overview

Removing all phantomjs and all references to it.

https://issues.newrelic.com/browse/NR-73497

### Testing

1. Pull the branch
2. Rlean all the `node_module` directories
3. Run `npm i`
4. Run `npm ls phantom`
5. Verify the dependency is not found in the project

All the tests should also pass with the `safe to test` label.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,11 +82,6 @@ tests (located in `/tools/jil`).
 _Before running tests locally, be sure to [install and build](#building) from the root directory to ensure all dependencies are loaded and the application is properly built._
 
 ### Running the test suite
-To run all applicable tests against [PhantomJS](http://phantomjs.org/), just type the following command with no additional arguments:
-
-```
-npm run test
-```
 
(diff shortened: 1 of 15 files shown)


bjfield (approved):
Install, build, and tests run as expected, and the rationale in the ticket makes sense. Also, PhantomJS is now unsupported and removing it will help keep dependencies up to date and free of distracting warnings—which is not much, but sometimes it's the little things.

Most of the changes in the code end up with a logical alternative. Switching the `browserList` method's `pattern` argument default to `chrome@latest` seems like a good choice.

The one thing I'd like to see is better README detail on how to test if not `npm run test`. However, updating the README.md is actually a chore I'd like to pick up separately, and it's beyond the scope of removing PhantomJS. I can spin that work up separately.

metal-messiah (approved):

PR #346, opened 2022-12-15
POC publish with environments

patrickhousley:
### Overview

Testing out publish workflow using github environments. This workflow doesn't actually do anything right now.

### Related Issue(s)

N/A

### Testing

Needs to be merged to main and manually executed.


.github/workflows/publish-environments.yml
```diff
@@ -0,0 +1,52 @@
...
+    container: node:16
+    steps:
+      - uses: actions/checkout@v3
+      - name: install
+        run: npm ci
+      - name: build:nr
+        run: npm run cdn:build:nr
+      - name: upload artifacts to NR
+        run: |
+          echo 'We would normally upload to CDN here'
```
metal-messiah, line 36: We should probably have the publish to internal prod step in here if it's all gonna be in one action

patrickhousley: Yea maybe. I just wanted to see how workflows worked with gh environments.

PR #347, opened 2022-12-15
Adding props to sauce connect

patrickhousley:
### Overview

Adding additional properties to the sauce connect launch script:

- tunnelName: tunnelIdentifier is deprecated in the sauce connect binary, this is just a rename
- logger: connecting the sauce connect launcher logger to console.log so we actually get output
- tunnelDomains: adding our internal test domain so only traffic for that domain is routed through the tunnel to the local server
  - Without this, SL forwards all network traffic through the tunnel. Setting this to our domain only will cut down on the amount of traffic coming through the tunnel and may help with performance and stability in the tests.

### Related Issue(s)

None

### Testing

Start sauce connect with `npm run sauce:connect -- -v` to see the log output. To test the change in general, run some of the JIL tests.

### DIFF
--- a/tools/jil/server/asset-server.js
+++ b/tools/jil/server/asset-server.js
@@ -152,7 +152,7 @@ class AgentInjectorTransform extends AssetTransform {
           if (typeof msg === 'object') {
             msg = JSON.stringify(msg)
           }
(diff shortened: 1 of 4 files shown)


bjfield (dismissed):
These are small useful changes that do not impact the tests themselves.

cwli24 (approved):

metal-messiah (approved):

PR #348, opened 2022-12-16
PR Build

metal-messiah:
### Overview

This PR makes the pr-build and ship more selective.  It should only build and ship PR builds if the PR is a new feature or bug.

### Testing

Once merged to main, adding "safe to test" label alone will not build and ship the PR. "bug", "MMF", "enhancement", or "standalone" must also be present on the PR to trigger the build.

### DIFF
--- a/.github/workflows/dev-pr.yml
+++ b/.github/workflows/dev-pr.yml
@@ -46,4 +46,4 @@ jobs:
             --purge-internal
       - name: Notify
         run: |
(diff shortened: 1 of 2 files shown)


bjfield (approved):
Build logic for pr-dev matches intention: executes the job if the workflow was dispatched directly; or if there's a `safe to test` label _and_ one or more of `standalone`, `MMF`, `bug`, `enhancement`.

patrickhousley (approved):

PR #351, opened 2022-12-16
remove metrics from lite build to support legacy account parity

metal-messiah:
### Overview

This PR addresses an issue where 1220+ agents have added metrics to the lite agent build, but legacy "lite" customers do not have access to the metrics endpoint, causing 403 network codes.  This is handled beyond the scope of the browser agent team, by assigning entitlements that are enforced at ingest time.

### Related Issue(s)

https://discuss.newrelic.com/t/bam-nr-data-net-post-returns-403-error/161276

### DIFF
--- a/cdn/agent-aggregator/util/features.js
+++ b/cdn/agent-aggregator/util/features.js
@@ -3,6 +3,9 @@ import agentIdentifier from '../../shared/agentIdentifier'
 import { getEnabledFeatures } from '@newrelic/browser-agent-core/src/common/util/enabled-features'
 import { Aggregator } from '@newrelic/browser-agent-core/src/common/aggregate/aggregator'
 
+/**
+ * A mapping of the agent features from camelCase to kabob, which is used in the directory names
+ */
 export const modules = {
   pageViewEvent: 'page-view-event',
   pageViewTiming: 'page-view-timing',
@@ -14,7 +17,7 @@ export const modules = {
   metrics: 'metrics'
 }
 
-const lite = [modules.pageViewEvent, modules.pageViewTiming, modules.metrics]
+const lite = [modules.pageViewEvent, modules.pageViewTiming] // lite agent cannot currently use metrics endpoint, because legacy lite customers are not provisioned access to that endpoint
 const pro = [...lite, modules.jsErrors, modules.ajax, modules.pageAction, modules.sessionTrace]
 const spa = [...pro, modules.spa]
 const worker = [ modules.metrics, modules.jsErrors, modules.ajax, modules.pageAction]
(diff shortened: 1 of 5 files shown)


metal-messiah:
Awaiting input from ingest and PM on how we want to proceed

metal-messiah:
I think in general the concern is that all customers are confused when *any* newrelic service that used to work is now randomly (in their perception) having errors. 

The 403 doesn't have any messaging tied to it so the customer just assumes the agent is broken. This is inefficient for everyone from the customer to support to us. 

I got this on ingests radar to add a 403 message at the very least, but further planning is needed. 

This pr would just align the lite loader with the behavior in 1216

cdn/agent-aggregator/util/features.js
```diff
@@ -14,7 +17,7 @@ export const modules = {
   metrics: 'metrics'
```
cwli24, line 17: Minor: can be removed

cdn/agent-aggregator/util/features.js
```diff
@@ -14,7 +17,7 @@ export const modules = {
...
 }
 
-const lite = [modules.pageViewEvent, modules.pageViewTiming, modules.metrics]
+const lite = [modules.pageViewEvent, modules.pageViewTiming] // lite agent cannot currently use metrics endpoint, because legacy lite customers are not provisioned access to that endpoint
 const pro = [...lite, modules.jsErrors, modules.ajax, modules.pageAction, modules.sessionTrace]
```
cwli24, line 21: You'll need to add metrics to Pro, to retain it for Pro & SPA, or they won't have them either.

cwli24 (changes requested):

cdn/agent-aggregator/util/features.js
patrickhousley: It's still used and can't be removed.

cdn/agent-loader/lite.js
```diff
@@ -2,16 +2,13 @@
  * Copyright 2020 New Relic Corporation. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-import { detectPolyfillFeatures } from './utils/feature-detection'
-const PolyfillFeatures = detectPolyfillFeatures();
```
patrickhousley, line 6: Why were these lines removed?

metal-messiah: these are used to report supportability metrics

patrickhousley: Oh

patrickhousley:
Need to add changelog entry. Otherwise, looks good once the unit tests are passing.

PR #353, opened 2022-12-20
Checkout & test future merge when adding "safe to test" - take #2

cwli24:
### Overview

Re-do of #337 that was unwounded on trunk due to failure.
The previous reusable workflow is converted to a composite action instead.

### Related Issue(s)

[NR-72380](https://issues.newrelic.com/browse/NR-72380)

### Testing

Since the new action does not yet exist in `main`, you can only test `workflow_dispatch` works. But you may have to add the proper `ref` to the initial checkout step for each job of the launched workflow: 
image

### DIFF
--- a/.github/actions/test-setup/action.yml
+++ b/.github/actions/test-setup/action.yml
@@ -11,7 +11,7 @@ runs:
     # For workflow_dispatch, it's assume that the dispatch branch is already checked out.
 
     - name: install
-      run: npm ci --cache ./.npm
+      run: npm ci --no-fund --no-audit --no-progress
       shell: bash
 
     - name: build
--- a/.github/actions/test-setup/action.yml
+++ b/.github/actions/test-setup/action.yml
@@ -1,0 +1,19 @@
+name: Testing setup
+descriptions: Checkout the appropriate commit and build it before testing
+
+runs:
+  using: "composite"
+  steps:
+    - if: github.event_name == 'pull_request_target'  # checkout the would-be merge instead of PR branch head
+      uses: actions/checkout@v3
+      with:
+        ref: refs/pull/${{ github.event.number }}/merge
+    # For workflow_dispatch, it's assume that the dispatch branch is already checked out.
+
+    - name: install
+      run: npm ci --cache ./.npm
+      shell: bash
+
+    - name: build
+      run: npm run build:all
+      shell: bash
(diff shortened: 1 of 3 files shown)


.github/actions/test-setup/action.yml
```diff
@@ -0,0 +1,19 @@
+name: Testing setup
+descriptions: Checkout the appropriate commit and build it before testing
+
+runs:
+  using: "composite"
+  steps:
+    - if: github.event_name == 'pull_request_target'  # checkout the would-be merge instead of PR branch head
+      uses: actions/checkout@v3
+      with:
+        ref: refs/pull/${{ github.event.number }}/merge
```
patrickhousley, line 10: See https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

patrickhousley: Also found https://github.com/actions/checkout/issues/696 where they are talking about how the context object is cached so your approach might be better.

.github/actions/test-setup/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    # For workflow_dispatch, it's assume that the dispatch branch is already checked out.
+
+    - name: install
+      run: npm ci --cache ./.npm
```
patrickhousley, line 14: Can we please change this to `npm ci --no-fund --no-audit --no-progress`

.github/actions/test-setup/action.yml
cwli24: We don't want to check out the pull request branch though

.github/actions/test-setup/action.yml
patrickhousley: Also, I think we should rename this to `jil-setup/action.yml` and only use it for jobs where we are running jil. We can add another action `ci-setup/action.yml` that can use `npm ci --no-fund --no-audit --no-progress --ignore-scripts`.

.github/actions/test-setup/action.yml
patrickhousley: Assuming the intention here is that we want to check out a simulation of the merge between the target branch and the pull request branch. If that is the case, my first comment would work but it caches the SHA and on re-runs it does not re-calculate that SHA. If the target branch has been merged into, those merged would be missed in the re-run. SO, I think your approach might solve that and would be better.

patrickhousley (approved):

metal-messiah (approved):

PR #357, opened 2023-01-04
Update babel targets and tests to match support statement (NEWRELIC-5417)

bjfield:
### Overview

Our working support statement is `last 10 versions of Chrome, Safari, Firefox, and Edge on mobile and desktop`. This PR:
- Updates the babel targets in the webpack configuration and JIL asset server to include missing ChromeAndroid and iOS specs.
- Modifies `sauce-browsers.js` to generate `browsers-supported.json` using the same browserslist source that we use to calculate version targets for babel transpilation--ensuring version-range parity.
- Enables Android functional and unit tests. Previously these were disabled, but [specifying Android 9 ensures that Sauce defaults to JSON Wire Protocol](https://changelog.saucelabs.com/en/update-to-google-chrome-version-100-on-android-emulators), which works with our current JIL.

### Related Issue(s)

[NEWRELIC-5417](https://issues.newrelic.com/browse/NEWRELIC-5417)

### Testing

Automated tests should pass prior to merge.

### Notes on reintroduced Android Chrome testing support

- This PR uses browerslist to derive version numbers for testing, and browserslist uses caniuse—which always [only has info about the latest ChromeAndroid version](https://github.com/browserslist/browserslist/issues/156). This means "Last 10" for ChromeAndroid is equivalent to "latest" from a transpilation perspective. So we can't trust browserslist to give us a complete version list for Android Chrome.
- Because Chrome version releases sync between desktop and mobile [within a day or two](https://en.wikipedia.org/wiki/Google_Chrome_version_history) in the real world, our support statement (including last 10 Chrome mobile) effectively means we support, and thus wish to test, the same versions of Chrome on Android mobile as on desktop.
- SauceLabs has [just one version](https://changelog.saucelabs.com/en/update-to-google-chrome-version-100-on-android-emulators) of ChromeAndroid on all their emulators: 100, and it's not the latest (108). So with Sauce the best we can do is test the version they offer.
- The Sauce api we use to select browsers does not give Chrome version numbers for android, but rather Android version numbers. So we just have to pick a browser-platform definition arbitrarily (version <= 9).
- Presently, I've confirmed that using an Android 9 browser-platform Sauce definition is compatible with JIL, whereas an Android 12 definition fails with W3C-specific errors. This PR specifies min and max versions for Android both as 9. When W3C compatibility updates are made for JIL, we can test Android 12. However, this will not impact the version of ChromeAndroid we are testing, as Sauce Labs uses the same version of ChromeAndroid on all its emulators (see above).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -235,60 +235,46 @@ jobs:
 
       - name: run tests
         run: node --max-old-space-size=8192 ./tools/jil/bin/cli.js -f merged -b ios@* -s -t 85000 --unit-only
-  # android-functional:
-    # if: contains(fromJson('["workflow_dispatch"]'), github.event_name) || contains(github.event.pull_request.labels.*.name, 'safe to test')
-  #   timeout-minutes: 30
-  #   continue-on-error: true
-  #   runs-on: ubuntu-latest
-  #   container: node:14
-  #   
-  #   env:
-  #     NEWRELIC_ENVIRONMENT: ci
-  #     JIL_SAUCE_LABS_USERNAME: ${{ secrets.JIL_SAUCE_LABS_USERNAME }}
-  #     JIL_SAUCE_LABS_ACCESS_KEY: ${{ secrets.JIL_SAUCE_LABS_ACCESS_KEY }}
-  #     NEW_RELIC_LICENSE_KEY: ${{ secrets.JIL_NODE_NEW_RELIC_LICENSE_KEY }}
+  android-functional:
+    if: contains(fromJson('["workflow_dispatch"]'), github.event_name) || contains(github.event.pull_request.labels.*.name, 'safe to test')
+    timeout-minutes: 30
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    container: node:14
+    
+    env:
+      NEWRELIC_ENVIRONMENT: ci
+      JIL_SAUCE_LABS_USERNAME: ${{ secrets.JIL_SAUCE_LABS_USERNAME }}
+      JIL_SAUCE_LABS_ACCESS_KEY: ${{ secrets.JIL_SAUCE_LABS_ACCESS_KEY }}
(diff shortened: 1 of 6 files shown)


bjfield:
Before tests can run on this PR, PR 350 needs to be merged to fix test-server issues.

patrickhousley (dismissed):

metal-messiah (approved):

bjfield:
Notes on reintroduced Android Chrome testing support:

- This PR uses browerslist to derive version numbers for testing, and browserslist uses caniuse—which always [only has info about the latest ChromeAndroid version](https://github.com/browserslist/browserslist/issues/156). This means "Last 10" for ChromeAndroid is equivalent to "latest" from a transpilation perspective. So we can't trust browserslist to give us a complete version list for Android Chrome.
- Because Chrome version releases sync between desktop and mobile [within a day or two](https://en.wikipedia.org/wiki/Google_Chrome_version_history) in the real world, our support statement (including last 10 Chrome mobile) effectively means we support, and thus wish to test, the same versions of Chrome on Android mobile as on desktop.
- SauceLabs has [just one version](https://changelog.saucelabs.com/en/update-to-google-chrome-version-100-on-android-emulators) of ChromeAndroid on all their emulators: 100, and it's not the latest (108). So with Sauce the best we can do is test the version they offer.
- The Sauce api we use to select browsers does not give Chrome version numbers for android, but rather Android version numbers. So we just have to pick a browser-platform definition arbitrarily (version <= 9).
- Presently, I've confirmed that using an Android 9 browser-platform Sauce definition is compatible with JIL, whereas an Android 12 definition fails with W3C-specific errors. This PR specifies min and max versions for Android both as 9. When W3C compatibility updates are made for JIL, we can test Android 12. However, this will not impact the version of ChromeAndroid we are testing, as Sauce Labs uses the same version of ChromeAndroid on all its emulators (see above).

cwli24 (approved):
GJ

PR #358, opened 2023-01-06
One agent poc + restructured

metal-messiah:
### Overview

This PR represents a `massive` change to the Browser Agent's:
* structure
* behavior
* deployment capability
* build output

As such, this is an `all hands on deck` PR that requires extreme scrutiny before merging.  An engineering meeting will be conducted around this PR, but the core concepts are as follows
* Reorganize `packages` into a src folder, instead utilizing package.json + build logic to distribute NPM.
    - cdn
        - entry points used by webpack to generate CDN assets
    - common
        - common files shared across cdn files, loaders, features, and other common files
    - features
        - feature `inst` and `agg` files, and related utility files such as base classes
    - loaders
        - entry points used by CDN files as well as intended to be exposed to NPM installations for custom usage.
* Create a centralized `loader` pattern, where nearly all implementations of a `loader` extend from the base-loader.
    - agent --> the base loader that relies on feature inst files to be fed as arguments
        - intended to be deployed to NPM
    - browser-agent --> an extension of the base `agent` that provides all features out of the box
        - intended to be deployed to NPM
    - micro-agent --> a custom implementation of the agent that bypasses importing inst files, and only imports a subset of feature agg files meant to be manually triggered (errors, pageActions, metrics)
    - worker-agent --> an extension of the base `agent` that is used for building the worker agent bundle
        - intended to be deployed to NPM
    - spa.js, pro.js, lite.js, worker.js, polyfills etc --> extensions of the base `agent` that provide a subset of features out of the box
        - intended to be deployed to CDN, NRDB, APM, Copy+Paste
* Remove the reliance on RUM call, and DOM Events to trigger features
    - Features now are responsible for importing and draining their own data, instead of relying on window.load + RUM call events.  -- note, Session Traces still rely on the RUM call result flags.
    - Features can be imported either by instrumentation file, or directly by aggregator file

### Related Issue(s)

https://issues.newrelic.com/browse/NR-62982
https://issues.newrelic.com/browse/NR-62982

### Link(s)
https://whimsical.com/package-structure-PsRh7W6zcZPFJAHdxVVoFU
https://whimsical.com/one-agent-directory-structure-oZt71zC2VoRsKUZhsb8uo
https://whimsical.com/one-agent-basic-behavior-SpGiNFWdWTfisje1eMdFUS

### Testing

Tests have been updated to accept the new format, and should execute and succeed as before. `package-tests` has been removed, and we will need to rethink this.  Because of the removal, `package-tests` will fail in the CI/CD of this PR.

### What this PR DOES NOT DO
* Define and execute NPM build + deploy logic, though it does have a scaffolding in place for some of the built files.
* Test NPM builds or deployments
* Define GH Actions to manage versioning or deploying NPM packages

### Comments
If you have the time, please read through the comments attached to this PR as a primer for the engineering discussion.  If it helps, add feedback or dialog in each (or new comments) to ensure we cover them during the engineering meeting.


babel-env-vars.js
```diff
@@ -0,0 +1,18 @@
+
+const pkg = require('./package.json')
+const fs = require('fs')
+
+const VERSION = fs.readFileSync('./VERSION', 'utf-8')
+
```
metal-messiah, line 6: This file is used to inject a "version" into the bundle.  This is complicated right now by the fact that NPM would build with package.json.version (0.0.9) and CDN will build with VERSION file's value (1222), as well as subversion attributes such as PROD, DEV, LOCAL, etc.. Once aligned, this process will get much simpler.

babel.npm.cjs.js
```diff
@@ -0,0 +1,19 @@
+
```
metal-messiah, line 1: Used to compile src/index.js for NPM

babel.npm.esm.js
```diff
@@ -0,0 +1,11 @@
+const babelEnv = require('./babel-env-vars')
```
metal-messiah, line 1: Used to compile src/index.js for NPM (esm)

jest.preset.js
```diff
@@ -1,3 +1,3 @@
-const nxPreset = require('@nrwl/jest/preset').default;
```
metal-messiah, line 1: unclear what the consequence of this is

package.json
```diff
@@ -1,41 +1,53 @@
 {
```
metal-messiah, line 1: The setup of package.json, as well as npmignore will need to be iterated on later.  This is a scaffolding.

package.json
```diff
@@ -1,41 +1,53 @@
...
+    ".": {
+      "import": "./dist/mjs/index.mjs",
+      "require": "./dist/cjs/index.cjs"
+    }
+  },
   "scripts": {
     "test": "node tools/jil/bin/cli.js",
-    "build:all": "npx nx run-many --target=build && npm run packages:bundle && npm run cdn:build:local",
+    "build:all": "npm run cdn:build:local && npm run build:npm",
+    "build:npm": "npm run npm:build:esm && npm run npm:build:cjs",
```
metal-messiah, line 21: build:all builds CDN (/build -- webpack) and builds NPM (/dist -- babel)

src/cdn/lite.js
```diff
@@ -0,0 +1,13 @@
+import { Agent } from "../loaders/agent";
```
metal-messiah, line 1: lite, pro, and spa loaders use `agent` loader.  This is the "custom" loader that you can feed features into.

src/cdn/pro.js
```diff
@@ -0,0 +1,21 @@
...
+new Agent({
+    features: [
+        InstrumentPageViewEvent,
+        InstrumentPageViewTiming,
+        InstrumentSessionTrace,
+        InstrumentXhr,
+        InstrumentMetrics,
+        InstrumentPageAction,
+        InstrumentErrors
+    ], loaderType: 'pro'
```
metal-messiah, line 20: `loaderType` is used to report a SM about what loader was used

src/common/config/state/runtime.js
```diff
@@ -14,6 +14,7 @@ const model = agentId => { return {
   customTransaction: undefined,
   disabled: false,
   features: {},
+  isolatedBacklog: false,
```
metal-messiah, line 17: isolatedBacklog is used by the EE in the MFE loader, to prevent EE context and buffers from bubbling up

src/common/util/feature-flags.js
```diff
@@ -15,12 +16,6 @@ export function activateFeatures (flags, agentIdentifier) {
     sharedEE.emit('feat-' + flag, [])
     activatedFeatures[flag] = true
   })
-  drainAll(agentIdentifier)
```
metal-messiah, line 18: activateFeatures (result of RUM call) is no longer responsible for draining.  Each feature buffers to a cache with the feature's name.  and each feature's agg file is responsible for draining its own cache

src/features/ajax/aggregate/index.js
```diff
@@ -48,21 +51,14 @@ export class Aggregate extends FeatureBase {
 
     if (allAjaxIsEnabled()) setDenyList(getConfigurationValue(agentIdentifier, 'ajax.deny_list'))
 
-    register('xhr', storeXhr, undefined, this.ee)
+    register('xhr', storeXhr, this.featureName, this.ee)
```
metal-messiah, line 54: This is an example of how each feature is now "bucketing" the EE buffers

src/features/ajax/aggregate/index.js
```diff
@@ -245,5 +241,7 @@ export class Aggregate extends FeatureBase {
       }
       return true
     }
+
+    drain(this.agentIdentifier, this.featureName)
```
metal-messiah, line 245: finally, the agg file drains it own bucket

src/features/ajax/instrument/index.js
```diff
@@ -13,18 +13,21 @@ import { wrapFetch, wrapXhr } from '../../../common/wrap'
...
 var handlers = ['load', 'error', 'abort', 'timeout']
 var handlersLen = handlers.length
 
 var origRequest = originals.REQ
 var origXHR = globalScope?.XMLHttpRequest
 
-export class Instrument extends FeatureBase {
-  constructor(agentIdentifier) {
-    super(agentIdentifier)
+export class Instrument extends InstrumentBase {
```
metal-messiah, line 27: InstrumentBase is responsible for importing its matching aggregator file, using `FEATURE_NAME`

src/features/jserrors/aggregate/index.js
```diff
@@ -199,7 +201,8 @@ export class Aggregate extends FeatureBase {
 
     // stn and spa aggregators listen to this event - stn sends the error in its payload,
     // and spa annotates the error with interaction info
-    handle('errorAgg', [type, bucketHash, params, newMetrics], undefined, undefined, this.ee)
+    handle('errorAgg', [type, bucketHash, params, newMetrics], undefined, FEATURE_NAMES.sessionTrace, this.ee)
+    handle('errorAgg', [type, bucketHash, params, newMetrics], undefined, FEATURE_NAMES.spa, this.ee)
```
metal-messiah, line 205: note that these buffers are storing to other buckets.  This is useful to know because these buckets belong to 'sessionTrace' and 'spa' but are being filled by data in jserrors

src/index.js
```diff
@@ -0,0 +1,13 @@
+export { Agent as CustomAgent } from './loaders/agent'
```
metal-messiah, line 1: This is the main entry for NPM, exposing all loaders and all feature inst files

src/loaders/agent.js
```diff
@@ -0,0 +1,56 @@
+// loader files
```
metal-messiah, line 1: this is the "custom" loader, where you can feed inst files as arguments

src/loaders/api/apiAsync.js
```diff
@@ -0,0 +1,96 @@
+import { FEATURE_NAMES } from "../features/features"
```
metal-messiah, line 1: this is the portion of the API that can be loaded async, as is what happens in current agent

src/loaders/browser-agent.js
```diff
@@ -0,0 +1,27 @@
+import { Agent } from "./agent";
```
metal-messiah, line 1: This represents the "prebuilt" "one-agent" type loader we could expose to NPM.  All features included, on/off but no bundle size change.

src/loaders/configure/configure.js
```diff
@@ -0,0 +1,37 @@
+import { setAPI } from '../api/api'
```
metal-messiah, line 1: This is similar to the former "configure" file, except that it can output the interface, so it can be stored to the class who calls it for NPM..  It can also set the window, as is seen with the CDN builds

example --
nr = new Agent()
nr.noticeError() ---> (instead of window.newrelic.noticeError())

src/loaders/configure/configure.js
```diff
@@ -0,0 +1,37 @@
...
+import { addToNREUM, gosCDN, gosNREUMInitializedAgents } from '../../common/window/nreum'
+import { setConfiguration, setInfo, setLoaderConfig, setRuntime } from '../../common/config/config'
+import { activateFeatures, activatedFeatures } from '../../common/util/feature-flags'
+import { isBrowserScope, isWorkerScope } from '../../common/util/global-scope'
+
+export function configure(agentIdentifier, opts = {}, loaderType) {
+    let { init, info, loader_config, runtime = {loaderType}, exposed = true } = opts
+    const nr = gosCDN()
+    let api = {}
+    if (!info) {
```
metal-messiah, line 11: This is where the decision is made to set the API to the window, or to the caller (class)

info is defined as an option, only in NPM. so if it is empty, we can assume the configs are on the page as is expected with CDN builds

src/loaders/micro-agent.js
```diff
@@ -0,0 +1,64 @@
+// loader files
```
metal-messiah, line 1: this represents the micro frontend loader, where all features are lazy-loaded, and only "manual" features are included.

src/loaders/worker-agent.js
```diff
@@ -0,0 +1,20 @@
+import { Agent } from "./agent";
```
metal-messiah, line 1: The worker agent also utilizes the "custom" agent

tests/assets/spa/external-scripts/script-with-ajax.html
```diff
@@ -22,7 +22,7 @@
     var args = [function(api) {
       spaApi = api
     }]
-    scopedEE.buffer([type])
+    scopedEE.buffer([type], 'spa')
```
metal-messiah, line 25: because events are now in "named" buffers, we also need to make sure tests that directly modify the buffers do this too

src/common/drain/drain.js
```diff
@@ -6,33 +6,66 @@
...
+import { featurePriority } from '../../loaders/features/features'
 
+const registry = {}
 
-// calls will need to update to call this more directly so we can explicitly pass in the ee and handler
-export function drain(agentIdentifier, group = 'feature') {
-  const baseEE = agentIdentifier ? ee.get(agentIdentifier) : ee
-  const handlers = defaultRegister.handlers
-  if (!baseEE.backlog || !handlers) return
+export function registerDrain(agentIdentifier, group) {
```
metal-messiah, line 13: Each feature `inst` will call `registerDrain` to help the drain module predict what all needs draining later on

src/common/drain/drain.js
```diff
@@ -6,33 +6,66 @@
...
-      mapOwn(handlerRegistrationList, function (i, registration) {
-        // registration is an array of: [targetEE, eventHandler]
-        registration[0].on(eventType, registration[1])
-      })
+/**
+ * Drain buffered events out of the EE. Each feature **should** have its events bucket into a "group".  The feature should drain this group explicitly
+ * @param {string} agentIdentifier 
+ * @param {string} group 
+ */
+export function drain(agentIdentifier = "", featureName = 'feature') {
```
metal-messiah, line 29: each feature `agg` will call drain().  But for race-condition purposes as well as preserving the logic chain with prev builds... it will wait until all the features that were imported are ready to drain.  It uses `registerDrain` which was called in the `inst` file to know what to expect at this stage. If for various reasons, there isnt a matching feature in the `registry`, it will just drain immediately.  This happens in tests, as well as in the MFE loader which bypasses inst files and directly imports agg files.

src/common/drain/drain.js
```diff
@@ -6,33 +6,66 @@
...
+  curateRegistry(agentIdentifier)
+  // if its not in the registry, that means the instrument file was bypassed.  This could happen in tests, or loaders that directly import the agg
+  if (!agentIdentifier || !registry[agentIdentifier].get(featureName)) return drainGroup(featureName)
+  registry[agentIdentifier].get(featureName).staged = true
+
+  const items = Array.from(registry[agentIdentifier])
+  if (items.every(([key, values]) => values.staged)) {
+    items.sort((a, b) => a[1].priority - b[1].priority)
```
metal-messiah, line 37: Its unclear whether the order of features drained has consequences or not, so manually enforcing this for now to match legacy build order.

src/common/drain/drain.js
```diff
@@ -6,33 +6,66 @@
...
+          // registration is an array of: [targetEE, eventHandler]
+          registration[0].on(eventType, registration[1])
+        })
+      })
+    }
+
+    delete handlers[group]
+    // Keep the group as a property so we know it was created and drained
+    baseEE.backlog[group] = null
+    baseEE.emit('drain-' + group, [])
```
metal-messiah, line 67: One nice thing here is that we can use an event to ensure that features do not start trying to harvest before they are appropriately drained.  (nearly) All features now subscribe to this drain-${name} event before starting their harvest cycle to prevent race conditions starting a harvest without drain having been called.

src/features/ajax/aggregate/index.js
```diff
@@ -59,7 +59,7 @@ export class Aggregate extends AggregateBase {
         getPayload: prepareHarvest
       }, this)
 
-      scheduler.startTimer(harvestTimeSeconds)
+      ee.on(`drain-${this.featureName}`, () => scheduler.startTimer(harvestTimeSeconds))
```
metal-messiah, line 62: example of waiting for drain to succeed before starting a harvest timer

src/features/utils/instrument-base.js
```diff
@@ -9,6 +10,8 @@ export class InstrumentBase extends FeatureBase {
       })
       this.hasAggregator = false
       this.auto = auto
+
+      if (auto) registerDrain(agentIdentifier, featureName)
```
metal-messiah, line 14: Auto is used in tests and other places to inform the class to/not to import the aggregator.  It is also used to inform the class whether a drain should be registered (to be fulfilled by the `agg` class)

src/loaders/configure/configure.js
```diff
@@ -25,7 +25,7 @@ export function configure(agentIdentifier, opts = {}, loaderType) {
     setRuntime(agentIdentifier, runtime)
 
 
-    setAPI(agentIdentifier, api)
+    setAPI(agentIdentifier, api, forceDrain)
```
metal-messiah, line 28: Some browser tests need the API to drain without loading any features.  This forces it to drain regardless of other things in the drain "registry"

tests/browser/utils/setup.js
```diff
@@ -1,10 +1,8 @@
-const { gosCDN } = require('../../../src/common/window/nreum')
 const { ee } = require('../../../src/common/event-emitter/contextual-ee')
 const {Aggregator} = require('../../../src/common/aggregate/aggregator')
 const { configure } = require('../../../src/loaders/configure/configure')
-
 export function setup(agentIdentifier = (Math.random() + 1).toString(36).substring(7)) {
-    const nr = configure(agentIdentifier)
+    const nr = configure(agentIdentifier, {}, 'browser-test', true)
```
metal-messiah, line 5: forceDrain on API buffers for browser tests

patrickhousley (commented):
We have one outstanding question around when the aggregators should be injected into the page. With this PR, all aggregators that do not require the page load event (only the pvt feature requires page load event) would get injected by webpack as soon as possible. If the pvt feature is present, harvesting will wait until the page load event happens. Otherwise, harvesting will be immediately started.

The concern is that we may negatively affect page performance by injecting our aggregators into the page before the page load event, more so if the pvt feature is not present because we would also start the harvesting logic.

That said, our aggregator chunks are really small and the logic in them is pretty simple. The question is, if we do inject and harvest before page load event, does it really make an impact on the page performance? And, should we wait until page load event to have webpack inject the aggregator chunks? The possible negative to this is, if a customer navigates away before the page load event happens, none of that data will be harvested and sent to NR1.

metal-messiah:
> We have one outstanding question around when the aggregators should be injected into the page. With this PR, all aggregators that do not require the page load event (only the pvt feature requires page load event) would get injected by webpack as soon as possible. If the pvt feature is present, harvesting will wait until the page load event happens. Otherwise, harvesting will be immediately started.
> 
> The concern is that we may negatively affect page performance by injecting our aggregators into the page before the page load event, more so if the pvt feature is not present because we would also start the harvesting logic.
> 
> That said, our aggregator chunks are really small and the logic in them is pretty simple. The question is, if we do inject and harvest before page load event, does it really make an impact on the page performance? And, should we wait until page load event to have webpack inject the aggregator chunks? The possible negative to this is, if a customer navigates away before the page load event happens, none of that data will be harvested and sent to NR1.

A decision was made to only lazy-load the agg modules after the load event has fired.  This matches the behavior of previous agent builds

metal-messiah:
Screen Shot 2023-01-24 at 8 51 19 PM

patrickhousley (approved):

src/loaders/features/enabled-features.js
```diff
@@ -1,20 +1,16 @@
...
-  "ajax",
-  "jserrors",
-  "metrics",
-  "page_action",
-  "page_view_event",
-  "page_view_timing",
-  "session_trace",
-  "spa"
-]
+const featureNames = [...asyncFeatures, ...syncFeatures]
```
cwli24, line 4: I don't get the value of separating them.

src/loaders/features/enabled-features.js
```diff
@@ -1,20 +1,16 @@
...
 
 function isEnabled(name, agentIdentifier) {
   return getRuntime(agentIdentifier).disabled !== true && getConfigurationValue(agentIdentifier, `${name}.enabled`) !== false
```
cwli24, line 7: I see this in places but it may be a relic. I don't think the runtime obj's `disabled` property ever gets set anywhere.

src/loaders/features/enabled-features.js
```diff
@@ -1,20 +1,16 @@
...
 }
 
+export function isAuto(name, agentIdentifier) {
```
cwli24, line 10: This doesn't seem be used anywhere.

cwli24 (changes requested):
Ongoing

src/loaders/agent.js
```diff
@@ -0,0 +1,56 @@
...
+            runtime: getRuntime(this.agentIdentifier)
+        }
+    }
+
+    start() {
+        try {
+            const enabledFeatures = getEnabledFeatures(this.agentIdentifier)
+            this.desiredFeatures.forEach(f => {
+                if (enabledFeatures[f.featureName]) {
+                    const dependencies = getFeatureDependencyNames(f.featureName)
```
cwli24, line 41: Just a 👍 on this. It's a cleaner iterative-style rework of the former system. May break if the `featurePriority` order that's tied to the dependency tree changes, but for now, this is satisfactory.

src/common/drain/drain.js
```diff
@@ -0,0 +1,83 @@
...
+
+import { ee } from '../event-emitter/contextual-ee'
+import { mapOwn } from '../util/map-own'
+import { registerHandler as defaultRegister } from '../event-emitter/register-handler'
+import { featurePriority } from '../../loaders/features/features'
+
+const registry = {}
+
+export function registerDrain(agentIdentifier, group) {
+  const item = { staged: false, priority: featurePriority[group] || 0 }
```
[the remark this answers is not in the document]
cwli24: I can also see how in the future there may be a potential bug with a (new) feature not calling registerDrain and hitting the "if not in registry, just drain" logic below. Something like that would be nice to highlight -- that all features must call this function.

src/common/drain/drain.js
cwli24: Most likely not, since they get buffered into EE then sent out on payloads that DEM don't mandate an ordering or statefulness to. But meh.

src/common/drain/drain.js
cwli24: Were features historically gated from harvesting until after RUM because of the other flags? Is this an old system? Are we able to just drain & harvest when features (and their dependecies) are ready, aside from session trace?

src/common/drain/drain.js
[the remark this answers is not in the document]
metal-messiah: The base instrument class calls registerDrain() automatically, so as long as a feature is extending instrument-base.js it should be handled.  Should add comment that "features should always use instrument-base.js` or something like that

src/common/event-emitter/contextual-ee.js
```diff
@@ -21,12 +23,19 @@ if (nr.ee) {
...
 
-function ee (old, debugId) {
+function ee(old, debugId) {
   var handlers = {}
   var bufferGroupMap = {}
   var emitters = {}
+  // backlog must be isolated if multiple agents can run on page. MFE loader packages sets this to true. only check this for feature emitters (length = 16)
+  var isolatedBacklog = false 
+  try{
+    isolatedBacklog = debugId.length !== 16 ? false : getRuntime(debugId).isolatedBacklog
```
cwli24, line 35: Somewhat of a magic number here. Is this corresponding to generateHexString for agentId?

src/loaders/features/enabled-features.js
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6480

src/loaders/features/enabled-features.js
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6480

src/common/event-emitter/contextual-ee.js
metal-messiah: yes, thats correct, checking if the EE's id is an agentIdentifier.  Some of our tests for instance do not namespace the EE with an agentID at all and just use the parent (which has no id)

metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6479

src/common/drain/drain.js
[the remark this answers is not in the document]
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6479

src/loaders/features/enabled-features.js
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6479

src/common/url/parse-url.js
```diff
@@ -25,13 +25,14 @@ export function parseUrl (url) {
...
+    urlEl.href = url
+  }
+  else {
     try {
-      urlEl = new URL(url, location.href);
-    } catch {
-      return ret; // IE doesn't support URL(), so for IE non-web env (e.g. workers), this func will break and return {}
+      urlEl = new URL(url, location.href)
+    } catch (err) {
+      return ret
```
cwli24, line 35: Why remove the comment? We may not support IE, but it doesn't hurt knowing customers still use it.

src/common/constants/environment-variables.js
```diff
@@ -0,0 +1,2 @@
+// process.env.BUILD_VERSION is replaced during webpack -or- npm build with environment vars supplied at build time
+export const VERSION = typeof process.env.BUILD_VERSION !== 'undefined' && process.env.BUILD_VERSION || ''
```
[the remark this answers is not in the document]
metal-messiah: This gets set in a helper that is used anywhere that babel is used (webpack, standalone babel scripts).  needs comments/ documentation
https://issues.newrelic.com/browse/NEWRELIC-6479

src/common/config/state/runtime.js
```diff
@@ -5,7 +5,7 @@ import { Configurable } from './configurable'
 import { gosNREUMInitializedAgents } from '../../window/nreum'
 import { getCurrentSessionIdOrMakeNew } from '../../window/session-storage'
 import { getConfigurationValue } from '../config'
-import globalScope from '../../util/global-scope';
+import { globalScope } from '../../util/global-scope';
```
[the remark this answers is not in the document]
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6480

src/loaders/configure/configure.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    setInfo(agentIdentifier, info)
+    setConfiguration(agentIdentifier, init || {})
+    setLoaderConfig(agentIdentifier, loader_config || {})
+    setRuntime(agentIdentifier, runtime)
+
+
+    setAPI(agentIdentifier, api, forceDrain)
+    gosNREUMInitializedAgents(agentIdentifier, nr, 'api')
```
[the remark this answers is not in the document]
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6478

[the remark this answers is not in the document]
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6479

src/common/url/parse-url.js
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-6479

src/common/util/feature-flags.js
cwli24: For commenting later, one here also to point out how PVE drain is separate from its agg module and why - thanks.

src/features/ajax/constants.js
```diff
@@ -0,0 +1,3 @@
+import { FEATURE_NAMES } from "../../loaders/features/features";
+
+export const FEATURE_NAME = FEATURE_NAMES.ajax
```
cwli24, line 3: BTW, I do think this model is onion-layering and confusing rather than just using FEATURE_NAMES.xyz directly in each aggregator file (that's already specific to each feature).

src/features/jserrors/aggregate/index.js
```diff
@@ -16,17 +16,18 @@ import { handle } from '../../../common/event-emitter/handle'
...
-import { FeatureBase } from '../../../common/util/feature-base'
+import { AggregateBase } from '../../utils/aggregate-base'
+import { FEATURE_NAME } from '../constants'
+import { drain } from '../../../common/drain/drain'
+import { FEATURE_NAMES } from '../../../loaders/features/features'
+import { onWindowLoad } from '../../../common/window/load'
 
-export class Aggregate extends FeatureBase {
+export class Aggregate extends AggregateBase {
+  static featureName = FEATURE_NAME
```
cwli24, line 28: This is redundant since you're already passing in `FEATURE_NAME` to the parent class who assigns it to `this.featureName`. (To access this, it'd be `Aggregate.featureName`.

cwli24: Issue exists in the other feature files too, btw.

src/features/ajax/constants.js
metal-messiah: imo we should try to isolate what is constant from those class files but these are subject and can be iterated on in the future for sure

src/features/jserrors/aggregate/index.js
metal-messiah: actually, could be a remnant from previous POC. needs validation

src/features/utils/feature-base.js
```diff
@@ -0,0 +1,48 @@
...
+import { gosCDN } from '../../common/window/nreum'
+
+export class FeatureBase {
+  constructor(agentIdentifier, aggregator, featureName) {
+    /** @type {string} */
+    this.agentIdentifier = agentIdentifier
+    /** @type {Aggregator} */
+    this.aggregator = aggregator
+    /** @type {ContextualEE} */
+    this.ee = ee.get(agentIdentifier, getRuntime(this.agentIdentifier).isolatedBacklog)
```
cwli24, line 13: (Btw, ee.get doesn't take a 2nd arg.. at least not yet.)

src/features/utils/aggregate-base.js
```diff
@@ -0,0 +1,7 @@
+import { FeatureBase } from "./feature-base";
+
+export class AggregateBase extends FeatureBase {
```
cwli24, line 3: Minor gripe about this serving no purpose.

metal-messiah: More of a pattern to follow for future modifications if necessary

src/features/utils/feature-base.js
```diff
@@ -0,0 +1,48 @@
...
+     */
+    this.blocked = false
+
+    this.checkConfiguration()
+  }
+
+  checkConfiguration() {
+    // try to support backwards compat with patter of some loader configs present after loader
+    // would need the same jsAttributes handling as the main branch
+    if (!isConfigured(this.agentIdentifier)) {
```
cwli24, line 29: I feel this issue/solution is in the wrong place, as it has nothing to do with _every_ feature. Better placed in the central import entry, subscribing to the window load event.

src/features/utils/instrument-base.js
```diff
@@ -0,0 +1,37 @@
...
+  constructor(agentIdentifier, aggregator, featureName, auto = true) {
+    super(agentIdentifier, aggregator, featureName)
+    this.completed = new Promise((resolve, reject) => {
+      this.resolve = resolve
+      this.reject = reject
+    })
+    this.hasAggregator = false
+    this.auto = auto
+
+    if (auto && featureName !== FEATURE_NAMES.sessionTrace) registerDrain(agentIdentifier, featureName)
```
cwli24, line 17: Since you have logic in drain to let things pass through the gate if they're a) tests, or b) "loaders that directly import [features'] aggregators",

1. the default for `auto` is false for pageaction (only at this time). Wouldn't this mean pageaction also fails to register itself and isn't harvest gated?
2. this seems to be the centralized (good) and only place `registerDrain` is called for all the features. That means `sessionTrace` would never register itself either.
3. given the exceptions to the drain-gating logic, can we be confident that in no other scenario, `auto` would be false even if instrumentation files were imported? Want to make sure gating is the **default** behavior in all other scenarios that's not (a) or (b).

src/features/utils/instrument-base.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  importAggregator() {
+    try {
+      if (this.hasAggregator || !this.auto) return
```
cwli24, line 22: Is `auto` for features being set somewhere else? Reading this and in relation to the previous above comment, I would've expected page actions to not import due to this short circuit. But your tests are passing, so I'm confused.

src/features/utils/feature-base.js
cwli24: This is also being ran by every feature's instrument **and** aggregate per.

src/loaders/api/api.js
```diff
@@ -3,15 +3,36 @@
...
+    'setErrorHandler', 'finished', 'addToTrace', 'inlineHit', 'addRelease',
+    'addPageAction', 'setCurrentRouteName', 'setPageViewName', 'setCustomAttribute',
+    'interaction', 'noticeError'
+  ]
+  funcs.forEach(f => {
+    nr[f] = (...args) => caller(f, ...args)
+  })
+
+  function caller(fnName, ...args) {
+    Object.values(nr.initializedAgents).forEach(val => {
```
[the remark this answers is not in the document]
cwli24: (in the case that the `!info` block is not executed)

src/features/jserrors/aggregate/index.js
cwli24: Actually I found the use on reviewing another file below.. I believe when you initialize a new `Agent` class and pass in all the feature **instrument** files, the ensuing dependencies & priority logic uses the static `.featureName` -- I think. Def could be combined though for sure.

src/loaders/features/featureDependencies.js
```diff
@@ -0,0 +1,26 @@
...
+
+export function getFrozenAttributes(feature) {
+    switch (feature) {
+        // right now, jserrors is the only feature that can have "on" or "off" page-level auto-instrumentation...
+        // page_action is always "off" (no instr)
+        // as new API/manual implementation methods are added, this list can likely be pruned
+        case FEATURE_NAMES.jserrors:
+            return []
+        default:
+            return ['auto']
```
cwli24, line 24: In relation to the issue I pointed out in _instrument-base.js_ about `auto` property, this is what concerns me: that customers can redefine it as off in the `init` block and then they no longer follow the drain-gating logic.

src/loaders/features/features.js
```diff
@@ -0,0 +1,37 @@
...
+    pageViewTiming: "page_view_timing",
+    sessionTrace: "session_trace",
+    spa: "spa"
+}
+
+export const asyncFeatures = [
+    FEATURE_NAMES.metrics
+]
+
+export const syncFeatures = [
```
cwli24, line 16: Probably already covered in a previous comment on commenting & docs, but doc here would be nice, as this is confusing to new folks who would think that these features actually operate in sync (block UI thread for long periods, don't use `setTimeout` or async functions, etc.)

tests/assets/js-error-attribute-precedence.html
```diff
@@ -11,11 +11,12 @@
 
     <script>
       newrelic.setCustomAttribute('localKey', 1)
+      setTimeout(function() {
       try {
         throw new Error("My error")
       } catch (e) {
         NREUM.noticeError(e)
-      }
+      }},100)
```
cwli24, line 19: Q: Just to double check, is there an edge case of our `noticeError` API not working on tick 0 of the event loop or something?

tools/jil/util/browser-list.js
```diff
@@ -148,7 +148,7 @@ function byVersion(left, right) {
 function cleanVersion(version) {
   // assign to high number, so that it is high in the list when sorted (i.e. beta is highest)
   if (!version || latestVersStringRe.test(version)) {
-    let prevVersionOffset = version?.split('-')[1];
+    let prevVersionOffset = !!version && version.split('-')[1];
```
cwli24, line 151: Q: Out of curiosity, what's wrong with the optional chaining? It's the same thing and works in Node 14.

cwli24 (changes requested):
👏 Great work. *standing ovation*

tools/jil/util/browser-list.js
metal-messiah: Jil was throwing errors with it for some reason at some point in the months-long POC 😓

tests/assets/js-error-attribute-precedence.html
metal-messiah: lets validate this

metal-messiah: it **should** work

src/loaders/features/featureDependencies.js
metal-messiah: thats a different auto.  This is a remnant that can be removed

src/loaders/api/api.js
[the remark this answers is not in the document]
metal-messiah: remnant from earlier POC.  `setTopLevelCallers()` needs to force NR to be top-level directly using `gosNREUM()`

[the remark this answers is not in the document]
metal-messiah: this is only relevant for MFE builds, but should be addressed in subsequent ticket

src/features/utils/instrument-base.js
metal-messiah: auto is a constructor arg, basically only used by tests at this point
```
  constructor(agentIdentifier, aggregator, featureName, auto = true) { ... }
  ```

File: src/features/utils/instrument-base.js (hunk 47 above)
Comment (metal-messiah) [reply src/features/utils/instrument-base.js:17]: the auto in the configs is a remnant from when we were trying to mash the MFE loader and regular loader into one thing and can be removed.  The "auto" used here is just a constructor argument used by tests to disable "auto" importing the agg which has side-effects that are obstructive in some tests.

File: src/features/utils/feature-base.js (hunk 46 above)
Comment (metal-messiah) [reply src/features/utils/feature-base.js:29]: the final check needs to happen at agg load time to support configs below the loader block.  Fine to move it to aggregateBase

File: src/features/utils/feature-base.js (hunk 44 above)
Comment (metal-messiah) [reply src/features/utils/feature-base.js:13]: remnant, can be removed in future ticket

Comment (metal-messiah) [reply src/features/utils/feature-base.js:13]: https://issues.newrelic.com/browse/NEWRELIC-6480

File: src/features/utils/instrument-base.js (hunk 47 above)
Comment (metal-messiah) [reply src/features/utils/instrument-base.js:17]: https://issues.newrelic.com/browse/NEWRELIC-6480

File: src/loaders/api/api.js (hunk 49 above)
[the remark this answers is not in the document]
Comment (metal-messiah) [reply src/loaders/api/api.js:27]: https://issues.newrelic.com/browse/NEWRELIC-6480

File: src/features/utils/feature-base.js (hunk 46 above)
Comment (metal-messiah) [reply src/features/utils/feature-base.js:29]: https://issues.newrelic.com/browse/NEWRELIC-6480

File: src/loaders/features/featureDependencies.js (hunk 50 above)
Comment (metal-messiah) [reply src/loaders/features/featureDependencies.js:24]: https://issues.newrelic.com/browse/NEWRELIC-6480

File: src/loaders/features/features.js (hunk 51 above)
Comment (metal-messiah) [reply src/loaders/features/features.js:16]: https://issues.newrelic.com/browse/NEWRELIC-6479

File: src/common/config/state/configurable.js [hunk 54]
```diff
@@ -0,0 +1,22 @@
...
+      const frozenAttrs = getFrozenAttributes(key)
+      if (frozenAttrs.length && value && typeof value === 'object'){
+        frozenAttrs.forEach(attr => {
+          if (attr in value) {
+            console.warn(`New Relic: "${attr}" is a protected attribute and can not be changed in feature ${key}.  It will have no effect.`)
+            delete value[attr]
+          }
+        })
+      }
+      if (!!value && typeof this[key] === 'object' && typeof value === 'object') this[key] = {...this[key], ...value}
```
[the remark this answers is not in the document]
metal-messiah: validate if this is needed

tests/assets/js-error-attribute-precedence.html
metal-messiah: validated, appears to work as intended

cwli24 (approved):

src/common/util/feature-flags.js
metal-messiah: [_https://issues.newrelic.com/browse/NEWRELIC-6479_](https://issues.newrelic.com/browse/NEWRELIC-6479)

PR #361, opened 2023-01-10
Adding CI workflow and miscellaneous maintenance

patrickhousley:
### Overview

Moving the linting and unit testing out of the current integration tests workflow to a new CI workflow. The new workflow will run on pull request opens, reopens, and syncs. It will also run on the main branch on push.

The new CI will execute the project linting, jest testing, and jil unit tests against a local selenium server for chrome and firefox. The workflow does not rely on any secrets and can be safely ran on PRs from forks.

Additional Changes:
- Updated Jest and jest configuration after project restructure
- Implemented prettier and reformatted all code
- Updated eslint and refreshed the configuration to use eslint default settings, standard js rules, and prettier rules
  - Rules that were failing have been turned off for now so there are no code changes in this PR
- Removed all remnants of typescript dependencies

### Related Issue(s)

Closes #232

### Testing

Let the PR check run.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: lts/*
+      - name: Install Dependencies
+        run: npm ci
+      - name: Build Agent
+        run: npm run build:all
+      - name: Run JIL Unit Tests
+        run: node ./tools/jil/bin/cli.js -f merged -b chrome@latest/linux -t 85000 --no-retry --selenium-server=selenium:4444 --unit-only
```
metal-messiah, line 75: do our concurrency rules work for selenium-server runs?

patrickhousley: No, not without a lot more work. Using selenium standalone it is recommended to only use one browser per container. If we want to load more than one, it is recommended to use a full selenium setup.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,104 @@
...
```
metal-messiah, line 77: `--no-retry` can sometimes be problematic when running the full list of tests

metal-messiah: That usually only applies to func tests though, so this should be ok since its just the browser tests

metal-messiah:
I know this is just introducing prettier with defaults, but this may actually be the best time to agree as a team on some config decisions so we don't have to format the whole code base twice if we change a prettier config.

metal-messiah:
> I know this is just introducing prettier with defaults, but this may actually be the best time to agree as a team on some config decisions so we don't have to format the whole code base twice if we change a prettier config.
> 
> For instance, i'm not a big fan of prettier's printWidth default. It breaks up functions into new lines too often. I would propose setting that to like, `120` or something larger than `80`
> 
> Lets make a decision as a group on the common settings like `semi`, `singleQuote`, `printWidth`, `trailingComma`

sounds like we can go with the defaults except for `printWidth` which lets just set to `120`.

CHANGELOG.md
```diff
@@ -6,103 +6,130 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 ## v1224
 
 ### Support SPA, XHR, and session trace features on Chrome for iOS.
```
bjfield, line 8:
```suggestion
### Support SPA, XHR, and session trace features on Chrome for iOS
```

patrickhousley: I don't want to make changes outside of what prettier is doing automatically in this PR.

.prettierrc
```diff
@@ -1,3 +1,3 @@
 {
-  "trailingComma": "es5"
+  "printWidth": 120
```
bjfield, line 2:
```suggestion
  "printWidth": 120,
  "singleQuote": true
```

metal-messiah (approved):

cwli24 (approved):

PR #363, opened 2023-01-11
[Patch] Distinguish reporting of windowUnload from pageHide to NR

cwli24:
### Overview

Under the previous PR, our `pageHide` and `unload` had become synonymous in that doc vis change would cause both to be triggered and sent. This was not the intention, as there is a distinct difference between their triggers.
image
Specifically, `pageHide` can be emitted multiple times in the lifespan of one page load/restore, but `unload` should only signify the end of a load/restore's span and occur once per.

**`subscribeToEOL` also had its limiter removed under the flag to, e.g., allow multiple final harvests as either event can cause one and the total number of "final harvest" is indeterministic. (It's also necessary to decouple the two events.)

### Related Issue(s)

Patch to work done in #308 

### Testing

When `init.allow_bfcache` is enabled, document vis change (tab, minimize, etc.) should emit `pageHide` still but _not_ `unload`. Only when the window pageHide event fires (navigating away, reload, etc.) should emit `unload` (**and** `pageHide`).

### DIFF
--- a/packages/browser-agent-core/src/features/page-view-timing/aggregate/index.js
+++ b/packages/browser-agent-core/src/features/page-view-timing/aggregate/index.js
@@ -26,7 +26,7 @@ export class Aggregate extends FeatureBase {
     this.clsSupported = false
     this.cls = 0
     this.clsSession = {value: 0, firstEntryTime: 0, lastEntryTime: 0}
-    this.userSessionEnded = false
+    this.curSessEndRecorded = false
 
     try {
       this.clsSupported = PerformanceObserver.supportedEntryTypes.includes('layout-shift')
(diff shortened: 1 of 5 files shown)


packages/browser-agent-core/src/features/page-view-timing/aggregate/index.js
```diff
@@ -122,28 +123,23 @@ export class Aggregate extends FeatureBase {
...
-   * Add the time of _document visibilitychange to hidden_ to the next PVT harvest.
+   * Add the time of _document visibilitychange to hidden_ to the next PVT harvest == NRDB pageHide attr.
    * @param {number} timestamp
    */
-  updateSessionEnd(timestamp) {
-    if (!this.userSessionEnded) { // TO DO: stage 2 - we don't want to capture this timing twice on page navigating away, but it should run again if we return to page and away *again*
-      this.addTiming('pageHide', timestamp, null, true)
-      this.userSessionEnded = true
+  endCurrentSession(timestamp) {
+    if (!this.curSessEndRecorded) { // TO DO: stage 2 - we don't want to capture this timing twice on page navigating away, but it should run again if we return to page and away *again*
```
metal-messiah, line 130: For Phase 2... Does this become something like... page is restored from cache or pageShow event, set curSessEndRecorded = false again?

cwli24: yeah, along with whatever else. CWV reset their internal metrics on pageshow.persisted

patrickhousley (approved):

cwli24:
Note: in reference to the diagram, the reason why window pageHide / recordPageUnload still causes our `pageHide` to be emitted as a byproduct is due to the timing in which these two event fires on page unloading.
1. window pagehide fires, then;  <-- this is also when analytics is sent
2. doc visibilitystate -> hidden
If our window pagehide listener doesn't _also assume and record_ `pageHide` aka vis change, then we'll miss sending it in the last outgoing analytics payload.

cwli24:
Include an "experimental" tag on the changelog for bfcache feature flag

metal-messiah (approved):

PR #367, opened 2023-01-13
NR-78867: Safe sessionStorage writes

patrickhousley:
### Overview

Wrapping our access of and write to sessionStorage in a `try..catch` to ensure it does not cause the agent to crash when sessionStorage is not available or when the quota has been exceeded or set to 0. Safari has been know to set the sessionStorage quota to 0 in private browsing windows. This was causing the agent to crash and the API to not be attached to the window.newrelic object.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-78867
https://issues.newrelic.com/browse/NR-78872

### Testing

A functional test has been added to ensure we no longer have this issue. To reproduce the error, comment out the `try..catch` wrapper added in this PR, rebuild, and load the new test page. After loading, check the window.newrelic object to see if the API methods were added.

### DIFF
--- a/packages/browser-agent-core/src/common/window/session-storage.js
+++ b/packages/browser-agent-core/src/common/window/session-storage.js
@@ -10,22 +10,30 @@ import { isBrowserScope } from '../util/global-scope'
 
 export { getCurrentSessionIdOrMakeNew };
 
-const ss = isBrowserScope ? window.sessionStorage : undefined;
 const SESS_ID = "NRBA_SESSION_ID"; // prevents potential key collisions in session storage
 
 /**
  * @returns {string} This tab|window's session identifier, or a new ID if not found in storage
  */
 function getCurrentSessionIdOrMakeNew() {
-  if (!ss) return null; // indicates sessionStoage API not avail
-
-  let sessionId;
-  if ((sessionId = ss.getItem(SESS_ID)) === null) {
-    // Generate a new one if storage is empty (no previous ID was created or currently exists)
-    sessionId = generateRandomHexString(16);
-    ss.setItem(SESS_ID, sessionId);
+  if (!isBrowserScope) {
+    return null;
   }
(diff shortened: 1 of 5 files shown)


metal-messiah (approved):

packages/browser-agent-core/src/common/window/session-storage.js
```diff
@@ -19,13 +19,20 @@ const SESS_ID = "NRBA_SESSION_ID"; // prevents potential key collisions in sessi
 function getCurrentSessionIdOrMakeNew() {
   if (!ss) return null; // indicates sessionStoage API not avail
 
-  let sessionId;
-  if ((sessionId = ss.getItem(SESS_ID)) === null) {
-    // Generate a new one if storage is empty (no previous ID was created or currently exists)
-    sessionId = generateRandomHexString(16);
-    ss.setItem(SESS_ID, sessionId);
+  try {
```
cwli24, line 22: Per huddle, the problematic line that triggers the DOM exception is actually above where ss is being defined as a side effect:
image

My suggestion is to wrap that instead and leave it as undefined if error thrown (as the getCurrent... function is already setup to handle a null `ss`) like:
```javascript
let ss;
try {
  if (isBrowserScope) ss = window.sessionStorage;
  // else, leave it undefined
} catch (err) {
  // exception thrown, then also leave it undefined
}
```

cwli24 (changes requested):

metal-messiah:
I used this verbiage in the release-notes:
```
### Fix check for sessionStorage object
Ensure the agent does not crash when sessionStorage is not available or when the quota has been exceeded or set to 0. Safari has been known to set the sessionStorage quota to 0 in private browsing windows.
```

Do you want to add that to the CHANGELOG.md file as well?

cwli24 (approved):

PR #368, opened 2023-01-13
use setToken response flags to kill instantiated features

metal-messiah:
### Overview
This PR uses the flags returned from the RUM call to disable in-progress features.  If the RUM reponse is 0, the feature will be "blocked" from aggregating and harvesting.  The feature will run automatically if enabled, even if RUM call is not made.  a 0 RUM response flag will block an inprogress feature, but unlike old builds of the agent, will not be a criteria for starting the feature. A token will return as "0" if ingest determines the account to lack entitlements to access the relevant feature endpoint.

### Links
This is a follow up and revision to PR #362 

### Testing
A new test has been added which mocks the RUM response and forces flags to be `0`.  The corresponding feature should not report if `0`


packages/browser-agent-core/src/features/ajax/aggregate/index.js
```diff
@@ -80,6 +79,7 @@ export class Aggregate extends FeatureBase {
 
       handle('bstXhrAgg', ['xhr', hash, params, metrics], undefined, undefined, ee)
 
+      if (this.blocked) return
```
cwli24, line 82: 1. If you're trying to read _jserrors aggregate_'s `blocked`, I don't think this will happen because jserroragg.blocked isn't a property on the inherited class.
2. I think there's more code below this that isn't specifically tied to sending 'xhr' payloads to /jserror endpoint, such as the SPA stuff. If that's true, this ought to be wrapped around pertinent code.

packages/browser-agent-core/src/common/util/feature-flags.js
```diff
@@ -10,6 +10,7 @@ export function activateFeatures (flags, agentIdentifier) {
   var sharedEE = ee.get(agentIdentifier)
   if (!(flags && typeof flags === 'object')) return
   mapOwn(flags, function (flag, val) {
+    if (!val) return sharedEE.emit('block-' + flag, [])
     if (!val || activatedFeatures[flag]) return
     sharedEE.emit('feat-' + flag, [])
```
cwli24, line 15: So as to not introduce redundancy, could you remove the `feat-` that's leftover for STN, and inverse the logic in STN to correspond to `block-stn` _not_ being sent instead?

cwli24 (changes requested):

packages/browser-agent-core/src/common/util/feature-flags.js
cwli24: nvm; ok with having the two "systems"

patrickhousley (approved):

cwli24 (approved):

PR #371, opened 2023-01-17
New INP & long task timings NEWRELIC-2439

cwli24:
### Overview
Two new types of timing is added to PageViewTiming feature to bolster existing web vitals and detect areas causing slow user responsiveness. Ingest action is required before they can be queried through NR dashboard and UI.

### Related Issue
Feature tickets: NEWRELIC-5157, NEWRELIC-5158
Public doc changes staged in: https://github.com/newrelic/docs-website/compare/develop...cwli24:docs-website:browser-1223-new-timings

### Testing
- New test added to pvt/timings.test.js for long tasks
- Existing test modified in pvt/timings.test.js for INP

#### Other side effects
- Defunct matchers and tests cleaned up
- CWV v3 now used via npm


cwli24:
First test results: [✔️](https://github.com/newrelic/newrelic-browser-agent/actions/runs/3943274099/jobs/6747923285)
Pushing rectified problem with an existing test assertion in timings.test.js

metal-messiah (commented):
Nice job on this PR!

packages/browser-agent-core/src/features/page-view-timing/instrument/long-tasks.js
```diff
@@ -0,0 +1,51 @@
+import { subscribeToEOL } from "@newrelic/browser-agent-core/src/common/unload/eol";
+
+export const onLT = (onReport) => {
```
metal-messiah, line 3: This could really benefit from jsdoc, as all these comments could come through with intellisense

patrickhousley: Can we rename this function to `onLongTask`

packages/browser-agent-core/src/features/page-view-timing/instrument/long-tasks.js
```diff
@@ -0,0 +1,51 @@
...
+					ltCtrId : entry.attribution[0].containerId,			// MDN: container's 'id' attribute
+					ltCtrName: entry.attribution[0].containerName		// MDN: container's 'name' attribute
+				});
+			}
+
+			onReport(metric);	// report every long task observed unconditionally
+		});
+	};
+
+	let po;
```
patrickhousley, line 27: Can this be renamed to `performanceObserver`? Abbreviated names make the code harder to read and there is no down side to using a longer, more descriptive name for variables, functions, etc.

tests/assets/js/some-long-task.js
```diff
@@ -0,0 +1,5 @@
+function someLongTask() {
+	const startTime = Date.now();
+	while (Date.now() - startTime < 60);  // long tasks are those >50ms which blocks main thread
```
patrickhousley, line 3: Is this definition of a long task the same for all browsers?

packages/browser-agent-core/src/features/page-view-timing/instrument/index.js
```diff
@@ -52,6 +54,16 @@ export class Instrument extends FeatureBase {
       documentAddEventListener(e, (...args) => this.captureInteraction(...args));
     })
 
+    /** Interaction-to-Next-Paint */
+    onINP(({name, value}) => {
```
manishco, line 58: Not sure if there is scope for change but as developer who is monitoring web performance we would love to know what element caused the INP issue and in some cases the event type. 

This information is available in the attribution API and wondering if the event target could be included. See docs: https://github.com/GoogleChrome/web-vitals#inp-attribution

e.g.
```
import { onINP } from 'web-vitals/attribution'; // <-- need to change import to get this data

onINP(({name, value, attribution }) => {
   // attribution.eventTarget the element interacted with
  // attribution.eventType I found this less useful but maybe others might find it useful 
})
```

tests/assets/js/some-long-task.js
manishco: This https://inp-demo.glitch.me/ is a good demo to play around with INP and they used the same approach. I believe it should be good enough.

packages/browser-agent-core/src/features/page-view-timing/instrument/index.js
cwli24: While I wholly agree with the reasoning, the application experience team is not quite ready to consume that data just yet and discussions to be had, but it's good to think about it in advance while designing the UI for displaying INP & other timings going forward.

Please consider creating a feature request on [New Relic JIRA board](https://issues.newrelic.com/projects/NEWRELIC) to the `Observability -> Application Experience Browser` team  _or_ through the feedback form in NR1 web UI: 
image
This would help the teams drive better feature improvements and user experience for all. Thank you!

tests/assets/js/some-long-task.js
cwli24: @patrickhousley yes, it's in the official definition of this timing API.

packages/browser-agent-core/src/features/page-view-timing/instrument/index.js
manishco: Thanks @cwli24

metal-messiah (commented):
I think everything makes sense here.  Barring the rebase i think this looks g2g.  Just want to confirm, pre-changes to dempla, does this new data just get dropped by ingest, or does ingest respond with any type of malformed request codes we should be aware of?

cwli24:
> I think everything makes sense here. Barring the rebase i think this looks g2g. Just want to confirm, pre-changes to dempla, does this new data just get dropped by ingest, or does ingest respond with any type of malformed request codes we should be aware of?

It indeed is just tossed into the bad bucket by consumer rather than transformed into timing nodes.

tools/jil/util/browser-matcher.js
```diff
@@ -238,6 +238,9 @@ features.addEventListenerOptions = new BrowserMatcher()
   .exclude('ios')
   .exclude('ie')
 
+features.supportsFirstInteraction = features.addEventListener
```
metal-messiah, line 241: Do we need this?

cwli24: Nope, I'll remove it. I think originally intended to be `firstInteractionDelay` matcher instead, but ended up unnecessary.. for now.

src/features/page_view_timing/instrument/long-tasks.js
```diff
@@ -0,0 +1,60 @@
...
+        })
+      }
+
+      onReport(metric)	// report every long task observed unconditionally
+    })
+  }
+
+  let observer
+  try {
+    if (PerformanceObserver.supportedEntryTypes.includes('longtask')) {
```
patrickhousley, line 38: Should we check if `PerformanceObserver` exists here? Thinking for IE11 support.

cwli24: Try block takes care of that; mirroring what core-vitals do so they don't bother with IE either (it's also excluded in PF tests).

patrickhousley: This feels like a code smell to me. We are using a `try..catch` instead of feature detection and the catch block is empty.

cwli24: I'd have to disagree there; it's pythonic in attempting to do something rather than waste time asking for permission. Both have merit, and I lean towards either depending on the situation. Ex., feature detect in metrics makes sense, and at the beginning of a whole module that assumes it, would be cleaner.

I think doing a check here would be tailored for IE, whereas this is expected to run just fine mostly, plus try-catching also includes any other exception, not just API support as a check does. Add to that it's consistent with PVT as a whole.

patrickhousley (approved):

PR #374, opened 2023-01-19
Bind navigator to sendBeacon to prevent illegal invocation errors on …

metal-messiah:
### Overview

Some browser versions will throw errors if sendBeacon does not have the navigator scope bound to it.  This is just a fail-safe action to try to support those browsers. See `Links` for more context 

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-4686

### Links 
[Issues with Navigator Context](https://xgwang.me/posts/you-may-not-know-beacon/#it-may-throw-error%2C-be-sure-to-catch)
[How next.js handled this](https://github.com/vercel/next.js/pull/26601/files)

### Testing

final-harvest.test.js tests the calling of sendBeacon on unload.  The behavior should be unchanged in our testing suite, and this ticket should just be treated as a "try our best to do-no-harm" effort and nothing more.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Fixed an issue where multiple custom interactions harvested at the same time wou
 ### Prevent ajax time slice metrics based on deny list
 Prevent time slice metric collection for ajax calls when such a call matches an entry in the ajax deny list.
 
-### Prevent ajax time slice metrics based on deny list
+### Bind navigator scope to sendBeacon
 Some browser versions will throw errors if sendBeacon does not have the navigator scope bound to it. A fail-safe action of binding the navigator scope to sendBeacon was added to try to support those browsers.
 
 ## v1223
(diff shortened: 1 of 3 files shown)


packages/browser-agent-core/src/common/util/submit-data.js
```diff
@@ -53,5 +53,8 @@ submitData.img = function img (url) {
  * Do NOT call this function outside of a guaranteed web window environment.
  */
 submitData.beacon = function (url, body) {
-  return window.navigator.sendBeacon(url, body)
+  // Navigator has to be bound to ensure it does not error in some browsers
+  // https://xgwang.me/posts/you-may-not-know-beacon/#it-may-throw-error%2C-be-sure-to-catch
+  const send = globalScope.navigator.sendBeacon.bind(navigator)
```
cwli24, line 58: (minor note: `navigator` cannot be reassigned & ought to be safe as is)
image

patrickhousley (approved):

packages/browser-agent-core/src/common/util/submit-data.js
cwli24: Update: actually can you just remove import/use of `globalScope`? That module is meant for APIs interchangeable between window and workers. In this case, the `navigator.sendBeacon` is only relevant to the window, so it should just be `window.navigator.sendBeacon.bind(navigator)`.

cwli24 (changes requested):

cwli24 (approved):

metal-messiah:
Changelog needed

PR #375, opened 2023-01-19
NR-80544: Initial WDIO implementation

patrickhousley:
### Overview

This is an initial implementation of WDIO within the project. It has a couple of example tests, starts the local testing server and supports both selenium and sauce. It also has a working example of how we might communicate between the test processes and the testing server using gRPC but this is not set in stone and might change.

When reviewing, keep in mind that nothing around how we run tests today should have changed. All JIL testing and CLI functionality should still work as it does today. The WDIO implementation has been isolated on purpose and may be renamed/moved later once we have a more fleshed out implementation.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-80544

### Testing

You are encouraged to run the existing JIL suite to make sure that has not been changed or broken. To run WDIO, run `npm run wdio`. Check `tools/wdio/args.mjs` for the supported CLI options.


package.json
```diff
@@ -44,9 +45,9 @@
     "cdn:watch:extension": "jung -r ./src -F '.*\\.test\\.js' --run -- npm run cdn:build:extension",
     "cdn:webpack": "npx webpack --progress --config ./webpack.config.js",
     "packages:bundle": "webpack -c tools/jil/webpack.modular.js",
-    "test-server": "node ./tools/jil/bin/server",
-    "sauce:connect": "node ./tools/jil/bin/sauce",
-    "sauce:get-browsers": "node ./tools/jil/util/sauce-browsers",
+    "test-server": "node ./tools/wdio/bin/server",
```
metal-messiah, line 48: Commenting this specifically for brevity other reviewers, running `test-server` will point at the wdio server by default and you will need to manually run `node tools/jil/...` to run the jil test server if needed

metal-messiah (approved):

package.json
patrickhousley: Oops I forgot to change that back to jil. It's the same server though, just through a different set of cli js files. Should work exactly the same. Thanks for catching that. Should I change it back?

bjfield: I see it's using test-server under the hood, but just for clarity for now, how about adding a script like "wdio-test-server" and restoring the old one?

PR #376, opened 2023-01-19
Adding testcases for BFC work NEWRELIC-5083

cwli24:
### Overview
Test that post-308, the agent indeed doesn't block pages from cacheing, and that pagehide/unload events are still emitted only once on initial load->unload page life.

#### Other minor effects
Page restore (from BF cache) metric changed to supportability from custom.

### Related Issue(s)

Addition to work done in #308 

### Testing

- `functional/pvt/bfcache.test.js` added to automation
- Refactored a couple of asset htmls


tools/testing-server/router.js
```diff
@@ -195,7 +195,7 @@ class RouterHandle {
 
   expectIns(appID) {
     // insights harvest at 30s
-    return this.expectBeaconRequest(this.beaconRequests.ins, 40000, appID)
+    return this.expectBeaconRequest(this.beaconRequests.ins, undefined, appID)
```
metal-messiah, line 198: Can you just pass in the timeout as an argument to the promise, instead of mutating the default timeout for the whole test, and changing the default behavior for all ins promises across the test suite?

```javascript
expectIns(appID, timeout) {
    // insights harvest at 30s
    return this.expectBeaconRequest(this.beaconRequests.ins, timeout || 40000, appID)
}
```

cwli24: Do note that based on my test, this is actually upper bounded by the `-t 32000` we set most tests to anyways. It's just that for some odd reason it's coded in such a way that if it's a hardcoded value rather than `undefined`, it won't respect the dynamic router timeout change.

metal-messiah (approved):

cwli24:
Rebased onto v1223 restructure

PR #378, opened 2023-01-20
Implement and verify support for Chrome on iOS (NEWRELIC-6276).

bjfield:
### Overview

This PR introduces support for Chrome for iOS by removing feature gates that previously restricted three features: XHR, SPA and STN (session trace). It also addresses one of the chromium bugs that led to the the feature gates to begin with.

Apart from removing feature gating, the operative change is to wrap the methods of `history` directly rather than wrapping the methods of that object's prototype. This was necessary because Chrome on iOS itself modifies `history` directly, thus ignoring our changes to the prototype.

For more discussion of the bugs that led to feature gating, as well as the rationale for the change in how we wrap history methods, see the comments on related JIRA tickets.

- Remove CriOS feature gate from `xhrWrappable` in `common/config/state/runtime.js`.
- Remove comment warning about CriOS and xhrWrappable in `features/spa/instrument/index.js`.
- Wrap `window.history` methods directly, not the prototype methods in `common/wrap/wrap-history.js`.
- Refactor `hasOwnProperty(...)` test for `window.history` methods in `browser/wrap-history.browser.js`.
- Add JSDoc comments to `common/wrap/wrap-history.js`.
- Update change log.

### Related Issue(s)

- [NEWRELIC-6276](https://issues.newrelic.com/browse/NEWRELIC-6276): Implement and verify support for Chrome on iOS.
- [NEWRELIC-6192](https://issues.newrelic.com/browse/NEWRELIC-6192): Validate status of bugs with Chrome on iOS.
- [Chrome Issue 471591](https://bugs.chromium.org/p/chromium/issues/detail?id=471591): Original bug that led to gating XHR.
- [Chrome Issue 783382](https://bugs.chromium.org/p/chromium/issues/detail?id=783382): Original bug that led to gating SPA, STN.

### Testing

After this PR's modifications, both JIL browser tests and a live implementation show the previously gated features (XHR, SPA, and STN) working properly on Chrome for iOS, with no consequences to existing browser support.

Because SauceLabs does not offer a browser-platform configuration for Chrome on iOS, testing must be semi- or fully-manual. It involves an iOS device connecting to JIL (for the suite of browser tests) as well as to an Angular server (for live testing), both running on a locally networked dev machine. The Angular server is the same one attached to PR 315.

Another form of testing is the inverse: ensuring other automated browser tests do not fail as a result of the changes. See the standard _Tests_ and _Polyfill Tests_ actions on this branch.
- Standard [Tests](https://github.com/newrelic/newrelic-browser-agent/actions/runs/3963303499)
- [Polyfill Tests](https://github.com/newrelic/newrelic-browser-agent/actions/runs/3964611717)

Please see the attached ticket and its predecessor for a more thorough discussion of testing decisions.

### DIFF
--- a/tests/browser/wrap-history.browser.js
+++ b/tests/browser/wrap-history.browser.js
@@ -23,7 +23,7 @@ jil.browserTest('history functions are wrapped', function (t) {
   t.end()
 })
 
-jil.browserTest('a new property is not shown in the history object', function (t) {
+jil.browserTest('two modified properties are shown on the history object', function (t) {
   // wrap
   const {setup} = require('./utils/setup')
   const {wrapHistory} = require('../../src/common/wrap/wrap-history')
@@ -31,17 +31,17 @@ jil.browserTest('a new property is not shown in the history object', function (t
   const {baseEE} = setup()
   wrapHistory(baseEE)
 
-  var proto = window.history.constructor && window.history.constructor.prototype
-  if (proto.pushState) {
-    t.equal(window.history.hasOwnProperty('pushState'), false)
-    t.equal(window.history.hasOwnProperty('replaceState'), false)
+
+  if (window.history && window.history.pushState) {
+    t.equal(window.history.hasOwnProperty('pushState'), true)
+    t.equal(window.history.hasOwnProperty('replaceState'), true)
   } else {
-    t.pass('not on constructor')
+    t.pass('no history functions')
   }
 
   t.end()
 })
-
+ 
 function isWrapped(fn) {
   return fn && (typeof fn['nr@original'] === 'function')
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## v1224
+
+### Support SPA, XHR, and session trace features on Chrome for iOS.
+Previously, the agent did not collect SPA browser interactions, XHR events, or session trace data in Chrome for iOS (which uses the webkit engine with modifications). The agent will now collect the same data in Chrome for iOS as in other supported browsers.
+
 ## v1223
 
 ### Refactor loader architecture for improved developer experience
--- a/src/common/config/state/runtime.js
+++ b/src/common/config/state/runtime.js
@@ -24,7 +24,7 @@ const model = agentId => { return {
   releaseIds: {},
   sessionId: getConfigurationValue(agentId, 'privacy.cookies_enabled') == true ?
     getCurrentSessionIdOrMakeNew() : null,  // if cookies (now session tracking) is turned off or can't get session ID, this is null
-  xhrWrappable: XHR && XHR_PROTO && XHR_PROTO['addEventListener'] && !/CriOS/.test(navigator.userAgent),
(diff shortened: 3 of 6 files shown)


packages/browser-agent-core/src/common/wrap/wrap-history.js
```diff
@@ -3,30 +3,45 @@
...
-    object = prototype
-  }
-
-  wrapFn.inPlace(object, [ 'pushState', 'replaceState' ], '-')
+  /**
+   * For objects that will be instantiated more than once, we wrap the object's prototype methods. The history object
+   * is instantiated only once, so we can wrap its methods directly--and we must wrap the history methods directly as
+   * long as [Chromium issue 783382](https://bugs.chromium.org/p/chromium/issues/detail?id=783382) remains unresolved.
+   */
+  wrapFn.inPlace(window.history, [ 'pushState', 'replaceState' ], '-')
```
patrickhousley, line 33: Maybe change this to `wrapFn.inPlace(globalScope.history`. `globalScope` should point to `window` here but the globalScope variable was created so we dont have window all over our code base.

patrickhousley (approved):
Just one opinion comment. Otherwise, looks good.

packages/browser-agent-core/src/common/wrap/wrap-history.js
cwli24: Just want to clarify that window was changed to `self` to accommodate workers without too much duplication/redundancy. And then `self` broke because of that ticket w/ customer redefining it. 
The point: keep `window` as `window` where relevant, and use `globalScope` when it's something shared with worker build/agent function. In this case, there's no `history` API in worker, so it doesn't quite make sense to ask `window or workerGlobalScope ? window.history : workerGlobalScope.history` @bjfield

packages/browser-agent-core/src/common/wrap/wrap-history.js
```diff
@@ -3,30 +3,45 @@
...
+/**
+ * Wrapper for pushState and replaceState methods of window.history object
+ */
+
+import { ee as globalEE } from '../event-emitter/contextual-ee'
+import { createWrapperWithEmitter as wfn } from './wrap-function'
 import { isBrowserScope } from '../util/global-scope'
 
-const wrapped = {}
+let alreadyWrapped = false
```
metal-messiah, line 14: this may have consequences in the niche scenario where multiple agents are on one page that share the same bundle, is there a reason this part was changed?  It used to have a namespace for each potential agent ID to separate that

tests/browser/wrap-history.browser.js
```diff
@@ -23,25 +23,6 @@ jil.browserTest('history functions are wrapped', function (t) {
   t.end()
 })
 
-jil.browserTest('a new property is not shown in the history object', function (t) {
```
metal-messiah, line 26: Instead of removing this test, can you just modify it to match the behavior you are expecting to see so that we can continue to validate this case for the future

bjfield: Done

packages/browser-agent-core/src/common/wrap/wrap-history.js
bjfield: I figured it was an unnecessary remnant. Makes sense that it would pertain to multi-agent scenarios.

cwli24 (approved):

PR #384, opened 2023-01-31
NR-74032: Prevent deny-list ajax time slice metrics

patrickhousley:
### Overview

Prevent time slice metrics collection for ajax calls that match an entry within the deny list.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-74032

### Testing

Added test cases for time slice metrics around ajax calls that should match the deny list.


tests/functional/stn/ajax.test.js
```diff
@@ -0,0 +1,114 @@
...
+    resourcePromise = router.expectResources()
+
+    return Promise.all([resourcePromise, clickPromise])
+  })
+  .then(([result]) => {
+    t.equal(result.res.statusCode, 200, 'server responded with 200')
+
+    const body = result.body
+    const harvestBody = JSON.parse(body).res
+    const loadNodes = harvestBody.filter(function (node) { return node.t === 'event' && node.n === 'load' || node.n === 'readystatechange' })
```
metal-messiah, line 47: Can you clarify what is being tested here?

Should you also check that nodes aren't created where n = `xmlhttprequest`, `Ajax`, `fetchStart`?

tests/functional/stn/ajax.test.js
```diff
@@ -0,0 +1,114 @@
...
+      page_view_timing: {
+        enabled: false
+      }
+    }
+  })
+
+  let loadPromise = browser.safeGet(assetURL)
+  let rumPromise = router.expectRum()
+  let resourcePromise = router.expectResources()
+  const ajaxPromise = router.expectBeaconRequest(
```
metal-messiah, line 80: can you use `expectXHRMetrics()` here so that we aren't crossing error payloads? (err, xhr, cm, sm all share that stream, so the above listener prevents those from polluting)

tests/functional/xhr/deny-list.test.js
```diff
@@ -0,0 +1,74 @@
...
+    scriptString: `
+      const xhr = new XMLHttpRequest();
+      xhr.open('GET', 'http://example.com/');
+      xhr.send();
+    `
+  }))
+
+  Promise.all([ajaxPromise, loadPromise, rumPromise])
+    .then(([response]) => {
+      const newAjaxPromise = router.expectBeaconRequest(
```
metal-messiah, line 24: same here, can we use `expectXhrMetrics()` for `xhr` in errors payloads to ensure other payloads don't come through?

tests/functional/xhr/deny-list.test.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+    },
+    scriptString: `
+      fetch('http://example.com').then(function() {}).catch(function() {})
+    `
+  }))
+
+  Promise.all([ajaxPromise, loadPromise, rumPromise])
+    .then(([response]) => {
+      const newAjaxPromise = router.expectBeaconRequest(
```
metal-messiah, line 59: here as well

metal-messiah:
Since this PR is unofficially fixing polyfills, would you mind updating the polyfills test to not run the "check failures" action?  it was giving false positives.

src/features/metrics/instrument/workers-helper.js
```diff
@@ -66,6 +66,10 @@ function insertSupportMetrics(report) {
      * @returns Proxy worker that intercepts the original constructor
      */
     function extendWorkerConstructor(origClass, workerType) {
+        if (typeof Proxy === 'undefined') {
```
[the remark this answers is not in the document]
patrickhousley: Correct.

tests/functional/stn/ajax.test.js
patrickhousley: I had to use `expectBeaconRequest` so I could set a custom timeout.

tests/functional/xhr/deny-list.test.js
patrickhousley: I had to use expectBeaconRequest so I could set a custom timeout.

tests/functional/stn/ajax.test.js
patrickhousley: This file was renamed. I only changed the file to add the second test below.

tests/functional/stn/ajax.test.js
metal-messiah: IMO we should instead just extend all the promises to accept+pass on a custom timeout, because the issue still remains that if the agent happens to send a supportability metric before the xhrMetrics payload, it will be an inconsistent test.

patrickhousley: Agree. That is done in the fastify rewrite of the test server.

metal-messiah (approved):

src/features/ajax/aggregate/index.js
```diff
@@ -75,13 +75,6 @@ export class Aggregate extends AggregateBase {
 
       handle('bstXhrAgg', ['xhr', hash, params, metrics], undefined, FEATURE_NAMES.sessionTrace, ee)
```
cwli24, line 76: Bit of a scope creep, but should the deny-list also block this from getting into session trace too?

patrickhousley: No. We talked about this and we decided that the info being in the session trace was acceptable.

cwli24 (approved):

PR #386, opened 2023-01-31
NEWRELIC-5498: Fixing custom interaction end time

patrickhousley:
### Overview

Resolving an issue where multiple successive custom interactions being saved and ended result in multiple interactions within the same harvest but only one interaction being saved in NR1. This was due to an issue in how the `customEnd` time was being calculated. The `firstTimestamp` time will now be set as an instance variable in the serializer and reset after a harvest completes. Subsequent interactions in the same harvest will re-use the same `firstTimestamp` to calculate the interactions `customEnd` time. The `customEnd` time of an interaction should now always match or be greater than the `end` time of the same interaction.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-5498

### Testing

Test case wrote where three custom interactions are ended in succession resulting in three interactions being harvested together. The customEnd time for the interactions will be equal to or greater than the interaction end time.

image


metal-messiah (approved):

patrickhousley:
The below error from the test run on this PR is fixed in https://github.com/newrelic/newrelic-browser-agent/pull/384

image

src/features/spa/aggregate/serializer.js
```diff
@@ -11,177 +11,180 @@ import { getInfo } from '../../../common/config/config'
...
 
   serializeMultiple(interactions, offset, navTiming) {
     const info = getInfo(this.sharedContext.agentIdentifier)
     var addString = getAddStringContext(this.sharedContext.agentIdentifier)
     var serialized = 'bel.7'
     interactions.forEach((interaction) => {
-      serialized += ';' + serializeInteraction(interaction.root, offset, navTiming, interaction.routeChange, addString, info)
+      serialized += ';' + this.serializeInteraction(interaction.root, offset, navTiming, interaction.routeChange, addString, info)
     })
+    this.firstTimestamp = undefined
```
cwli24, line 24: Could you just add a jsdoc or some comments for context on this variable? (Relating to this issue)

cwli24 (approved):
Thanks G

PR #389, opened 2023-02-02
Add a range of draft changes for discussion.

bjfield:
### Overview

This **DRAFT** PR adds JSDoc comments and an assortment of other changes aimed at code clarity and improved developer experience for contributors and users of the agent—with particular attention to recent architectural changes related to packaging and how agent features are composed and loaded.

### Related Issue(s)

[NEWRELIC-6479](https://issues.newrelic.com/browse/NEWRELIC-6479)

### Testing

Changes in this PR are confined to developer experience considerations and should not materially impact logic or functionality of the agent or the development environment. This intention should be confirmed by standard automated tests prior to merge.


src/common/url/parse-url.js
```diff
@@ -32,6 +32,7 @@ export function parseUrl (url) {
     try {
       urlEl = new URL(url, location.href)
     } catch (err) {
+      // IE doesn't support URL(), so for IE non-web contexts (e.g. workers), `parseUrl` will return the default `{}`.
```
patrickhousley, line 35: This is not entirely accurate. The if block checks for a browser scope and uses an anchor element which should be supported in all browsers we support + IE11. This else block is only going to get called in a now browser scope like a worker and we do not support workers in IE11.

patrickhousley (approved):
Looks good. Might want @metal-messiah and @cwli24 read through the comments too.

src/common/url/parse-url.js
bjfield: The comment is a bit confusing and not super relevant, so removing it. Thanks

patrickhousley: The catch block could still be considered relevant just in case what gets passed to `URL` constructor is not a valid URL. That will cause an exception to be thrown.

metal-messiah (approved):

PR #391, opened 2023-02-02
Better agent abort handling and clean up : NR-57403 & more

cwli24:
### Overview

The two purposes of this are:

1. Make the agent more responsible for resource consumed on end user systems in the case an internal exception occurred and the dirtying of globals shared with other frameworks and dev code.
2. Some more tidying following the arch restructure in #358 and CriOS re-discovery in #378, and address other tech debts to make the code easier to digest (contribute to) & maintain.

Additionally, some deprecated and old IE logic and XHR checks are removed, and so are remnants of worker work and the old runtime object's `features` system. 
The wrapping modules shared by different features are made to wrap/unwrap & function independently via counting the times they were called rather than just using a map of boolean flags.

### Related Issue(s)

NEWRELIC-6349
NEWRELIC-6637 -- Patch to _quiet_ the chunk 692 import errors from v1223 (due to uncaught Promise) is also included in this PR.
NR-57380
NR-57395
NR-57403

### Testing

Regression testing
Couple new unit(browser) tests around abort and wrap counts


src/common/wrap/wrap-promise.js
```diff
@@ -2,112 +2,103 @@
...
-          // eslint-disable-next-line
-          Promise.resolve(sub).then(setNrId(method === 'all'), setNrId(false))
-        })
+    function WrappedPromise (executor) {
+      var ctx = promiseEE.context()
+      var wrappedExecutor = promiseWrapper(executor, 'executor-', ctx, null, false)
 
-        var originalReturnValue = original.apply(OriginalPromise, arguments)
-        var promise = OriginalPromise.resolve(originalReturnValue)
+      const newCustomPromiseInst = Reflect.construct(prevPromiseObj, [wrappedExecutor], WrappedPromise);  // extend the orig Promise constructor as super
```
metal-messiah, line 42: This will likely need a polyfill `core-js/stable/reflect`

src/common/wrap/wrap-raf.js
```diff
@@ -5,25 +5,21 @@
...
-  wrapped[ee.debugId] = true
-  var wrapFn = wfn(ee)
+  if (!isBrowserScope || wrapped[ee.debugId]++) // Notice if our wrapping never ran yet, the falsey NaN will not early return; but if it has,
+    return ee;                                  // then we increment the count to track # of feats using this at runtime. (Check the env before wrapping!)
+  wrapped[ee.debugId] = 1;
 
-  var equestAnimationFrame = 'equestAnimationFrame'
+  var wrapFn = wfn(ee)
 
-  wrapFn.inPlace(window, [
```
metal-messiah, line 21: What is the purpose of not checking for these other properties? Have these all aligned to just use `requestAnimationFrame` across all browsers?

cwli24: You're talking about the 'ms' and 'webkit' prefixes, etc., right? If so, then it's similar to page-visibility.js `___hidden`, for really old versions & backwards compat.

src/features/ajax/instrument/index.js
```diff
@@ -22,35 +22,34 @@ var handlers = ['load', 'error', 'abort', 'timeout']
...
 
-export function getWrappedFetch(ee, handler) {
-  var wrappedFetch = wrapFetch(ee)
-  return wrappedFetch
+  /** Restoration and resource release tasks to be done if Ajax loader is being aborted. */
+  #abort() {
+    // (Much of this module affects specific XHR instances.)
+    unwrapFetch(this.ee);
+    unwrapXhr(this.ee);
+    this.abortHandler = null; // weakly allow this abort op to run only once
```
metal-messiah, line 51: If something does call `this.abortHandler` after being set to null it'll throw an error

```suggestion
    this.abortHandler = ( ) => {}; // weakly allow this abort op to run only once
```

src/features/jserrors/instrument/index.js
```diff
@@ -22,76 +21,66 @@ export class Instrument extends InstrumentBase {
...
+      handle('err', [err, now(), false, { unhandledPromiseRejection: 1 }], undefined, FEATURE_NAMES.jserrors, this.ee)
+    }, {signal: this.removeOnAbort?.signal});
 
-        if (agentRuntime.xhrWrappable) {
-          wrapXhr(this.ee)
-        }
+    wrapRaf(this.ee);
+    wrapTimer(this.ee);
+    wrapEvents(this.ee);
+    wrapXhr(this.ee);
```
metal-messiah, line 69: Do we still need to check `.xhrWrappable` here?  Its being checked in the other features still

src/features/ajax/instrument/index.js
cwli24: - You can see that it's also replaces a flag in JS error which would return `true` even for empty functions when it should return `false`.
- The existing places that calls it, calls it with optional chaining, so it's safe!

src/features/jserrors/instrument/index.js
cwli24: Yep probably should. Unlike you're ok with just removing `xhrWrappable` altogether. v1220 integration issues has made me maybe too fearful of assuming givens. 

How likely do you think the chances are someone modified their global `XMLHttpRequest.prototype` such that `addEventListener` is no longer a function? (I think the logic isn't specifically about addEventListener but just using it as a way to see if it's been tampered with at all.)

metal-messiah: Impossible to know, i would leave the check in, mostly for "do no harm"

cwli24:
@metal-messiah There are ~19 extra files since you last look over this, but that's just relocating the `wrap-____.browser.js` test files into a folder together.

src/features/utils/instrument-base.js
```diff
@@ -14,28 +14,32 @@ export class InstrumentBase extends FeatureBase {
...
-        try {
-          const { Aggregate } = await import(`../../features/${this.featureName}/aggregate`)
-          new Aggregate(this.agentIdentifier, this.aggregator)
-          this.resolve()
-        } catch (e) {
-          warn(`Failed to import aggregator class for ${this.featureName}`, e)
-        }
+    if (this.hasAggregator || !this.auto) return
+    this.hasAggregator = true
+    const waitForWindowLoad = async () => {
```
metal-messiah, line 27: This name change seems to indicate that this function will wait for the window to load.  This is a misnomer, and i think it made more sense the way it was before (lazyLoad some module when the function is called)

src/loaders/agent.js
```diff
@@ -45,11 +47,19 @@ export class Agent {
...
+            for (const featName in this.features) { // this.features hold only features that have been instantiated
+                this.features[featName].abortHandler?.();
+            }
+
+            const newrelic = gosNREUM();
+            delete newrelic.initializedAgents[this.agentIdentifier]?.['api'];   // prevent further calls to agent-specific APIs (see "configure.js")
+            delete newrelic.initializedAgents[this.agentIdentifier]?.[NR_FEATURES_REF_NAME];    // GC mem used internally by features
+            delete this.sharedAggregator;
+            // Keep the initialized agent object with its configs for troubleshooting purposes.
+            delete newrelic.ee?.get(this.agentIdentifier);  // clear our events storage
```
metal-messiah, line 62: should we call ee[namespace].abort() here?

cwli24: EE's Abort right now 1) is "global", and 2) sets flag and points ref to a new object. Whereas here we're outright deleting the entire ee object itself, so there's a slight distinction.

src/common/event-listener/event-listener-opts.js
```diff
@@ -1,21 +1,31 @@
...
-    get: function() {
-      supportsPassive = true
+  const options = {
+    get passive() { // this function will be called when the browser attempts to access the passive property
+      passiveSupported = true;
+      return false;
+    },
+    get signal() {
+      signalSupported = true;
+      return false;
```
patrickhousley, line 13: Why are getter methods setting a value?

patrickhousley: I think I understand this actually. This is more or less a trick to see if a browser supports these options by checking if these properties are accessed when we create an event listener. Pretty cool but I think we need better documentation here.

patrickhousley: Maybe a link to the MDN article would suffice.

src/common/harvest/harvest.js
```diff
@@ -208,27 +204,16 @@ export function getSubmitMethod(endpoint, opts) {
...
-    // number of data allowed before it starts failing, so we save it for
-    // unload data
-    if (xhrUsable) {  // this is practically every browser-version in use today, including all workers
-      useBody = true
-      method = submitData.xhr
-    } else if (endpoint === 'events' || endpoint === 'jserrors') {
-      method = submitData.img
-    } else {
-      return false
-    }
```
patrickhousley, line 231: Just a general question....with this change, does anything still use the img method?

src/common/wrap/index.js
```diff
@@ -1,46 +1,14 @@
...
+import {wrapHistory, unwrapHistory} from './wrap-history'
+import {wrapJsonP, unwrapJsonP} from './wrap-jsonp'
+import {wrapMutation, unwrapMutation} from './wrap-mutation'
+import {wrapPromise, unwrapPromise} from './wrap-promise'
+import {wrapXhr, unwrapXhr} from './wrap-xhr'
+import {wrapEvents, unwrapEvents} from './wrap-events'
+
+export {wrapEvents, wrapFetch, wrapHistory, wrapJsonP, wrapMutation, wrapPromise, wrapRaf, wrapTimer, wrapXhr};
+
+export {unwrapEvents, unwrapFetch, unwrapHistory, unwrapJsonP, unwrapMutation, unwrapPromise, unwrapRaf, unwrapTimer, unwrapXhr};
```
patrickhousley, line 14: 🤔 maybe combine these two exports into a single export? Or maybe use `export {wrapFetch, unwrapFetch} from './wrap-fetch'`. Personally, I hate barrel files like this because they can turn into points of circular reference leading to terrible errors to debug. However, it was already here. I think we should add a webpack plugin to detect circular dependencies if we can.

src/common/wrap/wrap-events.js
```diff
@@ -2,34 +2,38 @@
...
 import {ee as baseEE} from '../event-emitter/contextual-ee'
-import { createWrapperWithEmitter as wfn } from './wrap-function'
+import { createWrapperWithEmitter as wfn, unwrapFunction } from './wrap-function'
 import { getOrSet } from '../util/get-or-set'
 import { globalScope, isBrowserScope } from '../util/global-scope'
 
 const wrapped = {}
+var XHR = XMLHttpRequest
+var ADD_EVENT_LISTENER = 'addEventListener'
+var REMOVE_EVENT_LISTENER = 'removeEventListener'
```
patrickhousley, line 17: Use `const` or `let`

patrickhousley: Based on the capitalization of the variable names, I assume these are meant to be constants. That would make `const` the better choice. If they are getting re-assigned, then the name should be camelCased and `let` should be used.

src/common/wrap/wrap-events.js
```diff
@@ -2,34 +2,38 @@
...
   if ('getPrototypeOf' in Object) {
     if (isBrowserScope)
-      findAndWrapNode(document);
-    findAndWrapNode(globalScope);
-    findAndWrapNode(XHR.prototype)
+      findEventListenerProtoAndCb(document, wrapNode);
+    findEventListenerProtoAndCb(globalScope, wrapNode);
+    findEventListenerProtoAndCb(XHR.prototype, wrapNode);
     // eslint-disable-next-line
   } else if (XHR.prototype.hasOwnProperty(ADD_EVENT_LISTENER)) {
```
patrickhousley, line 34: This `else` block will never run based on browser support for `getPrototypeOf`

src/common/wrap/wrap-events.js
```diff
@@ -79,7 +76,32 @@ export function wrapEvents(sharedEE) {
...
 }
+/**
+ * Find the base prototype of 'object' that has its own "addEventListener" property, and run some function on it.
+ * @param {Object} object - the initial object to traverse prototype chain on
+ * @param {Function} cb - the function to run on the ancestral object once found, accepts an object as a arg
+ * @param {Array} rest - [optional] any additional arguments to pass to the cb
+ */
+function findEventListenerProtoAndCb(object, cb, ...rest) {
+  let step = object;
+  // eslint-disable-next-line
```
patrickhousley, line 87: Should not be adding new eslint disable line statements.

src/common/wrap/wrap-events.js
```diff
@@ -79,7 +76,32 @@ export function wrapEvents(sharedEE) {
...
+  while (step && !step.hasOwnProperty(ADD_EVENT_LISTENER)) { step = Object.getPrototypeOf(step) }
+  if (step) cb(step, ...rest);
+}
 
+export function unwrapEvents(sharedEE) {
+  const ee = scopedEE(sharedEE);
+  
+  // Don't unwrap until the LAST of all features that's using this (wrapped count) no longer needs this, but always decrement the count after checking it per unwrap call.
+  if (wrapped[ee.debugId]-- == 1) {
```
patrickhousley, line 96: This can be confusing and I would suggest we don't decrement the variable inside the logic check.

src/common/wrap/wrap-events.js
```diff
@@ -79,7 +76,32 @@ export function wrapEvents(sharedEE) {
...
+    [ADD_EVENT_LISTENER, REMOVE_EVENT_LISTENER].forEach(fn => {
+      if (typeof document === 'object')  findEventListenerProtoAndCb(document, unwrapFunction, fn); //==> unwrapFunction(findProto(document)?, fn);
```
patrickhousley, line 98: Above we use `isBrowserScope`. Why not do the same here?

src/common/wrap/wrap-events.js
```diff
@@ -79,7 +76,32 @@ export function wrapEvents(sharedEE) {
...
+/**
+ * Find the base prototype of 'object' that has its own "addEventListener" property, and run some function on it.
+ * @param {Object} object - the initial object to traverse prototype chain on
+ * @param {Function} cb - the function to run on the ancestral object once found, accepts an object as a arg
+ * @param {Array} rest - [optional] any additional arguments to pass to the cb
+ */
+function findEventListenerProtoAndCb(object, cb, ...rest) {
+  let step = object;
+  // eslint-disable-next-line
+  while (step && !step.hasOwnProperty(ADD_EVENT_LISTENER)) { step = Object.getPrototypeOf(step) }
```
patrickhousley, line 88: `getPrototypeOf` can throw a `TypeError` exception. 

```suggestion
  while (typeof step === 'object' && !Object.prototype.hasOwnProperty.call(step, ADD_EVENT_LISTENER)) {
    step = Object.getPrototypeOf(step)
  }
```

src/cdn/polyfills.js
```diff
@@ -7,6 +7,7 @@ import 'core-js/stable/object/assign'
 import 'core-js/stable/object/entries'
 import 'core-js/stable/object/values'
 import 'core-js/stable/map'
+import 'core-js/stable/reflect'
```
patrickhousley, line 10: Just curious why we need the reflect polyfill

src/common/wrap/wrap-history.js
```diff
@@ -21,20 +20,29 @@ const wrapped = {}
...
-  wrapFn.inPlace(window.history, [ 'pushState', 'replaceState' ], '-')
+  wrapFn.inPlace(window.history, HISTORY_FNS, '-')
 
   return ee
 }
+export function unwrapHistory(sharedEE) {
+  const ee = scopedEE(sharedEE);
 
+  // Don't unwrap until the LAST of all features that's using this (wrapped count) no longer needs this, but always decrement the count after checking it per unwrap call.
+  if (wrapped[ee.debugId]-- == 1) {
```
patrickhousley, line 41: Same here about decrementing a variable in the logic block.

src/common/wrap/wrap-fetch.js
```diff
@@ -60,24 +61,38 @@ export function wrapFetch(sharedEE){
...
     }
   }
 
   return ee
 }
+export function unwrapFetch(sharedEE) {
+  const ee = scopedEE(sharedEE);
 
+  // Don't unwrap until the LAST of all features that's using this (wrapped count) no longer needs this, but always decrement the count after checking it per unwrap call.
+  if (wrapped[ee.debugId]-- == 1) {
```
patrickhousley, line 87: Same here about decrementing a variable in the logic block.

src/common/harvest/harvest.js
cwli24: Looks like:

- IE on unload
- if harvest using sendBeacon fails
- by `inlineHit` api

So.. yes.

src/common/wrap/index.js
cwli24: Combined

src/common/wrap/wrap-events.js
cwli24: This had existed; will disabling make lint complain?

src/common/wrap/wrap-events.js
cwli24: Rather redundant, since the way things are setup, if it was never wrapped (checks `isBrowserScope` before doing so), then it wouldn't/shouldn't be unwrapped.

src/common/wrap/wrap-events.js
cwli24: Moved into `else` case.

src/common/wrap/wrap-events.js
cwli24: We just want to make sure `step` is an actual object right? Ok.

patrickhousley: Yup....and also eslint will start complaining about calling `hasOwnProperty` from the target object instead of using `Object.prototype.hasOwnProperty.call(....)`

src/common/wrap/wrap-events.js
patrickhousley: So this whole if logic check is redundant?

src/cdn/polyfills.js
cwli24: wrap-promise.js uses `Reflect.construct` now

src/common/wrap/wrap-promise.js
```diff
@@ -2,112 +2,104 @@
...
 
-    ;['all', 'race'].forEach(function (method) {
-      var original = OriginalPromise[method]
-      OriginalPromise[method] = function (subPromises) {
-        var finalized = false
-        mapOwn(subPromises, function (i, sub) {
-          // eslint-disable-next-line
-          Promise.resolve(sub).then(setNrId(method === 'all'), setNrId(false))
-        })
+    function WrappedPromise (executor) {
```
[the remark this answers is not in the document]
cwli24: ty sir 😓

src/cdn/polyfills.js
patrickhousley: Oh cool. Completely missed that.

src/common/wrap/wrap-events.js
cwli24: No, because this is checking if this is the (paired) "last" call or not. If so, then nothing is left that needs this to be wrapped, hence we unwrap it. If not, then we mark that there's 1 less dependency now.

metal-messiah (approved):

metal-messiah:
Changelog items needed for:
* Wrap Promise change
* Abort logic / unwrapping
* Messaging around blocked chunks
* General cleanup of unused code (optional)

PR #393, opened 2023-02-02
Clean up warning messages and test inconsistencies

metal-messiah:
### Overview
This PR cleans up warning messages about blocked scripts from CSP/ad blockers, and gets rid of false positives from cron tests

The warning no longer includes an error body + stack trace to diffuse tensions among customer that were artificially created by including that information in the message when it has nothing to do with the agent executing, and more to do with content-blocking on the page preventing lazy chunks from being imported.

Test workflows were modified to alleviate false positives that were occurring. Allowing a concrete pass-fail state is useful in these tests because they are directly linked to changelog badges, and are used to determine status for cutting a release.

### DIFF
--- a/src/features/utils/instrument-base.js
+++ b/src/features/utils/instrument-base.js
@@ -27,7 +27,7 @@ export class InstrumentBase extends FeatureBase {
           new Aggregate(this.agentIdentifier, this.aggregator)
           this.resolve()
         } catch (e) {
-          warn(`Failed to import aggregator class for ${this.featureName}`, e)
+          warn(`Something prevented the agent from being downloaded`)
         }
       }
       // theres no window.load event on non-browser scopes, lazy load immediately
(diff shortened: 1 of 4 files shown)


patrickhousley (approved):

src/features/utils/instrument-base.js
```diff
@@ -27,7 +27,7 @@ export class InstrumentBase extends FeatureBase {
           new Aggregate(this.agentIdentifier, this.aggregator)
           this.resolve()
         } catch (e) {
-          warn(`Failed to import aggregator class for ${this.featureName}`, e)
+          warn(`Something prevented the agent from being downloaded`)
```
bjfield, line 30: It's a nit, but there should probably be a period after "downloaded" here and elsewhere as a small mercy to grammar-anxious folks like me.

bjfield (approved):

PR #395, opened 2023-02-03
Centralize Stringify method and utilize circular handler

metal-messiah:
### Overview
This PR centralizes all use of `stringify()` to a single helper file, which uses the built-in `JSON.stringify()` with a handler function for circular logic.

### Related Issue(s)
https://issues.newrelic.com/browse/NEWRELIC-5610

### Testing
Tests have been changed some to account for data that used to get dropped or came back as undefined in circular object cases.


patrickhousley:
Running polyfill tests: https://github.com/newrelic/newrelic-browser-agent/actions/runs/4089069889

tests/browser/encode.browser.js
```diff
@@ -32,7 +32,12 @@ test('encode.obj', function (t) {
   var circular = {}
   circular.circular = circular
   var obj3 = {bar: ['a', circular, 'c']}
-  t.equal(encode.obj(obj3, 1000), '&bar=%5B%22a%22,null,%22c%22%5D', 'Handle bad objects')
+  t.equal(encode.obj(obj3, 1000), '&bar=%5B%22a%22,%7B%7D,%22c%22%5D', 'Handle bad objects')
+
+  var circular2 = []
+  circular2.circular = circular
```
cwli24, line 38: This isn't creating the circular structure I think you're testing for. It should be like:
```suggestion
  circular2.push(circular2);
```
So that you get `circular2 = [circular2(=circular2...)]

tests/browser/stringify.browser.js
```diff
@@ -30,31 +30,7 @@ test('fake stringify', function (t) {
...
-  t.equal(stringify(obj2), '{"other":222}', 'obj w/ prototype with bad toJSON')
-  t.equal(stringify(F), undefined, 'function with bad toJSON')
-  t.ok(stringify(obj3) === '{"stringified":"{\\"a\\":123,\\"c\\":\\"b\\",\\"f\\":\\"asdf\\",\\"n\\":null}"}', 'stringified object with bad toJSON')
-
-  delete Array.prototype.toJSON
-  delete String.prototype.toJSON
-  delete Number.prototype.toJSON
-  delete Date.prototype.toJSON
-  delete Object.prototype.toJSON
+  t.equal(stringify(a), '{}', 'Stringifying a circular object returns empty object')
```
cwli24, line 33: Have we verified with ingest or whoever parses the JSON how they want us to stringify these circ refs?

tests/browser/encode.browser.js
metal-messiah: ty

tests/browser/stringify.browser.js
metal-messiah: not even sure if we were able to report it before.  The previous test indicated that it would just be "undefined", not even a string

patrickhousley: Without this change, the data is not a valid json string. I think we are good to move forward with this.

patrickhousley (approved):

src/common/util/stringify.js
```diff
@@ -3,24 +3,24 @@
...
+  return (key, value) => {
+    if (typeof value === 'object' && value !== null) {
+      if (seen.has(value)) {
+        return
+      }
+      seen.add(value)
+    }
+    return value
+  }
 }
```
bjfield, line 20:
```suggestion

/**
 * The native JSON.stringify method augmented with a replacer function to handle circular references.
 * Circular references will be excluded from the JSON output rather than triggering errors.
 * @param {*} val - A value to be converted to a JSON string.
 * @returns {string} A JSON string representation of the value, with circular references handled.
 */
```

src/common/util/stringify.js
```diff
@@ -3,24 +3,24 @@
...
-var escapable = /[\\\"\x00-\x1f\x7f-\x9f\u00ad\u0600-\u0604\u070f\u17b4\u17b5\u200c-\u200f\u2028-\u202f\u2060-\u206f\ufeff\ufff0-\uffff]/g // eslint-disable-line
-var meta = {
-  '\b': '\\b',
-  '\t': '\\t',
-  '\n': '\\n',
-  '\f': '\\f',
-  '\r': '\\r',
-  '"': '\\"',
-  '\\': '\\\\'
+const getCircularReplacer = () => {
```
bjfield, line 8:
```suggestion
/**
 * Returns a function for use as a replacer parameter in JSON.stringify() to handle circular references.
 * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value MDN - Cyclical object value}
 * @returns {Function} A function that filters out values it has seen before.
 */
const getCircularReplacer = () => {
```

bjfield (approved):
I've suggested a couple optional JSDoc comments.

PR #396, opened 2023-02-07
Remove backticks from transpiled output (NR-86667)

bjfield:
### Overview

This PR removes backticks from transpiled output. Some customers assign our snippet to a variable for insertion in a specific place in their app, and to avoid conflicts with single and double quotes, they enclose the script in backticks (\`). However, since we started transpiling for newer browser versions, we have template literals in our output, which conflict with backticks. We also recently added an error message containing backticks.

This change results in a very slight increase in loader size. The SPA loader, for example, is 168 bytes larger.

It also looks like the package-lock.json on main was stale. I updated it in this PR so tests could run.

### Related Issue(s)

[NR-86667](https://issues.newrelic.com/browse/NR-86667)

### Testing

Standard automated tests apply.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16428,6 +16428,145 @@
         "which": "bin/which"
       }
     },
+    "node_modules/npm-run-all": {
+      "version": "4.1.5",
+      "resolved": "https://registry.npmjs.org/npm-run-all/-/npm-run-all-4.1.5.tgz",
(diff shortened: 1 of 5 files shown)


cwli24:
Has there been an actual issue ticket by a customer because they cannot do raw insertion for some reason?

cwli24 (approved):

patrickhousley (approved):

PR #398, opened 2023-02-09
NR-86614: Implementing centralized lazy loader

patrickhousley:
### Overview

This centralizes the lazy loading logic into a function that can be used by all agent types. This uses defined logic blocks and switch cases instead of template strings or string concatenation on purpose. Doing it this way ensures webpack does not need to "guess" what our chunks are and also allows us to easily name those chunks.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-86614

### Testing

All the tests cases should still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ Some older IE-related code is removed, and so is a query param in our page view 
 Related to failed imports, a "loading chunk 629" error was thrown despite the intention to just catch it as a warning. The cause was actually an uncaught promise in the code,
 which has now been fixed and the error silenced.
 
+### Resolving google indexing of agent relative paths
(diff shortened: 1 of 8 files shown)


patrickhousley:
We still need to look at a couple of issues, if they are issues.
1. There are still 4 chunks that have an ID and not a name. Where are these coming from?
2. There is no license file generated for non-polyfill output files.

metal-messiah (approved):
👏

metal-messiah:
> We still need to look at a couple of issues, if they are issues.
> 
> 1. There are still 4 chunks that have an ID and not a name. Where are these coming from?
> 2. There is no license file generated for non-polyfill output files.

These can be evaluated outside the scope of this ticket, if needed.

cwli24 (approved):

PR #407, opened 2023-02-16
Fix niche getter error in unhandledPromiseRejections - NR-89675

metal-messiah:
### Overview
This PR fixes an issue where niche browsers will throw an error when trying to set the message on the promise error.  We can avoid this by simply just returning the error if present without modifying it.

### Related Issue(s)
* https://newrelic-neworg.lightning.force.com/lightning/r/Case/5008W000011bs99QAA/view
* https://issues.newrelic.com/browse/NR-89675

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,6 +7,7 @@ module.exports = {
     'tests/assets/js/internal/**/*',
     'tests/assets/js/vendor/**/*',
     'tests/assets/modular/js-errors/js/vendor/**/*',
+    'tools/test-builds/**/*',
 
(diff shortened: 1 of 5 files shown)


cwli24:
Worth it to add a functional test that rejects with an non-writable object that would simulate this case?

metal-messiah:
> Worth it to add a functional test that rejects with an non-writable object that would simulate this case?

yeah I'll add one, good call

cwli24 (approved):

patrickhousley (approved):

PR #408, opened 2023-02-17
Defaulting harvest times to 5 seconds for integration tests

patrickhousley:
### Overview

Defaulting the harvest time to 5 seconds for all features when running integrations tests. This is a sweet spot where tests don't fail due to race conditions and timings but the tests are a bit faster.

### Related Issue(s)

N/A

### Testing

Run jil tests locally and make sure nothing fails.

### DIFF
--- a/tools/jil/util/browsers-supported.json
+++ b/tools/jil/util/browsers-supported.json
@@ -16,17 +16,17 @@
     },
     {
       "browserName": "chrome",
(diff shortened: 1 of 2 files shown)


metal-messiah (approved):

patrickhousley:
We will need to do more investigation around test failures in regards to the default harvest time. Closing for now.

Issue #409, opened 2023-02-17
Agent is not capturing/wrapping native async/await promises


patrickhousley:
I presumed you were already doing this and this bug was an edge case where this didn't work (hence my suggestion to add an option to disable) - but are you defining a Symbol.hasInstance on your custom promise, that returns true when both a native promise and mocked promise is passed

_Originally posted by @samdenty in https://github.com/newrelic/newrelic-browser-agent/issues/187#issuecomment-1435224768_

PR #410, opened 2023-02-17
NEWRELIC-6934: Re-wrap promise.prototype

patrickhousley:
### Overview

Issue 409 pointed out that `instanceof` the promise returned from `async` functions failed against the global (NR wrapped) Promise.

This led to our re-evaluation of the recent Promise wrapping re-do, which attempted to leave the original Promise unmodified and instead create a prototype chain. While we are able to replace the `fetch` promise, it would seem `async()` promise constructors are not replace-able, so the end result is that we switch back to directly modifying the original promise.prototype.

This ensures that all 3 (new Promise, fetch, and async function) promises are `instanceof` our wrapped promise and gets instrumented, once again.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-6934
#409

### Testing

Added an integration test to check the `instanceof` comparison.

### DIFF
--- a/src/common/wrap/wrap-promise.js
+++ b/src/common/wrap/wrap-promise.js
@@ -36,6 +36,15 @@ export function wrapPromise (sharedEE) {
     })
     WrappedPromise.toString = function () { return prevPromiseObj.toString() }
 
+    // Adds support for instanceof comparison for async functions
+    if (typeof Symbol === 'function' && typeof Symbol.hasInstance === 'symbol') {
+      Object.defineProperty(WrappedPromise, Symbol.hasInstance, {
+        value: function (instance) {
+          return instance instanceof prevPromiseObj
+        }
+      })
+    }
+
     function WrappedPromise (executor) {
       var ctx = promiseEE.context()
       var wrappedExecutor = promiseWrapper(executor, 'executor-', ctx, null, false)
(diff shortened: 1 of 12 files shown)


bjfield (changes requested):
Other than a comment that could be clarified in wrap-promise.js, looks great.

metal-messiah (approved):

src/common/wrap/wrap-promise.js
```diff
@@ -83,23 +88,32 @@ export function wrapPromise (sharedEE) {
...
-    // Mirror the orig Promise's prototype (catch, try, then, etc.) but with our own methods.
-    WrappedPromise.prototype = Object.create(prevPromiseObj.prototype)
-    WrappedPromise.prototype.constructor = WrappedPromise // calls to "new Promise()" should use our custom Promise
-    WrappedPromise.prototype.then = function (...args) {
+    /*
+     * Ideally, we create a new WrappedPromise.prototype chained off the original Promise's so that we don't alter it.
+     * However, there's no way to make the (native) promise returned from async functions use our WrappedPromise,
+     * so we have to modify the original prototype in essence. This ensures that "async()"" promise functions and runs the
+     * same instance methods as "new Promise()", and also that instanceof async() is the global Promise (see GH issue #409).
+     *  In addition, this affects the fetch() promise too.
```
bjfield, line 96:
```suggestion
     * so we have to modify the original prototype. This ensures that promises returned from async functions execute
     * the same instance methods as promises created with "new Promise()", and also that instanceof async() is
     * the global Promise (see GH issue #409). This also affects the promise returned from fetch().
```

bjfield (approved):

cwli24 (approved):

samdenty:
is this released yet?

patrickhousley:
Hey @samdenty not yet. We just merged the fix and are getting ready for the release of version 1227. We don't have an ETA just yet as we are also working on testing out release 1226 right now.

PR #411, opened 2023-02-17
Update builds to use SemVer - NEWRELIC-6928 NEWRELIC-6931

metal-messiah:
### Overview
This PR converts the internals of the browser agent to build and report semver in:
* Payloads to ingest
* Agent file names
* Workflows

### Related Issue(s)
NEWRELIC-6928
NEWRELIC-6931


.github/workflows/publish-cdn.yml
```diff
@@ -4,29 +4,9 @@ on:
...
-      - name: install
-        run: npm ci
-      - name: build:all
-        run: npm run build:all
-      - name: run tests
-        run: node ./tools/jil/bin/cli.js -f merged -b chrome@latest -s -t 85000 --concurrent=10
   publish-to-s3:
     runs-on: ubuntu-latest
     container: node:14
-    needs: [test]
```
patrickhousley, line 29: Just verifying here we are intentionally removing the smoke tests for the cdn publish.

.github/workflows/publish-release.yml
```diff
@@ -115,16 +119,20 @@ jobs:
       - uses: actions/checkout@v3
       - name: install
         run: npm ci
+      - name: get version
+        run: |
+          export PACKAGE_VERSION=$(node -e "(function () { console.log(require('./package.json').version) })()") \
+          export CDN_VERSION=$(cat VERSION)
```
patrickhousley, line 125: Does exporting the variable work in a later step. What I have seen, we need to use an output variable.

https://stackoverflow.com/a/75659156

tools/scripts/sync-version.js
```diff
@@ -0,0 +1,8 @@
+const pkg = require('../../package.json')
+const fs = require('fs')
+const path = require('path')
+
+const cdn_version = fs.readFileSync(path.join(__dirname, '../../', 'VERSION'), 'utf-8')
+const new_pkg_version = `0.${cdn_version.split('.').slice(0, 2).join('.')}`
```
patrickhousley, line 6: We may need to update this to account for the patch number from the `VERSION` file. Let's see what dempla and apm services say.

metal-messiah: DEM is 👍 on being able to parse this.  We should NOT publish pre-release versions to NRDB + connect for now.  prerelease versions should be for NPM only for now.

.github/workflows/publish-release.yml
metal-messiah: I figured it should work if its in the same job, since it doesn't spin up a new process, but we should verify

patrickhousley (approved):

.github/workflows/publish-release.yml
metal-messiah: moved to use @actions/core and reference in steps outputs

.github/workflows/publish-cdn.yml
metal-messiah: This is meant to be a "standalone" GHA only to be used to "re-run" if something goes wrong in the CI tied to cutting a release.

cwli24 (approved):

bjfield (approved):

PR #412, opened 2023-02-17
Revert XHR metrics collection - NEWRELIC-6935

metal-messiah:
### Overview
A recent change was made in 1224 to exclude XHR timeslice metrics collection if the url is in the deny-list.  This is particularly impactful as the BAM endpoints are in the deny-list by default, and ingest metrics are not representative of the full page impact without the agent calls.

This PR reverts the XHR metrics to report ALL ajax calls, regardless of deny-list status, with the expectation to follow up with a different method to limit timeslice metric collection to be discussed in the near-future.

### Related Issue(s)
https://issues.newrelic.com/browse/NEWRELIC-6935?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

### Testing
A test that was added for xhr collection testing was removed


patrickhousley (approved):

patrickhousley:
Need changelog

PR #417, opened 2023-02-23
[Patch] Remove unwrapping of global api & handlers on abort NEWRELIC-7013

cwli24:
### Overview

The previous PR had attempted to revert browser agent wrappings if and when its loading of components encounter exceptions, for example during import. The method in which that was done is simply restoring a copy of the global API or handler that existed immediately preceding the agent initialization.

However, this can cause an integration issue with other third party libraries and code that follows which wrap or references our wrapper functions because they also replaces or modify the globals with their internals, and our "global restores" that happen after that would effectively overwrite their wrapping.

With this patch, we will leave our wrappings in place to not break the integrations. Instead, we'll attempt to do our best to make them become a dummy code layer that does nothing if our agent aborts (already factored into current code).

### Related Issue(s)

NEWRELIC-7013
Following #391 

### Testing

Regression


patrickhousley:
Did you mean to leave in all the commented out code

cwli24:
> Did you mean to leave in all the commented out code

I did, but should I remove them? Are they for sure worthless at this point?

patrickhousley:
> > Did you mean to leave in all the commented out code
> 
> I did, but should I remove them? Are they for sure worthless at this point? @metal-messiah did say we could leave the abort handlers sys in place, so wasn't sure.

It's forever enshrined in the git repo so if we ever want to pull it back up, we can. I would delete it. There is no need in keeping commented code around with git.

patrickhousley (approved):
Just need a changelog entry but the code changes look fine.

metal-messiah (approved):

metal-messiah:
This is trivial and inconsequential, and more-so commenting for posterity, but do we/should we still increment a counter on these if we do not need to unwrap anymore? Approved with addition of changelog item

https://github.com/newrelic/newrelic-browser-agent/pull/417/files#diff-44610b3a5e0a1556ad22b44d1d2bff702a2ca7125d84335c716eeaecd3ee2702L23

PR #421, opened 2023-02-24
Fixing file mime types

patrickhousley:
### Overview

Fixing the mime type used when uploading to s3 and removed a couple of scripts that are no longer used.


metal-messiah (approved):

tools/scripts/upload-to-s3.js
```diff
@@ -3,6 +3,7 @@ var fs = require('fs')
 var path = require('path')
 var AWS = require('aws-sdk')
 var yargs = require('yargs')
+const mime = require("mime-types");
```
bjfield, line 6:
```suggestion
const mime = require('mime-types');
```

tools/scripts/publish-current.js
```diff
@@ -13,6 +13,7 @@ var AWS = require('aws-sdk')
 var request = require('request')
 var yargs = require('yargs')
 var loaders = require('jil/util/loaders.js')
+const mime = require("mime-types");
```
bjfield, line 16:
```suggestion
const mime = require('mime-types');
```

bjfield (approved):
I've confirmed that HTML and JSON mime types get picked up as expected.

```
ContentType: 'text/html',
Key: 'worker.stats.html'
...
ContentType: 'application/json',
Key: 'worker.stats.json'
```

If desired, I suggested a couple changes from double quotes to singles, but no big deal.

PR #426, opened 2023-03-01
NR-93365: Adding safety around interaction node cancelling

patrickhousley:
### Overview

Adding in some safety around the possibility that a SPA interaction node could `finish()` with a parent node that has been cancelled. This could cause an exception but the possibility of this happening is un-reproducible. This is a do-no-harm best-effort to ensure this error can never happen.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-93365
https://issues.newrelic.com/browse/NR-91546

### Testing

Added a jest test to reproduce and test the issue. I could not consistently reproduce the issue in integration tests.

### DIFF
--- a/src/features/spa/aggregate/interaction-node.js
+++ b/src/features/spa/aggregate/interaction-node.js
@@ -69,19 +69,12 @@ InteractionNodePrototype.finish = function finish (timestamp) {
   node.end = timestamp
 
   // Find the next parent node that is not cancelled
-  if (node.parent) {
-    let parent = node.parent
-    while (parent.cancelled) {
-      if (!parent.parent) break
-      parent = parent.parent
-    }
+  let parent = node.parent
+  while (parent?.cancelled) parent = parent.parent
 
-    // Assign the node to the non-cancelled parent node
-    if (!parent.cancelled) {
-      parent.children.push(node)
-    }
-    node.parent = null
-  }
+  // Assign the node to the non-cancelled parent node
+  if (parent) parent.children.push(node)
+  node.parent = null
 
   // Update the interaction remaining counter
   var interaction = this.interaction
--- a/src/features/spa/aggregate/interaction-node.js
+++ b/src/features/spa/aggregate/interaction-node.js
@@ -67,11 +67,23 @@ InteractionNodePrototype.finish = function finish (timestamp) {
   var node = this
   if (node.end) return
   node.end = timestamp
-  var parent = node.parent
-  while (parent.cancelled) parent = parent.parent
-  parent.children.push(node)
-  node.parent = null
 
+  // Find the next parent node that is not cancelled
+  if (node.parent) {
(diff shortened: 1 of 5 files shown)


cwli24:
Interaction-InterationNode on the list for an overhaul _someday_. I imagine a much larger undertaking with a re-arch design & draft with the entire team given it's the vertebrae of SPA.

src/features/spa/aggregate/interaction-node.js
```diff
@@ -67,11 +67,23 @@ InteractionNodePrototype.finish = function finish (timestamp) {
...
   node.end = timestamp
-  var parent = node.parent
-  while (parent.cancelled) parent = parent.parent
-  parent.children.push(node)
-  node.parent = null
 
+  // Find the next parent node that is not cancelled
+  if (node.parent) {
+    let parent = node.parent
+    while (parent.cancelled) {
```
cwli24, line 74: Just this would work, right?
```suggestion
    while (parent?.cancelled) {
```

patrickhousley: No because parent could still get set to a null value in the previous loop.

cwli24: Actually I'm suggesting this change instead overall:
```suggestion
    let parent = node.parent;
    while (parent?.cancelled) parent = parent.parent; // keep searching up the chain until we find a/the first ancestor that isn't cancelled
    if (parent) parent.children.push(node); // if parent isn't null, then we found one that isn't cancelled
    node.parent = null;
```

cwli24: I'm also quite confused by the original code, since if we _do_ find a cancelled ancestor, we say `ancestor -> node` but is why is `node.parent -> null` ?? Lol, any ideas?

patrickhousley: Yea that would work to. Ill commit that.

cwli24: > No because parent could still get set to a null value in the previous loop.

If `parent = null` from previous loop, then `parent?.cancelled` would be `undefined` and fail to continue looping though.

patrickhousley: > I'm also quite confused by the original code, since if we _do_ find a non-cancelled ancestor, we say `ancestor.children -> node` but is why is `node.parent -> null` ?? Lol, any ideas?

No, Jordan and I were confused by that too this morning when we were working through this.

cwli24: Maybe this is a bug... possible they meant `node.parent = parent`, whether that be a found node or `null`. "The world may never know."

cwli24 (approved):

cwli24:
Nice work on finding this

metal-messiah (approved):

PR #427, opened 2023-03-02
Collate all SM to page EoL NR-94597

cwli24:
### Overview

This PR is part of a larger effort to reduce the amount of network calls the agent makes and improve efficiency.

One of the features that is making a lot of requests, which factors into infrastructure costs, is the metrics module. Presently, metrics are reported on initialization, then set to an interval harvest that observes ongoing events like various API calls and usage.

Going forward, **we'll condense them into and only make _one_ metrics harvest** per our (post-BFC improved) EoL logic that signifies the potential end of user active sessions--in other words, at the end of a possible page lifecycle.

In addition, per previous work to prevent the agent from making harvest calls to endpoints that are 403-blocked, there was an overlooked gap in which the final harvests would still get sent irrespectively. That's also patched in this PR.

### Related Issue(s)

NR-94597 (partial)
Patch for #368 

### Testing

Regression


src/features/metrics/aggregate/index.js
```diff
@@ -15,27 +15,28 @@ export class Aggregate extends AggregateBase {
...
-    // if rum response determines that customer lacks entitlements for err endpoint, block it
+    // If RUM-call's response determines that customer lacks entitlements for the /jserror ingest endpoint, don't harvest at all.
     registerHandler('block-err', () => {
       this.blocked = true
-      scheduler.stopTimer()
+      if (scheduler) scheduler.aborted = true // RUM response may or may not have happened already before scheduler initialization below
     }, this.featureName, this.ee)
 
-    drain(this.agentIdentifier, this.featureName)
+    if (!this.blocked) {
```
metal-messiah, line 26: I'm not sure that there's really a point to wrap this with this if check.  The thing that blocks (`block-err`) will only trigger after draining (since its staged with registerHandler), which would always be after line 26 has already executed

src/features/metrics/aggregate/index.js
```diff
@@ -15,27 +15,28 @@ export class Aggregate extends AggregateBase {
...
+
+      this.singleChecks() // checks that are run only one time, at script load
+      this.eachSessionChecks() // the start of every time user engages with page
+
+      // *cli, Mar 23 - Per NR-94597, this feature should only harvest ONCE at the (potential) EoL time of the page.
+      scheduler = new HarvestScheduler('jserrors', { onUnload: () => this.unload() }, this)
+      scheduler.harvest.on('jserrors', () => ({ body: this.aggregator.take(['cm', 'sm']) }))
+    }
+
+    drain(this.agentIdentifier, this.featureName) // regardless if this is blocked or not, drain is needed to unblock other features from harvesting (counteract registerDrain)
```
metal-messiah, line 39: to continue that thought, it definitely will never have fired, because the drain call is inside the if block

cwli24: The drain is outside the `if` 👀

src/features/metrics/aggregate/index.js
cwli24: ok, figures. I thought `registerHandler` would be synchronous for some reason -- why isn't it actually? (Usually the handles are all in instrument and registers are all in aggs)

src/features/metrics/aggregate/index.js
metal-messiah: ah youre right.  It will still not have fired since drain is called below the if check tho

cwli24: > ah youre right. It will still not have fired since drain is called below the if check tho

You're talking about `block-err` (if registerHandler were immediate)? That's based on PVE though, which makes the RUM without being blocked by drain registry?

tests/functional/metrics.test.js
```diff
@@ -11,25 +11,21 @@ const fetchBrowsers = testDriver.Matcher.withFeature('fetchExt')
...
-    }))
-
-    Promise.all([metricsPromise, rumPromise, loadPromise])
-      .then(([{ request: data }]) => {
+    const loadPromise = browser.safeGet(router.assetURL('instrumented.html', { loader: loaderType }))
+
+    Promise.all([rumPromise, loadPromise])
+      .then(() => {
+        browser.get(router.assetURL('/'))
+        return router.expectMetrics(3000)
```
metal-messiah, line 26: These tests work great but i suspect that they would pass for metrics with an interval too (for instance if someone added it back in, the test would likely still pass).  

Should we add one more with a condition that it just checks that a payload isn't sent after x time? Or are we ok self-enforcing this one?

src/features/metrics/aggregate/index.js
metal-messiah: `block-err` wont fire it's callback(s) until the metrics feature is drained tho, since it was staged with registerHandler

src/features/metrics/aggregate/index.js
cwli24: Removed

tests/functional/metrics.test.js
cwli24: Since took out any sort of timer in the source code, the only way to test it would be having one hang for some arbitrary amount of time, which only lengthens test suite duration. Prob not worth adding it as discussed. 🤷‍♂️

cwli24:
Tests are all passing, just pushed VERSION file update & changelog.

src/features/metrics/aggregate/index.js
```diff
@@ -23,19 +30,11 @@ export class Aggregate extends AggregateBase {
     this.singleChecks() // checks that are run only one time, at script load
     this.eachSessionChecks() // the start of every time user engages with page
 
-    var harvestTimeSeconds = getConfigurationValue(this.agentIdentifier, 'metrics.harvestTimeSeconds') || 30
-
-    var scheduler = new HarvestScheduler('jserrors', { onUnload: () => this.unload() }, this)
+    // *cli, Mar 23 - Per NR-94597, this feature should only harvest ONCE at the (potential) EoL time of the page.
```
bjfield, line 33:
```suggestion
    // We are no longer harvesting periodically via `scheduler.startTimer` because per NR-94597 in March 2023, this
    // feature should only harvest ONCE at the (potential) EoL time of the page. (EoL collection is the default
    // behavior of HarvestScheduler even if `startTimer` is not invoked.)
```

bjfield (approved):

src/features/metrics/aggregate/index.js
cwli24: I think comments should serve to remind what the expected behavior is (readers can dig into historic info themselves) rather than cluttering current code w/ _how it was_ b/c they don't immediately see that change and just gets confused.

PR #428, opened 2023-03-04
ALL agents must still send initial connect call - NR-95039

metal-messiah:
### Overview
Recent revelations from ingest have indicated that the current architecture of downstream teams relies on all reporting apps to call the `/1` endpoint to mark the entity for internal storage and avoid garbage collection. Niche builds of the agent that did not include that initial call were subject to be garbage collected in NR1, and as such, we need to enforce that the `/1` call not be able to be disabled until another solution is reached.

### Related Issue(s)
https://issues.newrelic.com/browse/NR-95039

### Testing
Workers tests were modified to also expect a RUM call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 ## v1227
 
 ### Added INP and long tasks reporting
-The interaction-to-next-paint metric is now calculated and reported at the end of user sessions, via [Google CWV](https://github.com/GoogleChrome/web-vitals) library. In addition, long continuously executed and blocking scripts detected by the PerformanceLongTaskTiming API is also forwarded to New Relic.
+The interaction-to-next-paint metric is now calculated and reported at the end of user sessions, via [Google CWV](https://github.com/GoogleChrome/web-vitals) library. In addition, long continuously executed and blocking scripts detected by the PerformanceLongTaskTiming API is also forwarded to New Relic. This functionality is `off` by default, until a curated UI experience is created to utilize this data.
 
 ### Revert unwrapping of globals on agent abort
 Partial revert of graceful handling change made in v1225 that unwrapped modified global APIs and handlers, which caused integration issues with other wrapping libraries and code.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 ## v1227
 
 ### Added INP and long tasks reporting
-The interaction-to-next-paint metric is now calculated and reported at the end of user sessions, via [Google CWV](https://github.com/GoogleChrome/web-vitals) library. In addition, long continuously executed and blocking scripts detected by the PerformanceLongTaskTiming API is also forwarded to New Relic. This functionality is `off` by default, until a curated UI experience is created to utilize this data.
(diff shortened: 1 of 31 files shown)


src/features/page_view_event/aggregate/index.js
```diff
@@ -95,10 +96,10 @@ export class Aggregate extends AggregateBase {
     chunksForQueryString.push(param('ja', customJsAttributes === '{}' ? null : customJsAttributes))
 
     var queryString = fromArray(chunksForQueryString, agentRuntime.maxBytes)
-
-    submitData.jsonp(
+    const isValidJsonp = submitData.jsonp(
       this.getScheme() + '://' + info.beacon + '/' + protocol + '/' + info.licenseKey + queryString,
       jsonp
     )
+    if (!isValidJsonp) drain(this.agentIdentifier, this.featureName)
```
metal-messiah, line 103: Add comment why this is needed

src/features/utils/lazy-loader.js
```diff
@@ -33,26 +33,5 @@ export function lazyLoader (featureName, featurePart) {
       default:
         throw new Error(`Attempted to load unsupported agent feature: ${featureName} ${featurePart}`)
     }
```
metal-messiah, line 35: Add a comment in this file explaining why we should only import "Aggregate" files

bjfield (changes requested):
I've suggested a couple comment changes.

Tests pass, and changes make sense to me as summarized here:

* In `start-timing.js`, we now use the now-standard and worker-compatible `timeOrigin` property of the performance API, with `timing.navigationStart` as a fallback.
* To account for module workers not being able to handle JSONP, we must ask the pageViewEvent aggregator to drain manually in cases where the JSONP response cannot be relied upon to trigger drain automatically (when it finishes processing feature flags in the ingest response to the RUM call).
* The pageViewEvent aggregator must also reference `globalScope` instead of `window` to accommodate workers.
* The code in `lazy-loader.js` meant to allow for lazy-loading instrumentation (not just aggregators) was creating extra large outputs for worker builds, even though (and in part because) that code does not apply to workers. This code has been removed, and the way the micro-agent handles instrumentation modules will need to be revisited.
* Changes in `agent.js` ensure that the pageView event cannot under any circumstances be turned off or excluded.
* Tests have been modified to ensure RUM calls for workers.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 ## v1227
 
 ### Added INP and long tasks reporting
-The interaction-to-next-paint metric is now calculated and reported at the end of user sessions, via [Google CWV](https://github.com/GoogleChrome/web-vitals) library. In addition, long continuously executed and blocking scripts detected by the PerformanceLongTaskTiming API is also forwarded to New Relic.
+The interaction-to-next-paint metric is now calculated and reported at the end of user sessions, via [Google CWV](https://github.com/GoogleChrome/web-vitals) library. In addition, long continuously executed and blocking scripts detected by the PerformanceLongTaskTiming API is also forwarded to New Relic. This functionality is `off` by default, until a curated UI experience is created to utilize this data.
```
coreyarnold, line 9: I would break this apart or more clearly indicate that the functionality that is off is the long tasks. This currently reads that it might be LT and INP.

coreyarnold (approved):

bjfield (approved):

CHANGELOG.md
bjfield, line 9:
```suggestion
The interaction-to-next-paint metric is now calculated and reported at the end of user sessions, via [Google CWV](https://github.com/GoogleChrome/web-vitals) library. In addition, long continuously executed and blocking scripts detected by the PerformanceLongTaskTiming API is also forwarded to New Relic. The Long Task functionality is `off` by default, until a curated UI experience is created to utilize this data.
```

bjfield: Suggesting a change in a comment below.

PR #429, opened 2023-03-07
Run LT behind a config flag - NR-95944

metal-messiah:
### Overview
This PR puts the `Long Task` API behind a config flag under the `page_view_timing` setting.  It can be configured at `init.page_view_timing.long_task: bool` which defaults to `false`

### Related Issue(s)
https://issues.newrelic.com/browse/NR-95944

### Testing
Long task tests needed the flag to be set to run correctly now, so that adjustment has been made to the existing tests.

### DIFF
--- a/src/common/config/state/init.js
+++ b/src/common/config/state/init.js
@@ -19,7 +19,7 @@ const model = {
   metrics: { enabled: true },
   page_action: { enabled: true, harvestTimeSeconds: 30 },
   page_view_event: { enabled: true },
-  page_view_timing: { enabled: true, harvestTimeSeconds: 30 },
+  page_view_timing: { enabled: true, harvestTimeSeconds: 30, long_task: false },
   session_trace: { enabled: true, harvestTimeSeconds: 10 },
   spa: { enabled: true, harvestTimeSeconds: 10 }
 }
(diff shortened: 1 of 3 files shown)


src/common/config/state/init.js
```diff
@@ -19,7 +19,7 @@ const model = {
   metrics: { enabled: true },
   page_action: { enabled: true, harvestTimeSeconds: 30 },
   page_view_event: { enabled: true },
-  page_view_timing: { enabled: true, harvestTimeSeconds: 30 },
+  page_view_timing: { enabled: true, harvestTimeSeconds: 30, long_task: false },
```
bjfield, line 22: * The deviation to camelCase for the harvest time becomes conspicuous next to an additional snake_case property. Elsewhere, we have properties like `maxLCPTimeSeconds`, `initialHarvestSeconds`, `maxPayloadSize` etc. Do you know if there is some meaning or utility implied by these different conventions? If not, might be good to standardize all of them one way or the other at some point (not in this PR).
* I don't know if we are trying to keep `types.ts` in sync at all, but if we are, we could add `long_task?: boolean` and `allow_bfcache?: boolean` added to the definition for this model in that file. If we aren't keeping it in sync, might be good to remove it from the repo and let git preserve it for us.

metal-messiah: Lets add a ticket to standardize the configs, which should include updating the types.ts file. That will be a larger effort than here, because likely many tests would need to get changed if the config values change.  The types.ts file should be kept because that will be used by the NPM package when it gets deployed.

coreyarnold (approved):
Change looks good. 🚀

bjfield (approved):

PR #431, opened 2023-03-09
Fix offset time - NR-96691 - NR-57391

metal-messiah:
### Overview

Situation:

The agent sets an "offset" at the beginning of the lifecycle of the page. It uses this "offset" to determine when events happen on the page relative to the beginning of the page lifecycle.  In the old agent architecture, the reading of on-page config values happened AFTER the loader had initialized.  It's during this time, that the offset was calculated.  The new agent architecture reads the on-page configs BEFORE the loader initializes (and AFTER, if values are missing).  This means that an invalid default "offset" can now be stored on the runtime object BEFORE it has been appropriately calculated, creating an invalid offset usually by 0-200 ms.  This can cause the early-page timings to be negative.

Replication:

Build the agent locally, run `npm run test-server` and navigate to a page that tracks session_traces.  Check the session_traces payload, checking the first timings in the payload for negative "s" and "e" values.  You may need to refresh a few times.

AC:

The offset must be calculated and assigned to the runtime object BEFORE the rest of the agent runs.

Tests should be added to ensure that no timing values are negative in the following features:

session_traces
page_view_timing
page_view_event

### Related Issue(s)

NR-96691
NR-57391

### Testing
New tests have been added to ensure that RUM, PVT, and PVE timings do not contain negative values.


metal-messiah:
### Before
before
### After
after

src/common/timing/start-time.js
```diff
@@ -6,31 +6,29 @@
...
 // navigationStart is deprecated, but can still be used as a fallback
 // http://test.w3.org/webperf/specs/NavigationTiming/
 // http://blog.chromium.org/2010/07/do-you-know-how-slow-your-web-page-is.html
 function findStartWebTiming () {
   // note that we don't need to use a cookie to record navigation start time
   navCookie = false
   const timestamp = globalScope?.performance?.timeOrigin || globalScope?.performance?.timing?.navigationStart
-  if (!timestamp) return
-  return Math.round(timestamp)
+  if (!timestamp) return getLastTimestamp()
```
patrickhousley, line 32: Should the result of `getLastTimestamp` also go through `Math.floor`?

patrickhousley (approved):
Tested locally and working. Just one comment more as a question.

src/common/timing/start-time.js
metal-messiah: getLastTimestamp uses `new Date().getTime()` at its core, which only returns integers.  But for stability, I think it might be wise to just ensure that all these numbers get rounded.  Good call, will add a commit

src/common/timing/start-time.js
```diff
@@ -6,31 +6,34 @@
...
 // navigationStart is deprecated, but can still be used as a fallback
 // http://test.w3.org/webperf/specs/NavigationTiming/
 // http://blog.chromium.org/2010/07/do-you-know-how-slow-your-web-page-is.html
 function findStartWebTiming () {
   // note that we don't need to use a cookie to record navigation start time
   navCookie = false
-  const timestamp = globalScope?.performance?.timeOrigin || globalScope?.performance?.timing?.navigationStart
-  if (!timestamp) return
-  return Math.round(timestamp)
+  const timestamp = globalScope?.performance?.timeOrigin || globalScope?.performance?.timing?.navigationStart || getLastTimestamp()
```
cwli24, line 36: May I suggest doing this inside `now.js` file instead? I'm touching the same area with offset, and also am removing this function in the cwv branch

cwli24: Aside: `findStartTime()` is actually doing the same job as now.js with `setOffset` so it's also good to just trim one of them or use now.js inside findStartTime

cwli24:
Per slack discussion, this should be updated to coalesce all timing "offset" into `runtime.offset` and merge in NR-57391 work.

src/common/util/submit-data.js
```diff
@@ -20,7 +20,7 @@ submitData.jsonp = function jsonp (url, jsonp) {
         // for now theres no other way to execute the callback from ingest without jsonp, or unsafe eval / new Function calls
         // future work needs to be conducted to allow ingest to return a more traditional JSON API-like experience for the entitlement flags
         submitData.xhrGet(url + '&jsonp=' + jsonp)
```
cwli24, line 22: RE: existing code, was there a reason behind using `importScript` in this try-catch rather than just XHR regardless of env? All workers have XHR but not all have `importScript` function--can classic worker not just use XHR too?

src/features/page_view_event/aggregate/index.js
```diff
@@ -64,7 +64,7 @@ export class Aggregate extends AggregateBase {
 
     if (globalScope.performance && typeof (globalScope.performance.timing) !== 'undefined') {
       var navTimingApiData = ({
-        timing: addPT(globalScope.performance.timing, {}),
+        timing: addPT(getRuntime(this.agentIdentifier).offset, globalScope.performance.timing, {}),
```
cwli24, line 67: Very minor: there's a `agentRuntime` var above that's unused rn.

cwli24 (approved):
❤️ it

src/common/util/submit-data.js
metal-messiah: importScripts allows the JSONP response to execute (the flags returned by the rum response)

cwli24: Ah ok, just note that still doesn't work for module scripts though. Is the xhr in catch pointless (atm) then if it fails?

cwli24: Nvm, it's the whole "RUM needed to refresh index" isn't it.

metal-messiah: yeah, basically it tries to use a method that could utilize the jsonp response, but if it can't it falls back to just making the call to set the indexing and capture a "blind" PVE in that it can't get use the setToken call in a safe way

src/features/page_view_event/aggregate/index.js
metal-messiah: fixed, ty

PR #434, opened 2023-03-14
Prepare to Publish Agent to NPM - NEWRELIC-7455 NEWRELIC-7476

metal-messiah:
### Overview

This PR intends to do the following:
* Set up mechanisms to build and publish the browser agent repo to NPM under `@newrelic/browser-agent`
* Add testing assets to validate NPM builds locally (manually for now, to be used for automated testing later)

**note:** To support this, symlinks for `JIL` have been replaced with relative paths to the assets.

### Related Issue(s)
NEWRELIC-7455
NEWRELIC-7476

### Testing
Manual testing assets have been added under `tools/test-builds`.  Run `npm run build:all` to generate the test assets in the `tests/assets/test-builds` directory.  Run `npm run test-server` to interact with them.


package.json
```diff
@@ -6,26 +6,76 @@
   "description": "Tests for the New Relic JavaScript agent",
   "license": "Apache-2.0",
   "homepage": "https://docs.newrelic.com/docs/browser/browser-monitoring",
-  "main": "",
+  "main": "./dist/cjs/index.js",
   "types": "./types.ts",
-  "module": "./dist/mjs/index.mjs",
+  "module": "./dist/esm/index.js",
```
cwli24, line 11: `module` field proposal rejected (superceded by another proposal field `browser`) -- 
Can be removed with newer `exports` taking precedence over `main` (Node <=v10) / `module` / `browser` ~esbuild.github.io/api/#main-fields

src/index.js
```diff
@@ -1,29 +1,15 @@
-/**
- * @file Exposes the agent and available features in multiple forms.
- */
+export { CustomAgent } from './custom-agent'
+export { BrowserAgent } from './browser-agent'
+export { WorkerAgent } from './worker-agent'
```
cwli24, line 3: Can we not have these onion layers whose only purpose is to export another file of same name in the `loaders` folder? Rather just import from loader/___ directly.

package.json
```diff
@@ -6,26 +6,76 @@
...
+      "import": "./dist/esm/features/session_trace/index.js",
+      "require": "./dist/cjs/features/session_trace/index.js"
+    },
+    "./features/spa": {
+      "import": "./dist/esm/features/spa/index.js",
+      "require": "./dist/cjs/features/spa/index.js"
     }
   },
+  "engines": {
+    "node": ">=12.16.0 < 13.0.0 || >=13.7.0"
```
patrickhousley, line 67: This should be `>=12.17.0` https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions
```suggestion
    "node": ">=12.17.0 < 13.0.0 || >=13.7.0"
```

babel-env-vars.js
```diff
@@ -22,27 +22,38 @@ const VERSION = fs.readFileSync('./VERSION', 'utf-8')
...
  * @see https://babeljs.io/docs/en/babel-plugin-transform-inline-environment-variables
  */
-module.exports = ({ source, subversion, isSemver = true } = {}) => {
-  setBuildVersion(source, subversion, isSemver)
-  setBuildEnv(subversion)
+module.exports = ({ source, subversion, isSemver = true, distMethod = 'CDN' } = {}) => {
+  const envVarNames = []
+  envVarNames.push(setBuildVersion(source, subversion, isSemver))
+  envVarNames.push(setBuildEnv(subversion))
+  envVarNames.push(setDistMethod(distMethod))
```
cwli24, line 29:
```suggestion
  const envVarNames = [setBuildVersion(source, subversion, isSemver), setBuildEnv(subversion), setDistMethod(distMethod)]
```

package.json
metal-messiah: Does this still need resolution after the discussion earlier?

src/common/config/state/runtime.js
```diff
@@ -4,12 +4,13 @@ import { gosNREUMInitializedAgents } from '../../window/nreum'
 import { getCurrentSessionIdOrMakeNew } from '../../window/session-storage'
 import { getConfigurationValue } from '../config'
 import { globalScope } from '../../util/global-scope'
-import { BUILD_ENV, VERSION } from '../../constants/environment-variables'
+import { BUILD_ENV, DIST_METHOD, VERSION } from '../../constants/environment-variables'
```
cwli24, line 7: Since these are set at compile time.. do they belong in the `runtime` object, or does `info` make more sense?

metal-messiah: We should probably work to standardize the configs in the future, but imo 

info should be user configs related to account and agent data

init should be configurations of the agent and features 

runtime should be a place where state is kept and shared 

Definitely open to other ideas

package.json
metal-messiah: I think module was just included for bundlers that utilize it, but since we are sticking to requiring node versions that support export, I'm not opposed to removing it. I will just do that for now

src/common/config/state/runtime.js
cwli24: By that, it sounds like these should go in `init` then (`runtime` in my head are dynamic internal fields determined at run time--nothing statically known should be in there), but not important atm so resolving this.

src/loaders/api/api.js
```diff
@@ -26,15 +28,20 @@ function setTopLevelCallers (nr) {
...
+    return returnVals.length > 1 ? returnsVals : returnVals[0]
   }
 }
 
-export function setAPI (agentIdentifier, nr, forceDrain) {
+export function setAPI (agentIdentifier, forceDrain) {
   if (!forceDrain) registerDrain(agentIdentifier, 'api')
-  setTopLevelCallers(nr)
+  const apiInterface = {}
+  setTopLevelCallers()
```
cwli24, line 44: `setTopLevelCallers` is global scope while `setAPI` is agent scope. Can they be separated, and have _configure.js_ make the call to `setTopLevelCallers`?

tools/test-builds/browser-agent-wrapper/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "webpack --config webpack.config.js --mode=production --node-env=production"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@newrelic/browser-agent": "file:../../../temp/newrelic-browser-agent-0.1.229.tgz"
```
cwli24, line 12: Having to (remember) update the numbers here each version is going to be annoying, no?

tools/test-builds/browser-agent-wrapper/webpack.config.js
cwli24, line 1: Can this be tuned up? I'd prefer having our top-level _webpack.config_ include these builds, then have them injected into _instrumented.html_ or some common shell page like many of the existing tests. There's (too) many moving parts, duped code, and it's messy.

tools/test-builds/browser-agent-wrapper/webpack.config.js
```diff
@@ -0,0 +1,166 @@
...
+    <h1>This is a generic page that is instrumented by the NPM agent</h1>
+  </body>
+</html>`
+const workerHtmlTemplate = `<html>
+  <head>
+    <title>RUM Unit Test</title>
+    {init}
+    {config}
+    {worker-commands}
+    {script-injection}
```
cwli24, line 23: `script-injection` should go below the loader, like existing instrumented.html

cwli24 (commented):
A few more readability & maintenance things

tools/test-builds/browser-agent-wrapper/webpack.config.js
patrickhousley: There is no loader in the worker example html. The worker is loaded in the worker script, not the page.

cwli24: Meant below worker script in any case

patrickhousley: can you check now and see if I got it corrected

tools/test-builds/browser-agent-wrapper/package.json
metal-messiah: It automatically does it...

src/loaders/browser-agent.js
```diff
@@ -14,7 +14,7 @@ import { Instrument as InstrumentPageAction } from '../features/page_action/inst
  * The BrowserAgent class is the most convenient and reliable option for most use cases.
  */
 export class BrowserAgent extends Agent {
-  constructor (...args) {
+  constructor (args) {
```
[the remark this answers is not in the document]
metal-messiah: ty

src/loaders/api/api.js
metal-messiah: Yes i think that sounds reasonable

metal-messiah: fixed

tools/test-builds/browser-agent-wrapper/webpack.config.js
metal-messiah: Agree, this would be a larger effort than this PR.

src/loaders/api/api.js
```diff
@@ -26,15 +28,20 @@ function setTopLevelCallers (nr) {
...
 
   function caller (fnName, ...args) {
+    let returnVals = []
     Object.values(nr.initializedAgents).forEach(val => {
-      if (val.exposed && val.api[fnName]) val.api[fnName](...args)
+      if (val.exposed && val.api[fnName]) {
+        returnVals.push(val.api[fnName](...args))
+      }
     })
+    return returnVals.length > 1 ? returnsVals : returnVals[0]
```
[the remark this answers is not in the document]
metal-messiah: These are values that will return to customer API calls, which would break a lot of assumptions if we suddenly always returned an array.

tools/test-builds/browser-agent-wrapper/webpack.config.js
metal-messiah: https://issues.newrelic.com/browse/NR-102558

bjfield (approved):
The NPM package works as expected in a from-scratch front-end app scenario to compose an agent from desired features.

src/loaders/api/api.js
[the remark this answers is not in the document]
cwli24: Ah, right. In that case, they'd probably still expect a single value even with multiple initialized agents, but that's for later.

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## v1.229.0
+
+### Use semantic versioning scheme
+The agent will now utilize semantic versioning for subsequent releases. The previous version -- 1228 -- will carry forward into 1.229.0, and so on.
+
+### Use `web-vitals` library for internal timing calculations
+In an effort to align and standardize timings collecthe agent now uses the `web-vitals` library to track page timings such as CLS, LCP, FCP, TTFB and more. See https://www.npmjs.com/package/web-vitals for more information.
```
cwli24, line 12: `collecthe` - typo.

Also instead of "see https:// ....", you can just link to their repo like:
```[web-vitals](https://github.com/GoogleChrome/web-vitals)```

cwli24 (approved):

PR #435, opened 2023-03-15
Replace CWV metrics with GoogleChrome/web-vitals library NEWRELIC-6214

cwli24:
### Overview

Following an internal proposal, this work replaces existing implementation of the web vitals instrumentation with the [web-vitals](https://github.com/GoogleChrome/web-vitals) library API. This offloading makes our agent more robust and up-to-date in capturing the metrics and lets us rely on the BF cache and pre-rendering considerations, plus other side cases per metric, built into web-vitals.

Note that at this time however, we intentionally don't report timings again on BF restores, so that speed improvement is not factored into the dashboard on NR UI. The other significant or noteworthy change is deferral of loader code in the PVEvents & PVTiming features into post-window load event, which greatly reduced the size of our inline loader snippets.

### Related Issue(s)

NEWRELIC-4916
NEWRELIC-6214
NR-67469

### Testing
Web vitals matchers were recalibrated to have timings test run against the right supported versions.


src/features/page_view_event/aggregate/index.js
```diff
@@ -97,23 +97,19 @@ export class Aggregate extends AggregateBase {
       chunksForQueryString.push(param('perf', stringify(navTimingApiData)))
     }
 
-    if (globalScope.performance && globalScope.performance.getEntriesByType) {
+    try { // PVTiming sends these too, albeit using web-vitals and slightly different; it's unknown why they're duplicated, but PVT should be the truth
```
metal-messiah, line 100: This adds FP and FCP to the `PageView` event, while PVT adds it to the `PageViewTiming` events.  If web-vitals is intended to be the truth, this needs to get updated to use web-vitals here if it can (and fallback like PVT does)

cwli24: If we make it so either isn't aware which one runs first (and so creates the single value in `paintMetrics`), it gets ugly-looking real quick with refactoring onFirstPaint & onFCP out into its own module and passing more layers of callbacks. In majority cases, they'll end up with the same value as-is anyhow. I'm trying not to complicate it.

src/features/page_view_timing/aggregate/index.js
```diff
@@ -44,12 +45,24 @@ export class Aggregate extends AggregateBase {
...
-
-      this.addTiming(name.toLowerCase(), value)
-    })
+    if (iOS_below16) {
+      try {
+        if (!pageStartedHidden) { // see ios-version.js for detail on this following bug case; tldr: buffered flag doesn't work but getEntriesByType does
+          const paintEntries = performance.getEntriesByType('paint')
+          paintEntries.forEach(entry => {
+            if (entry.name === 'first-contentful-paint') {
+              this.addTiming('fcp', Math.floor(entry.startTime))
```
metal-messiah, line 54: Do we need to capture `fp` here as well?

cwli24: `first-paint` wasn't (and may still not be) supported in safari (or at least our past testing), so going to be a nope.

src/features/page_view_event/aggregate/index.js
metal-messiah: Can't you just call `onFCP(...)` in the PVE Agg as well?  That's what I was picturing for this

metal-messiah: If it's a hairy process lets just add a cleanup ticket for later. NBD

metal-messiah (approved):
Once you feel like the comments are resolved, you have my blessing

src/features/page_view_event/aggregate/index.js
cwli24: I suppose we _could_ but also their API isn't exactly synchronous, so we'd have to add an async-ness into sendRum. Still not a big deal but pointing out it's ugly with duplicate-looking but also not easily refactor-able code.

metal-messiah: > Lets just add a cleanup/consolidation ticket for later. NBD -- 👍

cwli24: Yeah, will add a ticket for this, because my attempt turned it into this monstrosity, and it didn't even pass `tests/functional/rum/paint-timing.test.js`... 
image

cwli24: https://issues.newrelic.com/browse/NR-101330

src/features/page_view_event/instrument/index.js
```diff
@@ -1,35 +1,38 @@
...
-    super(agentIdentifier, aggregator, FEATURE_NAME, auto)
+    super(agentIdentifier, aggregator, CONSTANTS.FEATURE_NAME, auto)
 
-    mark(agentIdentifier, 'starttime', getRuntime(agentIdentifier).offset)
-    mark(agentIdentifier, 'firstbyte', importTimestamp)
+    if (typeof PerformanceNavigationTiming !== 'undefined' && !isiOS) {
+      /* This block has been moved to the aggregate portion which is fetched and executed after page
+          load in an effort to reduce loader bundle size.
+      */
+    } else if (typeof PerformanceTiming !== 'undefined') {
```
patrickhousley, line 20: Can we clean this logic block up?

cwli24: Squashed into old else-if

patrickhousley (approved):
Nice job

PR #436, opened 2023-03-15
Fixed incorrect passing of options to BrowserAgent and WorkerAgent

e965:
### Overview

Obviously, the idea is that `args` will be an `options` object, and we pass it along with the spread operator. Unfortunately, JS [works a little differently](https://javascript.info/rest-parameters-spread), and `args` is an array of function arguments. Therefore, it becomes impossible to initiate new agent instances with the `options` that we are trying to pass there..

### DIFF
--- a/src/loaders/browser-agent.js
+++ b/src/loaders/browser-agent.js
@@ -1,5 +1,6 @@
 import { Agent } from './agent'
 
+import { generateRandomHexString } from '../common/ids/unique-id'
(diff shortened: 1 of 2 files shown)


metal-messiah:
Thanks for contributing this!  These classes will eventually be utilized in our upcoming initiative to release the agent repo to NPM. Thanks for finding and reporting this issue.

e965:
@metal-messiah Glad to hear it. I guess you can close the PR if you're going to fix this problem yourself.

PR #437, opened 2023-03-17
Run a basic test against <all> browsers - NEWRELIC-7583

metal-messiah:
### Overview
This PR adds a step in the sauce:get-browsers script to build a file of all desktop browsers.  It also adds a very basic test that can be run against by all browsers. It checks if there was an error on the page (from the loader), and then checks if a subsequent script can still run.

This is a basic test, *intended to be expanded upon over time*.

### Related Issue(s)

NEWRELIC-7583

### Testing

A new test has been added, a new browser definition file has been added (browsers-all.json), and a new param `-A` has been added to the server params, to run against `browsers-all.json`.  This runs against about ~250 browsers.  It is meant to include things to aid our understanding of the impact of running the agent in ALL browsers, regardless if supported or not, and should not necessarily block development, but aid in future decision making and communication.


bjfield (commented):
When we are clear to rebase, let me know and I'd be happy to do that and add a `graceful-error-test.yml`.

.github/workflows/tests-all.yml
```diff
@@ -0,0 +1,37 @@
+name: Smoke Test
```
patrickhousley, line 1: We already have a workflow with this name. This might cause issues in GHA.

tests/assets/graceful-error.html
```diff
@@ -0,0 +1,24 @@
...
+      window.onerror = function () {
+        window.errorSeen = true;
+      };
+    </script>
+    {init} {config} {loader} {script-injection}
+  </head>
+  <body>
+    this is a generic page that is instrumented by the JS agent
+    <script>
+      window.someOutput = true;
```
cwli24, line 21: This executes before the agent fully inits. If you want to be doubly sure, this should go in a `load` listener **+ some `setTimeout`**.

.github/workflows/tests-all.yml
metal-messiah:
```suggestion
name: Test All
```

tests/assets/graceful-error.html
metal-messiah:
```suggestion
      setTimeout(function(){window.someOutput = true;}, 1000)
```

bjfield (approved):

.husky/pre-commit
```diff
@@ -5,14 +5,15 @@
 export FORCE_COLOR=1
 
 npm run lint
-npm run unit -- --onlyChanged
+npm test
```
bjfield, line 8:
```suggestion
npm run test -- --onlyChanged
```

tests/assets/graceful-error.html
cwli24: That's still before load happens, no?

PR #438, opened 2023-03-17
Handle sendBeacon more gracefully - NR-95365

metal-messiah:
### Overview

Illegal Invocation errors still pop up from time to time on sendBeacon calls, even after work to bind scope to the beacon caller. There is nothing more we can do to get around this, besides failing gracefully and falling back to another call type.  This PR addresses that workflow.

### Related Issue(s)
Nr-95365

### DIFF
--- a/src/common/util/submit-data.js
+++ b/src/common/util/submit-data.js
@@ -84,7 +84,7 @@ submitData.img = function img (url) {
  * Send via sendBeacon. Do NOT call this function outside of a guaranteed web window environment.
  * @param {string} url
  * @param {string} body
- * @returns {function}
+ * @returns {boolean}
  */
 submitData.beacon = function (url, body) {
   // Navigator has to be bound to ensure it does not error in some browsers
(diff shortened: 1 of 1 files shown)


patrickhousley (approved):

src/common/util/submit-data.js
```diff
@@ -89,6 +89,13 @@ submitData.img = function img (url) {
 submitData.beacon = function (url, body) {
   // Navigator has to be bound to ensure it does not error in some browsers
   // https://xgwang.me/posts/you-may-not-know-beacon/#it-may-throw-error%2C-be-sure-to-catch
-  const send = window.navigator.sendBeacon.bind(navigator)
-  return send(url, body)
+  const send = window.navigator.sendBeacon.bind(window.navigator)
```
bjfield, line 92: Is the change of bind value to `window.navigator` a just-in-case-probably-should? Or does the scope matter somehow to the catch/fallback?

bjfield (approved):
Related tests pass locally and fallback logic (to harvest.js:143) looks correct.

PR #441, opened 2023-03-28
NEWRELIC-6739: Adding other asset sizes to report

patrickhousley:
### Overview

This adds two other assets sections to the PR report that list out the released and built sizes for non-loader assets. This should make it easier for us to maintain the performance docs until we automate those doc changes.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-6739

### Testing

Let the CI run and check the report comment on this PR.

### DIFF
--- a/tools/scripts/diff-sizes.mjs
+++ b/tools/scripts/diff-sizes.mjs
@@ -124,6 +124,17 @@ function generateDiffRows (assetSizes, buildType) {
   }).join('\n')
 }
 
+function generateOtherDiffRows (assetSizes, buildType, assetGroup) {
+  const targetBuildTypeSizes = assetSizes[buildType]
+  return Object.keys(targetBuildTypeSizes[assetGroup])
+    .filter(assetName => assetName.indexOf('nr-loader') === -1)
+    .map(assetName => {
+      const buildSizes = targetBuildTypeSizes[assetGroup][assetName]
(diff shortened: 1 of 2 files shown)


metal-messiah (approved):

tools/scripts/diff-sizes.mjs
```diff
@@ -136,7 +147,39 @@ ${generateDiffRows(assetSizes, 'standard')}
 ${generateDiffRows(assetSizes, 'polyfills')}
 ${generateDiffRows(assetSizes, 'worker')}
 
-<sub>This report only accounts for the minified loaders. Un-minified and async-loaded assets are not included.</sub>
+<details>
+<summary>Other Standard Assets</summary>
+
+## Released Assets
+
+| Asset Name | Asset Size |
```
cwli24, line 155: Can the "previous" and "new" sizes be in the same table like it is for the loaders? (Calculating diff isn't necessary.) Make it easier to read.

cwli24, line 155: Can the "previous" and "new" sizes be in the same table like it is for the loaders? (Calculating diff isn't necessary.) Make it easier to read.

patrickhousley: Not without a lot of extra work to find match the asset names across releases

patrickhousley: Not without a lot of extra work to find match the asset names across releases

cwli24 (approved):

PR #443, opened 2023-03-28
Add standard setUserId & expand setErrorHandler|setCustomAttribute APIs

cwli24:
### Overview

Per Errors Inbox work requests, this PR adds a standard way to tag user IDs onto outgoing payloads and for customers to provide their own fingerprinting for errors grouping.

In addition, session storage usage is buffed up with methods to store, and on subsequent page loads retrieve, custom attributes within the same tab session.

### Related Issue(s)

Features:
NEWRELIC-7755
NEWRELIC-7798
NEWRELIC-7887
NEWRELIC-7890
Bugs:
NR-100186

### Testing

Functional test cases added:

- `setUserId` appends/overwrites `enduser.id` on jserrors payloads, custom
- `setUserId` emits (new) supportability metric string on call
- `setErrorHandler` includes errors with the correct `group` label when retval is correctly structured, and exclude when `group` is an empty string
- `setUserId` persists and `setCustomAttribute` can persist in session storage
    - a `null` value provided to either of the them removes the former value


src/loaders/api/api.js
```diff
@@ -50,7 +50,7 @@ export function setAPI (agentIdentifier, nr, forceDrain) {
   var spaPrefix = prefix + 'ixn-'
 
   // Setup stub functions that queue calls for later processing.
-  mapOwn(asyncApiFns, function (num, fnName) {
+  asyncApiFns.forEach(fnName => {
```
metal-messiah, line 53: If we are going to arbitrarily rework using `mapOwn`, we should remove it completely from the file.

src/loaders/api/api.js
```diff
@@ -61,13 +61,23 @@ export function setAPI (agentIdentifier, nr, forceDrain) {
...
+    return apiCall(prefix, 'setPageViewName', true)()
   }
 
-  nr.setCustomAttribute = function (name, value) {
+  function appendJsAttribute (key, value, apiName = 'setCustomAttribute') {
     const currentInfo = getInfo(agentIdentifier)
-    setInfo(agentIdentifier, { ...currentInfo, jsAttributes: { ...currentInfo.jsAttributes, [name]: value } })
-    return apiCall(prefix, 'setCustomAttribute', true, 'api')()
+    setInfo(agentIdentifier, { ...currentInfo, jsAttributes: { ...currentInfo.jsAttributes, [key]: value } })
+    return apiCall(prefix, apiName, true)()
```
metal-messiah, line 70: just to confirm, was `api` buffer group removed because there is no async consumer of the ee buffer?

src/loaders/api/apiAsync.js
```diff
@@ -24,9 +24,9 @@ export function setAPI (agentIdentifier) {
   }
 
   // Hook all of the api functions up to the queues/stubs created in loader/api.js
-  mapOwn(api, function (fnName, fn) {
-    registerHandler('api-' + fnName, fn, 'api', instanceEE)
-  })
+  for (const [fnName, fnCall] of Object.entries(api)) {
```
metal-messiah, line 27: can we remove the `mapOwn` import if it is being removed?

tests/functional/api.test.js
```diff
@@ -352,3 +352,41 @@ testDriver.test('api is available when sessionStorage is not', function (t, brow
...
+    init: {
+      jserrors: {
+        harvestTimeSeconds: 2
+      }
+    },
+    scriptString: `
+    newrelic.setUserId(456);
+    newrelic.setUserId({'foo':'bar'});
+    newrelic.noticeError('fake1')
+    setTimeout(() => {  // there's a delay with errors grabbing the jsAttributes
```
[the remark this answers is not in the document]
cwli24: The comment? Without a `setTimeout`, what happens is that `fake1` error will have the `user123` ID. There's a race condition between `noticeError` and inner `appendJsAttribute`, in which the latter always win because it doesn't go thru the whole event-emitter system. So it looks sequential, but it really isn't; hence this way I can temporally isolate the two "sets" of tests.

src/loaders/api/api.js
cwli24: Righto

metal-messiah:
`setUserId emits (new) supportability metric string on call`

Can you please follow-up by submitting an Angler PR to consume this as well

metal-messiah:
> `setUserId emits (new) supportability metric string on call`
> 
> Can you please follow-up by submitting an Angler PR to consume this as well

Worth considering -- do we also want a SM created when an error gets a group assigned?

cwli24:
> > `setUserId emits (new) supportability metric string on call`
> > Can you please follow-up by submitting an Angler PR to consume this as well
> 
> Worth considering -- do we also want a SM created when an error gets a group assigned?

No, I don't see that count being relevant or useful, much like the numbers of excluded errors based on custom provided filter isn't noted today.

cwli24:
> `setUserId emits (new) supportability metric string on call`
> 
> Can you please follow-up by submitting an Angler PR to consume this as well

https://source.datanerd.us/agents/angler/pull/430

src/common/window/session-storage.js
```diff
@@ -6,34 +6,68 @@
  * @environment Browser script
  */
 import { generateRandomHexString } from '../ids/unique-id'
-import { isBrowserScope } from '../util/global-scope'
 
-export { getCurrentSessionIdOrMakeNew }
-
-const SESS_ID = 'NRBA_SESSION_ID' // prevents potential key collisions in session storage
+const AGENT_PREFIX = 'NRBA/'
```
metal-messiah, line 10: Do we care that the shape of this has changed?

`NRBA_SESSION_ID` to `NRBA/SESSION_ID`

Unclear if this outlined or defined in the docs anywhere in a concrete way.  If yes, we should match, if no, this is a non-issue.

src/features/jserrors/instrument/index.js
```diff
@@ -84,6 +84,8 @@ export class Instrument extends InstrumentBase {
    * @returns
    */
   onerrorHandler (message, filename, lineno, column, errorObj) {
+    if (typeof this.origOnerror === 'function') this.origOnerror(...arguments)
```
metal-messiah, line 87: I think this will need to be reverted, because if the customer defines their own handler to return `true` (ie to block the error from reaching the console) we should honor that.

src/loaders/api/api.js
```diff
@@ -2,20 +2,23 @@
...
+import { isBrowserScope, isWorkerScope } from '../../common/util/global-scope'
 import { warn } from '../../common/util/console'
 import { SUPPORTABILITY_METRIC_CHANNEL } from '../../features/metrics/constants'
+import { gosCDN } from '../../common/window/nreum'
+import { putInBrowserStorage, removeFromBrowserStorage } from '../../common/window/session-storage'
 
-function setTopLevelCallers (nr) {
+export const CUSTOM_ATTR_GROUP = 'CUSTOM/' // the subgroup items should be stored under in storage API
+
+export function setTopLevelCallers () {
```
metal-messiah, line 20: Is there a reason `setTopLevelCallers` got moved back to the API file? This fn keeps getting moved around PR to PR

src/loaders/api/api.js
```diff
@@ -64,20 +69,47 @@ export function setAPI (agentIdentifier, nr, forceDrain) {
...
+    } else {
+      setInfo(agentIdentifier, { ...currentInfo, jsAttributes: { ...currentInfo.jsAttributes, [key]: value } })
+      if (isBrowserScope && addToBrowserStorage) putInBrowserStorage(key, value, CUSTOM_ATTR_GROUP)
+    }
     return apiCall(prefix, apiName, true)()
   }
-  nr.setCustomAttribute = function (name, value) {
-    return appendJsAttribute(name, value)
+  apiInterface.setCustomAttribute = function (name, value, persistAttribute = false) {
+    if (typeof key !== 'string') {
```
metal-messiah, line 92: What is `key` here?

src/loaders/configure/configure.js
```diff
@@ -14,8 +14,13 @@ export function configure (agentIdentifier, opts = {}, loaderType, forceDrain) {
     loader_config = nr.loader_config
   }
 
+  info.jsAttributes ??= {}
   if (isWorkerScope) { // add a default attr to all worker payloads
-    info.jsAttributes = { ...info.jsAttributes, isWorker: true }
+    info.jsAttributes.isWorker = true
+  }
+  if (isBrowserScope) { // retrieve & re-add all of the persisted setCustomAttribute|setUserId k-v from previous page load(s)
+    let prevPageSessionJsAttrs = getAllStorageItemsOfGroup(CUSTOM_ATTR_GROUP)
```
metal-messiah, line 22: We are basically saying here that `sessionStorage` custom attributes take precedence over passed-in info.jsAttributes definitions.  Is this expected?

tests/functional/api.test.js
```diff
@@ -353,7 +354,29 @@ testDriver.test('api is available when sessionStorage is not', function (t, brow
...
+    })
+    .then(() => Promise.all([router.expectRum(), loadPromise])) // testing should've been erased from storage
+    .then(([{ request: { query } }]) => {
+      t.equal(query.ja, undefined, '3rd page load does not retain custom attribute after unsetting (set to null)')
+      t.end()
+    })
+    .catch(fail(t))
+})
+
+testDriver.test('setUserId adds correct (persisted) attribute to payloads', withUnload.and(notIE), function (t, browser, router) {
```
metal-messiah, line 379: Just for the record, does this not work with IE11?

metal-messiah: sessionStorage should work for IE so I am curious

src/common/window/session-storage.js
cwli24: Nope, our internal implementation detail

src/features/jserrors/instrument/index.js
cwli24: >if the customer defines their own handler to return `true`

You mean if they edit `window.onerror`, and not via `setErrorHandler(func)`? 
This is still a function:
```javascript
window.onerror = function() { return true }
```

src/loaders/api/api.js
cwli24: Yeah, aside from it being in the file still from the branch parent, that contains a list of API functions and would be more reasonably found in _api.js_ where all API-related setup lives.

tests/functional/api.test.js
cwli24: It couldn't functionally retrieve it from sessionStorage on `browser.refresh()`. Uncertain if it's a WD/Sauce issue or the api, but since IE still otherwise work, I kicked it aside.

src/loaders/configure/configure.js
cwli24: Wasn't on my mind, but now considering it, yes..? If someone wanted a custom attribute persisted, that implies it should have precedence and show up until such time they manually unset or change it.

It'll probably be one of divided opinion though. If I'm dev on team A working page A and I set it to persist, yes I'd want that value on refresh of page A over whatever my original info is. But if I'm dev on team B working on page B, and user navigates from A to B, I may not. So either,
- `persist` should only apply to the same page, rather than tab session, but this may defeat the perspective of tracking users thru multiple pages, OR
- team A takes the responsible of managing the unsetting themselves before user navigates to page B

src/common/window/session-storage.js
```diff
@@ -6,34 +6,68 @@
...
+/**
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem}
+ * @param {string} key
+ */
+function getFromBrowserStorage (key, subgroup = '') {
+  return window.sessionStorage.getItem(AGENT_PREFIX + subgroup + key)
+}
+export function removeFromBrowserStorage (key, subgroup = '') {
+  return window.sessionStorage.removeItem(AGENT_PREFIX + subgroup + key)
+}
```
patrickhousley, line 35: These function will need `try..catch` blocks. Safari throws an exception any time you try to use `sessionStorage` in a private browser window.

image

src/common/window/session-storage.js
```diff
@@ -6,34 +6,68 @@
...
 /**
- * @returns {string} This tab|window's session identifier, or a new ID if not found in storage
+ * @param {string} key
+ * @param {string} value
+ * @returns true, if item was set successfully. false, if storage is full or setItem threw an exception.
  */
-function getCurrentSessionIdOrMakeNew () {
-  if (!isBrowserScope) {
-    return null
+export function putInBrowserStorage (key, value, subgroup = '') {
```
patrickhousley, line 17: I am honestly debating if it merits making a storage wrapper or using something like [localforage](https://github.com/localForage/localForage).

src/features/jserrors/instrument/index.js
metal-messiah: Yes if they edit onerror before the agent is set, we should honor their return value (true/false)

What you have is always returning false

src/common/window/session-storage.js
```diff
@@ -6,34 +6,68 @@
...
+/**
+ * Look through the page's sessionStorage items and gather the ones that was set under some grouping label, if applicable.
+ * @param {string} [subgroup] - empty string by default, which will return all items set by browser agent
+ * @returns An object of the key-value items from putInBrowserStorage calls from agents with access to the same storage space.
+ */
+export function getAllStorageItemsOfGroup (subgroup = '') {
+  const prefixFilter = AGENT_PREFIX + subgroup
+  const groupItems = {}
+  try {
+    for (let i = 0; i < window.sessionStorage.length; i++) {
```
patrickhousley, line 64: If this method is called from a worker, it will throw an exception. Should we try to use a storage medium that can be supported from a worker?

src/common/window/session-storage.js
metal-messiah: Was having the same thoughts. Session replay May end up changing our definitions of how we manage a session so we should iterate on this concept. I don't want to bloat the scope here but it's on the to-do list

src/loaders/api/api.js
```diff
@@ -82,14 +141,14 @@ export function setAPI (agentIdentifier, forceDrain) {
     }
   }
 
-  mapOwn('actionText,setName,setAttribute,save,ignore,onEnd,getContext,end,get'.split(','), function addApi (n, name) {
+  void ['actionText', 'setName', 'setAttribute', 'save', 'ignore', 'onEnd', 'getContext', 'end', 'get'].forEach(name => {
```
patrickhousley, line 144: I don't think the `void` operator is necessary here.

src/loaders/api/apiAsync.js
```diff
@@ -24,9 +23,9 @@ export function setAPI (agentIdentifier) {
...
```
patrickhousley, line 26: Why not just use `Object.entries(api).forEach(...)`?

patrickhousley: https://jsbench.me/dplg3pwgei/1

tools/testing-server/plugins/agent-injector/script-transform.js
```diff
@@ -50,9 +50,21 @@ module.exports = function (request, reply, testServer) {
 
       // Replace {script-injection}
       if (chunkString.indexOf('{script-injection}') > -1) {
+        let scriptContent = ''
+        if (testServer.config.polyfills && request.query.scriptString) { // if the test is running in IE11, the test scriptString may need Babel transpilation
```
patrickhousley, line 54: Shouldn't need to check for scriptString in the query object. The logic check can be a little confusing to look at for those not familiar with the code.

tools/testing-server/plugins/browserify/browserify-transform.js
```diff
@@ -31,6 +31,8 @@ function browserifyScript (scriptPath, enablePolyfills) {
           '@babel/plugin-syntax-dynamic-import',
           '@babel/plugin-transform-modules-commonjs',
           '@babel/plugin-proposal-optional-chaining',
+          '@babel/plugin-proposal-nullish-coalescing-operator',
+          '@babel/plugin-proposal-logical-assignment-operators',
```
patrickhousley, line 35: Should these same transforms be added to the webpack and npm build?

src/loaders/configure/configure.js
metal-messiah: Yeah I think stick with behavior as currently defined for now until feedback is gathered

tools/testing-server/plugins/browserify/browserify-transform.js
cwli24: https://caniuse.com/?search=null%20coalescing%20assignment I don't think they do; it just isn't supported under Node v15

tools/testing-server/plugins/agent-injector/script-transform.js
cwli24:
```
const scriptStream = Readable.from(Buffer.from(request.query.scriptString, 'base64'))
```
Have trouble handling empty strings

src/loaders/api/api.js
cwli24: Eslint 🙃

src/common/window/session-storage.js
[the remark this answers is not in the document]
cwli24: Yes, but their speed is the same:
image
and that's the [official way of accessing](https://developer.mozilla.org/en-US/docs/Web/API/Storage/key) it -- unknown if `Object.keys(sessionStorage)` is widely acceptable.

cwli24: only called by _configure.js_ which uses `isBrowserScope`

src/common/window/session-storage.js
cwli24: Esp, `removeFromBrowserStorage` which doesn't have a catch in outer parts, thanks.

.eslintrc.js
```diff
@@ -78,6 +78,7 @@ module.exports = {
   ],
   rules: {
     'no-unused-vars': 'off',
+    'no-throw-literal': 0,
```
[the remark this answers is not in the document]
patrickhousley: The problem is this is applying to all files and not just the test files. If you need to apply a rule to just the tests, add the rule to the overides block that applies to the test files (line 61). I think you will need to update the files list to add the jil test files.

patrickhousley (approved):

.eslintrc.js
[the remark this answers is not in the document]
metal-messiah: Or just throw an error there instead of a string, but yeah this works.

src/features/jserrors/instrument/index.js
metal-messiah: ok in that case we will just keep an eye on this and iterate in the future if customers need a way to return "true" out of the top-level `onerror` method :+1:

metal-messiah (approved):
Pending smoke test and husky merge conflict, looks great. 👍 

Smoke test -- https://github.com/newrelic/newrelic-browser-agent/actions/runs/4640661193

PR #447, opened 2023-03-30
chore: Documenting unit testing

patrickhousley:
### Overview

Documentation for writing unit tests using jest.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7869

### DIFF
--- a/docs/unit-testing.md
+++ b/docs/unit-testing.md
@@ -32,7 +32,7 @@ Each test case should focus on testing a single scenario. This could be a single
 
 ## Importing source files
 
-There are two ways to import the source fill under test. The recommended way is to use an `import` at the top of the test file. However, when the source file under test is not stateless, you will need to use an async `import` statement within the test case or within a `beforeEach`.
+There are two ways to import the source file under test. The recommended way is to use an `import` at the top of the test file. However, when the source file under test is not stateless, you will need to use an async `import` statement within the test case or within a `beforeEach`.
 
 - **DO** use `import` at the top of the test file to pull in stateless source files
 - **DO** use `await import()` inside a test case or `beforeEach` to pull in a stateful source file
--- a/docs/unit-testing.md
+++ b/docs/unit-testing.md
@@ -4,7 +4,7 @@ We utilize [Jest](https://jestjs.io/docs/getting-started) for the purposes of wr
 
 ## Running Unit Tests
 
-After making sure you have the project dependencies installed, executing `npm test` will run all the unit tests. To run just a specific test, you can pass the path of the test file to the test command like so `npm test -- src/features/jserrors/aggregate/format-stack-trace.test.js`. There is also a watch command setup `npm run test:watch` if you are making multiple edits to tests and would like those tests to run after each saved change. To check the state of coverage, use `npm run test:coverage`.
+After making sure you have the project dependencies installed, execute `npm test` to run all the unit tests. To run just a specific test, you can pass the path of the test file to the test command like so: `npm test -- src/features/jserrors/aggregate/format-stack-trace.test.js`. There is also a watch command (`npm run test:watch`) for when you are making multiple edits to tests and would like those tests to run after each saved change. To check the state of coverage, use `npm run test:coverage`.
 
 ## Writing Tests - Basics
 
--- a/docs/unit-testing.md
+++ b/docs/unit-testing.md
@@ -23,6 +23,7 @@ Once you have a test file created, follow these best practices for creating the 
 - **DO** use `afterEach` to cleanup global state after each test
 - **DONT** use `beforeAll` or `afterAll`
 - **DO** use the `test` keyword to declare a test case
+- **DO** avoid promise chaining and use `async/await` to test asynchronous code
 - **DO** limit the focus of a test to a single scenario
 
 ### Focus each test case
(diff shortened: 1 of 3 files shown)


docs/unit-testing.md
```diff
@@ -0,0 +1,59 @@
...
+### Focus each test case
+
+Each test case should focus on testing a single scenario. This could be a single set of inputs followed by a single set of `expect` clauses. Tests that instantiate or call the code under test multiple times with different `expect` clauses should typically be split up into individual test cases. This will ensure these tests run quickly and, when a test fails, isolating the failure is much easier.
+
+## Import vs Require
+
+There are two ways to pull in the source fill under test in the test file. The recommended way is to use an `import` at the top of the test file. However, when the source file under test is not stateless, the source file will need to be pulled in using `require` within the test case or within a `beforeEach`.
+
+- **DO** use `import` at the top of the test file to pull in stateless source files
+- **DO** use `require` inside a test case or `beforeEach` to pull in a stateful source file
```
cwli24, line 37: With `import` working inside test functions, users can just use all (static/dynamic) imports, no? I rather stick to one than mix-and-match

patrickhousley: I am not sure if an async import in the test case would work. Something to try. Keep in mind, if it does work, I am pretty sure it would be an async call. Not a big deal, just something to remember.

cwli24: image
I did this right before my comment to check myself before I rekt myself 😆

patrickhousley: HA! Yea I am fine with that. I will make the change.

patrickhousley: I would prefer we use `async/await` for testing async code and avoid the promise chaining hell.

metal-messiah (approved):

docs/unit-testing.md
```diff
@@ -0,0 +1,60 @@
...
+Once you have a test file created, follow these best practices for creating the individual test cases:
+
+- **DONT** create a global `describe` wrapper to hold all the tests
+- **DO** create a `describe` to group tests that have similar pre/post test instructions
+- **DONT** save state between individual test cases
+- **DO** write and assume that each test case will be ran in isolation, parallelized, and share no local or global state
+- **DO** use `beforeEach` to instantiate variables and state before each test
+- **DO** use `afterEach` to cleanup global state after each test
+- **DONT** use `beforeAll` or `afterAll`
+- **DO** use the `test` keyword to declare a test case
```
metal-messiah, line 25: Curious where these best practices came from

patrickhousley: These are best practices that we as a team should agree to. They came from my personal experience with writing jest tests.

bjfield: These seem pretty reasonable. Couple questions:
- Why not create a global describe wrapper? If a file happens to contain only one describe wrapper, that's sort of the same thing as a global one in my mind. Seems like the idea is to use describe wrappers only for tests that share setup/tear-down or a clear descriptive relation, and if in doubt split the tests out.
- Why not use beforeAll and afterAll? Because of state spanning test cases?

docs/unit-testing.md
```diff
@@ -0,0 +1,60 @@
...
+- **DO** avoid promise chaining and use `async/await` to test asynchronous code
+- **DO** limit the focus of a test to a single scenario
+
+### Focus each test case
+
+Each test case should focus on testing a single scenario. This could be a single set of inputs followed by a single set of `expect` clauses. Tests that instantiate or call the code under test multiple times with different `expect` clauses should typically be split up into individual test cases. This will ensure these tests run quickly and, when a test fails, isolating the failure is much easier.
+
+## Importing source files
+
+There are two ways to import the source fill under test. The recommended way is to use an `import` at the top of the test file. However, when the source file under test is not stateless, you will need to use an async `import` statement within the test case or within a `beforeEach`.
```
bjfield, line 35:
```suggestion
There are two ways to import the source file under test. The recommended way is to use an `import` at the top of the test file. However, when the source file under test is not stateless, you will need to use an async `import` statement within the test case or within a `beforeEach`.
```

docs/unit-testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+- **DO** use `import` at the top of the test file to pull in stateless source files
+- **DO** use `await import()` inside a test case or `beforeEach` to pull in a stateful source file
+- **DO** reset the jest module cache between test cases using `afterEach` when testing a stateful source file
```
bjfield, line 39:
```suggestion
- **DO** use `import` at the top of the test file to pull in **stateless** source files.
- **DO** use `await import()` inside a test case or `beforeEach` to pull in a **stateful** source file.
- **DO** [reset the Jest module cache](https://jestjs.io/docs/jest-object#jestresetmodules) between test cases using `afterEach` when testing a **stateful** source file.
```

docs/unit-testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+```js
+// Example for pulling in a stateful source file
+
+let methodUnderTest;
```
bjfield, line 44:
```suggestion
let methodUnderTest
```

bjfield (changes requested):
Thanks for doing this, Patrick. Suggested some edits.

patrickhousley:
> These seem pretty reasonable. Couple questions:
> - Why not create a global describe wrapper? If a file happens to contain only one describe wrapper, that's sort of the same thing as a global one in my mind. Seems like the idea is to use describe wrappers only for tests that share setup/tear-down or a clear descriptive relation, and if in doubt split the tests out.
> - Why not use beforeAll and afterAll? Because of state spanning test cases?

- A global `describe` wrapper would be putting all `test` cases into one large `describe`. It's unnecessary and adds fluff to the console output. The root of a test file can contain N number of `test`, `describe`, `beforeEach`, `afterEach`, etc. You are right on the use of a `describe`; Use it to share setup and tear-down code.
- You got on `beforeAll` and `afterAll`. Exceptions can be made but I would be highly skeptical of the creating of shared state within unit tests.

metal-messiah (approved):

bjfield (approved):

Issue #454, opened 2023-04-03
Problems when adding a package to a project


e965:
Hi! I'm trying to add your package to my application, and I ran into some difficulties

1) There are no d.ts declarations for the agent modules, so it was possible to import it only through `// @ts-ignore` comment

2) When trying to import from the package itself (`import { BrowserAgent } from '@newrelic/browser-agent'`), the agent is imported as an interface, not as a class (obviously the [wrong declaration](https://github.com/newrelic/newrelic-browser-agent/blob/40168c71a6538a645237eab7f455113646bd1a97/types.ts#L1))

3) Even if I succeed to import the package, the following error is displayed, and so far I have not been able to progress beyond this step

```
✘ [ERROR] No matching export in "node_modules/@newrelic/browser-agent/dist/esm/common/wrap/wrap-events.js" for import "unwrapEvents"

    node_modules/@newrelic/browser-agent/dist/esm/common/wrap/wrap-xhr.js:10:21:
      10 │ import { wrapEvents, unwrapEvents } from './wrap-events';
         ╵                      ~~~~~~~~~~~~
```

4) I tried to go the workaround and properly redeclare the module, and do the import as described in the second point. But I am still encountering errors.

```
✘ [ERROR] Could not resolve "./custom-agent"

    node_modules/@newrelic/browser-agent/dist/esm/index.js:1:28:
      1 │ export { CustomAgent } from './custom-agent';
        ╵                             ~~~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "./browser-agent"

    node_modules/@newrelic/browser-agent/dist/esm/index.js:2:29:
      2 │ export { BrowserAgent } from './browser-agent';
        ╵                              ~~~~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "./worker-agent"

    node_modules/@newrelic/browser-agent/dist/esm/index.js:3:28:
      3 │ export { WorkerAgent } from './worker-agent';
        ╵                             ~~~~~~~~~~~~~~~~
```

Environment - Node 18, Vite 2.2.0, TypeScript 4.9.5 

Codesandbox with reproducabre example - https://codesandbox.io/p/sandbox/thirsty-https-f3s0q5?file=README.md

patrickhousley:
[...]

PR #455, opened 2023-04-03
fix: fix bad exports

patrickhousley:
### Overview

Fixing some bad exports in the index.js file and some left over from the revert of the unwrap code.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7903
https://github.com/newrelic/newrelic-browser-agent/issues/454

### Testing

Added a vite react wrapper project to the test builds and a functional test to make sure that wrapper generates a rum call.


patrickhousley:
Merged with #455

PR #457, opened 2023-04-04
Capture supportability metrics for console methods (NEWRELIC-7770)

bjfield:
### Overview

This PR adds supportability metrics for assorted console methods to inform future decisions around capturing logs.
- Wraps methods `debug`, `error`, `info`, `log`, `trace`,  and `warn` on `console` in the global scope.
- Captures **count** and **size** metrics as part of metrics feature instrumentation.
- New tests assess accurate number and size of each method.

Also adds a pre-commit hook in the husky config to update the caniuse-lite database, which is what browserslist targets are based on. After updating the DB we lost `ios_saf 14.5-14.8` as targets, which introduced the need to add browserify plugin support for `class-properties` and `private-methods`.

### Related Issue(s)

- [NEWRELIC-7770](https://issues.newrelic.com/browse/NEWRELIC-7770)
- [Angler PR 432](https://source.datanerd.us/agents/angler/pull/432)

### Testing

A new functional test has been added in `metrics.test.js`.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -7,9 +7,16 @@ export FORCE_COLOR=1
 npm run lint
 npm test -- --onlyChanged
 
+# Update version in package.json to match VERSION file
 npm run sync:version
 git add package.json
 
+# Include updated caniuse-lite database used by browserslist
+npx browserslist@latest --update-db
+git add package.json
+git add package-lock.json
+
+# Include fresh list of supported browsers from Sauce Labs based on defined browserslist
 npm run sauce:get-browsers
 git add tools/jil/util/browsers-supported.json
 git add tools/wdio/util/browsers-supported.json
--- a/src/features/page_view_event/instrument/index.js
+++ b/src/features/page_view_event/instrument/index.js
@@ -7,6 +7,9 @@ import { FEATURE_NAMES } from '../../../loaders/features/features'
 import { getRuntime } from '../../../common/config/config'
 import { onDOMContentLoaded, onWindowLoad } from '../../../common/window/load'
 import { now } from '../../../common/timing/now'
+import { wrapConsole } from '../../../common/wrap'
+import { stringify } from '../../../common/util/stringify'
+import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
 
 export class Instrument extends InstrumentBase {
   static featureName = CONSTANTS.FEATURE_NAME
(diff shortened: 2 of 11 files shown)


src/features/page_view_event/instrument/index.js
```diff
@@ -28,6 +31,29 @@ export class Instrument extends InstrumentBase {
     }
     // Else, inference: inside worker or some other env where these events are irrelevant. They'll get filled in with 0s in RUM call.
 
+    // For now we are just capturing supportability metrics on `console` usage to assess log forwarding feature.
+    const consoleEE = wrapConsole(this.ee)
```
metal-messiah, line 35: Since this is directly tied to sending a `sm` at the moment, we should put the wrapping and handling code in the metrics instrument file instead of the PVE instrument file to save space in the lite build.

src/features/page_view_event/instrument/index.js
```diff
@@ -28,6 +31,29 @@ export class Instrument extends InstrumentBase {
...
+          if (typeof arg === 'function' ||
+            (arg && arg.message && arg.stack) // Duck typing for Error objects
+          ) {
+            parsedArgs.push(arg.toString())
+          } else {
+            parsedArgs.push(arg)
+          }
+        }
+        const parsedArgsJSON = stringify(parsedArgs)
+        handle(SUPPORTABILITY_METRIC_CHANNEL, [`Generic/Console/${method}`, parsedArgsJSON.length], undefined, FEATURE_NAMES.metrics, consoleEE)
```
metal-messiah, line 53: Can we group these SM's a little differently to help distinguish them from `Generic` SMs?

`Console/${method}/Seen` or something similar is a pattern we have used -- https://source.datanerd.us/agents/angler/blob/master/src/main/resources/metric_names.txt

bjfield: Changed, makes sense. There seem to be a lot of specific things under Generic. What's the intent of that category?

src/features/page_view_event/instrument/index.js
bjfield: This has been moved.

.husky/pre-commit
```diff
@@ -5,11 +5,20 @@
...
+npm run test
+
+# Update version in package.json to match VERSION file
 npm run sync:version
 git add package.json
 
+# Include updated caniuse-lite database used by browserslist
+npx browserslist@latest --update-db
+git add package.json
+git add package-lock.json
```
bjfield, line 19: Is it a problem to add package.json and package-lock.json here? Could see this being an issue if somebody has messed with their files for other reasons.

patrickhousley (approved):

.husky/pre-commit
metal-messiah: Our pre-commit hook is starting to get pretty cumbersome.  Future effort to speed that up

metal-messiah (approved):

PR #459, opened 2023-04-07
SMs - SessionTrace duration and endpoint payload sizes per page

metal-messiah:
#### This PR is in support of Session Replay COGS investigation.  

Aspect 1: The intent is to capture the length end-to-end of session traces, to be able to evaluate what the "average" session trace duration is today per "page session".

Aspect 2: This PR begins tallying bytesSent for each endpoint, which will give us raw figures per endpoint per "page session"

SMs added:
* `PageSession/Endpoint/1/BytesSent`
* `PageSession/Endpoint/Resources/BytesSent`
* `PageSession/Endpoint/Events/BytesSent`
* `PageSession/Endpoint/Jserrors/BytesSent`
* `PageSession/Feature/SessionTrace/DurationMs`

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,7 +5,7 @@
 export FORCE_COLOR=1
 
 npm run lint
-npm run unit -- --onlyChanged
+npm test
 
 npm run sync:version
 git add package.json
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,7 +5,7 @@
 export FORCE_COLOR=1
 
 npm run lint
-npm test
+npm test -- --onlyChanged
 
 npm run sync:version
 git add package.json
--- a/src/features/metrics/aggregate/index.js
+++ b/src/features/metrics/aggregate/index.js
@@ -96,6 +96,7 @@ export class Aggregate extends AggregateBase {
     // TODO - these SMs are to be removed when we implement the actual resources feature
     try {
       if (this.resourcesSent) return
+      const agentRuntime = getRuntime(this.agentIdentifier)
       // make sure this only gets sent once
       this.resourcesSent = true
       // differentiate between internal+external and ajax+non-ajax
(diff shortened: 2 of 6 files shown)


.husky/pre-commit
```diff
@@ -5,7 +5,7 @@
 export FORCE_COLOR=1
 
 npm run lint
-npm run unit -- --onlyChanged
+npm test
```
patrickhousley, line 8: This should be `npm test -- --onlyChanged` so the pre-commit hook only runs the tests for files that changed. As we add more tests, running them all will take more time.

patrickhousley (approved):

src/features/metrics/aggregate/index.js
```diff
@@ -113,6 +114,19 @@ export class Aggregate extends AggregateBase {
...
+
+      // Capture Agent Bytes Sent
+      Object.keys(agentRuntime.bytesSent).forEach(endpoint => {
+        this.storeSupportabilityMetrics(
+          `PageSession/Endpoint/${endpoint.charAt(0).toUpperCase() + endpoint.slice(1)}/BytesSent`,
+          agentRuntime.bytesSent[endpoint]
+        )
+      })
+
+      // Capture STN metrics if avail
```
bjfield, line 126:
```suggestion
      // Capture metrics for session trace if active (`ptid` is set when returned by replay ingest.)
```

bjfield:
Thoughts on why/not these SMs warrant adding tests?

bjfield (approved):
I noticed there's not a test, but was able to verify manually that the metrics are being captured in this build.
- PageSession/Endpoint/1/BytesSent (RUM)
- PageSession/Endpoint/.../BytesSent (endpoints)
- PageSession/Feature/SessionTrace/DurationMs

.husky/pre-commit
metal-messiah:
```suggestion
npm test -- --onlyChanged
```

metal-messiah:
> Thoughts on why/not these SMs warrant adding tests?

I think we need to re-evaluate the usefulness of testing supportability metrics collection specifics (beyond just testing that triggering the metrics feature produces a harvest).  Another time another place

PR #462, opened 2023-04-10
fix: build time env variables

patrickhousley:
### Overview

Changing how the source code imports environment variables. By default, the source code will use pre-filled values and the package.json. This is used when consumers of the NPM package import from our source code instead of from the CJS or ESM package.

For webpack and the npm builds, we will use the existing `transform-inline-environment-variables` to inline the environment variables and a new plugin `./tools/scripts/babel-plugin-transform-import` to swap out import statements. The webpack config has been updated to pass a babel environment that will result in the use of the `env.cdn.js` file during the webpack build. The npm builds have been updated to pass a bebel environment that weill result in the use of the `env.npm.js` file during the NPM package build.

Additionally, the babel configurations have been centralized and babel environments are used to apply overrides. The build will also now rely on the `.browserslistrc` file except for the webpack IE11 (polyfills) build. That build overrides the targets to support IE11.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7989

### Testing

The builds should still function and produce little to no change in the output. The test cases for npm importing from `src` have also been re-enabled.


metal-messiah:
> If you are wondering why we need the `env.npm.js` file, know that the NPM builds output to a directory one level deeper than the source code. This results in the import of the package.json in the `env.js` file being invalid. The other option for this was to override that import statement as well. Either option works.

I'm good with either but it might make for the most consistency to have "one" pattern that happens (ie import overrides)

metal-messiah (approved):

tools/scripts/babel-plugin-transform-import.js
```diff
@@ -0,0 +1,55 @@
...
+    if (importPath.match(regex)) {
+      newPath = importPath.replace(regex, state.opts[key])
+    }
+  }
+
+  return newPath
+}
+
+module.exports = function ({ types }) {
+  const visitor = {
```
bjfield, line 17:
```suggestion
  /**
   * A visitor object containing methods for modifying specific types of nodes encountered in the Abstract Source Tree.
   * The state parameter is used to pass along the current state of the plugin as it traverses the AST. We use it
   * specifically for access to plugin options.
   *
   * @type {Object}
   * @property {Function} CallExpression - A function to modify CallExpression nodes.
   * @property {Function} ImportDeclaration - A function to modify ImportDeclaration nodes.
   * @property {Function} ExportNamedDeclaration - A function to modify ExportNamedDeclaration nodes.
   * @property {Function} ExportAllDeclaration - A function to modify ExportAllDeclaration nodes.
   */
  const visitor = {
```

tools/scripts/babel-plugin-transform-import.js
```diff
@@ -0,0 +1,55 @@
...
+      if (path.node.source) {
+        path.node.source.value = replacePath(path.node.source.value, state)
+      }
+    },
+    ExportNamedDeclaration (path, state) {
+      if (path.node.source) {
+        path.node.source.value = replacePath(path.node.source.value, state)
+      }
+    },
+    ExportAllDeclaration (path, state) {
```
bjfield, line 41: `ExportAllDeclaration` and `CallExpression` don't end up calling `replacePath` currently. Do we need these?

tools/scripts/babel-plugin-transform-import.js
```diff
@@ -0,0 +1,55 @@
+function replacePath (importPath, state) {
```
bjfield, line 1:
```suggestion
/**
 * @file This custom babel plugin swaps out import statements. Whenever an import path is encountered, it
 * is checked against key-value pairs of RegEx patterns defined in the plugin config options block.
 * To understand the code (which uses the Visitor pattern to traverse an Abstract Source Tree), see the Babel
 * [plugin handbook](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md).
 */

/**
 * Replaces part(s) of an import path based on key-value RegEx pairs from plugin state options.
 *
 * @param {string} importPath - The original import path to modify.
 * @param {Object} state - A state object containing options to use for replacement.
 * @returns {string} - The modified import path with the replaced value.
 */
function replacePath (importPath, state) {
```

bjfield (approved):
Approved with optional comment suggestions

tools/scripts/babel-plugin-transform-import.js
patrickhousley: They do call `replacePath` if an `export` or `require` statement is hit in the AST.

PR #464, opened 2023-04-13
feat: Add new stateful Session Manager - NEWRELIC-7986

metal-messiah:
The agent will now manage the end-user session lifecycle statefully, across page views and tab changes.
---

### Overview

This is a PR for designing and implementing a `Session Manager` which is a self-contained class meant to store data in a stateful way that automatically handles its own lifecycle to match the details in the definition of a session.

### Related Issue(s)
NEWRELIC-8005
NEWRELIC-8006
NEWRELIC-8007
NEWRELIC-8008

### Testing
Jest and WDIO tests have been added
`npm run wdio -- -b '*@*' --concurrent=30 tests/specs/session-manager.e2e.js `


src/common/session/session-entity.js
```diff
@@ -0,0 +1,225 @@
...
+const PREFIX = 'NRBA'
+
+export class SessionEntity {
+  /**
+   * Create a self-managing Session Entity. This entity is scoped to the agent identifier which triggered it, allowing for mutliple simultaneous session objects to exist.
+   * The key is inteded to be unique, if not, existing entities in the same agent ID will overlap and overwrite.
+   * The value can be overridden in the constructor, but will default to a unique 16 character hex string
+   * expiresMs and inactiveMs are used to "expire" the session, but can be overridden in the constructor. Pass 0 to disable expiration timers.
+   */
+  constructor ({ agentIdentifier, key, value = generateRandomHexString(16), expiresMs = 14400000, inactiveMs = 1800000 }) {
```
patrickhousley, line 21: Extract the magic numbers for `expiresMs` and `inactiveMs` to a constant.

src/common/session/session-entity.js
```diff
@@ -0,0 +1,225 @@
...
+import { ee } from '../event-emitter/contextual-ee'
+import { subscribeToVisibilityChange } from '../window/page-visibility'
+import { Timer } from '../timer/timer'
+import LocalStorage from '../storage/local-storage.js'
+import FPC from '../storage/first-party-cookies'
+import { getConfiguration } from '../config/config'
+
+const PREFIX = 'NRBA'
+
+export class SessionEntity {
```
patrickhousley, line 14: This class should probably be broken up. Create a `SessionManager` that is responsible for creating and managing a session, the timers, and where a session is stored. A `SessionEntity` is just a `key=value` pair tied to a specific `SessionManager`. The concern here is this class is trying to do too much and, for multiple `key=value` pairs, we will have multiple timers running that are not in sync.

src/common/storage/first-party-cookies.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+  remove (key) {
+    try {
+      return document.cookie = `${key}=; expires=Thu, 01 Jan 1970 00:00:00 UTC; Domain=${domain}; Path=/`
+    } catch (err) {
+      return () => {}
+    }
+  },
+  setDomain (d) {
+    domain = d
```
patrickhousley, line 20: Use a class instead of a stateful module. This would make it a lot easier to unit test.

src/common/session/session-entity.js
metal-messiah: We should discuss because I would argue that at some point we probably should be able to have multiple session entities with different timer rules

src/features/utils/lazy-loader.js
```diff
@@ -11,27 +31,27 @@ import { FEATURE_NAMES } from '../../loaders/features/features'
...
-        return import(/* webpackChunkName: "page_view_timing-aggregate" */ '../page_view_timing/aggregate')
-      case FEATURE_NAMES.sessionTrace:
-        return import(/* webpackChunkName: "session_trace-aggregate" */ '../session_trace/aggregate')
-      case FEATURE_NAMES.spa:
-        return import(/* webpackChunkName: "spa-aggregate" */ '../spa/aggregate')
-      default:
-        throw new Error(`Attempted to load unsupported agent feature: ${featureName} ${featurePart}`)
-    }
+export function lazyLoader (agentIdentifier, featureName) {
+  checkSession(agentIdentifier)
```
patrickhousley, line 35: Why do we need to check session in the lazy loader? Should checking session be done in the agent class? Should that be a function of the agent class?

src/common/timer/timer.js
```diff
@@ -0,0 +1,41 @@
+export class Timer {
```
patrickhousley, line 1: Having directories that only contain a single file is a little annoying. Just a personal opinion.

src/common/session/session-entity.js
metal-messiah: currently, the session-entity is tied to one key, but can have many values (the main value, but other custom values too).  You could have multiple session-entities, with distinct values and timers and rules.  we'll need to discuss the shortcomings of this

patrickhousley: Ok so maybe I was thinking of the key in the wrong way. I think we need some docs around what constitutes a session key or entity.

src/features/utils/lazy-loader.js
metal-messiah: We could load this in a different async module if needed, not sure what the benefits of that are

src/common/session/session-entity.js
```diff
@@ -0,0 +1,225 @@
...
+        // easiest way to get the root domain to store to the cookie is through user input
+        FPC.setDomain(sessionConfig.domain)
+        this.storage = FPC
+      } else {
+        this.storage = LocalStorage
+      }
+    } catch (e) {
+      // storage is inaccessible
+      warn('Storage API is unavailable. Session information will not operate correctly.', e)
+      return Object.assign(this, { key, value, sessionReplayActive: false, sessionTraceActive: false, isNew: true, read: () => this, write: (vals) => Object.assign(this, vals), reset: () => new SessionEntity(this) })
```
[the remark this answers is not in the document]
metal-messiah: I replaced a short-circuit with a `LocalMemory` wrapper that workers and other things could fall back to use.  It simply holds an object in state that is read from or written to.

src/common/session/session-entity.js
```diff
@@ -0,0 +1,223 @@
...
+  compress (obj) {
+    // no compression applied at this time
+    // could use bel-serializer encoding here if prioritized
+    return obj
+  }
+
+  decompress (obj) {
+    // no need to decompress if we aren't compressing
+    return obj
+  }
```
patrickhousley, line 207: Can we remove this code since it's not being used? Seems a waste of bytes and stack.

patrickhousley (approved):

bjfield (approved):

cwli24 (changes requested):
1st batch

src/common/config/state/configurable.js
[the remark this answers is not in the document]
patrickhousley: I wonder if there is a better way to do this config stuff where we define and utilize a json schema. Then our configurations also become self-documenting.

src/common/config/state/init.js
```diff
@@ -13,6 +15,12 @@ const model = {
     cors_use_tracecontext_headers: undefined,
     allowed_origins: undefined
   },
+  session: {
+    domain: new URL(globalScope.location).host, // used by first party cookies to set the top-level domain (if subdomains === true)
```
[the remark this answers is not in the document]
patrickhousley: If we do default this, I am not sure why we would need to use `URL` and not just use `globalScope.location?.host`. We would need to processes the value to get the top most domain or, as @cwli24 points out, just make the `domain` a configurable option in the UI. When we set a cookie with a domain, the browser should prepend a period automatically making the cookie accessible for all sub-domains of the configured domain. `window.cookie = 'test=test;domain=example.com` will set the cookie domain to `.example.com`

src/common/storage/first-party-cookies.js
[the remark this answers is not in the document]
patrickhousley: FPC limitations are a bit more complicated. Each browser can have it's own limit to the number of cookies per domain. Check out [this](http://browsercookielimits.iain.guru/) site in different browsers AND [RFC-6265](https://www.rfc-editor.org/rfc/rfc6265#section-6.1) which sets a minimum limit on a browsers limit of cookies per domain (50).

In reality, the `4096 bytes per cookie` and limitations upon header sizes by web servers comes into play here more than anything. For example, Apache HTTPD's default limit is [8190 bytes](https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestfieldsize). Other servers have different values, and customers can change it. If a request to a customer's server exceeds this limit because our cookie is too big or because we are one of a million third-parties that are adding cookies to the customer's domain, the customer's server will usually drop the request or respond with a HTTP error potentially breaking their site.

In reality, this is all fuel for really making FPC something the customer MUST enable and configure, IMHO.

src/common/config/state/configurable.js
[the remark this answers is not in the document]
metal-messiah: Thats not much different than what happens today (a model with defaults getting extended by new values), its just abstracted a bit with the class, but i agree it could use some polishing

src/common/session/session-entity.js
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * Fetch the stored values from the storage API tied to this entity
+   * @returns {Object}
+   */
+  read () {
+    try {
+      const val = this.storage.get(this.lookupKey)
+      if (!val) return {}
+      // TODO - decompression would need to happen here if we decide to do it
+      const obj = typeof val === 'string' ? JSON.parse(val) : val
```
cwli24, line 106: This further enforces that really it's `localStorage` & `FP cookies`, and that `localMemory` isn't really a "storage API option".
The first two only ever store strings (will forcefully convert), _and_ "local memory" aka just-an-object is no different from how the current state of data lives in an initialized agent.

Following my other comment about storing all the key-values inside one central object, I'm thinking you could set/update/remove value in this intermediary, and only ever read (on vis to visible) and write (on vis to hidden) _once_, akin to React's virtual DOM.

cwli24: Let me know if I'm confusing this with the storage medium for which customer-defined data would persist. I assume it's through this interface.

src/common/harvest/harvest-scheduler.js
```diff
@@ -76,8 +85,8 @@ export class HarvestScheduler extends SharedContext {
         }
       }
     } else {
-      const runAfterSending = opts?.unload ? undefined : onHarvestFinished // don't bother running onFinish handler if this is the final harvest
-      this.harvest.sendX(this.endpoint, opts, runAfterSending)
+      // const runAfterSending = opts?.unload ? undefined : onHarvestFinished // don't bother running onFinish handler if this is the final harvest
+      this.harvest.sendX(this.endpoint, opts, onHarvestFinished)
```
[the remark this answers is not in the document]
metal-messiah: why does it matter if the page thinks about the runAfterSending part?

If the page actually offloaded, it will never get to the point that it can call onHarvestFinished since it uses sendBeacon

`if (cbFinished && method === submitData.xhr) { ...` 

If it didnt "really" offload, it will call the onHarvestFinished() cb since it used xhr.

src/common/session/session-entity.js
```diff
@@ -0,0 +1,219 @@
...
+import { isBrowserScope } from '../util/global-scope'
+import { DEFAULT_EXPIRES_MS, DEFAULT_INACTIVE_MS, PREFIX } from './constants'
+import { LocalMemory } from '../storage/local-memory'
+import { InteractionTimer } from '../timer/interaction-timer'
+import { wrapEvents } from '../wrap'
+
+export class SessionEntity {
+  /**
+   * Create a self-managing Session Entity. This entity is scoped to the agent identifier which triggered it, allowing for mutliple simultaneous session objects to exist.
+   * The key is inteded to be unique, if not, existing entities in the same agent ID will overlap and overwrite.
```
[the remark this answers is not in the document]
metal-messiah: This is a remnant comment that has no meaning now.  Basically we need to be aware that -- as of right now -- there is one "namespace" an agent can store data in LS.  If there are two agents on one page, they could overwrite each other since they would both use the same namespace in LS by default.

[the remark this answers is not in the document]
metal-messiah: example --
agent 1 --> NRBA_SESSION = abcd1234
agent 2 --> NRBA_SESSION = zyxw9876

One is gonna overwrite the other in that scenario, we just need to be aware of that

src/common/session/session-entity.js
metal-messiah: i'm not really following what you are saying here

src/common/storage/local-memory.js
```diff
@@ -0,0 +1,30 @@
...
+    try {
+      return this.state[key]
+    } catch (err) {
+      return ''
+    }
+  }
+
+  set (key, value) {
+    try {
+      if (value === undefined || value === null) return this.remove(key)
```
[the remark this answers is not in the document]
metal-messiah: We should ponder about this.  What would we ingest as "custom data" for an undefined value?  Can ingest handle that as a case?

src/common/storage/local-storage.js
```diff
@@ -0,0 +1,28 @@
...
+      // Cast it back to undefined if it doesnt exist
+      return localStorage.getItem(key) || undefined
+    } catch (err) {
+      return ''
+    }
+  }
+
+  set (key, value) {
+    try {
+      if (value === undefined || value === null) return this.remove(key)
```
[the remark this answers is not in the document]
metal-messiah: Same as above: as long as we have a valid way to ingest undefined data, this seems ok.  I'm not currently positive how that gets represented

src/cdn/polyfills.js
```diff
@@ -15,3 +15,4 @@ import 'core-js/stable/map'
 import 'core-js/stable/reflect'
 import 'core-js/stable/set'
 import 'core-js/stable/weak-set'
+import 'core-js/features/url'
```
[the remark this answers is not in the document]
metal-messiah: good catch, removed

src/loaders/api/api.js
```diff
@@ -81,12 +80,10 @@ export function setAPI (agentIdentifier, forceDrain) {
     const currentInfo = getInfo(agentIdentifier)
     if (value === null) {
       delete currentInfo.jsAttributes[key]
-      if (isBrowserScope) removeFromBrowserStorage(key, CUSTOM_ATTR_GROUP) // addToBrowserStorage flag isn't needed to unset keys from storage
     } else {
       setInfo(agentIdentifier, { ...currentInfo, jsAttributes: { ...currentInfo.jsAttributes, [key]: value } })
-      if (isBrowserScope && addToBrowserStorage) putInBrowserStorage(key, value, CUSTOM_ATTR_GROUP)
     }
-    return apiCall(prefix, apiName, true)()
+    return apiCall(prefix, apiName, true, !!addToBrowserStorage || (value === null && 'session'))(key, value)
```
cwli24, line 86:
```suggestion
    return apiCall(prefix, apiName, true, !!addToBrowserStorage || value === null)(key, value)
```

test.txt
cwli24, line 1: Accidental commit

src/common/window/page-visibility.js
```diff
@@ -10,8 +10,8 @@ import { documentAddEventListener } from '../event-listener/event-listener-opts'
  * @param {boolean} [toHiddenOnly=false] - only execute the 'cb' when the vis is changing to the hidden state; no arg is passed to 'cb' if used
  * @returns void
  */
-export function subscribeToVisibilityChange (cb, toHiddenOnly = false) {
-  documentAddEventListener('visibilitychange', handleVisibilityChange)
+export function subscribeToVisibilityChange (cb, toHiddenOnly = false, capture, abortSignal) {
```
cwli24, line 13:
```suggestion
export function subscribeToVisibilityChange (cb, toHiddenOnly = false, capture = false, abortSignal = undefined) {
```
```suggestion
export function subscribeToVisibilityChange (cb, toHiddenOnly = false, capture, abortSignal) {
```
Prob better to turn into an object param though.

src/common/timer/interaction-timer.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+
+      // // watch for the vis state changing.  If the page is hidden, the local inactivity timer should be paused
+      // // if the page is brought BACK to visibility and the timer hasnt "naturally" expired, refresh the timer...
+      // // this is to support the concept that other tabs could be experiencing activity.  The thought would be that
+      // // "backgrounded" tabs would pause, while "closed" tabs that "reopen" will just instantiate a new SessionEntity class if restored
+      // // which will do a "hard" check of the timestamps.
+      subscribeToVisibilityChange((state) => {
+        if (state === 'hidden') this.pause()
+        else this.refresh()
```
cwli24, line 31: Should this not be `resume` ?

tests/functional/worker/all-uncategorized.test.js
```diff
@@ -193,7 +193,7 @@ function harvestSessionIsNullWhenEnabled (type, browserVersionMatcher) {
       Promise.all([ajaxPromise, loadPromise, router.expectRum()])
         .then(([{ request: { query } }]) => {
           t.equal(query.ck, '0', "The cookie flag ('ck') should equal 0.")
-    		t.equal(query.s, '0', "The session id attr 's' should be 0.")
+    		t.ok(query.s, "The session id attr 's' should exist and have a truthy value.")
```
cwli24, line 196: Why is worker sending session id now?

src/features/utils/agent-session.js
```diff
@@ -0,0 +1,67 @@
...
+  })
+
+  // The first time the agent runs on a page, it should put everything
+  // that's currently stored in the storage API into the local info.jsAttributes object
+  if (isBrowserScope) {
+    // retrieve & re-add all of the persisted setCustomAttribute|setUserId k-v from previous page load(s)
+    const customSessionData = agentRuntime.session?.read?.()?.custom
+    const agentInfo = getInfo(agentIdentifier)
+    if (customSessionData) {
+      setInfo(agentIdentifier, { ...agentInfo, jsAttributes: { ...agentInfo.jsAttributes, ...customSessionData } })
```
cwli24, line 46: Since `agentInfo` is already a reference to the object in `_cache`, can't you just do:
```suggestion
      Object.assign(agentInfo.jsAttributes, customSessionData)
```

cwli24 (changes requested):
Here's rest

src/common/event-listener/event-listener-opts.js
```diff
@@ -32,10 +32,10 @@ export function eventListenerOpts (useCapture, abortSignal) {
...
 /** Do not use this within the worker context. */
-export function windowAddEventListener (event, listener, capture = false) {
-  window.addEventListener(event, listener, eventListenerOpts(capture))
+export function windowAddEventListener (event, listener, capture = false, abortSignal) {
+  window.addEventListener(event, listener, eventListenerOpts(capture, abortSignal))
 }
 /** Do not use this within the worker context. */
-export function documentAddEventListener (event, listener, capture = false) {
-  document.addEventListener(event, listener, eventListenerOpts(capture))
+export function documentAddEventListener (event, listener, capture = false, abortSignal) {
```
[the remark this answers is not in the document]
metal-messiah: Defaulting to undefined is pointless imo. That's basic JS behavior, we don't need to add bytes to represent that.

src/common/timer/interaction-timer.js
```diff
@@ -0,0 +1,43 @@
...
+    this.onRefresh = opts.onRefresh
+
+    if (!opts.refreshEvents) opts.refreshEvents = ['click', 'keydown', 'scroll']
+    this.abortController = opts.abortController
+
+    if (isBrowserScope && opts.ee) {
+      if (opts.ee) {
+        const debouncedRefresh = debounce(this.refresh.bind(this), 500, { leading: true })
+        opts.ee.on('fn-end', (evts) => {
+          if (opts.refreshEvents.includes(evts?.[0]?.type) || '') {
```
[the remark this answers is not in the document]
metal-messiah: Nope

src/common/timer/interaction-timer.js
```diff
@@ -0,0 +1,43 @@
...
+  constructor (opts, ms) {
+    super(opts, ms)
+    this.onRefresh = opts.onRefresh
+
+    if (!opts.refreshEvents) opts.refreshEvents = ['click', 'keydown', 'scroll']
+    this.abortController = opts.abortController
+
+    if (isBrowserScope && opts.ee) {
+      if (opts.ee) {
+        const debouncedRefresh = debounce(this.refresh.bind(this), 500, { leading: true })
```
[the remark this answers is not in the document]
metal-messiah: What's the scenario you are thinking of?

src/common/timer/interaction-timer.js
```diff
@@ -0,0 +1,43 @@
...
+
+      // // watch for the vis state changing.  If the page is hidden, the local inactivity timer should be paused
+      // // if the page is brought BACK to visibility and the timer hasnt "naturally" expired, refresh the timer...
+      // // this is to support the concept that other tabs could be experiencing activity.  The thought would be that
+      // // "backgrounded" tabs would pause, while "closed" tabs that "reopen" will just instantiate a new SessionEntity class if restored
+      // // which will do a "hard" check of the timestamps.
+      subscribeToVisibilityChange((state) => {
+        if (state === 'hidden') this.pause()
+        else this.refresh()
+      }, false, false, this.abortController?.signal)
```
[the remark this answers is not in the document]
metal-messiah: Timers are local to the page. We don't want them to arbitrarily clear the session if the page is open on another tab. They need to be paused to prevent cross contamination if multiple pages are open

src/loaders/api/api.js
metal-messiah: This is the second weird selector. I wonder if the lint tools are doing something weird

src/common/window/page-visibility.js
metal-messiah: Agree about the object param. Don't agree with defaulting to undefined, JS defaults args to undefined already.

src/common/timer/interaction-timer.js
metal-messiah: We are treating "focusing" the page as a new interaction in this case. So for now there's not really a "resume" timer case

tests/functional/worker/all-uncategorized.test.js
metal-messiah: Is there a reason it shouldn't? We didn't before because we were limited. The limitation is gone so I guess it's a question of "why not"?

src/features/utils/agent-session.js
metal-messiah: Yes probably, the current weird model wrapper makes it a little weird feeling to do that since it's a pseudo getter/setter so that feels dirty but we should align on an easy way to maintain that state over time

src/common/config/state/init.js
```diff
@@ -13,6 +15,12 @@ const model = {
     cors_use_tracecontext_headers: undefined,
     allowed_origins: undefined
   },
+  session: {
+    domain: undefined, // used by first party cookies to set the top-level domain (if subdomains === true)
+    subdomains: false, // "true" triggers Session Manager to use First Party Cookies
```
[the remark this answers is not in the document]
metal-messiah: It can be represented however we want at the moment. Could completely kick the FPC bucket down the road if we want

src/common/session/session-entity.js
metal-messiah: "Session" is a session entity instance, with a key value and potentially user-defined data. 

Later on something else in the agent like "CheckoutSession" or "ClickSession" or whatever could be added to the agent with completely different rules from a "Session", but with the concept that it expires in a certain amount of time. 

A "session entity" in general (beyond the concept of a "session"), is just a key and a value -- stored somewhere (local storage by default)... but it has some extra "housekeeping" data saved with it like timestamps and can have custom user data added to it through the newrelic api

src/common/event-listener/event-listener-opts.js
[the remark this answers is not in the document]
cwli24: It seems minify reduces it to `void 0`, so then I rather have it as an `options` object which will still be more readable and reduce or keep the amt of bytes as you said.

src/common/harvest/harvest-scheduler.js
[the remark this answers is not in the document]
metal-messiah: Need to remove lingering commented out code

src/common/storage/first-party-cookies.js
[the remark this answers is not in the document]
metal-messiah: Needs a ticket added to handle compression/decomp in FPC, as well as read/write to storage API vs reading local memory

src/common/storage/local-memory.js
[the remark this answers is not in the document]
metal-messiah: Will re-eval this later.  Behvaior would need to change for both setCustomAttribute as well as session storage item

src/common/storage/local-storage.js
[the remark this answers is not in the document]
metal-messiah: See above comments `local-memory`

src/common/session/session-entity.js
```diff
@@ -0,0 +1,219 @@
...
+      delete this.value
+      const newSess = new SessionEntity({
+        agentIdentifier: this.agentIdentifier,
+        key: this.key,
+        storageAPI: this.storage,
+        expiresMs: this.expiresMs,
+        inactiveMs: this.inactiveMs
+        // value: value === '0' ? value : undefined // add this back in if we have to send '0' for disabled cookies
+      })
+      Object.assign(this, newSess)
```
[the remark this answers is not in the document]
metal-messiah: extract the constructor logic to a re-callable method, and just call that to overwrite the properties instead of the weird new instance behavior.

src/common/session/session-entity.js
```diff
@@ -0,0 +1,219 @@
...
+
+    // the set-up of the timer used to expire the session due to "inactivity" at a certain time
+    // this gets ignored if the value is falsy, allowing for session entities that do not expire
+    // this gets "refreshed" when "activity" is observed
+    this.inactiveMs = inactiveMs
+    if (inactiveMs) {
+      this.inactiveAt = initialRead?.inactiveAt || this.getFutureTimestamp(inactiveMs)
+      this.inactiveTimer = new InteractionTimer({
+        onEnd: () => this.reset(),
+        onRefresh: () => this.refresh(),
```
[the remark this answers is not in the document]
metal-messiah: this is fine, add bind()

src/common/timer/interaction-timer.js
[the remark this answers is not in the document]
metal-messiah: will add check to read on vis change "shown"

src/loaders/api/api.js
metal-messiah: will fix

test.txt
metal-messiah: will remove

src/common/window/page-visibility.js
metal-messiah: will kick the can, but should do the object params

src/common/timer/interaction-timer.js
metal-messiah: We only ever pause or refresh the timer at the current moment.  New visibility changes to "visible" are treated as an interaction with the page.

tests/functional/worker/all-uncategorized.test.js
metal-messiah: do not spawn session entity if cookies disabled in settings

src/common/config/state/configurable.js
```diff
@@ -3,24 +3,31 @@ import { warn } from '../../util/console'
...
+      const frozenAttrs = getFrozenAttributes(key)
+      if (frozenAttrs.length && value && typeof value === 'object') {
+        frozenAttrs.forEach(attr => {
+          if (attr in value) {
+            warn(`"${attr}" is a protected attribute and can not be changed in feature ${key}.  It will have no effect.`)
+            delete value[attr]
+          }
+        })
+      }
+      if (typeof state[key] === 'object' && typeof value === 'object') state[key] = setValues(value, state[key])
```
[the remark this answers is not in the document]
metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-8459

PR #468, opened 2023-04-14
chore: jest tests for event-emitter

patrickhousley:
### Overview

Wrote jest tests for the event-emitter files.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7976


src/common/event-emitter/contextual-ee.test.js
```diff
@@ -0,0 +1,320 @@
...
+    expect(result.debugId).toEqual(childName)
+  })
+
+  test('it always returns the same child event-emitter once created for the given name', async () => {
+    const { ee } = await import('./contextual-ee')
+
+    const childName = faker.datatype.uuid()
+    const result = ee.get(childName)
+
+    expect(result).toEqual(ee.get(childName))
```
metal-messiah, line 61: This test seems like its testing the same thing as the test on `44`

src/common/event-emitter/handle.test.js
```diff
@@ -0,0 +1,56 @@
...
+  const eventContext = {}
+  const eventGroup = faker.datatype.uuid()
+
+  handle(eventType, eventArgs, eventContext, eventGroup)
+
+  expect(handleEE.buffer).toHaveBeenCalledWith([eventType], eventGroup)
+  expect(handleEE.emit).toHaveBeenCalledWith(eventType, eventArgs, eventContext)
+})
+
+test('it should use the provided scoped event-emitter', async () => {
```
metal-messiah, line 36: nice

metal-messiah (approved):

src/common/event-emitter/contextual-ee.test.js
patrickhousley: No, this test is calling `ee.get` again and comparing the object to the original result. We are testing that the child EE is only created once.

metal-messiah: Lines 48 and 51 test the same thing that lines 59 and 61 do.

The entire test from 55 to 62 (lets call it test B) is exactly the same as lines 43 to 53 (test A), except that test A just has two other cases it also checks.

This means that test B is a subset of test A, making test B redundant.  Obviously this isn't a big deal, but it's unnecessary because it doesn't tell us anything new from test A.

patrickhousley: Ah yea you are correct.

bjfield (commented):
I haven't done a full compare of the prior tests, but it seems like the new tests do not have as much 1:1 coverage of the previously tested behavior as some other of your Jest conversions. Before I dig deeper, any comments on what you intended in terms of coverage match?

bjfield (approved):
Converted tests appear to provide as good or better coverage than previous tests.

PR #469, opened 2023-04-17
Remove deprecated timing APIs - NEWRELIC-8115

metal-messiah:
### Overview
This PR removes the deprecated performance.timing and performance.navigation API references, in favor of the performance -- Navigation Timing API data.  This has a benefit of no longer needing to calculate an offset against the timings, and they will never be negative.

### Related Issue(s)
NEWRELIC-8115

### Testing
Existing tests should still apply.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10370,9 +10370,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001478",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001478.tgz",
-      "integrity": "sha512-gMhDyXGItTHipJj2ApIvR+iVB5hd0KP3svMWWXDvZOmjzJJassGLMfxRkQCSYgGd2gtdL/ReeiyvMSFD1Ss6Mw==",
+      "version": "1.0.30001480",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001480.tgz",
+      "integrity": "sha512-q7cpoPPvZYgtyC4VaBSN0Bt+PJ4c4EYRf0DrduInOz2SkFpHD5p3LnvEpqBp7UnJn+8x1Ogl1s38saUxe+ihQQ==",
       "dev": true,
       "funding": [
         {
@@ -36240,9 +36240,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001478",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001478.tgz",
-      "integrity": "sha512-gMhDyXGItTHipJj2ApIvR+iVB5hd0KP3svMWWXDvZOmjzJJassGLMfxRkQCSYgGd2gtdL/ReeiyvMSFD1Ss6Mw==",
+      "version": "1.0.30001480",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001480.tgz",
+      "integrity": "sha512-q7cpoPPvZYgtyC4VaBSN0Bt+PJ4c4EYRf0DrduInOz2SkFpHD5p3LnvEpqBp7UnJn+8x1Ogl1s38saUxe+ihQQ==",
       "dev": true
     },
     "capital-case": {
--- a/src/features/session_trace/aggregate/index.js
+++ b/src/features/session_trace/aggregate/index.js
@@ -75,7 +75,7 @@ export class Aggregate extends AggregateBase {
     this.laststart = 0
 
     registerHandler('feat-stn', () => {
-      this.storeTiming(window.performance.timing)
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -187,6 +187,7 @@
     "jest": "^28.1.1",
     "jest-environment-jsdom": "28.1.1",
     "jquery": "1.11.3",
+    "jung": "^2.1.0",
```
patrickhousley, line 190: Where is this being used?

metal-messiah: its used in a few of the package.json scripts.  I use it for `npm run cdn:watch` and it started failing because the dependency got removed, just added it back

src/features/page_view_event/aggregate/index.js
```diff
@@ -89,12 +89,17 @@ export class Aggregate extends AggregateBase {
...
+    if (globalScope.performance) {
+      try {
+        const navTimingApiData = globalScope?.performance?.getEntriesByType('navigation')?.[0]
+        const perf = ({
+          timing: addPT(agentRuntime.offset, navTimingApiData, {}),
+          navigation: addPN(navTimingApiData, {})
+        })
+        chunksForQueryString.push(param('perf', stringify(perf)))
+      } catch (err) {
+        // performance API failed for some reason
```
patrickhousley, line 101: For these empty catch blocks, I really think we should be doing something with the error. Either capturing and reporting an internal agent error or maybe capturing and reporting a supportability metric around certain browser API failures.

src/common/timing/nav-timing.js
```diff
@@ -27,44 +27,43 @@ var DOM_CONTENT_LOAD_EVENT = 'domContentLoadedEvent'
...
-  var relativeValue
-  if (typeof (value) === 'number' && (value > 0)) {
-    relativeValue = Math.max(Math.round(value - offset), 0)
-    obj[prop] = relativeValue
+function handleValue (value, obj, prop) {
+  if (typeof (value) === 'number' && (value >= 0)) {
+    obj[prop] = Math.round(value)
   }
-  navTimingValues.push(relativeValue)
+  navTimingValues.push(value)
```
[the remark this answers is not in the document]
bjfield: @patrickhousley When you say "state" of `navTimingValues` do you mean adding (e.g.) `expect(navTimingValues).toEqual(...)` checks similar to the output checks, or something more sophisticated like `navTimingValues` length changing predictably across multiple calls to `addPT` and `addPN`? Looks like those methods are only called once when sending the RUM call.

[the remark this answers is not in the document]
patrickhousley: I am specifically talking about unit tests that ensure that when one of the timing methods is called and timings are processed, that the module state is also updated. When, where, and how that exported variable is used is outside the scope of a unit test.

src/features/page_view_event/aggregate/index.js
metal-messiah: internal error sounds good

src/common/timing/nav-timing.test.js
```diff
@@ -0,0 +1,91 @@
...
+    expect(output4).toEqual(expected)
+  })
+})
+
+describe('addPN()', () => {
+  test('an output object is populated with valid values', () => {
+    const output = addPN(testValues, {})
+    expect(output).toEqual(expectedPN)
+  })
+})
```
bjfield, line 91:
```suggestion

  test('adds entries to navTimingValues', () => {
    const beforeLength = navTimingValues.length
    addPN(offset, { testValues }, {})
    const afterLength = navTimingValues.length
    expect(afterLength - beforeLength).toEqual(2)
  })
})
```

src/common/timing/nav-timing.test.js
```diff
@@ -0,0 +1,91 @@
...
+    expected = { ...expectedPT }
+    delete expected.le
+    expect(output3).toEqual(expected)
+
+    const output4 = addPT(offset, { ...testValues, loadEventEnd: null }, {})
+    expected = { ...expectedPT }
+    delete expected.le
+    expect(output4).toEqual(expected)
+  })
+})
```
bjfield, line 84:
```suggestion

  test('rounds values to integers', () => {
    const output = addPT(offset, { unloadEventStart: 3.14159 }, {})
    expect(output.u).toEqual(3)
  })

  test('adds entries to navTimingValues', () => {
    const beforeLength = navTimingValues.length
    addPT(offset, { testValues }, {})
    const afterLength = navTimingValues.length
    expect(afterLength - beforeLength).toEqual(21) // 20 + value of n
  })
})
```

src/common/timing/nav-timing.js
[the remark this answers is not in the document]
bjfield: Okay, I've suggested tests for that. The only way `navTimingValues` is expected to be useful (as far as I can tell) is to provide a slice of the PT values (not PN), and this usage assumes `addPT` is called first so that the PT values fall in the correct portion of the array (`1, 21`). So technically we wouldn't need to test if `addPN` updates the array, just `addPT`—and maybe `addPN` should not in fact update the array as it does. But I've suggested both tests.

bjfield (approved):
I've suggested optional tests for `navTimingValues`.

PR #471, opened 2023-04-18
chore: lint and wdio fixes

patrickhousley:
### Overview

Fixing lint errors and wdio ie@11 execution.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-87779
https://issues.newrelic.com/browse/NR-87780
https://issues.newrelic.com/browse/NR-87759
https://issues.newrelic.com/browse/NR-87770

### DIFF
--- a/src/common/browser-version/ios-version.js
+++ b/src/common/browser-version/ios-version.js
@@ -4,7 +4,8 @@ export const isiOS = /(iPad|iPhone|iPod)/g.test(navigator.userAgent)
 
 // Shared Web Workers introduced in iOS 16.0+ and n/a in 15.6-
 export const iOS_below16 = isiOS && Boolean(typeof SharedWorker === 'undefined')
-/* ^ It was discovered in Safari 14 (https://bugs.webkit.org/show_bug.cgi?id=225305) that the buffered flag in PerformanceObserver
-	did not work. This affects our onFCP metric in particular since web-vitals uses that flag to retrieve paint timing entries.
-	This was fixed in v16+.
+/**
+ * ^ It was discovered in Safari 14 (https://bugs.webkit.org/show_bug.cgi?id=225305) that the buffered flag in PerformanceObserver
+ * did not work. This affects our onFCP metric in particular since web-vitals uses that flag to retrieve paint timing entries.
+ * This was fixed in v16+.
 */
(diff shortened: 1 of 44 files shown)


tools/wdio/plugins/jil-commands.mjs
```diff
@@ -4,8 +4,12 @@
 export default class TestingServerWorker {
   async before (capabilities, context, browser) {
     browser.addCommand('waitForFeature', async function (feature) {
+      const command = `window.NREUM && window.NREUM.activatedFeatures && window.NREUM.activatedFeatures['${feature}']`
       await browser.waitUntil(
-        () => browser.execute((feat) => `window.NREUM && window.NREUM.activatedFeatures && window.NREUM.activatedFeatures['${feat}']`, feature),
+        // () => browser.execute((feat) => `window.NREUM && window.NREUM.activatedFeatures && window.NREUM.activatedFeatures['${feat}']`, feature),
+        () => browser.execute(function (feat) {
```
metal-messiah, line 10: Does `browser.execute(...)` need to get transpiled, or do we need to enforce ES5 syntax?

src/common/browser-version/ios-version.js
```diff
@@ -4,7 +4,8 @@ export const isiOS = /(iPad|iPhone|iPod)/g.test(navigator.userAgent)
 
 // Shared Web Workers introduced in iOS 16.0+ and n/a in 15.6-
 export const iOS_below16 = isiOS && Boolean(typeof SharedWorker === 'undefined')
-/* ^ It was discovered in Safari 14 (https://bugs.webkit.org/show_bug.cgi?id=225305) that the buffered flag in PerformanceObserver
-	did not work. This affects our onFCP metric in particular since web-vitals uses that flag to retrieve paint timing entries.
-	This was fixed in v16+.
+/**
```
cwli24, line 7: What's with this? The double asterisks are associated with JSDocs and will become the description of whatever follows it, whereas the single asterisks are for multi-line comments and don't get auto tagged as jsdoc

tools/wdio/plugins/jil-commands.mjs
patrickhousley: WDIO does not transpile the function code automatically. We could override the `execute` command to add transpiling. If we want to do that, I can add that to this PR.

patrickhousley: Looking through the code of wdio, I don't think its a good idea to override the `execute` command. I think we should just adhere to ES5 syntax to support IE. Or we could create a new command that assumes a function being passed in and babelifies it.

https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/execute.ts
https://github.com/webdriverio/webdriverio/blob/main/packages/devtools/src/commands/executeScript.ts

bjfield (approved):
Lint changes look fine and IE@11 works for WDIO runs.

metal-messiah (approved):

PR #473, opened 2023-04-18
Upload "fuzzy" versions to NRDB - NEWRELIC-7720

metal-messiah:
### Overview
As part of "Version Pinning" MMF, this PR introduces a naming scheme that will "overwrite" in NRDB on each publish.  
* A "latest" minor on a major version --> 1.x.x
* A "latest" patch on a minor version --> 1.230.x

-- note -- 
The "fuzzy" versions above only generate when running `npm run cdn:build:prod` which is the script tied to building production code.

### Related Issue(s)
NEWRELIC-7720

### Testing
Uploading of these assets has been manually validated at the "science" endpoint for RPM Grape.  A "dev" path in `upload-to-nr.js` was added to represent that.


webpack.config.js
```diff
@@ -173,6 +175,73 @@ const standardConfig = merge(commonConfig, {
   target: 'web'
 })
 
+const fuzzyConfig = merge(commonConfig, {
```
bjfield, line 178: Using webpack to create these fuzzy files instead of copying the explicitly versioned files to the additional filename formats has a couple possible downsides:
- It's slightly slower (no big deal but it's a little something).
- The output is not exactly the same for the fuzzies as for the primary (maybe no big deal, but it makes it harder to validate that it's actually the same build). E.g., in my test I found that `installedChunks` variable is different in the outputs of the primary config vs the outputs of the fuzzyConfig, making the latter files a few bytes larger than their non-fuzzy counterparts.
- It adds another configuration to maintain.

bjfield (commented):
I've focused on reviewing the generation of the fuzzy files, not the CDN/NRDB end of things (harder to test). I've verified that the correct files are created (24 and only 24 new files, all with the correct names). Elsewhere I posed a question about how the files are created.

webpack.config.js
metal-messiah: @bjfield this has been updated to a file-clone script, would you mind taking another look?

patrickhousley (approved):

tools/scripts/clone.js
```diff
@@ -0,0 +1,45 @@
+const fs = require('fs')
+const path = require('path')
+
+const buildDir = path.resolve(__dirname, '../../build/')
+const builtFileNames = fs.readdirSync(buildDir)
+
+function prependSemicolon (fileName, text) {
+  const newText = ';' + text
```
bjfield, line 8: Clone works great, except this ends up prepending a double ;; to the output when running `npm run cdn:build:prod`.

bjfield (approved):

tools/scripts/clone.js
```diff
@@ -0,0 +1,42 @@
...
+    const content = f
+    if (fileName.startsWith('nr-loader') && fileName.endsWith('.js')) {
+      const pieces = fileName.split(/-(?=\d)/).map(x => x.split('.'))
+      if (!pieces[1]) return Promise.resolve()
+      pieces[1] = pieces[1].map((v, i) => i === 2 ? 'x' : v).join('.')
+      const minor = pieces.join('-')
+      return fs.promises.writeFile(`${buildDir}/${minor}`, content)
+    }
+    return Promise.resolve()
+  }))
```
cwli24, line 36: Bit confusing if someone had no idea what this was about.
```suggestion
  const fuzzyMajor = await Promise.all(files.map((f, i) => {
    const fileName = builtFileNames[i]
    // Assuming the filename contains a semantic version pattern, "-#.#.#.", replace the minor and patch numbers.
    const allMinorAndPatch = fileName.replace(/(^nr-loader.*-\d+\.)(\d+)\.(\d+)(.*\.js$)/, "$1x.x$4")
    if (allMinorAndPatch !== fileName) { // we only get the same string back if the filename didn't have that pattern, in which case we'll skip this
      return fs.promises.writeFile(`${buildDir}/${allMinorAndPatch}`, f)
    }
    return Promise.resolve()
  }))

  const fuzzyMinor = await Promise.all(files.map((f, i) => {
    const fileName = builtFileNames[i]
    // Assuming the filename contains a semantic version pattern, "-#.#.#.", replace the patch number.
    const allPatch = fileName.replace(/(^nr-loader.*-\d+\.\d+\.)(\d+)(.*\.js$)/, "$1x$3")
    if (allPatch !== fileName) { // we only get the same string back if the filename didn't have that pattern, in which case we'll skip this
      return fs.promises.writeFile(`${buildDir}/${allPatch}`, f)
    }
    return Promise.resolve()
  }))
```
image
image

cwli24 (approved):

metal-messiah:
⚠️ This needs confirmation from API-E that the API changes have been deployed to production before merging

PR #479, opened 2023-04-20
patch: Omit CLS of 0s if not reported by web-vitals

cwli24:
### Overview
The problematic way in which we report CLS today intensified with the shift to web-vitals + deferring of it introduced in [v1.229.0](https://github.com/newrelic/newrelic-browser-agent/releases/tag/v1.229.0) that reduced sample noise and delayed CLS calculation.

Without diving into multiple factors and complications, in summary, we internally track `cls` throughout the page life as it updates, and it starts off as set to `0`. In edge cases, this `0` becomes the "final" cls value of the page, with the way we interpret or limit (our) `pageHide` today. As a result, this skews and dilutes many monitoring data set with 0s for CLS and affects their scores.

As a temporary patch, we're excluding these false positive 0s CLS, while we work on a solution to re-capture the full population of user samples. The following cases have been observed and addressed:

1. When document becomes hidden _before_ the window `load` event fires.
    - This affected pages with long load times, for which the user could/would hide the page before it fully loads.
    - CLS values in these cases are **omitted**, and no final-cls is reported at all.
2. When the page unloads, e.g. user navigates to a different url, without ever hiding the page.
    - This is entirely user-behavior based and widely affected all pages.
    - CLS values in these cases are **re-adjusted**, and a final-cls is reported regardless of whichever happens first: page unloading or document hiding.

In addition, as part of this change, some of the 0s from CLS being attached to all `PageViewTiming` events are also omitted in practice. A couple timings such as `fp`  that previously reported `cls = 0` will no longer do so.
(However, NRQL queries should still filter `cumulativeLayoutShift` by `WHERE timingName = 'pageHide'` for now until we fully resolve this.)

Finally, we also removed a minor, redundant harvest call whenever the user leaves page session.

### Related Issue(s)
Discussions:
- NEWRELIC-6143
- Idea talk with UI during #363 to emit a `pageHide` per vis change rather than only once

GTSE // Complaints:
- NR-107111
- NR-107172
- NR-108958
- NR-108969
- ... other Salesforce tickets


src/features/page_view_timing/aggregate/index.js
```diff
@@ -27,11 +27,7 @@ export class Aggregate extends AggregateBase {
     this.timings = []
     this.timingsSent = []
     this.curSessEndRecorded = false
-
-    try { // we (only) need to track cls state because it's attached to other timing events rather than reported on change...
-      this.clsSupported = PerformanceObserver.supportedEntryTypes.includes('layout-shift')
```
metal-messiah, line 32: Does web-vitals already do this check?

cwli24: Yep it does; only reports when its own compatibility is met

patrickhousley (approved):

metal-messiah (approved):
Approved, pending tests are made to be consistent in Chrome

tests/functional/pvt/timings.test.js
```diff
@@ -320,39 +316,7 @@ function runPvtInStnTests (loader) {
 }
 
 function runClsTests (loader) {
-  testDriver.test(`LCP for ${loader} agent collects cls attribute`, supportsCLS, function (t, browser, router) {
```
metal-messiah, line 323: this seems like it should collect a CLS again (but with a positive attribute)

tests/functional/pvt/timings.test.js
```diff
@@ -308,7 +303,8 @@ function runPvtInStnTests (loader) {
         return Promise.all([resourcesPromise, clickPromise])
       })
       .then(([{ request: resourcesResult }]) => {
-        const expectedPVTItems = ['fi', 'fid', 'lcp', 'pageHide', 'fcp', 'load', 'unload']
+        const expectedPVTItems = ['fi', 'fid', 'lcp', 'fcp', 'load', 'unload']
```
metal-messiah, line 306: pageHide should be good again, right?

tests/functional/pvt/timings.test.js
```diff
@@ -397,73 +361,17 @@ function runClsTests (loader) {
...
-        t.ok(unload, 'there should be an unload timing')
+        t.ok(pagehide, 'there should be a pageHide timing')
         t.equal(cls.value, 0, 'cls value should be a perfect score of 0')
 
         t.end()
       })
       .catch(fail(t))
   })
 
-  testDriver.test(`First interaction ${loader} agent collects cls attribute`, supportsCLS, function (t, browser, router) {
```
metal-messiah, line 411: This also seems like it should be able to get a CLS again (needs validation)

tests/functional/pvt/timings.test.js
cwli24: This test would be remove anyways in the future?

PR #482, opened 2023-04-22
Use efficient stack-based size function for console SM

bjfield:
### Overview

- The new `approximateSize` function uses an efficient stack-based algorithm to estimate the size of an object.
  - Compared with the `JSON.stringify().length` method of sizing, this function evaluates objects with a large number of UUID-sized keys and values (tens of thousands of nodes) in about 32% of the time.
  - The difference is even more stark for objects with very large string values (as with the functional tests for the console supportability metrics). This efficiency is possible because the goal is not serialization—just estimating size.
  - In most cases, the calculated size is identical to the length of a string produced by `JSON.stringify()`.
- To better represent real-world objects, the new `generateObject` test function produces an object of the same size as before with more nodes, more symmetrically distributed. `JSON.stringify()` takes 30-40% longer for the new test objects. These differences are reflected in benchmark comments.

### Related Issue(s)

- PR https://github.com/newrelic/newrelic-browser-agent/pull/483
- [NEWRELIC-8224](https://issues.newrelic.com/browse/NEWRELIC-8224) - _Remove supportability metrics for console observability_

### Testing

Compare the updated test benchmarks in `tests/specs/console-performance.e2e.js`.


src/common/util/approximate-size.js
```diff
@@ -0,0 +1,47 @@
...
+      // assuming zero bytes for circular values (but does count the key)
+      seen.add(value)
+      if (Array.isArray(value)) { // arrays and other iterables
+        size += 2 // []
+        let commas = -1
+        for (const v of value) {
+          stack.push(v)
+          commas++
+        }
+        if (commas > 0) size += commas
```
patrickhousley, line 29: Duplicated line with 40. This can be moved after line 42 and line 40 can be removed.

src/common/util/approximate-size.js
```diff
@@ -0,0 +1,47 @@
...
+    else if (value === false) size += 5
+    else if (typeof value === 'string') size += value.length + 2 // + ""
+    else if (typeof value === 'number') size += String(value).length
+    else if (value instanceof Date) size += 26 // ISO format + ""
+    else if (typeof value === 'object' && !seen.has(value)) {
+      // assuming zero bytes for circular values (but does count the key)
+      seen.add(value)
+      if (Array.isArray(value)) { // arrays and other iterables
+        size += 2 // []
+        let commas = -1
```
patrickhousley, line 24: Duplicated line with 34. This can be moved after line 21 and line 34 can be removed.

src/common/util/approximate-size.test.js
```diff
@@ -0,0 +1,83 @@
...
+          { id: 0, name: 'Allison' },
+          { id: 1, name: 'Shahram' },
+          { id: 2, name: 'Dominic' }
+        ]
+      }
+    ]
+    const stringifySize = JSON.stringify(data).length
+    const thisSize = approximateSize(data)
+    expect(thisSize).toBe(stringifySize)
+  })
```
patrickhousley, line 82: Add another test for esoteric data types. We never know what the customer is going to console log.

```suggestion
  })
  it('should support esoteric data types', () => {
    const dataTypes = [
      0,
      // eslint-disable-next-line no-new-wrappers
      new Number(0),
      Infinity,
      NaN,
      'asdf"\b\t\n\f\r\\',
      new Date(),
      {},
      new Error(),
      /./,
      false,
      // eslint-disable-next-line no-new-wrappers
      new Boolean(false),
      class FooBar {},
      class {},
      function () {},
      () => {},
      Symbol.for('test'),
      JSON.stringify({ foo: 'bar' })
    ]
    const thisSize = approximateSize(dataTypes)
    expect(thisSize).toBeGreaterThan(0)
  })
```

src/common/util/approximate-size.js
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param {Object} object - The object for which size will be calculated.
+ * @returns {number} The approximate size of the object.
+ */
+export function approximateSize (object) {
+  const seen = new WeakSet()
+  const stack = [object]
+  let size = 0
+  while (stack.length) {
+    const value = stack.pop()
```
patrickhousley, line 13: You may want to wrap the innards of the while loop in a `try..catch` to prevent errors from this function possibly breaking a customer's site.

```javascript
while (stack.length) {
  try {
    ...
  } catch (_) {
    continue
  }
}
```

src/common/util/approximate-size.js
bjfield: Good catch.

src/common/util/approximate-size.js
bjfield: Consolidated.

patrickhousley (approved):
This looks great. The only thing I think we should revisit later is the performance tests. We could probably use more of those and we may want to set the "limits" to match our performance expectations and not be driven by how long the tests take in certain browsers.

bjfield:
Created Jira issue [NEWRELIC-8256](https://issues.newrelic.com/browse/NEWRELIC-8256) to follow up on the performance test idea at some point.

metal-messiah:
Closing as "will not do" for now.

PR #484, opened 2023-04-24
fix: npm package version reporting

patrickhousley:
### Overview

Fixing the version reported by the agent when using the npm package. The version the NPM package reports on BAM calls should match the version that is reported by other installation methods.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-8150

### Testing

Run the full build and use the local test server to verify the version on the BAM calls.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,10 +1,11 @@
 const process = require('process')
+const fs = require('fs')
 
 module.exports = function (api) {
   api.cache(true)
 
   if (!process.env.BUILD_VERSION) {
-    process.env.BUILD_VERSION = require('./package.json').version
+    process.env.BUILD_VERSION = process.env.VERSION_OVERRIDE || fs.readFileSync('./VERSION', 'utf-8')?.trim()
   }
 
   const presets = [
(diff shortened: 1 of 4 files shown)


bjfield (approved):
Verified version in RUM calls for both NPM and regular builds when set via VERSION file or VERSION_OVERRIDE. VERSION_OVERRIDE supersedes VERSION file.

cwli24:
Is the _env.js_ file useless/unused then?

patrickhousley:
No, the `env.js` file is the default. If a consumer of the NPM package imports from the agent `src` directory, that is the constants they will get.

patrickhousley:
No. We cant import node globals in `env.js`. It will break the customer's app.

cwli24:
> No. We cant import node globals in `env.js`. It will break the customer's app.

But if they import from src then, wouldn't they face the same issue? If so, is that just unavoidable?

patrickhousley:
Correct. We cannot fix this issue for customers that import from the `src` directory. They will have to wait until we sync the version numbers.

cwli24 (approved):

PR #486, opened 2023-04-24
fix: cwv add error prevention logic

patrickhousley:
### Overview

Adding in logic checks for object passed to the CWV callbacks to prevent accessing properties of undefined.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-8039

### Testing

Integration tests should pass.

### DIFF
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -66,7 +66,7 @@ export class Aggregate extends AggregateBase {
 
     /* First Input Delay (+"First Interaction") - As of WV v3, it still imperfectly tries to detect document vis state asap and isn't supposed to report if page starts hidden. */
     onFID(({ name, value, entries }) => {
-      if (pageStartedHidden || this.alreadySent.has(name) || !Array.isArray(entries) || entries.length === 0) return
+      if (pageStartedHidden || this.alreadySent.has(name) || entries.length === 0) return
       this.alreadySent.add(name)
 
       // CWV will only report one (THE) first-input entry to us; fid isn't reported if there are no user interactions occurs before the *first* page hiding.
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -50,7 +50,7 @@ export class Aggregate extends AggregateBase {
         if (!pageStartedHidden) { // see ios-version.js for detail on this following bug case; tldr: buffered flag doesn't work but getEntriesByType does
           const paintEntries = performance.getEntriesByType('paint')
           paintEntries.forEach(entry => {
-            if (entry?.name === 'first-contentful-paint') {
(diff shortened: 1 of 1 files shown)


src/features/page_view_timing/aggregate/index.js
```diff
@@ -66,7 +66,7 @@ export class Aggregate extends AggregateBase {
 
     /* First Input Delay (+"First Interaction") - As of WV v3, it still imperfectly tries to detect document vis state asap and isn't supposed to report if page starts hidden. */
     onFID(({ name, value, entries }) => {
-      if (pageStartedHidden || this.alreadySent.has(name)) return
+      if (pageStartedHidden || this.alreadySent.has(name) || !Array.isArray(entries) || entries.length === 0) return
```
cwli24, line 69:
```suggestion
      if (pageStartedHidden || this.alreadySent.has(name) || entries.length === 0) return
```
entries is always guaranteed to be `[]` at least

src/features/page_view_timing/aggregate/index.js
```diff
@@ -94,7 +94,7 @@ export class Aggregate extends AggregateBase {
       if (lcpEntry.url) {
         attrs['elUrl'] = cleanURL(lcpEntry.url)
       }
-      if (lcpEntry.element && lcpEntry.element.tagName) {
+      if (lcpEntry.element?.tagName) {
         attrs['elTag'] = lcpEntry.element.tagName
       }
```
cwli24, line 99: `onLCP` can trigger with just a value but empty entries `[]`. Rather than quitting, it's better to check if `lcpEntry` is `undefined`, and if so, skip over ln 89-99.

(I still don't know how that happens that isn't something wrong with their instrumentation. `onLCP` only verifiably trigger with empty entry on b/f cache restore.)

src/features/page_view_timing/aggregate/index.js
```diff
@@ -50,7 +50,7 @@ export class Aggregate extends AggregateBase {
         if (!pageStartedHidden) { // see ios-version.js for detail on this following bug case; tldr: buffered flag doesn't work but getEntriesByType does
           const paintEntries = performance.getEntriesByType('paint')
           paintEntries.forEach(entry => {
-            if (entry.name === 'first-contentful-paint') {
+            if (entry?.name === 'first-contentful-paint') {
```
cwli24, line 53: This shouldn't be possible to be `null|undefined`, given `getEntriesByType` returns an empty array at minimum

src/features/page_view_timing/aggregate/index.js
cwli24: No, skipping the entire callback for FID is fine, because it doesn't have a similar case of defined value w/ no entries in the src code. Unsure what we'd even report without an user-input entry to reference.

bjfield (approved):

PR #489, opened 2023-04-27
Add fallbacks for performance timings

metal-messiah:
### Overview
This PR adds fallbacks to the old deprecated timings APIs for the edge of our supported browser list.

### DIFF
--- a/src/common/timing/nav-timing.js
+++ b/src/common/timing/nav-timing.js
@@ -27,11 +27,15 @@ var DOM_CONTENT_LOAD_EVENT = 'domContentLoadedEvent'
 
 export var navTimingValues = []
 
-const PNT_TYPES = {
-  navigate: undefined,
-  reload: 1,
-  back_forward: 2,
-  prerender: 3
+function getPntType (type) {
+  if (typeof type == 'number') return type
+  const types = {
+    navigate: undefined,
+    reload: 1,
+    back_forward: 2,
+    prerender: 3
+  }
+  return types[type]
 }
 
 export function addPT (offset, pt, v = {}, isL1Api = false) {
@@ -63,7 +67,7 @@ export function addPT (offset, pt, v = {}, isL1Api = false) {
 
 // Add Performance Navigation values to the given object
 export function addPN (pn, v) {
-  handleValue(PNT_TYPES[pn.type], v, 'ty')
+  handleValue(getPntType(pn.type), v, 'ty')
   handleValue(pn.redirectCount, v, 'rc')
   return v
 }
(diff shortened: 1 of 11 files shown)


cwli24 (changes requested):

cwli24:
Does the unit test file need to include tests for the old API epoch values?

src/common/timing/nav-timing.js
```diff
@@ -68,11 +68,20 @@ export function addPN (pn, v) {
...
-    val = offset ? Math.max(Math.round(value - offset), 0) : value
-    obj[prop] = Math.round(val)
+function handleValue (value, obj, prop, isOldApi) {
+  /*
+  For L2 Timing API, the value will already be a relative-to-previous-document DOMHighResTimeStamp.
+  For L1 (deprecated) Timing, the value is an UNIX epoch timestamp, which we will convert to a relative time using our offset.
+  //
+  PNT.type is reported as 0, 1, 2 -- 0 values must be honored to allow this
+  */
+  if (typeof value === 'number' && value >= 0) {
```
cwli24, line 78: Backwards compat `value > 0` needed

src/common/timing/nav-timing.js
```diff
@@ -34,30 +34,30 @@ const PNT_TYPES = {
   prerender: 3
 }
 
-export function addPT (offset, pt, v = {}, isLegacy) {
+export function addPT (offset, pt, v = {}, isL1Api = false) {
   if (!pt) return
   v.of = offset
   handleValue(0, v, 'n')
```
cwli24, line 40: This `0` needs to be replaced with `v.of` when `value > 0` is enforced to keep 'n' in object.

metal-messiah: good catch

src/common/timing/nav-timing.js
```diff
@@ -27,43 +27,60 @@ var DOM_CONTENT_LOAD_EVENT = 'domContentLoadedEvent'
...
+  handleValue(pt.domComplete, v, 'dc', isL1Api)
+  handleValue(pt[LOAD_EVENT + START], v, 'l', isL1Api)
+  handleValue(pt[LOAD_EVENT + END], v, 'le', isL1Api)
   return v
 }
 
 // Add Performance Navigation values to the given object
 export function addPN (pn, v) {
-  handleValue(pn.type, v, 'ty')
+  handleValue(PNT_TYPES[pn.type], v, 'ty')
```
cwli24, line 66: The L1 timing which have numeric `.type` would get undefined here instead of passing through

cwli24 (approved):

patrickhousley (approved):

PR #495, opened 2023-04-28
chore: Enable no-control-regex lint rule

patrickhousley:
### Overview

Enabling the `no-control-regex` lint rule and placing ignore statements in the post build cleanup script where we purposefully use the control regex to remove these characters from our build files.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-87767


.eslintrc.js
```diff
@@ -106,7 +106,6 @@ module.exports = {
     'no-use-before-define': 'off',
     'valid-typeof': 'off',
     'no-undef': 'off',
-    'no-control-regex': 'off',
```
metal-messiah, line 109: Is there a point to this rule if the only file affected is ignored?

patrickhousley: Ensures we don't do this in our source code.

metal-messiah (approved):

bjfield (approved):

PR #497, opened 2023-04-28
chore: Enable n/no-deprecated-api lint rule

patrickhousley:
### Overview

Enabling the `n/no-deprecated-api` lint rule and replacing the `util._extend` call with the `deepmerge-ts` library.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-87765


metal-messiah (approved):

bjfield (approved):
Confirmed that the outputs of `deepmerge` and `util._extend` are equivalent, as implemented.

PR #499, opened 2023-04-28
chore: Enable dot-notation lint rule

patrickhousley:
### Overview

Enabling the `dot-notation` lint rule and running auto fixes.
### Related Issue(s)

https://issues.newrelic.com/browse/NR-87755

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -99,7 +99,6 @@ module.exports = {
     'no-new': 'off',
     'no-useless-return': 'off',
     'no-useless-constructor': 'off',
-    'dot-notation': 'off',
     'spaced-comment': 'off',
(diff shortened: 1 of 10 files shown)


cwli24 (commented):
Is this really even worth it? Does it know not to touch something like `arr['foo.bar']`, since that's distinctly different from `arr.foo.bar` ?

metal-messiah (approved):

patrickhousley:
> Is this really even worth it? Does it know not to touch something like `arr['foo.bar']`, since that's distinctly different from `arr.foo.bar` ?

eslint is smart enough not to error on `arr['foo.bar']`

bjfield (approved):

PR #500, opened 2023-04-28
chore: enable no-useless-constructor lint rule

patrickhousley:
### Overview

Enabling the `no-useless-constructor` lint rule and resolving lint errors. The `AggregateBase` class was empty and so was removed. It can be added back later if and when it becomes needed for centralizing some logic across all the aggregators.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-87754

### DIFF
--- a/src/common/util/obfuscate.js
+++ b/src/common/util/obfuscate.js
@@ -8,10 +8,6 @@ var fileProtocolRule = {
   replacement: 'file://OBFUSCATED'
 }
 export class Obfuscator extends SharedContext {
-  constructor (parent) {
-    super(parent) // gets any allowed properties from the parent and stores them in `sharedContext`
-  }
-
   shouldObfuscate () {
     return getRules(this.sharedContext.agentIdentifier).length > 0
   }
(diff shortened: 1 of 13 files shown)


cwli24 (approved):
✅  upon tests passing

src/common/util/obfuscate.js
```diff
@@ -8,10 +8,6 @@ var fileProtocolRule = {
   replacement: 'file://OBFUSCATED'
 }
 export class Obfuscator extends SharedContext {
-  constructor (parent) {
-    super(parent) // gets any allowed properties from the parent and stores them in `sharedContext`
-  }
-
   shouldObfuscate () {
     return getRules(this.sharedContext.agentIdentifier).length > 0
```
metal-messiah, line 12: `this.sharedContext.agentIdentifier` will be undefined without the constructor passing args into super

metal-messiah (changes requested):
We either need the call to `super()` or the dependent functions need the agent ID passed in as an arg

src/common/util/obfuscate.js
cwli24: Is this not superfluous since there's a [default constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/constructor#description) if one isn't provided

metal-messiah: looks like it, TIL

metal-messiah (approved):

PR #506, opened 2023-05-03
CI Experiments - NEWRELIC-8153

metal-messiah:
### Overview
This PR adds CI tooling to allow for deployments of experimental code to the staging NR1 environment without needing to merge to main.

### Related Issue(s)
NEWRELIC-8153

### Testing
Local testing:
* pull in the branch and install
* run this command in the terminal for a dry run of the publish action ( I have uploaded a dummy build to the CDN for this purpose )
```shell
node ./tools/scripts/upload-ab-to-s3.js --env staging --appId 123456 --licenseKey 987654321 --bucket nr-browser-agent --sha abcdefg --workflow workflow_dispatch --role <lookup role value in shared vault> --dry
```

Send me a message if you need help configuring AWS locally

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -6,7 +6,6 @@ on:
       - main
   workflow_dispatch:
   workflow_call:
-    secrets: inherit
 
 jobs:
   publish-to-s3:
--- a/.github/workflows/publish-experiment.yml
+++ b/.github/workflows/publish-experiment.yml
@@ -29,3 +29,7 @@ jobs:
           node ./tools/scripts/fastly.js \
             --fastly-key ${{ secrets.FASTLY_PURGE_KEY }} \
             --purge-path experiments/$GITHUB_REF_NAME/nr-loader-spa.min.js
+  run-dev-process:
+    needs: [publish-experiment-to-s3]
+    uses: ./.github/workflows/publish-dev.yml
+    secrets: inherit


tools/scripts/upload-ab-to-s3.js
```diff
@@ -205,3 +206,8 @@ function getOpenPrNums () {
     })
   })
 }
+
+async function getExperiments () {
+  const objects = await getListOfObjects(bucket, 'experiments/')
+  return objects.filter(({ Key }) => Key.includes('nr-loader-spa.min.js')).map(({ Key }) => Key)
```
patrickhousley, line 212: So are we just going with a flat folder for experiments for now? Not supporting multiple experiments?

metal-messiah: it will be a single nesting under `experiments/`

So basically `experiments/<branch-name>/nr-loader-spa.min.js` for each branch name that triggers an action.


The response from "getExperiments" should be a list of full truncated paths
```javascript
[
    "experiments/initial-test/nr-loader-spa.min.js",
    "experiments/rrweb/nr-loader-spa.min.js",
    // etc...
 ]
 ```

File: tools/scripts/upload-ab-to-s3.js [hunk 2]
```diff
@@ -98,27 +98,28 @@ function getIdFromUrl (url) {
...
-
-    console.log(`found ${contents.length} valid PR builds in CDN`)
-    let output = `window.NREUM=${JSON.stringify(config)};`
-    output += `
+  connectToS3(role, dry).then(async () => {
+    const filePaths = [
+      next, // defaults to dev build
+      ...(env !== 'dev' ? [current] : []), // defaults to current build
+      ...((env === 'dev' && await getOpenPrNums()) || []).map(num => `https://js-agent.newrelic.com/pr/PR-${num}/nr-loader-spa.min.js`),
+      ...((env === 'staging' && await getExperiments()) || []).map(s3ObjectName => `https://js-agent.newrelic.com/${s3ObjectName}`)
```
metal-messiah, line 106: Should we do this for `dev` too?  Could get rid of the PR build system completely and treat them the same between dev and staging

metal-messiah: PR builds do have a place i guess, since they don't always work as expected.  We should probably make sure we treat "experiments" as something polished enough to be in staging

patrickhousley (approved):

bjfield (approved):

cwli24 (approved):
:angry_old_man_yelling_at_Github_Actions:

PR #516, opened 2023-05-05
feat: Allow custom error grouping

bjfield:
The `setErrorHandler` method now supports applying custom fingerprints to exceptions. Events with the same fingerprint appear grouped together in the New Relic One errors inbox.
---

### Overview

This dummy PR is for documentation purposes. The branch makes a minimal, arbitrary change so the title and header (above) get picked up in change logs and release notes for v1.232.0.

### Related Issue(s)

https://github.com/newrelic/newrelic-browser-agent/pull/443/

### Testing

N/A


cwli24 (approved):
For clarity of general audience, this is actually available on the agent side for v1.230.0 and above, but NR1 was not yet able to process it at the time. It's working now, so those lower versions would also accept this.

PR #518, opened 2023-05-06
feat: adding query parameter supportability metrics

patrickhousley:
Adding supportability metrics to capture legacy network method usage, query parameter size statistics, and custom data size statistics.
---

### Overview

Adding supportability metrics (SMs) to capture the number of times the img network method is used, the bytes sent to each endpoint as query parameters, and the size of custom data.

The SM for RUM bytes sent `PageSession/Endpoint/1/BytesSent` has been updated to always send `0` since we do not send anything in the body of the RUM call. That will be updated later when we migrate the RUM call to a POST.

The SM for the img method does not account for the unload call on purpose because that is too late in the page and harvest lifecycle to account for it.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-8477
https://issues.newrelic.com/browse/NEWRELIC-8478
https://issues.newrelic.com/browse/NEWRELIC-8479

### DIFF
--- a/src/common/config/state/runtime.js
+++ b/src/common/config/state/runtime.js
@@ -6,7 +6,9 @@ import { BUILD_ENV, DIST_METHOD, VERSION } from '../../constants/env'
 
 const model = {
   buildEnv: BUILD_ENV,
-  bytesSent: {},
+  bytesSent: {}, // Used for SM to capture body bytes sent per endpoint
+  queryBytesSent: {}, // Used for SM to capture query parameter bytes sent per endpoint
+  imgMethodCount: 0, // Used for SM to capture count of time img network method is used
   customTransaction: undefined,
   disabled: false,
   distMethod: DIST_METHOD,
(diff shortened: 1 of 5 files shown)


src/common/config/state/runtime.js
```diff
@@ -6,7 +6,9 @@ import { BUILD_ENV, DIST_METHOD, VERSION } from '../../constants/env'
 
 const model = {
   buildEnv: BUILD_ENV,
-  bytesSent: {},
+  bytesSent: {}, // Used for SM to capture body bytes sent per endpoint
+  queryBytesSent: {}, // Used for SM to capture query parameter bytes sent per endpoint
+  imgMethodCount: 0, // Used for SM to capture count of time img network method is used
```
metal-messiah, line 11: I'm debating if we should keep storing things like this in the runtime.

If we send imgMethodCount = 2 on page hide, then the page is restored, imgMethodCount = might become 4 and send, skewing the average.

It might be safer/better to just invoke the SM subscriber directly, which will unload at page hide time and reset its state -- making each pagehide/pageshow an independent count.... Lets discuss

src/features/metrics/aggregate/index.js
```diff
@@ -97,6 +98,7 @@ export class Aggregate extends FeatureBase {
     try {
       if (this.resourcesSent) return
```
metal-messiah, line 99: This flag means that the metrics will only count up to the first unload (which could happen at vis change).  Is this what we want?

metal-messiah (approved):

bjfield (approved):

PR #519, opened 2023-05-07
chore: wdio browser matching

patrickhousley:
Adding support for skipping tests in wdio based on the browser being used.
---

### Overview

Adding support for skipping tests in wdio.

The matcher code was rewrote from what was used in JIL.
- The code is not split out into multiple files instead of being in one large file.
- The matching/testing logic is simpler while still being flexible with the spec string
- Only a single pre-defined matcher was carried over for now
- Was able to do a little cleanup on the wdio configs due to this change

Moved the sauce labs code to a shared folder and updated the package.json scripts.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7589

### Testing

Use the below example test to create different spec matchers.

Example test file:
```javascript
import SpecMatcher from '../../tools/browser-matcher/spec-matcher.mjs'
import { notInternetExplorer } from '../../tools/browser-matcher/common-matchers.mjs'

const notChrome = new SpecMatcher()
  .exclude('chrome')

describe('newrelic api', () => {
  let testHandle

  beforeEach(async () => {
    testHandle = await browser.getTestHandle()
  })

  afterEach(async () => {
    await testHandle.destroy()
  })

  test(notInternetExplorer)('not ie', async () => {
    const url = await testHandle.assetURL('instrumented.html', {
      loader: 'spa'
    })

    await Promise.all([
      testHandle.expectRum(),
      browser.url(url)
    ])
    const result = await browser.execute(function () {
      return typeof window.newrelic.addToTrace === 'function'
    })

    expect(result).toEqual(true)
  })

  test(notChrome)('not chrome', async () => {
    const url = await testHandle.assetURL('instrumented.html', {
      loader: 'spa'
    })

    await Promise.all([
      testHandle.expectRum(),
      browser.url(url)
    ])
    const result = await browser.execute(function () {
      return typeof window.newrelic.addToTrace === 'function'
    })

    expect(result).toEqual(true)
  })
})
```


tools/browser-matcher/common-matchers.mjs
```diff
@@ -0,0 +1,5 @@
+import SpecMatcher from './spec-matcher'
+
+// Does the browser have a reliable 'unload' event callback?
+export const notInternetExplorer = new SpecMatcher()
+  .exclude('ie')
```
metal-messiah, line 5: Are we blending common matchers with bespoke ones in each test, or adding all matchers here?

docs/integration-testing.md
```diff
@@ -0,0 +1,70 @@
...
+
+To run the tests against a remote Selenium instance, such as a Selenium Grid service, use the `--selenium-host <host>` and `--selenium-port <port>` arguments.
+
+## WDIO CLI
+
+The project wraps the WDIO execution in a custom CLI. When executing WDIO, you are not running the CLI that ships with WDIO. The customer CLI arguments are defined [here](../tools/wdio/args.mjs). The custom CLI will take arguments, construct a WDIO configuration, and execute WDIO.
+
+## Writing Tests
+
+All tests should exist in the path `tests/specs` for WDIO to pick them up. Tests should be grouped by the functionality being tested or scenario being ran. WDIO will launch a new browser for each file, execute all tests of that file in the browser, and close the browser. This means more files will launch more browsers improving test isolation but also increasing the amount of time it takes to run the full test suite. Keep this in mind when writing integration tests.
```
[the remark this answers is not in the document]
bjfield: Have we talked about the naming of this directory? (Doesn't apply to this PR but) wondering if `tests/e2e` or even `tests/wdio` would be more clear. OTOH I suppose eventually there will only be /assets and /specs in the tests dir.

docs/integration-testing.md
```diff
@@ -0,0 +1,70 @@
...
+
+- **DO** create a global `describe` wrapper to hold all the tests. The name of this wrapper will be used to name the test result in SauceLabs making it easier to review test failures.
+- **DO** create a `describe` to group tests that have similar pre/post test instructions.
+- **DON'T** save state between individual test cases.
+- **DO** load a new page to start each test.
+- **DO** use `beforeEach` to instantiate variables and state before each test.
+- **DO** use `afterEach` to cleanup global state after each test.
+- **DON'T** use `beforeAll` or `afterAll`.
```
bjfield, line 53:
```suggestion
- **DON'T** use `beforeAll` or `afterAll`, which can result in poor state isolation.
```

docs/integration-testing.md
```diff
@@ -0,0 +1,70 @@
...
+- **DO** use the `it` keyword to declare a test case that has no browser limitation.
+- **DO** use the `test()` function to declare a test case that has browser limitations.
+- **DO** avoid promise chaining and use `async/await` to test asynchronous code.
+- **DO** create custom commands in `tools/wdio/plugins/jil-commands.mjs` instead of writing utility methods for common testing functionality.
```
bjfield, line 57: Might should be renamed `custom-commands.mjs`?

tools/wdio/runner.mjs
```diff
@@ -30,9 +30,17 @@ if (['trace', 'debug', 'info'].indexOf(wdioConfig.logLevel) > -1) {
 // Clear the JIL CLI params before starting wdio so they are not passed to worker processes
 process.argv.splice(2)
 
+const jsonReplacer = (_, value) => {
+  if (value instanceof RegExp) {
+    return value.toString()
+  }
```
cwli24, line 36: This won't work to get a RegExp object back when deserializing; I've tried in JIL before (after spending too much 💀 time).
image

If can't use serialize-anything, you can take from the example in https://github.com/newrelic/newrelic-browser-agent/blob/ba7d17cd3cf3b0a665d8e1a4e1b1bba17d19e1df/tools/testing-server/utils/url.js#L30-L36 , but still needs to de-serialize it with replace (see init-transform.js, L27-30).

docs/integration-testing.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Tests for specific browsers
+
+In some cases, a test may only apply to or not work expectedly in certain browser. We have created a new global `test()` that can be used in place of `it`. The `test` method takes a browser matcher and returns a higher-order function. If the browser matcher matches the browser the test is currently running inside, the higher-order function will call the `it` global method to register the test case with the WDIO framework. Below is an example.
```
bjfield, line 61:
```suggestion
In some cases, a test may apply only to a subset of browsers. We have created a new global `test()` that can be used in place of `it`. The `test` method takes a browser matcher and returns a higher-order function. If the browser matcher matches the browser the test is currently running inside, the higher-order function will call the `it` global method to register the test case with the WDIO framework. Below is an example.
```

docs/integration-testing.md
```diff
@@ -0,0 +1,70 @@
...
+
+```javascript
+import Matcher from '../../tools/jil/util/browser-matcher'
```
cwli24, line 64: Does wdio `test()` work with the old `Matcher` from JIL?

cwli24 (changes requested):

tools/wdio/runner.mjs
patrickhousley: This works for this purpose because of how WDIO parses the configuration. The point of the `replacer` is to simply convert the `RegExp` to a string representation which WDIO supports. We do not ever parse the config after starting WDIO, WDIO does all that so we cannot use `serialize-anything`.

docs/integration-testing.md
patrickhousley: No, thanks for catching this. I forgot to update this doc.

tools/wdio/plugins/browser-matcher.mjs
```diff
@@ -0,0 +1,43 @@
...
+export default class BrowserMatcher {
+  #browserSpec
+
+  async beforeSession (_, capabilities) {
+    this.#browserSpec =
+      `${this.#getBrowserName(capabilities)}@${capabilities.browserVersion || capabilities.version}`
+    global.test = this.#test.bind(this)
+  }
+
+  #test (matcher) {
```
bjfield, line 14: I like the higher order function approach. I think this would be more clear if instead of `test` we had `inBrowsers` or `withBrowsersMatching` or similar. Seems like `test` is not very descriptive. I don't feel strongly since it's documented well in your .md, but something to consider if you want.

docs/integration-testing.md
```diff
@@ -0,0 +1,70 @@
...
+
+**Note** the team is in the process of migrating our integration tests from an in-house tool called JIL to WDIO. This doc will cover WDIO only.
+
+## Running Tests - SauceLabs
+
+To run the integration tests with SauceLabs, you must have two environment variables set: `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY`. You can find more information about these variables [here](https://docs.saucelabs.com/basics/environment-variables/).
+
+The integration tests will also need to be able to communicate to a mock BAM server running on the local machine. This is accomplished with a SauceLabs [tunnel](https://docs.saucelabs.com/secure-connections/sauce-connect/setup-configuration/basic-setup/). The tunnel can be started separately from the integration tests allowing multiple test runs to use the same tunnel and speeding up the test execution. To start the tunnel from the project, make sure all the project dependencies are installed and execute `npm run sauce:connect`.
```
bjfield, line 11:
```suggestion
The integration tests will also need to be able to communicate to a mock BAM (data collection) server running on the local machine. This is accomplished with a SauceLabs [tunnel](https://docs.saucelabs.com/secure-connections/sauce-connect/setup-configuration/basic-setup/). The tunnel can be started separately from the integration tests allowing multiple test runs to use the same tunnel and speeding up the test execution. To start the tunnel from the project, make sure all the project dependencies are installed and execute `npm run sauce:connect`.
```

bjfield (approved):

docs/integration-testing.md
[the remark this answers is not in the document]
patrickhousley: No we have not talked about it. I have no opinion either way honestly.

docs/integration-testing.md
patrickhousley: Agreed. I think I want to remove all references to `JIL` at some point.

tools/browser-matcher/common-matchers.mjs
patrickhousley: I suggest we create shared/common matchers in this file and import them into the tests. If a test needs a one-off matcher, it can be created in that test.

tools/browser-matcher/common-matchers.mjs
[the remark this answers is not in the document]
patrickhousley: It will be.

tools/wdio/plugins/browser-matcher.mjs
patrickhousley: I like that idea. What does everyone else think? @metal-messiah @cwli24

tools/browsers-lists/browsers-list.mjs
```diff
@@ -0,0 +1,26 @@
+import { parseSpecString } from '../browser-matcher/spec-parser.mjs'
+import SpecMatcher from '../browser-matcher/spec-matcher.mjs'
+
+export default function browsersList (browsersList, spec = 'chrome@latest') {
+  let { browserName, specOperator, browserVersion } = parseSpecString(spec, null)
```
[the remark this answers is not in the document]
patrickhousley: I think I got this addressed. Please check and let me know if there is anywhere else I may have missed the spec string having multiple comma delimited values.

tools/wdio/plugins/browser-matcher.mjs
metal-messiah: Agreed, `test` overlaps too much with the other mocha/jesty type command syntax

tools/browsers-lists/browsers-list.mjs
```diff
@@ -1,26 +1,30 @@
 import { parseSpecString } from '../browser-matcher/spec-parser.mjs'
 import SpecMatcher from '../browser-matcher/spec-matcher.mjs'
+import browsersSupported from './browsers-supported.json' assert { type: "json" }
```
metal-messiah, line 3: What version of Node do the GH actions run in?  Can they support this syntax?

tools/browser-matcher/matcher-rule.mjs
```diff
@@ -0,0 +1,104 @@
...
+   * @param {string} spec The browser spec to compare against the base browser spec
+   */
+  test (spec) {
+    const { browserName, _, browserVersion } = parseSpecString(spec)
+
+    if (this.#matcherType === MATCHER_TYPE.EXCLUDE && this.#browserName !== browserName) {
+      return true
+    } else if (this.#matcherType === MATCHER_TYPE.INCLUDE && this.#browserName !== browserName) {
+      return false
+    }
```
cwli24, line 66: There's a conflict of default behavior here. For example,

- If I `SpecMatcher().exclude('safari@10')` then `.test('chrome')` it will pass.
- If I `SpecMatcher().exclude('safari@10').include('safari@11')`, then `.test('chrome')` it will **fail**

You should pick a default when this rule doesn't concern `browserName`, regardless if it's an inclusion or exclusion: either return `true` and be inclusive, or return `false` and be exclusive.

tools/browsers-lists/browsers-list.mjs
```diff
@@ -0,0 +1,30 @@
...
+
+      if (!browserVersion) {
+        browserVersion = 999999
+      }
+
+      const matcher = new SpecMatcher()
+        .include(`${browserName}${specOperator}${browserVersion}`)
+
+      return Array.from(browsersList[browserName] || [])
+        .filter(sauceBrowser => matcher.test(`${browserName}@${sauceBrowser.version}`))
```
cwli24, line 28: Suggestion: pass in name & version as 2 args rather than make it a string to be re-parsed by each & every `MatcherRule.test`:
`const { browserName, _, browserVersion } = parseSpecString(spec)`

tools/browsers-lists/browsers-list.mjs
```diff
@@ -0,0 +1,30 @@
...
+        const specs = Array.from(browsersList[browserName] || [])
+          .sort((a, b) => Number(b.version) - Number(a.version))
+        return [specs[0]]
+      }
+
+      if (!specOperator) {
+        return Array.from(browsersList[browserName] || [])
+      }
+
+      if (!browserVersion) {
```
cwli24, line 20: Comment: hopefully no one specifies an operator but leaves version blank, like `chrome@` ... 🙃

tools/browsers-lists/browsers-list.mjs
cwli24, line 1: `*`, `*@latest`, or `*@*` is not yet accepted since `browserName === '*'` is not considered.

tools/browsers-lists/browsers-list.mjs
```diff
@@ -0,0 +1,30 @@
...
+        browserVersion = 999999
+      }
+
+      const matcher = new SpecMatcher()
+        .include(`${browserName}${specOperator}${browserVersion}`)
+
+      return Array.from(browsersList[browserName] || [])
+        .filter(sauceBrowser => matcher.test(`${browserName}@${sauceBrowser.version}`))
+    })
```
cwli24, line 29: I'm not sure if this makes sense inside `flatMap`. This is what `browsersList` should be doing:

1. For every spec in specString, separated by `,`, create a MatcherRule for that spec, add it to the SpecsMatcher.
2. Filter thru the entire browsersSupported list with the final SpecsMatcher, and these are the browser-versions that should be tested in this execution.

Right?

tools/browser-matcher/spec-parser.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * @example
+ * // throws an error because `#` is not a valid operator
+ * parseSpecString('ie#11/windows')
+ * @param {string} spec A string representing a browser spec
+ * @returns {object} The parsed spec
+ */
+export function parseSpecString (spec) {
+  // Drop the platformName if it exists
+  const [browserInfo, platform] = spec.split('/')
+  const [specOperator] = browserInfo.match(/(@|<=|>=|<|>)/) || []
```
cwli24, line 24: In JIL, the spec string is actually like `safari@>=15.1,chrome@100,edge@<105`, etc, whereas the `@` operator here is intended to imply `==` rather than permanently existing as part of the spec.

I suppose it's fine to make it the users' (team) responsibility to know the new format, but also want to say this _could_ be hardier to account for JIL's format by checking for a 2nd operator.

tools/browsers-lists/browsers-list.mjs
patrickhousley: Latest LTS version.

tools/browsers-lists/browsers-list.mjs
patrickhousley: No, the browsersList is only responsible for taking in a comma delimited string of browser specs and returning an array of capabilities that can be used in WDIO to connect to SauceLabs or Selenium. It does not have anything to do with browser matching within a test.

tools/browser-matcher/matcher-rule.mjs
patrickhousley: Those tests make sense. If the matcher is only excluding `safari@10` then `chrome` should pass.  The second test is excluding `safari@10` and only including `safar@11` so unless you pass Safari 11, it shouldn't pass. The default is already set....it matches the rules supplied, period. If you look at a lot of the predefined rules for JIL, they start with `include('*')` or `exclude('*')` for this very reason.

tools/browsers-lists/browsers-list.mjs
patrickhousley: I am pretty sure this is the exact way it worked for the existing JIL browsersList function but I will check.

patrickhousley: Yup confirmed. The current browserlist in jil processes each value in the comma delimited list independently. So this is matching that behavior and I think thats what we would want anyways.

cwli24: image
You're right about existing behavior that's broken: ran JIL with `-b "chrome@<113,chrome@>107"` and it still gave me all 4 because it doesn't consider the rules in tandem. It does look like de-dupe happens through the Set.

cwli24: This should be fixed though, no? Thoughts? @bjfield @metal-messiah

patrickhousley: I think the whole point of passing multiple browsers to the CLI via a comma delimited list is to run tests against each of those browser specs individually. Additionally, we don't ever pass multiple in any CI and I never do it personally so maybe this is a non-issue.

patrickhousley: I will fix the dedupe for sure.

tools/browsers-lists/browsers-list.mjs
patrickhousley: Not supported in the spec matcher by design.

tools/browser-matcher/spec-parser.mjs
patrickhousley: If we want to restore the syntax, I can. Right now, the string just needs an operator not an at sign and operator.
@metal-messiah @bjfield do you want me to restore the original syntax?

metal-messiah: I'm good either way

metal-messiah (approved):

PR #520, opened 2023-05-08
feat: Minor changes to SMs

metal-messiah:
Add internal metrics to track the duration of the average session to improve estimation for future efforts.
---

### Overview

- Fixes some missed cases with the session length supportability metrics.
- Creates and uses a specific `session-entity.html` test asset instead of generic `instrumented.html`.

### DIFF
--- a/src/common/session/session-entity.js
+++ b/src/common/session/session-entity.js
@@ -51,8 +51,6 @@ export class SessionEntity {
     this.expiresMs = expiresMs
     this.inactiveMs = inactiveMs
 
-    console.log(expiresMs, inactiveMs)
-
     this.ee = ee.get(agentIdentifier)
 
     wrapEvents(this.ee)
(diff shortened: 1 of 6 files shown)


patrickhousley (approved):

bjfield (approved):

src/common/session/session-entity.js
```diff
@@ -39,13 +39,18 @@ export class SessionEntity {
     if (!isBrowserScope) this.storage = new LocalMemory()
     else this.storage = storageAPI
 
+    this.sync(model)
```
cwli24, line 42: Should `inactiveAt` and `expiresAt` also default to meaningful values here, if those properties are getting unnecessarily set once here and again in the following blocks?

metal-messiah: Probably.  This is mostly to make sure keys checks always have a property exist, but open to opinions on what those meaningful defaults would be for those timestamps

PR #521, opened 2023-05-08
feat: Migrate RUM network call from GET to POST

patrickhousley:
When making its initial network call (the RUM call), the agent will now use the POST method instead of the GET method and will not depend on JSON-P to trigger follow-on events. This change facilitates setting custom attributes in the request body, reducing the likelihood of long GET URLs.
---

### Overview

Migrating the rum api from a JSON-P GET call to a POST call.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7999

### Testing

Integration tests should be ran and should pass.

**DO NOT MERGE** until the DEMPLA team completes the downstream requirements.


src/features/page_view_event/aggregate/index.js
```diff
@@ -112,32 +113,22 @@ export class Aggregate extends FeatureBase {
...
-    )
-    // Usually `drain` is invoked automatically after processing feature flags contained in the JSONP callback from
-    // ingest (see `activateFeatures`), so when JSONP cannot execute (as with module workers), we drain manually.
-    if (!isValidJsonp) drain(this.agentIdentifier, CONSTANTS.FEATURE_NAME)
+    harvester.send('rum', { qs: queryParameters, body }, { needResponse: true }, undefined, ({ responseText }) => {
+      try {
+        activateFeatures(JSON.parse(responseText), this.agentIdentifier)
+        drain(this.agentIdentifier, this.featureName)
+      } catch (err) {
+        handle('ierr', [err, performance.now(), true], undefined, this.featureName, this.ee)
```
metal-messiah, line 129: Just FYI, This won't do anything if RUM fails (since errors wont drain)

src/features/page_view_event/aggregate/index.js
```diff
@@ -112,32 +113,22 @@ export class Aggregate extends FeatureBase {
...
-    agentRuntime.bytesSent[protocol] = (agentRuntime.bytesSent[protocol] || 0) + queryString?.length || 0
-
-    const isValidJsonp = submitData.jsonp(
-      this.getScheme() + '://' + info.beacon + '/' + protocol + '/' + info.licenseKey + queryString,
-      jsonp
-    )
-    // Usually `drain` is invoked automatically after processing feature flags contained in the JSONP callback from
-    // ingest (see `activateFeatures`), so when JSONP cannot execute (as with module workers), we drain manually.
-    if (!isValidJsonp) drain(this.agentIdentifier, CONSTANTS.FEATURE_NAME)
+    harvester.send('rum', { qs: queryParameters, body }, { needResponse: true }, undefined, ({ responseText }) => {
```
metal-messiah, line 124: Do we need to handle retries / status codes under the new strat?

src/common/harvest/harvest.js
```diff
@@ -88,7 +88,7 @@ export class Harvest extends SharedContext {
 
     var agentRuntime = getRuntime(this.sharedContext.agentIdentifier)
 
-    if (!payload.body) { // no payload body? nothing to send, just run onfinish stuff and return
+    if (!payload.body && !payload.qs) { // no payload body? nothing to send, just run onfinish stuff and return
```
metal-messiah, line 91: Do we have cases where a request can be sent without a body?

cwli24: Why's payload w/ `qs` allowed through?

patrickhousley: Before this change, the page view event aggregator called directly to the send data module. I would like to remove that direct tying and just use the harvest module but if there are no custom attributes, then the body is empty for the rum post call. How to handle that?

metal-messiah: Theres some ways but IDK if I'm in love with any of them.  Could set up rules for each endpoint for instance, which might be needed for other features down the line re: retries, codes body vs query, etc

cwli24: Though I get making PVE "in line" with the others, not sure it's worth it to create a "harvest scheduler" for something only sent once on load then becomes irrelevant, unless the query params are moving to the body too.
(Metrics now only sent once too, but it re-uses the harvest's EoL logic & does send stuff in body.)

src/features/page_view_event/aggregate/index.js
patrickhousley: We do not currently handle retries for RUM right now. I have moved this discussion to slack and will open tickets as necessary and the work will be completed separately.

src/common/harvest/harvest.js
cwli24: Maybe rum consumer accepts & does nothing with an empty `{}` object for body

patrickhousley: The logic in the harvest module is to not send a body if it is empty. That would also affect every other endpoint.

src/common/harvest/harvest.js
```diff
@@ -263,7 +262,7 @@ function createAccumulator () {
   var accumulator = {}
   var hasData = false
   return function (key, val) {
-    if (val !== null && val !== undefined && val.length) {
+    if (val !== null && val !== undefined && val.toString()?.length) {
```
metal-messiah, line 265: Does this handle empty objects appropriately? (all objects.toString() will just create a string of `[object Object]` so just double checking this is intended

tests/functional/spa/basic.test.js
```diff
@@ -26,7 +26,7 @@ testDriver.test('capturing SPA interactions', supported, function (t, browser, r
       let interactionTree = (body && body.length ? body : querypack.decode(query.e))[0]
 
       t.equal(interactionTree.trigger, 'initialPageLoad', 'initial page load should be tracked with an interaction')
-      t.equal(interactionTree.children.length, 0, 'expect no child nodes')
+      t.ok(interactionTree.children[0].path.startsWith('/1/'), 'expect rum call child node')
```
metal-messiah, line 29: Do we know why this wasn't previously captured in a SPA page load but now it is?

src/common/harvest/harvest.js
```diff
@@ -98,7 +98,7 @@ export class Harvest extends SharedContext {
 
     var agentRuntime = getRuntime(this.sharedContext.agentIdentifier)
 
-    if (!payload.body) { // no payload body? nothing to send, just run onfinish stuff and return
+    if (!payload.body && ((opts && !opts.sendEmptyBody) || !opts)) { // no payload body? nothing to send, just run onfinish stuff and return
```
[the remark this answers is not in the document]
patrickhousley: The previous behavior was to short-circuit if there is no body. The new behavior short-circuits if the body is empty and opts are null or undefined or options.sendEmptyBody is null, undefined, or false.

src/common/harvest/harvest.js
patrickhousley: `mapOwn` takes care of empty objects.

src/common/harvest/harvest.js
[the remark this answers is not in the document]
metal-messiah: opts will never be undefined though, since theres a default to set it to `{}` in the params, so just double checking here

tests/functional/spa/basic.test.js
metal-messiah: TIL, but page view event didn't use the `img` method before did it?  It looks like it injected a script elem onto the page. But that's also a valid reason for sure, since neither of those cases are "ajax" events.

src/common/harvest/harvest.js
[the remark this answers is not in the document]
metal-messiah: Thanks for proof-checking that.  I think its mostly just the way it's put together that makes it a little difficult to comprehend the intention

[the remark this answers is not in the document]
metal-messiah: Does this cover the same ground?

```suggestion
    if (!payload.body && !opts?.sendEmptyBody) { // no payload body? nothing to send, just run onfinish stuff and return
```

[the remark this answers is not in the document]
metal-messiah: Screenshot 2023-06-01 at 1 47 56 PM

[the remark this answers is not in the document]
patrickhousley: Yup

tests/functional/spa/basic.test.js
patrickhousley: HAHA!! Yea I was confusing two different things. SO I did track that down. Our JSONP wrapper expects a parameter in the URL of `callback` or `cb`. Our rum call used `jsonp` as the query parameter. I honestly doubt our JSONP wrapping is effective.

metal-messiah (approved):

src/features/page_view_event/aggregate/index.js
```diff
@@ -67,27 +66,27 @@ export class Aggregate extends AggregateBase {
...
+      ...(
+        Object.entries(this.aggregator.get('measures') || {}).reduce((aggregator, [metricName, measure]) => {
+          aggregator[metricName] = measure.params?.value
+          return aggregator
+        }, {})
+      ),
+      xx: info.extra,
+      ua: info.userAttributes,
+      at: info.atts
+    }
```
bjfield, line 84:
```suggestion
    }

    var customJsAttributes = stringify(info.jsAttributes)
    queryParameters.ja = customJsAttributes === '{}' ? null : customJsAttributes
```

bjfield: There's one way to fix the functional tests.

patrickhousley: The `ja` attribute was moved to the body on purpose.

src/common/harvest/harvest.js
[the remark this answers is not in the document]
cwli24: Oh it does now; anyways that was more of an internal comment so reader knows what branch is being used by what. That if case is no different from the else case, so they could really be merged.

bjfield (approved):

PR #522, opened 2023-05-09
fix: Identify inline stack trace URLs more precisely (NEWRELIC-7001)

bjfield:
URLs in `JavaScriptError` stack traces were sometimes misidentified as inline scripts when they shared a partial URL with the page on which the agent was loaded. The agent will now identify same-page scripts more precisely.
---

### Overview

- Move the `canonicalizeUrl` function into its own file with tests.
- Canonicalize URLs not in the aggregator as before but when the stack trace is first computed.
- Add JSDoc blocks in affected files and rename some variables for clarity.
- Modify fastly configuration to pick up implicit `index.html` files at directory paths.
- Create a test asset at a directory path.
  - This is for functional testing of errors in scripts at sub-paths of the loader origin.
- Define `initialLocation` as a new export from `global-scope` module.
- Add end-to-end tests to verify <inline> behavior.

### Related Issue(s)

[NEWRELIC-7001](https://issues.newrelic.com/browse/NEWRELIC-7001)

### Testing

- All automated tests should still pass.
- A new e2e test should pass: `tests/specs/stack-trace.e2e.js`.


src/features/jserrors/aggregate/compute-stack-trace.js
```diff
@@ -66,9 +67,34 @@ var gecko = /^\s*(?:(\S*|global code)(?:\(.*?\))?@)?((?:file|http|https|chrome|s
...
+/**
+ * Attempts to compute a stack trace for the given exception.
+ * @param {Error} ex - The exception for which to compute the stack trace.
+ * @param {string} loaderOriginUrl - The origin URL of the agent loader, for identifying inline scripts.
+ * @returns {StackInfo} A stack trace object containing information about the frames on the stack.
+ */
+export function computeStackTrace (ex, loaderOriginUrl) {
   var stack = null
 
+  _loaderOriginUrl = loaderOriginUrl
```
metal-messiah, line 96: It feels like we are storing some flavor of the origin in a lot of places. This probably isn't a big deal, but just worth calling it out.

metal-messiah: Actually, it looks like `canonicalizeUrl` will call cleanUrl.  Can we just pass in the origin directly instead of keeping a "cleaned url" in the agg file and passing it around?

src/common/url/canonicalize-url.js
```diff
@@ -0,0 +1,26 @@
...
+import { cleanURL } from './clean-url'
+
+/**
+ * Converts a URL to its basic form without a query string or fragment. If the resulting URL is the same as the
+ * loader's origin URL, returns '<inline>'.
+ * @param {string} url - The URL to be canonicalized.
+ * @param {string} loaderOriginUrl - The origin URL of the agent loader, used for inline detection.
+ * @returns {string} The canonicalized URL, or '<inline>' if the URL matches the loader origin URL.
+ */
+export function canonicalizeUrl (url, loaderOriginUrl) {
```
patrickhousley, line 15: Couldn't we just use `window?.location.href || navigation?.currentEntry.url` as the value for loaderOriginUrl and not have to pass that value in?

bjfield: Now pulling from globalScope.

src/features/jserrors/aggregate/compute-stack-trace.js
bjfield: Now getting current page location directly within canonicalizeUrl.

src/common/url/canonicalize-url.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param {string} loaderOriginUrl - The origin URL of the agent loader, used for inline detection.
+ * @returns {string} The canonicalized URL, or '<inline>' if the URL matches the loader origin URL.
+ */
+export function canonicalizeUrl (url) {
+  if (typeof url !== 'string') return ''
+
+  const cleanedUrl = cleanURL(url)
+  const cleanedGlobalScopeUrl = cleanURL(initialLocation)
+
+  // If the URL matches the origin URL of the loader, we assume it originated with an inline script.
```
patrickhousley, line 22:
```suggestion
  // If the URL matches the origin URL of the loader, we assume it originated within an inline script.
```

src/common/util/global-scope.js
```diff
@@ -20,6 +20,8 @@ export let globalScope = (() => {
   throw new Error('New Relic browser agent shutting down due to error: Unable to locate global scope. This is possibly due to code redefining browser global variables like "self" and "window".')
 })()
 
+export let initialLocation = '' + globalScope.location
```
patrickhousley, line 23: I don't think we need the ability to set this variable anywhere, right?

```suggestion
export const initialLocation = '' + globalScope.location
```

patrickhousley (approved):

metal-messiah (approved):

PR #527, opened 2023-05-12
chore: Session trace spring cleaning NEWRELIC-8524

cwli24:
NA
---

### Overview

Dusting of the trace feature that hopefully makes it easier to read & understand. Unnecessary code were removed, logic checks updated for reliability, and variables renamed for clarity.

### Related Issue(s)

NEWRELIC-8524


src/features/session_trace/instrument/index.js
```diff
@@ -71,57 +58,23 @@ export class Instrument extends InstrumentBase {
       handle('bstHist', [location.pathname + location.hash, this.startPath, this.time], undefined, FEATURE_NAMES.sessionTrace, thisInstrumentEE)
     })
 
-    if (supportsPerformanceObserver()) {
-      // capture initial resources, in case our observer missed anything
-      handle(BST_RESOURCE, [window.performance.getEntriesByType('resource')], undefined, FEATURE_NAMES.sessionTrace, thisInstrumentEE)
```
metal-messiah, line 76: `[window.performance.getEntriesByType('resource')]` is older and more universally supported than the performance observer callback subscription.  For example, IE supports getEntriesByType, but not PerformanceObserver

tests/assets/stn/ensure-passive.html
```diff
@@ -21,6 +21,11 @@
     </script>
     {config}
     {loader}
+    <script>
+      try {
+        document.addEventListener('scroll', function(e){}, {capture:false, passive:true})
```
metal-messiah, line 26: This test was meant to ensure that the listeners the agent sets up are passive.  This test is pointless the way it's constructed if stn does not have the listeners added

src/features/session_trace/aggregate/index.js
```diff
@@ -200,31 +173,54 @@ export class Aggregate extends FeatureBase {
...
     if (t && t instanceof XMLHttpRequest) {
-      var params = this.ee.context(t).params
+      const params = this.ee.context(t).params
       if (!params || !params.status || !params.method || !params.host || !params.pathname) return 'xhrOriginMissing'
       origin = params.status + ' ' + params.method + ': ' + params.host + params.pathname
     } else if (t && typeof (t.tagName) === 'string') {
       origin = t.tagName.toLowerCase()
       if (t.id) origin += '#' + t.id
-      if (t.className) origin += '.' + slice(t.classList).join('.')
+      if (t.className) origin += '.' + t.classList.value.replaceAll(' ', '.') // p.s. IE11 may not be happy about DOMTokenList.value (?)
```
metal-messiah, line 223: replaceAll isn't supported in ie11

src/features/session_trace/aggregate/index.js
```diff
@@ -237,43 +233,43 @@ export class Aggregate extends FeatureBase {
...
   }
 
+  // Processes all the PerformanceResourceTiming entries captured (by observer).
   storeResources (resources) {
     if (!resources || resources.length === 0) return
 
     resources.forEach((currentResource) => {
-      var parsed = parseUrl(currentResource.name)
-      var res = {
+      if (currentResource.fetchStart <= this.laststart) return // don't recollect already-seen resources
```
metal-messiah, line 253: this was comparing `.fetchStart | 0` before, is this operation no longer necessary?

src/features/session_trace/instrument/index.js
```diff
@@ -71,57 +58,23 @@ export class Instrument extends InstrumentBase {
...
-    }
-
-    function observeResourceTimings () {
-      var observer = new PerformanceObserver((list, observer) => { // eslint-disable-line no-undef
-        var entries = list.getEntries()
-
+    let observer
+    try {
+      // Capture initial resources and watch for future ones. Don't defer this given there's a default cap on the number of buffered entries.
+      observer = new PerformanceObserver((list) => { // eslint-disable-line no-undef
```
metal-messiah, line 64: If we are going to use the PO with buffered true, does this still need to be in the instr file?

src/features/session_trace/instrument/index.js
```diff
@@ -71,57 +58,23 @@ export class Instrument extends InstrumentBase {
...
-
-  /** Restoration and resource release tasks to be done if Session trace loader is being aborted. Unwind changes to globals. */
-  #abort () {
-    window.performance.removeEventListener(RESOURCE_TIMING_BUFFER_FULL, this.onResourceTimingBufferFull, false)
-    // The doc interaction noOp listeners are harmless--cannot buffer data into EE.
-    this.abortHandler = undefined // weakly allow this abort op to run only once
-  }
-
-  noOp (e) { /* no-op */ }
+    // document.addEventListener('scroll', noOp, eventListenerOpts(false))
```
metal-messiah, line 73: what was the point of these?

src/features/session_trace/instrument/index.js
metal-messiah: this would also prevent the need from passing it around between inst and agg

src/features/session_trace/instrument/index.js
metal-messiah: i guess that wasn't running anyway, was it?

src/features/session_trace/instrument/index.js
cwli24: This way better ensures there's no gap in the resource collected for **heavy** pages, since the buffer can be filled, and overflow won't be captured w/o an observer.

src/features/session_trace/aggregate/index.js
cwli24: Oh thanks for noticing that. I'm not sure when the old `PT.fetchStart` would ever be undefined, but I'll just add it back here anyways.

src/features/session_trace/aggregate/index.js
cwli24: The `.value` actually isn't supported by ie either (DOMTokenList.value). This is annoying bc none of the other shorthand replacements like `new Array(t.classList)` works in ie11 either.
I guess I'll just make this an old `for...in` loop.

src/features/session_trace/instrument/index.js
patrickhousley: I think we might be able to find some reductions of our bundle sizes if were were to 1. pass the agent into the instrument file when instantiating and 2. pass the instrument into the aggregator when instantiating. Just in general, we have a lot of code dedicated to getting the correct event emitter, passing around the agent identifier, etc. Realistically, if the Agent class is instantiated for every agent on the page and manages all the instances of "things" passed down into features, we could probably strip out some code in the agent.

src/features/session_trace/instrument/index.js
patrickhousley: If these are not needed, can we remove the commented out code?

tests/assets/stn/ensure-passive.html
metal-messiah: yes

src/features/session_trace/aggregate/index.js
```diff
@@ -4,124 +4,95 @@
...
   static featureName = FEATURE_NAME
-  constructor (agentIdentifier, aggregator) {
+  constructor (agentIdentifier, aggregator, argsArray) {
     super(agentIdentifier, aggregator, FEATURE_NAME)
 
     // Very unlikely, but in case the existing XMLHttpRequest.prototype object on the page couldn't be wrapped.
     if (!getRuntime(agentIdentifier).xhrWrappable) return
 
-    const handlerCache = new HandlerCache()
+    this.resourceObserver = argsArray?.[0]
```
metal-messiah, line 40: This is trivial, but if we are going to do pass things from inst to agg can we pass them as a dict instead of an array, for ease of readability later?

So instead of `argsArray?.[0]` it'd be something like `args.resourceObserver`

src/features/session_trace/instrument/index.js
cwli24: Removed

src/features/session_trace/aggregate/index.js
cwli24: Changed

cwli24:
Smoke test ✔️ after PR comments <26b1ddf>

src/cdn/polyfills.js
```diff
@@ -7,6 +7,8 @@ import 'core-js/stable/promise'
 import 'core-js/stable/array/includes'
 import 'core-js/stable/array/from'
 import 'core-js/stable/array/find'
+import 'core-js/stable/array/flat'
+import 'core-js/stable/array/flat-map'
```
metal-messiah, line 11: Where are these being used?

metal-messiah (approved):

metal-messiah:
approved once appropriate review of polyfills included in this PR is conducted

src/cdn/polyfills.js
cwli24: Confirmed a-ok

PR #529, opened 2023-05-15
chore: add support for saucelabs extended debugging

patrickhousley:
Adding support for SauceLabs extended debugging feature.
---

### Overview

This PR adds support for SauceLabs [extended debugging](https://docs.saucelabs.com/insights/debug/) feature. It also adds an extra log entry when running tests using JIL to show which test is related to which testId (license key). This will make it easier to track down tests that are failing due to BAM expect timeouts. The expects have also been updated to provide each function a proper name so the log entry for a timeout will show which BAM endpoint timed out.

### Related Issue(s)

N/A

### DIFF
--- a/tests/assets/worker/js/module-worker.js
+++ b/tests/assets/worker/js/module-worker.js
@@ -3,9 +3,8 @@
 onmessage = function (e) {
   if (e.data.type === 'startAgent') {
     self.NREUM = e.data.payload
-    import('../../../../../../../../web-worker-agent').then(() => {
-      self.postMessage({ type: 'ready' })
-    })
+    importScripts('/web-worker-agent') // JIL's endpoint to fetch the single bundle web worker pkg
+    self.postMessage({ type: 'ready' })
   } else if (e.data.type === 'command') {
     // Let errors go unhandled so bad commands crashes the tests for troubleshooting.
     let retVal = eval(e.data.fn) // run the literal string cmd
--- a/tests/assets/worker/js/module-worker.js
+++ b/tests/assets/worker/js/module-worker.js
@@ -3,8 +3,10 @@
 onmessage = function (e) {
   if (e.data.type === 'startAgent') {
     self.NREUM = e.data.payload
-    importScripts('/web-worker-agent') // JIL's endpoint to fetch the single bundle web worker pkg
-    self.postMessage({ type: 'ready' })
+    // Module workers do not support importScripts, need to import directly
+    import('../../../../../../../../web-worker-agent').then(() => {
+      self.postMessage({ type: 'ready' })
+    })
   } else if (e.data.type === 'command') {
     // Let errors go unhandled so bad commands crashes the tests for troubleshooting.
     let retVal = eval(e.data.fn) // run the literal string cmd
--- a/tests/functional/final-harvest.test.js
+++ b/tests/functional/final-harvest.test.js
@@ -267,13 +267,13 @@ testDriver.test('final harvest sends resources', reliableResourcesHarvest.and(st
   let url = router.assetURL('final-harvest.html')
   let loadPromise = browser.safeGet(url).catch(fail)
   let rumPromise = router.expectRum()
+  let resourcesPromise = router.expectResources()
 
-  Promise.all([rumPromise, loadPromise])
-    .then(() => {
-      t.equal(router.requestCounts.bamServer.resources, 1, 'resources harvest is sent on startup')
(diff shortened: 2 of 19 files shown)


tests/assets/worker/js/module-worker.js
```diff
@@ -3,6 +3,7 @@
 onmessage = function (e) {
   if (e.data.type === 'startAgent') {
     self.NREUM = e.data.payload
+    // Module workers do not support importScripts, need to import directly
     import('../../../../../../../../web-worker-agent').then(() => {
```
metal-messiah, line 7: I know this is unrelated to this PR, but this import path is goofy.

patrickhousley: I know, it was so confusing and I changed it...which broke the tests. It's actually an import path relative to the script url when loaded in the browser.

metal-messiah (approved):

tests/functional/final-harvest.test.js
```diff
@@ -283,9 +283,8 @@ testDriver.test('final harvest sends resources', reliableResourcesHarvest.and(st
 
       return Promise.all([resourcesPromise, domPromise])
     })
-    .then((results) => {
-      t.equal(router.requestCounts.bamServer.resources, 2, 'received second resources harvest')
-      t.ok(results[0].request.body, 'received res harvest')
+    .then(([{ request: { body } }]) => {
+      t.ok(body, 'received res harvest')
```
cwli24, line 287:
```suggestion
      t.ok(body, 'received second res harvest')
```
Makes test failures more clear on where problem is

tools/jil/runner/index.js
```diff
@@ -134,6 +134,10 @@ function loadBrowsersAndRunTests () {
     } else {
       let sauceCreds = getSauceLabsCreds()
       connectionInfo = `http://${sauceCreds.username}:${sauceCreds.accessKey}@ondemand.saucelabs.com/wd/hub`
+
+      if (config.sauceExtendedDebugging && desired.browserName === 'chrome') {
```
cwli24, line 138: > Google Chrome (no older than 3 versions from latest) or Firefox browser (versions 53 and above) as the test browser

If the version is older than 3 from latest, there's no problem, right? What about including `firefox` too?

tools/wdio/config/sauce.conf.mjs
```diff
@@ -30,7 +30,14 @@ function sauceCapabilities () {
       platform: undefined,
       version: undefined,
       'sauce:options': !jilArgs.sauce
-        ? { tunnelName: getSauceConnectTunnelName() }
+        ? {
+            tunnelName: getSauceConnectTunnelName(),
+            ...(() => {
+              if (jilArgs.sauceExtendedDebugging && sauceBrowser.browserName === 'chrome') {
```
cwli24, line 36: Also related to the other comment here.

tools/jil/runner/index.js
patrickhousley: >Extended Debugging generates additional assets that impact test performance and is not intended for everyday testing.

I didn't even see the line below including the extendedDebugging attribute. Thanks for pointing that out. Given the above note from the SL docs, I don't think we should have extended debugging on all the time. In fact, I wonder if having it turned off by default during CI will help speed the CI tests up any. Maybe just a small boost. I'll fix this in a bit.

tools/browsers-lists/extended-debugging.mjs
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  if (browserName === 'firefox' && (Number(browserVersion) >= 53 || Number(version) >= 53)) {
+    return true
+  }
+
+  const latestChrome = browsersAll.chrome
+    .reduce((aggregator, sauceBrowser) => aggregator = Math.max(Number(sauceBrowser.version), Number(sauceBrowser.browserVersion), aggregator), 0)
+  return (Number(browserVersion) || Number(version)) >= latestChrome - 2
+}
```
cwli24, line 19: Side comment: it may be faster to just sift thru `browsers-supported` instead for what would practically be the same result. Also, `firefox` < 53 would fall down here though it would realistically never fulfill this condition, of course.

cwli24 (approved):

tools/browsers-lists/extended-debugging.mjs
patrickhousley: Good point.

Issue #526, opened 2023-05-11
X-NewRelic-ID: undefined being set on same origin XHR requests


asos-rupeshghelani:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

The agent is setting a request header unnecessarily when making same-origin XHR requests.

This [code](https://github.com/newrelic/newrelic-browser-agent/blob/6c5cbf0a2628d44a5792be9daea47818310117d4/src/features/ajax/instrument/index.js#L102) `if ('xpid' in loader_config && this.sameOrigin) {`  is always returning `true`, because the `xpid` key is [always present](https://github.com/newrelic/newrelic-browser-agent/blob/6c5cbf0a2628d44a5792be9daea47818310117d4/src/common/config/state/loader-config.js#L10).

And because `xpid` is `undefined`,  the header value is blindly set to `undefined`.

I expect a more robust check that `xpid` is set to a valid value.

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

## Steps to Reproduce

[NOTE]: # ( Please be as specific as possible. )

1. load browser agent Pro or higher with default config
2. make an XHR request to the same-origin
3. inspect the request headers of the XHR

## Expected Behavior
`X-NewRelic-ID` HTTP request header should not be set

## Actual Behavior
`X-NewRelic-ID: undefined ` HTTP request header is set

[NOTE]: # ( Tell us what you expected to happen. )

## Relevant Logs / Console output

[NOTE]: # ( Please provide specifics of the local error logs, Browser Dev Tools console, etc. if appropriate and possible. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )

* ex: Browser name and version: any 
* ex: Operating System and version: any

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

<img width="849" alt="xpid_is_undefined" src="https://github.com/newrelic/newrelic-browser-agent/assets/34572581/9d5997b9-5e62-4eb9-b62c-6c4cd540800b">

<img width="601" alt="Fullscreen_11_05_2023__09_40" src="https://github.com/newrelic/newrelic-browser-agent/assets/34572581/3af8f327-668e-4a44-91eb-fa06b07b4525">

bjfield:
[...]

PR #531, opened 2023-05-16
fix: Add X-NewRelic-ID header only if defined - NEWRELIC-8521

bjfield:
The agent will now set the `X-NewRelic-ID` header on XHR requests only when a truthy `xpid` value is explicitly set in the loader configuration. Previously the header was added even if undefined.
---

### Overview

As described in the bug linked below, the `X-NewRelic-ID` header was getting set even when the xpid configuration value had not been set. This is likely an artifact of an earlier agent architecture in which there was no default model for the configuration, which meant `xpid` was not set except manually. Now with a model, `xpid` always exists with at least a default value of `undefined`.

### Related Issue(s)

- https://github.com/newrelic/newrelic-browser-agent/issues/526
- [NEWRELIC-8521](https://issues.newrelic.com/browse/NEWRELIC-8521)

### Testing

- Automated tests should pass.
- The linked bug provides manual replication steps.

### DIFF
--- a/src/features/ajax/instrument/index.js
+++ b/src/features/ajax/instrument/index.js
@@ -99,7 +99,7 @@ function subscribeToEvents (agentIdentifier, ee, handler, dt) {
 
   function onOpenXhrEnd (args, xhr) {
     var loader_config = getLoaderConfig(agentIdentifier)
-    if ('xpid' in loader_config && this.sameOrigin) {
+    if (loader_config.xpid && this.sameOrigin) {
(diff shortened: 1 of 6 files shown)


metal-messiah:
Thanks for the quick work on this.  To prevent repeating mistakes, bugs like this could benefit from a quick test add.  undefined header should not get set, something to that effect.  Would you mind adding one?

patrickhousley (approved):
Tests look good.

metal-messiah (approved):

PR #532, opened 2023-05-16
feat: Update agent internals in early preparation for new features

metal-messiah:
The agent code base now contains the non-functional foundations of session replay mechanics. Note that this code is dormant and disabled. It will not download or execute until the session replay feature and its downstream dependencies are in a finished state. 
---

### Overview
This PR contains internal changes to support upcoming features, and introduces some [WIP] feature modules.  These modules will not be included in any agent builds until a complete UX is designed and implemented.

Important notes:
* The feature will not run, as it is not included in the CDN loader classes. Also -- For future privacy considerations for customers, this feature will only download and execute any recording code AFTER a check for customer entitlements and enablement has been conducted on the backend. This means that the recoding functionality will have a slight delay from the beginning of the page lifecycle, and could theoretically miss recording something very early in the lifecycle of the page.  This is a known and acknowledged risk, which is outweighed by privacy considerations.
* The entitlement check has been mocked in the local test server as `sr: 1`.  
* As of today, the sampling decision is rolled locally for both error and full replay states. This may change in future states.

### Testing
Testing has not been adapted to this feature yet.  Along with the tests themselves, testing architecture needs to be added to support this feature, namely mock route handling for the `/blob` endpoint.


src/common/session/session-entity.js
```diff
@@ -8,21 +8,29 @@ import { DEFAULT_EXPIRES_MS, DEFAULT_INACTIVE_MS, PREFIX } from './constants'
...
+// this is what can be stored in local storage (not enforced but probably should be)
+// these values should sync between local storage and the parent class props
 const model = {
   value: '',
   inactiveAt: 0,
   expiresAt: 0,
   updatedAt: Date.now(),
-  sessionReplayActive: false,
+  sessionReplay: 0,
   sessionTraceActive: false,
```
patrickhousley, line 24: Should this also be a zero to match the data types we expect to get back from the rum call?

metal-messiah: This is probably a little confusing and could use some polishing, but this value is actually going to map to the "mode" of the session replay which (as of right now, could change) is the result of 2 local sample decisions

```
const MODE = {
  OFF: 0,
  FULL: 1,
  ERROR: 2
}
```

src/common/session/session-entity.js
```diff
@@ -31,22 +39,22 @@ export class SessionEntity {
...
   constructor (opts) {
+    this.state = {}
     this.setup(opts)
   }
 
   setup ({ agentIdentifier, key, value = generateRandomHexString(16), expiresMs = DEFAULT_EXPIRES_MS, inactiveMs = DEFAULT_INACTIVE_MS, storageAPI = new LocalMemory() }) {
     if (!agentIdentifier || !key) throw new Error('Missing Required Fields')
     if (!isBrowserScope) this.storage = new LocalMemory()
-    else this.storage = storageAPI
+    else this.storage = storageAPI // new LocalStorage()
```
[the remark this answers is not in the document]
metal-messiah: will remove

src/common/timer/interaction-timer.js
```diff
@@ -6,8 +6,9 @@ import { isBrowserScope } from '../util/global-scope'
 export class InteractionTimer extends Timer {
   constructor (opts, ms) {
     super(opts, ms)
-    this.onRefresh = opts.onRefresh
     this.onPause = opts.onPause
+    this.onRefresh = opts.onRefresh
+    this.onResume = opts.onResume
```
patrickhousley, line 11: Maybe check if `opts.*` is a function or assign a `noop`.

```suggestion
    this.onPause = typeof opts.onPause === 'function' ? opts.onPause : () => { /* noop */ }
    this.onRefresh = typeof opts.onRefresh === 'function' ? opts.onRefresh : () => { /* noop */ }
    this.onResume = typeof opts.onResume === 'function' ? opts.onResume : () => { /* noop */ }
```

src/common/timer/interaction-timer.js
```diff
@@ -42,7 +43,10 @@ export class InteractionTimer extends Timer {
       subscribeToVisibilityChange((state) => {
         if (state === 'hidden') this.pause()
         // vis change --> visible is treated like a new interaction with the page
-        else this.refresh()
+        else {
+          this.onResume?.()
```
patrickhousley, line 47: Would not need optional chaining if applying above comment on using `noop` functions.

src/common/timer/interaction-timer.js
```diff
@@ -71,5 +75,6 @@ export class InteractionTimer extends Timer {
   //   if (!this.remainingMs || !this.isValid()) return
   //   this.timer = this.create(this.cb, this.remainingMs)
   //   this.remainingMs = undefined
+  //   this.onResume?.()
```
patrickhousley, line 78: If this is not being used, maybe we remove this commented out code and remove the `opts.onResume` in the constructor.

src/common/util/feature-flags.js
```diff
@@ -12,7 +12,10 @@ const bucketMap = {
   stn: [FEATURE_NAMES.sessionTrace],
   err: [FEATURE_NAMES.jserrors, FEATURE_NAMES.metrics],
   ins: [FEATURE_NAMES.pageAction],
-  spa: [FEATURE_NAMES.spa]
+  spa: [FEATURE_NAMES.spa],
+  sr: [FEATURE_NAMES.sessionReplay],
+  srf: [FEATURE_NAMES.sessionReplay],
+  sre: [FEATURE_NAMES.sessionReplay]
```
patrickhousley, line 18: Should probably add comments on how these three features are mutually exclusive and what each one "means".

src/common/util/submit-data.js
```diff
@@ -11,7 +11,7 @@ export const submitData = {}
  * @param {string} jsonp
  * @returns {Element}
  */
-submitData.jsonp = function jsonp (url, jsonp) {
+submitData.jsonp = function jsonp ({ url, jsonp }) {
```
patrickhousley, line 14: Need to update the jsdocs.

src/common/util/submit-data.js
```diff
@@ -46,7 +46,7 @@ submitData.xhrGet = function xhrGet (url) {
  * @param {boolean} sync
  * @returns {XMLHttpRequest}
  */
-submitData.xhr = function xhr (url, body, sync, method = 'POST') {
+submitData.xhr = function xhr ({ url, body, sync, method = 'POST', gzipped, licenseKey }) {
```
patrickhousley, line 49: Need to update the jsdocs.

src/common/util/submit-data.js
```diff
@@ -57,7 +57,13 @@ submitData.xhr = function xhr (url, body, sync, method = 'POST') {
     // do nothing
   }
 
-  request.setRequestHeader('content-type', 'text/plain')
+  if (gzipped) {
+    request.setRequestHeader('content-type', 'application/json')
+    request.setRequestHeader('X-Browser-Monitoring-Key', licenseKey)
```
patrickhousley, line 62: Is this still necessary? Adding this header causes CORS options calls to be made to our edge network and is not compatible with `sendBeacon`.

src/common/util/submit-data.js
```diff
@@ -74,7 +80,7 @@ submitData.xhr = function xhr (url, body, sync, method = 'POST') {
  * @param {string} url
  * @returns {Element}
  */
-submitData.img = function img (url) {
+submitData.img = function img ({ url }) {
```
patrickhousley, line 83: Need to update the jsdocs.

src/common/util/submit-data.js
```diff
@@ -86,7 +92,7 @@ submitData.img = function img (url) {
  * @param {string} body
  * @returns {boolean}
  */
-submitData.beacon = function (url, body) {
+submitData.beacon = function ({ url, body }) {
```
patrickhousley, line 95: Need to update the jsdocs.

src/features/jserrors/instrument/index.js
```diff
@@ -124,6 +124,8 @@ function notice (err, doNotStamp, ee) {
   // this is to preserve existing behavior
   var time = (!doNotStamp) ? now() : null
   handle('err', [err, time], undefined, FEATURE_NAMES.jserrors, ee)
+  // emit to session replay feature if present
+  handle('err', [err, time], undefined, FEATURE_NAMES.sessionReplay, ee)
```
patrickhousley, line 128: Just wondering if we are expecting any alteration of the err object by the session replay feature like the spa feature does.

src/features/page_view_event/aggregate/index.js
```diff
@@ -135,10 +135,10 @@ export class Aggregate extends FeatureBase {
     // Capture query bytes sent to RUM call endpoint (currently `1`) as a supportability metric. See metrics aggregator (on unload).
     agentRuntime.queryBytesSent[protocol] = (agentRuntime.queryBytesSent[protocol] || 0) + queryString?.length || 0
 
-    const isValidJsonp = submitData.jsonp(
-      this.getScheme() + '://' + info.beacon + '/' + protocol + '/' + info.licenseKey + queryString,
+    const isValidJsonp = submitData.jsonp({
+      url: this.getScheme() + '://' + info.beacon + '/' + protocol + '/' + info.licenseKey + '?' + queryString,
```
patrickhousley, line 139: Just double checking the `queryString` does not already have the `?` character.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    })
+  }
+
+  /** Store a payload in the buffer (this.events).  This should be the callback to the recording lib noticing a mutation */
+  store (event, isCheckout) {
+    if (this.blocked) return
+    const eventBytes = stringify(event).length
+    /** The estimated size of the payload after compression */
+    const payloadSize = this.getPayloadSize(eventBytes)
+    // Vortex will block payloads at a certain size, we might as well not send.
```
patrickhousley, line 243: Can we split the payload up and send it in multiple chunks?

src/features/session_replay/constants.js
```diff
@@ -0,0 +1,3 @@
+import { FEATURE_NAMES } from '../../loaders/features/features'
+
+export const FEATURE_NAME = FEATURE_NAMES.sessionReplay
```
patrickhousley, line 3: Nothing to do here but constants files like this that just re-export a specific constant from another import increase bundle size needlessly.

src/features/utils/aggregate-base.js
```diff
@@ -0,0 +1,16 @@
+import { registerHandler } from '../../common/event-emitter/register-handler'
+import { FeatureBase } from './feature-base'
+
+export class AggregateBase extends FeatureBase {
```
patrickhousley, line 4: This class was originally removed because it didn't actually do anything. Nice to see it coming back with functionality. Are all the feature aggregate classes updated to extend this?

src/features/utils/instrument-base.js
```diff
@@ -42,16 +44,26 @@ export class InstrumentBase extends FeatureBase {
   importAggregator () {
     if (this.hasAggregator || !this.auto) return
     this.hasAggregator = true
-
+    let session, agentSessionFile
+    if (getConfigurationValue(this.agentIdentifier, 'privacy.cookies_enabled') === true) {
```
patrickhousley, line 48: We really should create a new setting for this instead of overloading the meaning of `cookies_enabled`.

src/features/utils/instrument-base.js
```diff
@@ -42,16 +44,26 @@ export class InstrumentBase extends FeatureBase {
...
+      agentSessionFile = import(/* webpackChunkName: "session-manager" */ './agent-session')
```
patrickhousley, line 49:
```suggestion
      agentSessionImport = import(/* webpackChunkName: "session-manager" */ './agent-session')
```

src/features/utils/instrument-base.js
```diff
@@ -42,16 +44,26 @@ export class InstrumentBase extends FeatureBase {
...
     const importLater = async () => {
       /**
        * Note this try-catch differs from the one in Agent.start() in that it's placed later in a page's lifecycle and
        * it's only responsible for aborting its one specific feature, rather than all.
        */
       try {
         // The session entity needs to be attached to the config internals before the aggregator chunk runs
-        const { setupAgentSession } = await import(/* webpackChunkName: "session-manager" */ './agent-session')
-        setupAgentSession(this.agentIdentifier)
+        if (agentSessionFile && !session) {
```
patrickhousley, line 58:
```suggestion
        if (agentSessionImport && !session) {
```

src/features/utils/instrument-base.js
```diff
@@ -42,16 +44,26 @@ export class InstrumentBase extends FeatureBase {
...
+          const { setupAgentSession } = await agentSessionFile
+          session = setupAgentSession(this.agentIdentifier)
+        }
+        if (!shouldImportAgg(this.featureName, session)) {
+          drain(this.agentIdentifier, this.featureName)
+          return
+        }
+
         // import and instantiate the aggregator chunk
         const { lazyLoader } = await import(/* webpackChunkName: "lazy-loader" */ './lazy-loader')
```
patrickhousley, line 68: hahaha....we should probably rename that file to `lazy-feature-loader`

src/common/util/feature-flags.js
metal-messiah: I think these flags are going to change / maybe go away, in waiting mode there

src/loaders/api/api.js
```diff
@@ -154,6 +154,7 @@ export function setAPI (agentIdentifier, forceDrain) {
     if (typeof err === 'string') err = new Error(err)
     handle(SUPPORTABILITY_METRIC_CHANNEL, ['API/noticeError/called'], undefined, FEATURE_NAMES.metrics, instanceEE)
     handle('err', [err, now(), false, customAttributes], undefined, FEATURE_NAMES.jserrors, instanceEE)
+    handle('err', [err, now(), false, customAttributes], undefined, FEATURE_NAMES.sessionReplay, instanceEE)
```
patrickhousley, line 157: Just curious....why do we have to dispatch the error twice? Why cant the session replay just listing on the errors channel?

src/common/util/submit-data.js
metal-messiah: What exactly do you mean?

src/features/jserrors/instrument/index.js
metal-messiah: this is a good question.  In one way or another, we will need to decorate the error object that gets harvested with a flag indicating a session recording was occurring when the error was captured. This could be done in a contextual way like mentioned above, or by checking the session manager at harvest time in the error feature

src/features/session_replay/aggregate/index.js
metal-messiah: up to a certain point.  The most likely case of exceeding the limit is that the "initial snapshot", which is a single rrweb event, exceeds the limit.  Splitting that object further would be much more complex

src/features/jserrors/instrument/index.js
patrickhousley: We should look for a way to do this "decorating" across features without creating dependency between features. I'm looking at you spa feature!!!

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    this.events.push(event)
+    this.payloadBytesEstimation += eventBytes
+
+    // We are making an effort to try to keep payloads manageable for unloading.  If they reach the unload limit before their interval,
+    // it will send immediately.  This often happens on the first snapshot, which can be significantly larger than the other payloads.
+    if (payloadSize > IDEAL_PAYLOAD_SIZE) {
+      // if we've made it to the ideal size of ~64kb before the interval timer, we should send early.
+      this.scheduler.runHarvest()
```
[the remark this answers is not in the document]
metal-messiah: Definitely blurred.  Not sure theres a right or wrong way, but we should align on a standard way, and also consider doing these sorts of operations for each harvest.  This currently only happens in AJAX (and now SR)

src/features/session_replay/constants.js
metal-messiah: all the features do this, so we should add this as a cleanup item

src/features/utils/instrument-base.js
metal-messiah: yeah this setting is obscured rn.  In the UI, it is represented a "session tracking" but passes through the config as "cookies_enabled"

src/loaders/api/api.js
metal-messiah: The main (only?) reason would be that -- If jserrors feature drains before sessionReplay sets up its listeners, the data would be missed

src/common/timer/interaction-timer.js
metal-messiah: might need to be renamed, but it is an event that is needed in the session replay feature, to know that a session is "unpaused"

src/features/page_view_event/aggregate/index.js
metal-messiah: correct

src/loaders/api/api.js
patrickhousley: Hmm...could we track that all features have been loaded and instantiated before allowing the drain to commence? I think the best way to have cross-feature functionality is for featureA to listen for featureB events. Using the event emitter removes the direct dependency between the two features but it does complicate the event emitter a bit.

src/common/harvest/harvest.js
```diff
@@ -97,24 +101,28 @@ export class Harvest extends SharedContext {
 
     if (!opts) opts = {}
 
-    var url = this.getScheme() + '://' + info.errorBeacon + '/' + endpoint + '/1/' + info.licenseKey + this.baseQueryString()
-    if (payload.qs) url += encodeObj(payload.qs, agentRuntime.maxBytes)
+    var url = customUrl || this.getScheme() + '://' + info.errorBeacon + '/' + endpoint + '/1/' + info.licenseKey
 
+    var baseParams = includeBaseParams ? this.baseQueryString() : ''
```
cwli24, line 106: doesn't `includeBaseParams` get passed down from harvest send as `undefined` forcibly so the default never applies?

src/common/harvest/harvest.js
```diff
@@ -97,24 +101,28 @@ export class Harvest extends SharedContext {
...
     var body
-    var fullUrl = url
-    if (useBody && endpoint === 'events') {
-      body = payload.body.e
-    } else if (useBody) {
-      body = stringify(payload.body)
-    } else {
-      fullUrl = url + encodeObj(payload.body, agentRuntime.maxBytes)
-    }
+    var fullUrl = url + '?' + baseParams + params
```
cwli24, line 115: `'?'` not needed. `baseQueryString` will have `?` prepended already

src/common/harvest/harvest.js
```diff
@@ -97,24 +101,28 @@ export class Harvest extends SharedContext {
...
+
+    if (!gzip) {
```
cwli24, line 117: Compression shouldn't have anything to do with the content of the request body?

src/common/util/feature-flags.js
cwli24: should be just `sr`, adding `feat-srf` and `feat-sre` overly complicates

cwli24 (commented):
There'll likely be more, but here's some for consideration

src/common/util/feature-flags.js
metal-messiah: agreed

src/common/util/submit-data.js
metal-messiah: we should discuss, because i'm in the opposite camp.  I think named args, esp. when theres like 6-7 of them, is extremely messy and makes passing optional arguments a huge pain in comp. with object args.

For instance, the send* calls all can have many arguments but only 1 is required.

src/features/jserrors/instrument/index.js
metal-messiah: agreed, changed to use the errorAgg call to match the others

src/loaders/api/api.js
metal-messiah: yeah, i removed this code but this is a good call out.  I think we actually **do** already wait for all the features to be loaded before draining anyway, so this was likely redundant.

src/common/config/state/configurable.js
```diff
@@ -1,32 +1,25 @@
...
-      if (frozenAttrs.length && value && typeof value === 'object') {
-        frozenAttrs.forEach(attr => {
-          if (attr in value) {
-            warn(`"${attr}" is a protected attribute and can not be changed in feature ${key}.  It will have no effect.`)
-            delete value[attr]
-          }
-        })
+    // allow getters and setters to pass from model to target
+    const output = Object.create(
+      Object.getPrototypeOf(model),
```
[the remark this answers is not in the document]
metal-messiah: was needed for the getters/setters mentioned above, but if we do go back away from that will not be needed

src/common/harvest/harvest-scheduler.js
```diff
@@ -68,22 +68,44 @@ export class HarvestScheduler extends SharedContext {
...
-      var retry = submitMethod.method === submitData.xhr
+      retry = submitMethod.method === submitData.xhr
       var payload = this.opts.getPayload({ retry: retry })
+
       if (payload) {
         payload = Object.prototype.toString.call(payload) === '[object Array]' ? payload : [payload]
-        for (var i = 0; i < payload.length; i++) {
-          this.harvest.send(this.endpoint, payload[i], opts, submitMethod, onHarvestFinished)
-        }
+        harvests.push(...payload)
```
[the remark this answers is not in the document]
metal-messiah: This was already handled this way ==> payload was always coerced into an array, as the AJAX feature already puts into an array

src/common/util/submit-data.js
patrickhousley: StandardJS doesn't have any opinion on this that I can find. I prefer using an options object because it makes adding parameters easier and it's easier to tell what the parameters are for when reading a method invocation without having to go lookup that method in the code. However, it may add overhead in memory usage, garbage collection, and larger bundles.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+/** The "mode" with which the session replay is recording */
+const MODE = {
+  OFF: 0,
+  FULL: 1,
+  ERROR: 2
+}
+/** Vortex caps payload sizes at 1MB */
+const MAX_PAYLOAD_SIZE = 1000000
+/** Unloading caps around 64kb */
+const IDEAL_PAYLOAD_SIZE = 64000
```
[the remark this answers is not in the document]
cwli24: In other words, let SR do XHR at 1MB cap on interval & disregard unload. Tell harvest scheduler **not** to add `subscribeToEOL` for this feature via a param in `opts`.

src/common/util/submit-data.js
```diff
@@ -57,7 +57,13 @@ submitData.xhr = function xhr (url, body, sync, method = 'POST') {
     // do nothing
   }
 
-  request.setRequestHeader('content-type', 'text/plain')
+  if (gzipped) {
+    request.setRequestHeader('content-type', 'application/json')
```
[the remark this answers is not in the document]
metal-messiah: replaced with an argument for `headers` to be passed in

src/common/harvest/harvest-scheduler.js
```diff
@@ -68,22 +68,44 @@ export class HarvestScheduler extends SharedContext {
...
       }
-    } else {
-      this.harvest.sendX(this.endpoint, opts, onHarvestFinished)
     }
 
+    let send = args => this.harvest.sendX(args)
```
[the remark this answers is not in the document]
metal-messiah: for scoping this. could also use bind, but its the same thing.

src/common/harvest/harvest-scheduler.js
```diff
@@ -68,22 +68,44 @@ export class HarvestScheduler extends SharedContext {
...
+        endpoint: this.endpoint,
+        payload,
+        opts,
+        submitMethod,
+        cbFinished: onHarvestFinished,
+        includeBaseParams: this.opts.includeBaseParams,
+        customUrl: this.opts.customUrl,
+        gzip: this.opts.gzip
+      })
+    })
```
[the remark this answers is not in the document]
metal-messiah: disagree.  If some (most) of those 9 params becomes optional, you end up with

send('endpoint', undefined, undefined, undefined, true, undefined, false, etc..)

Which imo is much more difficult to both extend and read.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+      gzip: true
+    }, this)
+
+    // Wait for an error to be reported.  This currently is wrapped around the "Error" feature.  This is a feature-feature dependency.
+    // This was to ensure that all errors, including those on the page before load and those handled with "noticeError" are accounted for. Needs evalulation
+    registerHandler('err', (e) => {
+      this.hasError = true
+      // run once
+      if (this.errorNoticed) return
+      // if the error was noticed AFTER the recorder was imported....
```
cwli24, line 90: If the error was noticed _before_ `initializeRecording` imports rrweb's record, the mode should still switch to FULL. One such case can happen due to race condition since that function is async.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    /** The estimated size of the payload after compression */
+    const payloadSize = this.getPayloadSize(eventBytes)
+    // Vortex will block payloads at a certain size, we might as well not send.
+    if (payloadSize > MAX_PAYLOAD_SIZE) {
+      return this.abort()
+    }
+    // Checkout events are flags by the recording lib that indicate a fullsnapshot was taken every n ms. These are important
+    // to help reconstruct the replay later and must be included.  While waiting and buffering for errors to come through,
+    // each time we see a new checkout, we can drop the old data.
+    if (this.mode === MODE.ERROR && !this.errorNoticed && isCheckout) {
```
cwli24, line 250: Again, since the mode should always be FULL as soon as the first error occurs, `errorNoticed` is redundant here. That is, on `isCheckout`, you just have to check if the mode is still ERROR--if it is, then there couldn't have been an error noticed yet.

cwli24: Given it's redundant in both of the places it's used, it can actually be removed.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    // Wait for an error to be reported.  This currently is wrapped around the "Error" feature.  This is a feature-feature dependency.
+    // This was to ensure that all errors, including those on the page before load and those handled with "noticeError" are accounted for. Needs evalulation
+    registerHandler('err', (e) => {
+      this.hasError = true
+      // run once
+      if (this.errorNoticed) return
+      // if the error was noticed AFTER the recorder was imported....
+      if (this.mode === MODE.ERROR && !!recorder) {
+        this.errorNoticed = true
+        this.scheduler.runHarvest({ needResponse: true })
```
cwli24, line 93: You have needResponse here, but I don't see where the response is used.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+        this.stopRecording()
+        this.mode = MODE.FULL
+        this.startRecording()
+        this.scheduler.startTimer(this.harvestTimeSeconds)
```
cwli24, line 97: You need to `stopTimer` of the existing before this too, or I think this new startTimer won't take effect.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+
+        const { session } = getRuntime(this.agentIdentifier)
+        session.write({ ...session.state, sessionReplay: this.mode })
```
cwli24, line 100: `session.write` happens 3 times in this file; is it necessary if the session itself (should) handles writing `agentRuntime` to storage on user nav away?

The point of the intermediary local mem is to be recipient of all the modification and then flushing it once to long term storage at EoL.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    this.stopRecording = recorder({
+      emit: this.store.bind(this),
+      blockClass,
+      ignoreClass,
+      maskTextClass,
+      blockSelector,
+      maskInputOptions,
+      maskTextSelector,
+      maskAllInputs,
+      ...(this.mode === MODE.ERROR && { checkoutEveryNms: 30000 })
```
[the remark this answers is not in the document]
cwli24: Also, please extract 30000 to a constant at top of file.

src/common/harvest/harvest.js
```diff
@@ -41,13 +41,15 @@ export class Harvest extends SharedContext {
...
-  sendX (endpoint, opts, cbFinished) {
+  sendX (spec) {
+    const { endpoint, opts } = spec
     var submitMethod = getSubmitMethod(endpoint, opts)
     if (!submitMethod) return false
     var options = {
       retry: submitMethod.method === submitData.xhr
     }
-    return this.obfuscator.shouldObfuscate() ? this.obfuscateAndSend(endpoint, this.createPayload(endpoint, options), opts, submitMethod, cbFinished) : this._send(endpoint, this.createPayload(endpoint, options), opts, submitMethod, cbFinished)
+    var caller = this.obfuscator.shouldObfuscate() ? (...args) => this.obfuscateAndSend(...args) : (...args) => this._send(...args)
```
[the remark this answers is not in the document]
metal-messiah: would need bind with your implementation.  

basically `=>` works in much the same way as manually forcing `this` via bind, in that it doesn't inherently have any `this` in the function body, instead inheriting it from above

src/common/harvest/harvest.js
metal-messiah: includeBaseParams uses the default (true) for every call except session replay at the moment

src/common/harvest/harvest.js
metal-messiah: felt it was better to not hide it in that `baseParams` call which may or may not happen now

src/common/harvest/harvest.js
metal-messiah: outdated, will be changed

cwli24 (commented):
More comments; can't link pending one until I publish the batch 💢

src/common/harvest/harvest.js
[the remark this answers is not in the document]
cwli24: Functionally, yes. But if you entertain the polyfill build, .bind requires no transpilation, whereas the arrow function does, as well as the spread operator.

src/common/timer/interaction-timer.js
```diff
@@ -42,7 +43,10 @@ export class InteractionTimer extends Timer {
       subscribeToVisibilityChange((state) => {
         if (state === 'hidden') this.pause()
         // vis change --> visible is treated like a new interaction with the page
-        else this.refresh()
+        else {
+          this.onResume?.()
+          this.refresh()
```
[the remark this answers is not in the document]
metal-messiah: `onResume()` is the callback passed in from outer scope.  `timer.refresh()` should not always trigger `onResume()`.  Right now, only a vis change (visible) should trigger `onResume` ( but things like click, scroll, etc can also trigger timer.refresh() )

test.txt
cwli24, line 1: Dummy file

tests/assets/instrumented.html
cwli24, line 1: Accidental modification committed

src/features/session_replay/aggregate/index.js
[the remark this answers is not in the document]
metal-messiah: 100% disagree.  In order for a session replay to be successful, it absolutely needs to be able to capture up until a page unloads, in order to be able to be a smooth replay from one page to the next.  It's quite critical that *all* payloads from SR get harvested.  None of the other features are nearly as dependent on that fact as SR is, as all other features are just aggregated samples.

The one exception is now STN, which should probably mimic this behavior in some way (like you said, oversight)

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    /** A value which increments with every new mutation node reported. Resets after a harvest is sent */
+    this.payloadBytesEstimation = 0
+
+    /** The method to stop recording. This defaults to a noop, but is overwritten once the recording library is imported and initialized */
+    this.stopRecording = () => { /* no-op until set by rrweb initializer */ }
+
+    // The SessionEntity class can emit a message indicating the session was cleared and reset (expiry, inactivity). This feature must abort and never resume if that occurs.
+    this.ee.on('session-reset', () => {
+      this.abort()
+    })
```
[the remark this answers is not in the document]
metal-messiah: fresh recordings should only be determined at page load time for new sessions the time being

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    if (session.state.sessionReplay === MODE.FULL || (session.isNew && fullSample)) this.mode = MODE.FULL
+    else if (session.state.sessionReplay === MODE.ERROR || errorSample) this.mode = MODE.ERROR
+
+    console.log('MODE', this.mode, '-- entitlements, error, full', entitlements, errorSample, fullSample)
+
+    // FULL mode records AND reports from the beginning, while ERROR mode only records (but does not report).
+    // ERROR mode will do this until an error is thrown, and then switch into FULL mode.
+    // If an error happened before we've gotten to this stage, we can simply just run in FULL mode.
+    if (this.mode === MODE.FULL || (this.mode === MODE.ERROR && this.errorNoticed)) {
+      this.mode = MODE.FULL
```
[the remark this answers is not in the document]
metal-messiah: you have to also consider the gap between an error throwing on the same page before the load event and the agg file instantiating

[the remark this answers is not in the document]
metal-messiah: as currently written, the handler will only set to full mode if the recorder lib has been imported, could change

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+      // We only report (harvest) in FULL mode
+      this.scheduler.startTimer(this.harvestTimeSeconds)
+      this.recording = true
+    }
+    // We record in FULL or ERROR mode
+    if (this.mode !== MODE.OFF) {
```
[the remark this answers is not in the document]
metal-messiah: its likely redundant in some fashion

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    console.log('MODE', this.mode, '-- entitlements, error, full', entitlements, errorSample, fullSample)
+
+    // FULL mode records AND reports from the beginning, while ERROR mode only records (but does not report).
+    // ERROR mode will do this until an error is thrown, and then switch into FULL mode.
+    // If an error happened before we've gotten to this stage, we can simply just run in FULL mode.
+    if (this.mode === MODE.FULL || (this.mode === MODE.ERROR && this.errorNoticed)) {
+      this.mode = MODE.FULL
+      // We only report (harvest) in FULL mode
+      this.scheduler.startTimer(this.harvestTimeSeconds)
+      this.recording = true
```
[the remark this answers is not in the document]
metal-messiah: good catch

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+    // We record in FULL or ERROR mode
+    if (this.mode !== MODE.OFF) {
+      const { record } = await import(/* webpackChunkName: "recorder" */'rrweb')
+      recorder = record
+      this.startRecording()
+      const { gzipSync, strToU8 } = await import(/* webpackChunkName: "compressor" */'fflate')
+      gzipper = gzipSync
+      u8 = strToU8
```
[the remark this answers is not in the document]
metal-messiah: does this work to set top level vars?

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+      // We only report (harvest) in FULL mode
+      this.scheduler.startTimer(this.harvestTimeSeconds)
+      this.recording = true
+    }
+    // We record in FULL or ERROR mode
+    if (this.mode !== MODE.OFF) {
+      const { record } = await import(/* webpackChunkName: "recorder" */'rrweb')
+      recorder = record
+      this.startRecording()
+      const { gzipSync, strToU8 } = await import(/* webpackChunkName: "compressor" */'fflate')
```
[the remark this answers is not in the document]
metal-messiah: probably not anything official, but will be needed by GA time.  long story short is: rrweb snapshots likely exceed 1MB ingest limit for most sites with no easy course of action other than compression

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+
+  /** Abort the feature, once aborted it will not resume */
+  abort () {
+    this.blocked = true
+    this.stopRecording()
+    const { session } = getRuntime(this.agentIdentifier)
+    session.write({ ...session.state, sessionReplayActive: false })
+  }
+
+  /** Extensive research has yielded about an 88% compression factor on these payloads.
```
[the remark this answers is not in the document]
metal-messiah: added link to nr1 dashboard

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,292 @@
...
+  constructor (agentIdentifier, aggregator) {
+    super(agentIdentifier, aggregator, FEATURE_NAME)
+
+    /** Each page mutation or event will be stored (raw) in this array. This array will be cleared on each harvest */
+    this.events = []
+    /** The interval to harvest at.  This gets overridden if the size of the payload exceeds certain thresholds */
+    this.harvestTimeSeconds = getConfigurationValue(this.agentIdentifier, 'session_replay.harvestTimeSeconds') || 60
+    /** Set once the recorder has fully initialized after flag checks and sampling */
+    this.initialized = false
+    /** Set once an error has been detected on the page.  Is cleared after a harvest */
```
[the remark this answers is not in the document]
metal-messiah: ty

src/features/session_replay/aggregate/index.js
metal-messiah: cleaned up and removed, ty

src/features/session_replay/aggregate/index.js
metal-messiah: good catch, copy/pasta issue

src/features/session_replay/aggregate/index.js
metal-messiah: there's no timer at all if this code is executed, this is starting the first timer here

src/features/session_replay/aggregate/index.js
metal-messiah: good call, makes sense.  will just update state of local module, which will sync its state to local storage when the time comes (ie vis change)

src/features/utils/aggregate-base.js
```diff
@@ -0,0 +1,16 @@
...
+  waitForFlags (flagNames = []) {
+    return Promise.all(flagNames.map(fName => new Promise((resolve) => {
+      registerHandler(`feat-${fName}`, () => {
+        resolve({ name: fName, value: true })
+      }, this.featureName, this.ee)
+      registerHandler(`block-${fName}`, () => {
+        resolve({ name: fName, value: false })
```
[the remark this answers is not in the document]
metal-messiah: A lot we could do here, needs some alignment so its generic enough for all the features.

tests/assets/instrumented.html
metal-messiah: ty

src/features/session_replay/aggregate/index.js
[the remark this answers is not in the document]
metal-messiah: changed to get rid of `errorNoticed`

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,304 @@
...
+    // we are not actively recording SR... DO NOT import or run the recording library
+    // session replay samples can only be decided on the first load of a session
+    // session replays can continue if already in progress
+    if (!session.isNew && session.state.sessionReplay === MODE.OFF) return
+
+    if (session.state.sessionReplay === MODE.FULL) this.mode = MODE.FULL
+    else if (session.isNew && fullSample) this.mode = MODE.FULL
+    else if (session.state.sessionReplay === MODE.ERROR) this.mode = MODE.ERROR
+    else if (errorSample) this.mode = MODE.ERROR
+    else return
```
cwli24, line 146: This is still really confusing. How about this?
```suggestion
    if (!session.isNew) { // inherit the mode of the existing session
        this.mode = session.state.sessionReplay
    } else { // determine the mode the new session should start in
        if (fullSample) this.mode = MODE.FULL // full mode has precedence over error mode
        else if (errorSample) this.mode = MODE.ERROR
        // If neither are selected, then the mode stays off.
    }
    if (this.mode === MODE.FULL) return // finally, if the mode is off, don't start rrweb
```

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,304 @@
...
+        this.mode = MODE.full
+        if (recorder) {
+          this.scheduler.runHarvest()
+          this.stopRecording()
+          this.mode = MODE.FULL
+          this.startRecording()
+          this.scheduler.startTimer(this.harvestTimeSeconds)
+
+          const { session } = getRuntime(this.agentIdentifier)
+          session.state.sessionReplay = this.mode
```
cwli24, line 105: L96 already set the mode to full in this case, so L100 can be removed.
L104-105 should be moved up & out of this inner block: if L96 has updated the local module mode var, it should also be saved to the agent's runtime.

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,304 @@
...
+
+    recorder = (await import(/* webpackChunkName: "recorder" */'rrweb')).record
+    this.startRecording()
+    const { gzipSync, strToU8 } = await import(/* webpackChunkName: "compressor" */'fflate')
+    gzipper = gzipSync
+    u8 = strToU8
+
+    this.isFirstChunk = !!session.isNew
+
+    session.state.sessionReplay = this.mode
```
cwli24, line 166: This should be moved earlier; see https://github.com/newrelic/newrelic-browser-agent/pull/532/files#r1199100432

src/features/session_replay/aggregate/index.js
metal-messiah: If rrweb was blocked in some way, we wont want future pages to inherit the state. could use alternative approach somewhere, but thats my main concern

src/features/session_replay/aggregate/index.js
metal-messiah: again, we just need to consider that we probably dont want to do this until we know rrweb made it to the page

metal-messiah:
Passing test with new changes --> https://github.com/newrelic/newrelic-browser-agent/actions/runs/5040202999

bjfield (approved):

src/features/session_replay/aggregate/index.js
```diff
@@ -0,0 +1,303 @@
+// import { record } from 'rrweb'
```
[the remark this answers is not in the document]
metal-messiah: Thanks, I forgot about the disclaimer tags

patrickhousley (approved):
Verified the session replay feature does not load in the agent.

PR #534, opened 2023-05-17
feat: Include bundled agent loaders in npm package

patrickhousley:
The npm package will now include pre-bundled agent loaders, which may be included directly in HTML files without needing to import and configure the agent in JavaScript source code.
---

### Overview

Added the `build` directory to the npm package and update the readme to include instructions for using the bundled loaders.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-8217

### Testing

No testing added. Will be highly dependent on the customer's build toolchain.


README.md
```diff
@@ -170,6 +170,51 @@ import { Agent } from '@newrelic/browser-agent/loaders/agent'
...
+
+<!-- Include the lite agent -->
+<script src="/node_modules/@newrelic/browser-agent/build/nr-loader-spa-1.232.1.min.js"></script>
+<!-- Include the pro agent -->
+<script src="/node_modules/@newrelic/browser-agent/build/nr-loader-full-1.232.1.min.js"></script>
+<!-- Include the pro+spa agent -->
+<script src="/node_modules/@newrelic/browser-agent/build/nr-loader-rum-1.232.1.min.js"></script>
+```
+
+The version number in the file name will change depending on the version of the npm package that is installed. It is recommended that you setup your build tooling to copy the correct file from the `/node_modules/@newrelic/browser-agent/build/` directory into a `static` directory where you normally host static files. You can rename the file at the same time to make the `script` element static but remember to manage cache busting for your users.
```
metal-messiah, line 194: Should we just build the agent without the version suffix before shipping NPM so that this is not an issue customers would need to deal with?

bjfield: They could use e.g. `nr-loader-spa-1.x.x.min.js` and that might last them until the next breaking change.

bjfield (approved):
The build directory appears to be created properly in the pack. Not sure if we need the webpack stats in the build dir but I don't think it's bad to have them there.

README.md
metal-messiah: This actually might be a good case for generating x.x.x which was removed from the script

tools/testing-server/plugins/agent-injector/config-transform.js
```diff
@@ -30,7 +30,7 @@ function getConfigContent (request, reply, testServer) {
     }
   })()
   const config = {
-    agent: `${testServer.assetServer.host}:${testServer.assetServer.port}/build/nr.js`,
+    agent: `${testServer.assetServer.host}:${testServer.assetServer.port}/cdn/build/nr.js`,
```
cwli24, line 33: Is this correct? No `/dist`?

README.md
```diff
@@ -170,6 +170,51 @@ import { Agent } from '@newrelic/browser-agent/loaders/agent'
...
+<script src="/node_modules/@newrelic/browser-agent/dist/cdn/nr-loader-spa-1.x.x.min.js"></script>
+<!-- Include the pro agent -->
+<script src="/node_modules/@newrelic/browser-agent/dist/cdn/nr-loader-full-1.x.x.min.js"></script>
+<!-- Include the pro+spa agent -->
+<script src="/node_modules/@newrelic/browser-agent/dist/cdn/nr-loader-rum-1.x.x.min.js"></script>
+```
+
+The version number in the file name will change depending on the version of the npm package that is installed. It is recommended that you setup your build tooling to copy the correct file from the `/node_modules/@newrelic/browser-agent/dist/cdn/` directory into a `static` directory where you normally host static files. You can rename the file at the same time to make the `script` element static but remember to manage cache busting for your users.
+
+Also included in the bundle are un-minified and polyfilled versions of each of the agents. The polyfilled versions of the agents are built to ES5 standards. When using the polyfilled versions of the agents, remember to either include your own polyfills in the HTML above the loader `script` or use our `polyfills` script.
```
cwli24, line 196: Doing a cdn prod build, the `/dist/cdn` folder is 10.2 MB. Is this all going into our current npm pkg, which is only 1.7 MB? Any downside aside from devs having to download more bytes on their machines?
Is there a way to only include the loaders without the other chunks?

webpack.config.js
```diff
@@ -52,22 +52,22 @@ switch (PUBLISH) {
...
+    MAP_PATH = '\n//# sourceMappingURL=http://bam-test-1.nr-local.net:3333/dist/cdn/[url]'
     break
   default:
     // local build
     PATH_VERSION = ''
     SUBVERSION = 'LOCAL'
-    PUBLIC_PATH = '/build/'
-    MAP_PATH = '\n//# sourceMappingURL=http://bam-test-1.nr-local.net:3333/build/[url]'
+    PUBLIC_PATH = '/dist/cdn/'
+    MAP_PATH = '\n//# sourceMappingURL=http://bam-test-1.nr-local.net:3333/dist/cdn/[url]'
```
cwli24, line 70: Maybe related to the npm size thread, can local be kept in a local `/build` dir?

tools/testing-server/plugins/agent-injector/config-transform.js
patrickhousley: Probably should be `/dist/...` but i don't think the agent property is even used anymore is it.

README.md
patrickhousley: There's always a way but I don't really think its necessary or an issue.

tools/testing-server/plugins/agent-injector/config-transform.js
cwli24: Not in prod, no. Doesn't seem it appears in any tests either, so can prob remove.

patrickhousley:
Closing this PR. We have discussed internally whether we want to include the bundles in the npm package and have concluded, at least for now, that we will not be including them. If customers would like to use a `<script src=""` element to pull in the browser agent, our recommendation is to use our public CDN.

While we desire that customers using the Copy/Paste method of installing the browser agent copy the snippet from New Relic and embed the loader in their HTML, we understand that one size does not fit all. If you would like to use a script src tag, below is an example.

```
<script src="https://js-agent.newrelic.com/nr-loader-spa-1.239.1.min.js"></script>
```

All the loaders follow the same naming pattern `nr-loader-[agentType]-[version].min.js` where `[agentType]` can be `rum` for the lite agent, `full` for the pro agent, and `spa` for the spa agent. The `[version]` value will always match the version in our [published releases](https://github.com/newrelic/newrelic-browser-agent/releases). You may also leave off the `.min` part if you would like an unminified version of the agent loader.

Keep in mind that this is not an officially supported installation mechanism because it does require the browser to download an additional resource before the agent can wrap globals and start intercepting events for harvesting. However, we do our best to try and capture as much information as possible to provide a complete observability picture. We do not currently recommend using `async` or `defer` for the agent loader script. Either of these could exacerbate the issue of missing data and, if the agent is not started before page load, may result in the agent not collecting any data. These attributes could also change the order in which globals are wrapped leading to strange interactions with other third-party libraries.

PR #536, opened 2023-05-21
chore: adding wdio workflows

patrickhousley:
Adding wdio workflows and fixing tests to run in all supported browsers.
---

### Overview

- Added two new wdio workflows; one to run tests on all browsers and one to run tests for a specific browser
  - The all browsers workflow re-uses the single browser workflow
  - Both workflows are re-usable
- Added a new nightly workflow to centralize jobs we run every night
  - Added call to wdio single browser shared workflow to run tests on latest chrome
  - We can relocation nightly jobs later as we fix our workflows to be reusable
- Fixed all the wdio tests so the run and pass in all browsers.
  - Safari 15 just fails with the same network errors that was reported to Sauce Labs in December. I have reached out to support again but don't expect any new update. We may need to remove Safari 15 from our testing or look at switch service providers.
  - Centralized our default config used to create test urls
  - Moved agent session functions to shared custom wdio commands
  - Removed all use of `waitForFeature` and added a deprecation notice in the jsdoc.
  - General cleanup of tests to reduce the amount of code in each test where possible. For example, combining the `browser.url` and `testHandle.assetURL` calls. Expects and `browser.url` are combined into an overall `Promise.all` with destructing to eliminate unnecessary lines of code.
- Session manager tests:
  - Cleaned up the `describe` and `it` names to remove unnecessary characters
  - Add reliance on shared custom commands
  - Blocked multi-tab test for mobile devices that do not support opening new tabs or windows
- Stack trace tests:
  - Blocking tests from running on non-chromium browsers. Non-chromium browsers do not produce the expected stack trace. These tests could probably be moved to unit tests since the whole premise of the test is that the URL for sub-resources does not contain `<inline>`.
-  Moved the mootools tests to wdio and created a re-usable testing method that we can use with other third-party libraries to ensure loading of the library does not prevent the agent from reporting. We could easily add tests for jquery, etc.
- Browser matcher fixes:
  - Fixed the matcher rule class to properly convert the browser version to a number before doing comparison
  - Fixed the spec matcher so it did not aggregate the results of each rule. Instead, as long as one rule matches, then the spec matcher will return true. This ensures the exclusive nature of the spec matcher.
- Browsers lists fixes:
  - Extracted out the browserName and browserVersion methods to a utils file. These methods take a sauceBrowser definition and ensures the browserName and browserVersion is correctly parsed.
  - Updated browsers lists file to remove an unnecessary filter and to use the shared browserName and browserVersion methods when doing comparisons.
- Test server fixes:
  - Fixed the debug script that gets injected when the `-d` flag is passed so it is compatible with IE@11
  - Centralized the config that is used when creating the fastify instances (reduces code duplication)
  - Fixed expects that have a timeout so the timeout is not hit after the expect was already resolved. If an expect resolves, it should no longer result in an unexpected call to reject an already resolved promise
- WDIO fixes:
  - Removed all reference to JIL in variable names, comments, etc
  - Fixed error in sauce config to remove additional properties from the browser definition when running on mobile browsers
  - Added a new setting so job names in Sauce Labs start with `Browser Agent: ` to help use differentiate our tests from others in the company
  - Fixed the browser matcher so it uses the shared browserName and browserVersion methods to properly extract the name and version of the sauce browser definition
  - Renamed the event used when reporting test results to New Relic so we can tell the difference between tests ran in JIL and WDIO
  - Fixed an issue in the test connector where unhandled promise rejections could occur if an expect failed after a test had already failed. The scenario here is that a test sets up multiple expects, something causes the test to fail leading to the testHandle being destroyed, but the expects are still pending and timeout. When a testHandle is destroyed, if it has any pending expects, those are immediately cancelled.
  -  Changed the config serialization to use `serialize-anything`. The previous serialization relied on `JSON.stringify` which could not handle things like functions.
  - Added a new global `beforeEach` and `afterEach` that handles the creation and destruction of the `testHandle` and the destruction of the agent session between each test.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-7593

Before we resolve the above ticket, we may want to talk through how we want to trigger these tests.

### Testing

Pull the browser and run `npm run wdio`. All the tests should run and pass. You can use `-b` to try other browsers, combinations of browsers, etc. The only thing that probably will not pass is Safari 15.


.github/workflows/wdio-all-browsers.yml
```diff
@@ -0,0 +1,64 @@
...
+        required: true
+      JIL_NODE_NEW_RELIC_LICENSE_KEY:
+        required: true
+
+env:
+  BUILD_NUMBER: PR${{ github.event.number }}-job-${{ github.run_number }}-attempt-${{ github.run_attempt }}
+
+jobs:
+  chrome:
+    if: contains(fromJson('["workflow_dispatch"]'), github.event_name)
```
metal-messiah, line 19: None of the jobs will run on `workflow_call`

patrickhousley: Thanks for catching that

.github/workflows/wdio-single-browser.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: npm ci
+      - name: Build agent
+        run: npm run build:all
+      - name: Run WDIO Tests
+        run: |
+          node ./tools/wdio/bin/cli.js \
+            -s \
+            -b ${{ inputs.browser-target }} \
+            --concurrent 10 \
+            ${{ runner.debug && '-v -L -D -d' }} \
```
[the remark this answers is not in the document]
patrickhousley: It's a feature of github workflows. When you go to rerun a workflow, there is a checkbox you can select to enable extended debugging.

tests/assets/third-party-compatibility/jspdf/2.5.1.html
```diff
@@ -5,27 +5,28 @@
...
     {init}
     {loader}
     {config}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jspdf/2.5.1/jspdf.umd.js" integrity="sha512-Bw9Zj8x4giJb3OmlMiMaGbNrFr0ERD2f9jL3en5FmcTXLhkI+fKyXVeyGyxKMIl1RfgcCBDprJJt4JvlglEb3A==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/html2canvas/1.4.1/html2canvas.js" integrity="sha512-sn/GHTj+FCxK5wam7k9w4gPPm6zss4Zwl/X9wgrvGMFbnedR8lTUSLdsolDRBRzsX6N+YgG6OWyvn9qaFVXH9w==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
   </head>
   <body>
     <div id="content">
-      this is a generic page that is instrumented by the JS agent
+      This is a generic page that is instrumented by the JS agent and includes mootools to ensure compatibility.
```
metal-messiah, line 17: NBD but this page doesn't have mootools on it

tests/assets/third-party-compatibility/mootools/1.6.0-nocompat.html
```diff
@@ -0,0 +1,37 @@
...
+  PDX-License-Identifier: Apache-2.0
+-->
+<html>
+  <head>
+    <title>Mootools 1.6.0-nocompat</title>
+    {init}
+    {config}
+    {loader}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mootools/1.6.0/mootools-core.js" integrity="sha512-nxKp6INemMtUWHV+BToAOXm2ZV4+LCc/sw+4j2fyVJG088hHf7Rt/h8qCYT7bTsEtqkDCXo6sbtFlkro3Zo3HA==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mootools-more/1.6.0/mootools-more.js" integrity="sha512-76y+QT6I9ZvXkATH8rhr5xc+1ODnMdWJ1+EwITq0VxUzLMCKlWmHVISENgt4tHwr522eEAYH2bIhDky1X+JQPA==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
metal-messiah, line 13: Is `mootools-more` something we need to be cognizant of in addition to the base library?

tests/assets/third-party-compatibility/requirejs/2.3.6.html
```diff
@@ -0,0 +1,25 @@
...
+-->
+<html>
+  <head>
+    <title>RequireJS 2.3.6</title>
+    <!--  RequireJS is a JavaScript file and module loader.
+          The data-main attribute tells require.js to load JQuery after
+          RequireJS loads. For other typical usages, see the docs:
+          https://requirejs.org/docs/start.html  -->
+    <!-- <script data-main="js/vendor/jquery-3.2.1" src="js/vendor/require-2.3.6.js"></script> -->
+    <script data-main="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.0/jquery.js" src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.js" integrity="sha512-vRqhAr2wsn+/cSsyz80psBbCcqzz2GTuhGk3bq3dAyytz4J/8XwFqMjiAGFBj+WM95lHBJ9cDf87T3P8yMrY7A==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
metal-messiah, line 14: in the same vein as mootools, do we need to draw a line in the sand for requireJS versions?

tests/assets/third-party-compatibility/mootools/1.6.0-nocompat.html
patrickhousley: It contains functionality that was broken out of the core mootools. The recent issues we were investigating included the more library so I added it here.

tests/assets/third-party-compatibility/requirejs/2.3.6.html
patrickhousley: Not sure. This is the latest version of requirejs. I don't know how far back we want to, care to, can test.

tests/specs/api.e2e.js
```diff
@@ -1,17 +1,7 @@
-const { getTime } = require('../functional/uncat-internal-help.cjs')
-const { getErrorsFromResponse } = require('../functional/err/assertion-helpers')
+import { config } from '../util/defaults'
```
metal-messiah, line 1: Instead of necessitating this in every file, wouldn't it be more graceful to just change the default behavior to use this "default" config here --> https://github.com/newrelic/newrelic-browser-agent/pull/536/files#diff-51da539cea54b4b3bf642582b196cf964b82b2b219d472c1dc68d574da65013eR246-R254

Then the only time you pass a differing config in a test is to override the default.

Thoughts?

patrickhousley: Yup could do that

tests/specs/session-manager.e2e.js
```diff
@@ -9,247 +12,284 @@ describe('newrelic session ID', () => {
...
+      await Promise.all([
+        testHandle.expectRum(),
+        browser.url(await testHandle.assetURL('session-entity.html', config)) // Setup expects before loading the page
+      ])
 
-      const ls1 = await browser.execute(getLocalStorageData)
+      const { localStorage: ls1 } = await browser.getAgentSessionInfo()
       expect(ls1).toEqual(anySession())
 
-      await browser.execute(function () { window.open('/') })
```
metal-messiah, line 86: I think `window.open` already worked on all envs.  Are we limiting ourselves by doing this around a browser matcher arbitrarily?

patrickhousley: In my testing, `window.open` does not work in mobile browsers.

patrickhousley: So I did some more testing and here are my findings. While the original test appears to pass on ios, a new window/tab is not actually being opened. I have also tried creating a new link element and clicking it using `browser.execute` and `$().click()`. Nothing I can find can actually force ios safari to open a new tab or window.

metal-messiah (approved):

bjfield (approved):

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,20 @@
...
+    - cron: '0 4 * * *'
+  workflow_dispatch:
+
+env:
+  BUILD_NUMBER: Nightly-job-${{ github.run_number }}-attempt-${{ github.run_attempt }}
+
+jobs:
+  wdio-smoke:
+    name: 'WDIO: Smoke Test'
+    if: contains(fromJson('["schedule"]'), github.event_name) || contains(fromJson('["workflow_dispatch"]'), github.event_name)
```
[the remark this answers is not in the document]
patrickhousley: Probably not needed at all. I will remove. Just a remnant of copy/paste

PR #539, opened 2023-05-23
feat: Capture metrics for usage of MooTools and certain polyfills

patrickhousley:
New supportability metrics check for MooTools library usage and presence of customer-defined polyfills for certain native JavaScript features, such as function binding and ES6 Array prototype methods.
---

### Overview

- Moved the framework-detection script into the metrics aggregator folder since that is the only place it is used.
- Added mootools detection to the framework-detection script.
- Added a new script to detect polyfills for ES6+ globals and natives that we commonly rely upon within the agent.
  - The polyfills script will be overwritten in the ES5 webpack build so an empty array is always returned. This is because the usage of that build includes all the polyfills we are checking for. It seems redundant to report on the polyfills we supply with the ES5 build.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-119275

### Testing

Build the agent and start the local test server. The test server already has a few pages for the different frameworks to demonstrate the framework supportability metrics.

The easiest way to test the polyfill metrics is to disable the file swap for the ES5 build in `babel.config.js` and rebuild the agent. Launch the test server and use IE11 in SauceLabs to test the polyfill metrics.


bjfield (approved):
By testing I've verified that the agent now identifies all the specified polyfills, as well as MooTools (1.5.1 nocompat and 1.6.0 nocompat). We can't know how many pages use incompatible MooTools builds because the agent breaks before it can capture the SM, but this does give us some idea of the relative popularity of compatible MooTools. In conjunction with other metrics sources about usage of frameworks, we may be able to extrapolate what percentage of MooTools instrumentation we aren't seeing due to breakage of our agent. E.g. MooTools is A% as popular as JQuery but we see only B% as much of MooTools as JQuery. Similarly, the polyfill metrics tell us how many non-breaking polyfill instances we run into, but not the instances that break our agent. Just noting these known limitations.

patrickhousley:
Test [run](https://github.com/newrelic/newrelic-browser-agent/actions/runs/5074759250) passes except Safari 15.

PR #541, opened 2023-05-25
chore: Add route and compression handling to test-server

metal-messiah:
Add route and compression handling to test-server in preparation for SR.  SR will report compressed data payloads in a non-standard way, by including a query param which notes its compression.  GZIP handling as well as preprocessing were added to the testing server to facilitate this.
---

### Overview
Add route and compression handling to test-server in preparation for SR.  A positive side-effect of this is that the testing-server can now return compressed scripts and pages, which may decrease the total burden on sauce-connection issues.

### Related Issue(s)
NEWRELIC-8807

### Testing
The test server has a new route `/blob` which can handle compressed payloads and expects data shaped as defined in the `/blob` endpoint spec found here --> https://newrelic.atlassian.net/l/cp/CjDxzmo3


src/features/session_replay/aggregate/index.js
```diff
@@ -153,7 +149,7 @@ export class Aggregate extends AggregateBase {
       if (fullSample) this.mode = MODE.FULL // full mode has precedence over error mode
       else if (errorSample) this.mode = MODE.ERROR
       // If neither are selected, then don't record (early return)
-      return
+      else return
```
patrickhousley, line 152: Same difference

tools/testing-server/routes/mock-apis.js
```diff
@@ -110,7 +110,9 @@ module.exports = fp(async function (fastify, testServer) {
   fastify.get('/js', (request, reply) => {
     reply.type('text/javascript').send('console.log(\'hi\')')
   })
-  fastify.get('/text', (request, reply) => {
+  fastify.get('/text', {
+    compress: false
+  }, (request, reply) => {
```
patrickhousley, line 115: Should we just add this to all the mock api routes?

src/features/session_replay/aggregate/index.js
metal-messiah: Not exactly. The else is Basically saying if it's "off" then return (ie don't run the rest of the function)

If we always return it'll never run

metal-messiah: Not exactly. The else is Basically saying if it's "off" then return (ie don't run the rest of the function)

If we always return it'll never run

tools/testing-server/routes/mock-apis.js
metal-messiah: Probably tbh

src/features/session_replay/aggregate/index.js
patrickhousley: Ah lol, I didn't see the rest of the code.

patrickhousley (approved):

bjfield (approved):
Using `rrweb-record.html?loader=experimental` and some console logging, I verified that the agent is sending gzipped content and the blob test endpoint is unzipping the body.

PR #543, opened 2023-05-25
chore: Check configuration when importing Aggregators

metal-messiah:
Check and reset configuration when importing Aggregators.
---

### Overview
The configuration check lived in `FeatureBase`, which meant it was called before and after the load event.  This should really only need to be called after the load event directly, so it has been moved to the `AggregateBase` class.

### Related

https://github.com/newrelic/newrelic-browser-agent/pull/546 addresses a "configurable" warning due to an unused userAgent wrapper on the runtime, which led to the configuration check cleanup in this PR as well.

### Testing
Tests should continue to pass. Though a minor change, this PR does impact agent code.


patrickhousley (approved):
Tests for `browser/xhr/fetch.browser.js` pass in my local.

bjfield:
Moved userAgent portion out into https://github.com/newrelic/newrelic-browser-agent/pull/546.

PR #544, opened 2023-05-25
fix: Refactor usage of Array.from to address MooTools conflict

bjfield:
To work around a conflict with some versions of the MooTools library that mutate the native Array.from method, the agent will now use alternative methods.
---

### Overview

In the 1.232.1 release, we removed lodash.slice, which was being used to convert iterables to arrays—using Array.from instead. It was discovered that some older versions of the MooTools library mutate the Array.from method in a way that caused pages to break when both libraries were used in combination. This PR:

- Replaces instances of `Array.from` with spread operators (which will be transpiled by babel as needed).
- ~~Removes the `core-js/stable/array/from` polyfill (which was included only with polyfill builds of the agent).~~
  - Babel transpiles spreads to Array.from, requiring the polyfill.

### Related Issue(s)

[NR-119295](https://issues.newrelic.com/browse/NR-119295)

### Testing

- Standard tests should pass.
- Various builds of MooTools should be tested manually.


patrickhousley (approved):

bjfield:
Manual verification:
- No console errors appear for these `nocompat` builds of MooTools: `1.3.0`, `1.3.1`, `1.3.2`, `1.4.1`, `1.4.2`, `1.4.3`, `1.4.4`, `1.4.5`, `1.5.0`, `1.5.1`, `1.5.2`, `1.6.0`.
- The reported error (`Uncaught (in promise) TypeError: Failed to execute 'removeEventListener' on 'EventTarget': 2 arguments required, but only 0 present.`) does not appear for any builds, although an `Uncaught TypeError: t is not a function` error does appear for builds with the compatibility layer. This other error does not break the web page, is not addressed by this issue, and is expected for all MooTools `compat` builds.

src/common/drain/drain.js
```diff
@@ -55,7 +55,7 @@ export function drain (agentIdentifier = '', featureName = 'feature') {
 
   // Only when the event-groups for all features are ready to drain (staged) do we execute the drain. This has the effect
   // that the last feature to call drain triggers drain for all features.
-  const items = Array.from(registry[agentIdentifier])
+  const items = Array.prototype.slice.call(registry[agentIdentifier])
```
metal-messiah, line 58: This is not going to work to convert a Set to an Array

patrickhousley: Ah-ha it is a set. I was wondering about that. Yea this will break stuffs.

babel.config.js
```diff
@@ -44,6 +44,9 @@ module.exports = function (api) {
       ]
     },
     'webpack-ie11': {
+      assumptions: {
+        iterableIsArray: false
```
patrickhousley, line 48: TIL

PR #545, opened 2023-05-26
feat: Bringing session to Session Trace NEWRELIC-8662

cwli24:
Session trace becomes more align with the new interpretation of an user's session. When the session tracking setting is enabled, this feature will now use stored session info to make capturing decisions. The new behavior also serves to enable coupling it with the upcoming and experimental Session Replay feature.
---
### Overview

Session concept is incorporated into the trace feature such that when session tracking is enabled, it adopts and propagate the same operational mode across pages within a session. The former behavior (standalone) is maintained when it's disabled, and it's important to note that Trace does not necessarily require the existence of session to function. Even with session tracking is enabled, if for whatever reason Replay isn't, backward compatibility is maintained as much as possible to ensure that existing case today (prior to mainstream Replay release) doesn't break.

### Related Issue(s)

NEWRELIC-8662

### Testing

Workflows to verify:

- With session tracking **disabled**,
  - RUM response flag of `stn: 0` still blocks _all_ harvests.
  - Flag of `stn: 1` still sends all harvests.
  - Feature duration is capped at 10 minutes and all post-initial harvests have 30+ nodes.
- With session tracking **enabled**,
  - Every new page runs harvest with the same `session` ID but different `ptid`.
  - There's no duration limit, and post-initial harvests can occur with < 30 nodes while active on page.
  - When a session resets, there's a non-interval outgoing payload with whatever's recorded since last interval.
    - The following interval payload (on same page) has a new `session` ID but same `ptid`.
  - On brand new sessions,
    - The session object's `sessionTraceMode` gets the right MODE value corresponding to the related truth table in ticket/JIRA Feature's CDD.
    - In all cases where it should "collect_trace", all harvests are sent regardless of `stn` flag response.
    - In the single case where it and SR are off, no harvests are sent (`stn` is assumed 0).
    - As of now, the `sessionTraceMode` should _always_ match the `sessionReplay` mode of runtime state.
  - On existing sessions aka when a new page is open after the previous step,
    - `sessionTraceMode` of the new page matches that of the previous page.
    - The mode is respected; as of now, either the full lifecycle of events should be captured & harvested if FULL or ERROR, or nothing should be harvested if OFF.


src/common/aggregate/aggregator.js
```diff
@@ -10,6 +10,10 @@ export class Aggregator extends SharedContext {
   constructor (parent) {
     super(parent)
     this.aggregatedData = {}
+
+    /* Unfortunately, there's no other good medium through which SR can signal it's ready to other features that depends on it.
```
metal-messiah, line 14: The event emitter?

cwli24: It switches to unbuffered when everything drains, and there's no guarantee that `registerHandler` from another feature agg on an event emitted by SR agg will win the race, I think.

src/common/harvest/harvest-scheduler.js
```diff
@@ -26,10 +27,16 @@ export class HarvestScheduler extends SharedContext {
     // unload if EOL mechanism fires
     subscribeToEOL(this.unload.bind(this), getConfigurationValue(this.sharedContext.agentIdentifier, 'allow_bfcache')) // TO DO: remove feature flag after rls stable
 
-    // unload if session resets
-    this.sharedContext?.ee.on('session-reset', this.unload.bind(this))
+    /* Flush all buffered data if session resets and give up retries. This should be synchronous to ensure that the correct `session` value is sent.
+      Since session-reset generates a new session ID and the ID is grabbed at send-time, any delays or retries would cause the payload to be sent under
+      the wrong session ID. */
```
metal-messiah, line 32: good point

src/features/session_trace/aggregate/index.js
```diff
@@ -45,54 +50,57 @@ export class Aggregate extends FeatureBase {
...
+      this.isStandalone = true
+      registerHandler('rumresp-stn', (on) => {
+        if (on === true) this.startTracing(handlerCache)
+        else handlerCache.decide(false)
+      }, this.featureName, this.ee)
+    } else {
+      const doStuffByMode = (sessionEntity) => { // take care that the state obj reference is always subject to change but session mgr keeps one instance
+        switch (sessionEntity.state.sessionTraceMode) {
+          case MODE.ERROR:
+            registerHandler('errorAgg', () => sessionEntity.state.sessionTraceMode = MODE.FULL, this.featureName, this.ee) // switch to full capture when any error is encountered
```
metal-messiah, line 69: Make sure to trust the errorAgg that was set up before draining or you might miss an error

metal-messiah, line 69: sessionEntity state should be a reflection of the aggregator state.... they should both be set to "FULL" at the same time

```javascript
sessionEntity.state.sessionTraceMode = MODE.FULL;
this.mode = MODE.FULL;
```

src/features/session_trace/aggregate/index.js
```diff
@@ -45,54 +50,57 @@ export class Aggregate extends FeatureBase {
...
+            subscribeToVisibilityChange(visState => { // on user return, check if mode transitioned error -> full since pause (e.g., on another page), and if so, also switch here
```
metal-messiah, line 70: Maybe we should use the storage API to handle this so we don't miss data from waiting for vis change

metal-messiah: Leave as vis change for now

src/features/session_trace/aggregate/index.js
```diff
@@ -106,6 +114,58 @@ export class Aggregate extends FeatureBase {
...
+        if (now() > MAX_TRACE_DURATION) { // been collecting for over the longest duration we should run for, empty trace object so ST has nothing to send
+          scheduler.stopTimer()
+          this.trace = {}
+          return
+        }
+        // Only harvest when more than some threshold of nodes are pending, after the very first harvest.
+        if (this.ptid && this.nodeCount <= REQ_THRESHOLD_TO_SEND) return
+      } else {
+        // With sessions on harvest intervals, visible pages will send payload regardless of pending nodes but backgrounded pages will still abide by threshold.
+        if (this.ptid && document.visibilityState === 'hidden' && this.nodeCount <= REQ_THRESHOLD_TO_SEND) return
```
metal-messiah, line 162: Lets not limit the background interval by nodeCount for now, to ensure we always send all data when a replay is in progress.

This might change later

cwli24:
Trace needs to **end** (new function) when session-reset is emitted at this time.

src/features/session_trace/aggregate/index.js
```diff
@@ -27,15 +29,17 @@ const toAggregate = {
...
 }
-const MAX_TRACE_DURATION = 15 * 60 * 1000 // 15 minutes
+const MAX_TRACE_DURATION = 10 * 60 * 1000 // 10 minutes
+const REQ_THRESHOLD_TO_SEND = 30
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
   constructor (agentIdentifier, aggregator, argsObj) {
     super(agentIdentifier, aggregator, FEATURE_NAME)
+    this.agentRuntime = getRuntime(agentIdentifier)
```
patrickhousley, line 39: This should probably be moved to the `FeatureBase`. Better yet, the agent instance should be passed down into the instrument and aggregate classes upon initialization so we don't have to use utility functions like this. But I digress.

.eslintrc.js
```diff
@@ -106,6 +106,7 @@ module.exports = {
     'valid-typeof': 'off',
     'no-undef': 'off',
     'no-return-assign': 'off',
+    'no-fallthrough': ['error', { commentPattern: 'fallthrough' }],
```
patrickhousley, line 109: Where do we need this exception to the rule?

src/features/utils/instrument-base.js
```diff
@@ -43,46 +52,53 @@ export class InstrumentBase extends FeatureBase {
...
+
+          argsObjFromInstrument.sessionTrackingOn = true // let relevant features know the manager was initialized successfully and should be found on runtime obj
+
+          if (this.featureName === FEATURE_NAMES.sessionReplay) {
+            if (!shouldImportSR(sessionManager)) {
+              drain(this.agentIdentifier, this.featureName)
+              return
+            }
+          }
+        }
```
patrickhousley, line 77: Need an else block to prevent the session replay feature from starting when sessions are disabled.

```suggestion
        } else if (this.featureName === FEATURE_NAMES.sessionReplay) {
          drain(this.agentIdentifier, this.featureName)
          return
        }
```

.eslintrc.js
cwli24: No longer applicable. Will remove

src/common/aggregate/aggregator.js
```diff
@@ -10,6 +10,10 @@ export class Aggregator extends SharedContext {
...
+      Here, the central aggregator shared across all feature instrument & aggregate components is used for that signal. */
+    this.sessionReplayInitialized = new Promise(resolve => this.onReplayReady = resolve)
```
metal-messiah, line 16: This still feels like a really weird place to put this.  I think the shared aggregator can be stripped out as (most) of the features do not utilize it, so that would make this a casualty of that effort

metal-messiah: If the shared promise thing is the way you want to do this instead of using the event emitter, we should really have some other layer defined in the `agent.js` that handles these sort of cross-feature links.

cwli24: You mean move this to Trace instrument and add a handler subscribed to a new event emitted by Replay aggregate after it starts recording, and pass this promise from Trace instrument -> aggregate -> `getSessionReplayMode` ?

cwli24: Promise can also live in `replay-mode.js` but the handler tying event to resolve needs to happen pre-load. So will need to add a stateful init to the module so that it sets up the handler & promise with the instruments.

src/features/session_trace/aggregate/index.js
cwli24: Would go in its own PR to switch all the feature modules' getRuntime -> this.agentRuntime.

patrickhousley: I agree. Something to discuss internally.

tests/functional/stn/xhr-runtime-patched.test.js
```diff
@@ -8,11 +8,18 @@ const testDriver = require('../../../tools/jil/index')
...
 
 testDriver.test('posts session traces when xhr is runtime patched', supported, function (t, browser, router) {
-  t.plan(2)
-
   let rumPromise = router.expectRum()
   let resourcePromise = router.expectResources()
-  let loadPromise = browser.get(router.assetURL('xhr-runtime-patched.html')).waitForFeature('loaded')
+  let loadPromise = browser.get(router.assetURL('instrumented.html', {
+    init: {
+      privacy: { cookies_enabled: false }
```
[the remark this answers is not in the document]
metal-messiah: I'm wondering instead of cookies_enabled false, if we should just clear the localStorage in the script string?  Nervous that we are no longer testing our prime case since cookies_enabled will be true for most customers

src/common/aggregate/aggregator.js
patrickhousley: I agree this probably shouldn't be in the aggregator class. It sounds like we can't use the event emitter because we can't guarantee the load order and that the subscribe will be in place when the event is dispatched indicating the session replay feature loaded. Is that correct? If so, I wonder if we could take this as a need to enhance the event emitter to allow for replaying of events for certain streams? Or, maybe the base feature class can have an `isInitialized` deferred object that can be resolved? Something along the lines of setting up lifecycle events for features. I digress, these would all be later PRs and work.

patrickhousley (approved):
Approved but we should solve the aggregator class discussion before merging.

patrickhousley (approved):
Approved but we should solve the aggregator class discussion before merging.

src/common/aggregate/aggregator.js
cwli24: We settled on creating a new `sharedXYZ` which will basically be a side-by-side obj to the sharedAggregator. Just use it for whatever inter-feature comm needs to happen.

src/common/util/feature-flags.js
```diff
@@ -30,7 +30,8 @@ export function activateFeatures (flags, agentIdentifier) {
 
         handle('rumresp-' + flag, [Boolean(num)], undefined, feat, sharedEE) // this is a duplicate of feat-/block- but makes awaiting for 1 event easier than 2
       })
-    } else if (num) handle('feat-' + flag, [], undefined, undefined, sharedEE) // not sure what other flags are overlooked, but there's a test for ones not in the map
+    } else if (num) handle('feat-' + flag, [], undefined, undefined, sharedEE) // not sure what other flags are overlooked, but there's a test for ones not in the map --
+    // ^^^ THIS DOESN'T ACTUALLY DO ANYTHHING AS UNDEFINED/FEATURE GROUP ISN'T DRAINED
```
metal-messiah, line 34: Should we add this here if it doesn't do anything?  Can we address this separately?

src/common/session/session-entity.js
```diff
@@ -21,10 +25,9 @@ const model = {
   expiresAt: 0,
   updatedAt: Date.now(),
   sessionReplay: 0,
```
metal-messiah, line 27:
```suggestion
  sessionReplay: MODE.OFF,
```

src/features/session_trace/aggregate/index.js
```diff
@@ -44,65 +48,131 @@ export class Aggregate extends AggregateBase {
...
+        case true:
+          this.startTracing(operationalGate)
+          break
+        case MODE.OFF:
+        case false:
+        default: // this feature becomes "off" (does nothing & nothing is sent)
+          operationalGate.decide(false)
+          break
+      }
+    }
```
metal-messiah, line 69: Is there a reason i'm missing to be overly explicit with the case statements here?

```suggestion
const controlTraceOp = (traceMode) => {
  if (traceMode) this.startTracing(operationalGate)
  else operationalGate.decide(false)
}
```

src/features/session_trace/aggregate/index.js
```diff
@@ -44,65 +48,131 @@ export class Aggregate extends AggregateBase {
...
+    if (!sessionEntity) {
+      // Since session manager isn't around, do the old Trace behavior of waiting for RUM response to decide feature activation.
+      this.isStandalone = true
+      registerHandler('rumresp-stn', (on) => controlTraceOp(on), this.featureName, this.ee)
```
metal-messiah, line 73: Should probably just align on one way for waiting for flags.  Doesn't have to be done in this PR, but just below this you are using `waitForFlags()`.  For consistency would make sense to use that everywhere

metal-messiah: https://issues.newrelic.com/browse/NEWRELIC-9281

tests/functional/pvt/timings.test.js
```diff
@@ -22,25 +22,25 @@ function fail (t) {
...
+// runWindowLoadTests('rum')
+// runPageHideTests('spa')
+// runPageHideTests('rum')
+// runClsTests('spa')
+// runClsTests('rum')
+// runCustomAttributeTests('spa')
+// runCustomAttributeTests('rum')
+// runLcpTests('rum')
+// runLcpTests('spa')
+// runPvtInStnTests('spa')
```
metal-messiah, line 43: These need to be uncommented

src/features/session_trace/aggregate/index.js
cwli24: This was originally specific to MODE enum. Might separate true/false apart in the future, or actually have more to-dos for MODE.ERROR, which originally _was_ the case.

src/features/session_trace/aggregate/index.js
cwli24: tbh I used waitForFlags bc I needed _both_ flags in that case

cwli24:
Persistent safari 15 on Sauce tunnel error, but can just treat as all tests passing.

metal-messiah (approved):
Behavior appears to match logic tree on test-server for all cases

metal-messiah:
Approved pending passing JIL, jest, and WDIO tests

PR #546, opened 2023-05-26
fix: Address "configurable" warnings arising from user-agent module

bjfield:
The unused userAgent property and corresponding module have been removed from the runtime object due to "configurable" warnings being thrown in cases attempting to overwrite the runtime configuration.
---

### Overview

The runtime object contained a direct module import of a wrapper around the userAgent. This is unused and unnecessary, and caused warnings to throw in cases attempting to overwrite the runtime configuration. This has been removed.

### Related Issue(s)

N/A

### Testing

Tests should continue to pass. Run this PR locally and go to any asset file where the `info` configurations are provided after the loader code, such as the JSPDF or RequireJS assets, and ensure the warnings no longer throw.


patrickhousley (approved):
Tests passing.

metal-messiah (approved):

PR #549, opened 2023-05-26
chore: Experimental builds for local testing

metal-messiah:
Add another loader type, `experimental` which only builds for local tests.  This enables a local testing environment which supports features which have yet to be GA'd.
---

### Overview
This PR adds a new loader type for use in local testing. This can be accessed by supplying the loader type `experimental` into a functional or WDIO test's configuration.  This will enable the local testing of features which are not included in the GA CDN products.

### Testing
Run local build, then open the test-server.  Go to a page and add `?loader=experimental` to the URL.  This replicates what the test-server does to run tests with a certain loader type.

### DIFF
--- a/src/cdn/experimental.js
+++ b/src/cdn/experimental.js
@@ -1,0 +1,39 @@
+/**
+ * @file Creates a "SPA" agent loader bundle composed of the core agent and all available feature modules.
+ */
+/*
+ * Copyright 2023 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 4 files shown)


patrickhousley (approved):
You may need to update `tools/wdio/util/loaders.js` and `tools/jil/util/loaders.js`.

metal-messiah:
> You may need to update `tools/wdio/util/loaders.js` and `tools/jil/util/loaders.js`.

Its probably important not to update those, since it looks like those are references by the deploy scripts

PR #554, opened 2023-05-31
chore: Add jest tests for common/util/traverse (NEWRELIC-7856)

bjfield:
Includes unit tests for the common/util/traverse module and refactors the `applyFnToProps` function for simplicity.
---

### Related Issue(s)

[NEWRELIC-7856](https://issues.newrelic.com/browse/NEWRELIC-7856)

### Testing

- `npm test src/common/util/traverse.test.js`
- Various browser tests use the `obfuscateAndSend` function, which uses `applyFnToProps`.
- This makes use of the `array/includes` core-js polyfill.


src/common/util/traverse.js
```diff
@@ -3,34 +3,27 @@
...
-  }
 
-  function shouldIgnore (key) {
-    var ignore = false
-    for (var i = 0; i < ignoreKeys.length; i++) {
-      if (ignoreKeys[i] === key) {
-        ignore = true
-        break
+  for (var property in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, property)) {
```
patrickhousley, line 19: Could use `Object.keys()` to do the loop. It will iterate over own enumerable properties of the object.

src/common/util/traverse.js
```diff
@@ -3,34 +3,27 @@
...
-    return obj
-  }
 
-  function shouldIgnore (key) {
-    var ignore = false
-    for (var i = 0; i < ignoreKeys.length; i++) {
-      if (ignoreKeys[i] === key) {
-        ignore = true
-        break
+  for (var property in obj) {
```
patrickhousley, line 18:
```suggestion
  for (const property in obj) {
```

src/common/util/traverse.test.js
```diff
@@ -0,0 +1,57 @@
+import { applyFnToProps } from './traverse'
+
+describe('applyFnToProps', () => {
+  it('should return the object unchanged if it is not an object', () => {
```
patrickhousley, line 4:
```suggestion
  it('should return the input unchanged if it is not an object', () => {
```

src/common/util/traverse.test.js
```diff
@@ -0,0 +1,57 @@
...
+    const result = applyFnToProps('not an object', jest.fn())
+    expect(result).toEqual('not an object')
+  })
+
+  it('should return the object unchanged if it is null', () => {
```
patrickhousley, line 9:
```suggestion
  it('should return the input unchanged if it is null', () => {
```

src/common/util/traverse.test.js
```diff
@@ -0,0 +1,57 @@
...
+    const result = applyFnToProps(null, jest.fn())
+    expect(result).toBeNull()
+  })
```
patrickhousley, line 12: This could be wrote using `test.each`

```suggestion
  test.each(['not an object', null, undefined])('should return input unchanged when input is not object', (input) => {
    const result = applyFnToProps(input, jest.fn())
    expect(result).toEqual(input)
  })
```

src/common/util/traverse.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+  it('should apply the provided function only to properties of the specified type', () => {
```
patrickhousley, line 14: Use `test` instead of `it`

src/common/util/traverse.js
bjfield: Changed

metal-messiah (approved):

PR #560, opened 2023-05-31
chore: Add jest tests for common/util/get-or-set (NEWRELIC-7848)

bjfield:
Includes jest unit tests for the `common/util/get-or-set` module, and a new JSDoc block.
---

### Related Issue(s)

[NEWRELIC-7848](https://issues.newrelic.com/browse/NEWRELIC-7848)

### Testing

`npm test src/common/util/get-or-set.test.js`

### DIFF
--- a/src/common/util/get-or-set.test.js
+++ b/src/common/util/get-or-set.test.js
@@ -1,0 +1,54 @@
+import { getOrSet } from './get-or-set'
+
+describe('getOrSet', () => {
+  it('should return the current value of an existing property', () => {
+    const obj = { foo: 'bar' }
+    const prop = 'foo'
+    const getVal = jest.fn()
+
+    const result = getOrSet(obj, prop, getVal)
+
+    expect(result).toBe('bar')
+    expect(getVal).not.toHaveBeenCalled()
+  })
+
+  it('should set and return the value from getVal if the property does not exist', () => {
+    const obj = {}
+    const prop = 'foo'
+    const getVal = jest.fn(() => 'baz')
+
+    const result = getOrSet(obj, prop, getVal)
+
+    expect(result).toBe('baz')
+    expect(getVal).toHaveBeenCalled()
+    expect(obj.foo).toBe('baz')
+  })
+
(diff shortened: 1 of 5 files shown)


src/common/util/get-or-set.test.js
```diff
@@ -0,0 +1,54 @@
...
+    const result = getOrSet(obj, prop, getVal)
+
+    expect(result).toBe('bar')
+    expect(getVal).not.toHaveBeenCalled()
+  })
+
+  it('should set and return the value from getVal if the property does not exist', () => {
+    const obj = {}
+    const prop = 'foo'
+    const getVal = jest.fn(() => 'baz')
```
patrickhousley, line 18: You could use `jest.fn().mockReturnValue('baz')`. Just fyi

src/common/util/get-or-set.test.js
```diff
@@ -0,0 +1,54 @@
...
+  it('should set the property from getVal if Object.defineProperty and Object.keys are not supported', () => {
+    const obj = {}
+    const prop = 'foo'
+    const getVal = jest.fn(() => 'baz')
+
+    const result = getOrSet(obj, prop, getVal)
+
+    expect(result).toBe('baz')
+    expect(obj.foo).toBe('baz')
+  })
```
patrickhousley, line 53: This test is not exercising the fallback in the function like the name is implying. Change the test:

```suggestion
  it('should set the property from getVal if Object.defineProperty and Object.keys are not supported', () => {
    const obj = {}
    const prop = 'foo'
    const getVal = jest.fn(() => 'baz')

    const originalFn = Object.defineProperty
    Object.defineProperty = null

    const result = getOrSet(obj, prop, getVal)

    expect(result).toBe('baz')
    expect(obj.foo).toBe('baz')
    expect(Object.prototype.propertyIsEnumerable.call(obj, 'foo')).toBe(true)

    Object.defineProperty = originalFn
  })
```

bjfield: Oops! Didn't finish that one. Thank you.

metal-messiah (approved):

PR #561, opened 2023-05-31
fix: Address early return preempting XHR harvest re-schedule

bjfield:
A early return condition in the harvest rescheduling logic in version 1.233.0 was preventing Ajax event harvest from being rescheduled in cases where no data had accumulated.
---

### Overview

This PR addresses a problem where, once the harvest scheduler runs for the Ajax feature and the Ajax feature has no accumulated data, the Ajax feature no longer performs future harvests. The issue appears to be [line 92](https://github.com/newrelic/newrelic-browser-agent/blob/main/src/common/harvest/harvest-scheduler.js#L92) in the harvest scheduler where it early-returns when there is no data to send, but does not schedule the next harvest.

### Related Issue(s)

[NEWRELIC-8876](https://issues.newrelic.com/browse/NEWRELIC-8876)

### Testing

- All tests should run.
- `npm test src/common/harvest/harvest-scheduler.test.js`
- Manual verification of event harvest behavior should be confirmed.

### DIFF
--- a/src/common/harvest/harvest-scheduler.js
+++ b/src/common/harvest/harvest-scheduler.js
@@ -89,7 +89,12 @@ export class HarvestScheduler extends SharedContext {
       const retry = submitMethod.method === submitData.xhr
       var payload = this.opts.getPayload({ retry: retry })
 
-      if (!payload) return
+      if (!payload) {
+        if (this.started) {
+          this.scheduleHarvest()
+        }
+        return
+      }
 
       payload = Object.prototype.toString.call(payload) === '[object Array]' ? payload : [payload]
       harvests.push(...payload)
--- a/src/common/harvest/harvest-scheduler.test.js
+++ b/src/common/harvest/harvest-scheduler.test.js
@@ -2,7 +2,7 @@ import { setConfiguration } from '../config/state/init'
 import { HarvestScheduler } from './harvest-scheduler'
 
 describe('runHarvest', () => {
-  it('should re-schedule Ajax harvest even if there is no accumulated data', () => {
+  it('should re-schedule harvest even if there is no accumulated data', () => {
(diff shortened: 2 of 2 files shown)


patrickhousley (approved):

src/common/harvest/harvest-scheduler.test.js
```diff
@@ -0,0 +1,25 @@
+import { setConfiguration } from '../config/state/init'
+import { HarvestScheduler } from './harvest-scheduler'
+
+describe('runHarvest', () => {
+  it('should re-schedule Ajax harvest even if there is no accumulated data', () => {
```
cwli24, line 5: Omit "Ajax"; event endpoints isn't specific to ajax harvests

src/common/harvest/harvest-scheduler.js
```diff
@@ -89,7 +89,12 @@ export class HarvestScheduler extends SharedContext {
       const retry = submitMethod.method === submitData.xhr
       var payload = this.opts.getPayload({ retry: retry })
 
-      if (!payload) return
+      if (!payload) {
+        if (this.started) {
+          this.scheduleHarvest()
+        }
+        return
```
cwli24, line 96: Minor but can you refactor this & L127-129 into
```javascript
function scheduleNextHarvestIfActive() {
  if (this.started) {
    this.scheduleHarvest()
  }
}
```
Can insert this below `function onHarvestFinished (result)` on L132

bjfield: I think this is probably the right way to do it but it actually makes the file a couple lines longer. For the sake of time and to avoid an error, I'd like to leave it as is for now.

cwli24 (approved):

PR #564, opened 2023-06-01
Patching Fix Insufficient Entropy in cryptiles

wica-sufatmawati:
cryptiles is a package for general crypto utilities. Affected newrelic-browser-agent of this project are vulnerable to Insecure Randomness. The randomDigits() method is supposed to return a cryptographically strong pseudo-random data string, but it was biased to certain digits. An attacker could be able to guess the created digits.

```js

exports.randomDigits = function (size) {

    const buffer = exports.randomBits(size * 8);
    const digits = [];
    for (let i = 0; i < buffer.length; ++i) {
        digits.push(Math.floor(buffer[i] / 25.6));
```
## CVE Weaknesses
CVE-2018-1000620
CWE-331
`CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -83,7 +83,7 @@
         "path-browserify": "^1.0.1",
         "preprocessify": "0.0.6",
         "process": "^0.11.10",
-        "request": "2.74.0",
+        "request": "2.88.2",
         "saucelabs": "^7.2.1",
         "semver": "^5.3.0",
         "serialize-anything": "^1.2.3",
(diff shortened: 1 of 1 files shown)


patrickhousley:
@wica-sufatmawati thanks for the pull request. This PR only has changes to the `package-lock.json` file. This file get regenerated often and does not need manual updating. Additionally, the CVE does not apply to a dependency of the built code, only to a dependency used during testing. For these reasons, I am closing this pull request.

PR #568, opened 2023-06-02
chore: fix open handles warnings

patrickhousley:
Fixing jest warning about open handles remaining after test runs.
---

### Overview

Fixing the warnings jest produces about open handles being left after running the submitData tests.

### Related Issue(s)

N/A

### Testing

Run `npm test -- --detectOpenHandles ./src/common/util/submit-data` and verify there are no open handle warnings.

### DIFF
--- a/src/common/util/submit-data.test.js
+++ b/src/common/util/submit-data.test.js
@@ -3,30 +3,31 @@
  * @jest-environment-options {"html": "<html><head><script></script></head><body></body></html>", "url": "https://example.com/"}
  */
 
+import { faker } from '@faker-js/faker'
 import { submitData } from './submit-data'
+import * as globalScope from './global-scope'
 
-const mockWorkerScope = jest.fn().mockImplementation(() => false)
 jest.mock('./global-scope', () => ({
   __esModule: true,
   get isWorkerScope () {
-    return mockWorkerScope()
+    return false
   }
 }))
 
 const url = 'https://example.com/api'
 
-beforeEach(() => {
+afterEach(() => {
   jest.restoreAllMocks()
-  mockWorkerScope.mockReturnValue(false)
 })
 
 describe('submitData.jsonp', () => {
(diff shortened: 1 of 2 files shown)


src/common/util/submit-data.test.js
```diff
@@ -35,125 +35,141 @@ describe('submitData.jsonp', () => {
...
-    delete global.importScripts
+    expect(global.importScripts).toHaveBeenCalledWith(url + '&jsonp=' + jsonp)
   })
 
-  test('should fall back to an xhrGet call and return false when called from a worker scope', () => {
-    mockWorkerScope.mockReturnValueOnce(true)
-
-    const jsonp = 'callback'
-
+  test('should fall back to an xhrGet call and return false importScripts throws an error', () => {
```
bjfield, line 47:
```suggestion
  test('should fall back to an xhrGet call and return false when importScripts throws an error', () => {
```

src/common/util/submit-data.test.js
```diff
@@ -3,30 +3,30 @@
  * @jest-environment-options {"html": "<html><head><script></script></head><body></body></html>", "url": "https://example.com/"}
  */
 
+import { faker } from '@faker-js/faker'
```
bjfield, line 6: As a principle for future reference, I'm curious about this choice. It seems like using faker instead of primitives makes test runs indeterministic and increases run time by a tiny fraction.

bjfield (approved):
Confirmed that this addresses the open handles.

src/common/util/submit-data.test.js
patrickhousley: Using fake data is better in unit tests to catch possible issues caused by hard coded values. It adds a small amount of overhead but it can help prevent nasty bugs later in production where a code path was mistakenly tied to a magic value unintentionally.

metal-messiah (approved):

PR #569, opened 2023-06-03
chore: unit tests

patrickhousley:
Adding more unit tests. Cleaning up a little bit for existing tests.
---

### Overview

More unit tests and a little cleanup work.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-8893
https://issues.newrelic.com/browse/NEWRELIC-7849
https://issues.newrelic.com/browse/NEWRELIC-7855

### Testing

Run `npm test`


src/features/utils/instrument-base.test.js
```diff
@@ -0,0 +1,187 @@
...
+  const windowLoadCallback = jest.mocked(onWindowLoad).mock.calls[0][0]
+  await windowLoadCallback()
+
+  expect(getConfigurationValue).toHaveBeenCalledWith(agentIdentifier, 'privacy.cookies_enabled')
+  expect(setupAgentSession).toHaveBeenCalledWith(agentIdentifier)
+  expect(drain).toHaveBeenCalledWith(agentIdentifier, FEATURE_NAMES.sessionReplay)
+  expect(lazyFeatureLoader).not.toHaveBeenCalled()
+  expect(mockAggregate).not.toHaveBeenCalled()
+})
+
```
bjfield, line 169: Given that the "warn-don't-error" behavior is important, maybe we need a test like "should catch and warn when the session manager fails to import"—applying to the catch condition on `instrument-base.js:53`.

src/features/utils/lazy-feature-loader.test.js
```diff
@@ -0,0 +1,37 @@
+import { faker } from '@faker-js/faker'
+import { FEATURE_NAMES } from '../../loaders/features/features'
+import { lazyFeatureLoader } from './lazy-feature-loader'
+
+// Use enableAutomock to make it easier to mock all the things that get imported by the aggregators
+jest.enableAutomock()
```
bjfield, line 6: Is the purpose of this to prevent executing lazy-loaded aggregate modules?

bjfield (approved):

src/features/utils/instrument-base.test.js
patrickhousley: I was not able to find a way to `spyOn` the `import` function without it causing issues but maybe I can try again.

src/features/utils/lazy-feature-loader.test.js
patrickhousley: Pretty much. If you run just this test file with coverage, you shouldn't get any coverage on the feature aggregates. Mocking in these tests was a bit of trial and error.

src/features/utils/instrument-base.test.js
bjfield: Yeah, that's why I left you a comment instead of a code suggestion! I don't think it's worth the effort from a cost-benefit perspective. Resolving this conversation.

src/common/util/obfuscate.js
```diff
@@ -5,7 +5,7 @@ import { warn } from './console'
 
 var fileProtocolRule = {
   regex: /^file:\/\/(.*)/,
-  replacement: 'file://OBFUSCATED'
+  replacement: atob('ZmlsZTovL09CRlVTQ0FURUQ=')
```
metal-messiah, line 8: Just want to confirm, was the only issue previously presented just that the presence of the string `file://` in the source code was triggering code security monitors?

metal-messiah (approved):

src/common/util/obfuscate.js
patrickhousley: Yes and `atob` is supported by all browsers IE10+.

PR #574, opened 2023-06-07
feat: Decorate error objects to facilitate future UI experiences

metal-messiah:
Error payloads will now be decorated with a real timestamp and feature flags to facilitate error usages in future NR1 UI experiences.
---

### Overview

This PR adds a real timestamp and a `hasReplay` flag to the error object.  The timestamp is based on the first occurrence of an error (since multiple matching errors are aggregated). 

### Related Issue(s)
NEWRELIC-8990

### Testing
A new wdio test has been added to validate these properties.  Existing JIL and browser tests should continue to pass.


bjfield (approved):
The line is blurry for me as to whether this should be a "feat". I'd be more inclined to make it a chore because even though src is changed, the changes are small and independent of other logic in the agent.

metal-messiah:
> The line is blurry for me as to whether this should be a "feat". I'd be more inclined to make it a chore because even though src is changed, the changes are small and independent of other logic in the agent.

I was leaning toward "feat" due to the fact that this impacts customer ingest data (though very subtly)

bjfield:
> I was leaning toward "feat" due to the fact that this impacts customer ingest data (though very subtly)

Okay, yes, I think that’s the right call.

patrickhousley (approved):

tests/specs/js-errors/error-payload.e2e.js
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  it('should set a timestamp, tied to the FIRST error seen', async () => {
+    await browser.url(await browser.testHandle.assetURL('instrumented.html')) // Setup expects before loading the page
+      .then(() => browser.waitForAgentLoad())
+
+    const [before, after] = await browser.execute(function () {
+      var a = 0
+      var before = Date.now()
+      while (a++ < 20) newrelic.noticeError(new Error('test'))
```
cwli24, line 45: I'm not sure this test makes a lot of sense or is purposeful. The first error will have this param, but the rest 19 falls into sending `params.browser_stack_hash` instead.

I think really you should check that throwing an error (first) has this property with a reasonably correct timestamp (is not past the `after`), and then throwing the same error again _doesn't_ carry this property.

Also, note `noticeError` records the timestamp in a separate `handle('err',...)` than jserror module's `fn-err` cb and wrapped global handler -- both more likely to be what's hit if users run into unexpected errors: so it's better to test with `throw new Error(...)`.

src/features/jserrors/aggregate/index.js
```diff
@@ -171,6 +171,7 @@ export class Aggregate extends AggregateBase {
     if (!this.stackReported[bucketHash]) {
       this.stackReported[bucketHash] = true
       params.stack_trace = truncateSize(stackInfo.stackString)
+      params.firstSeenAt = agentRuntime.offset + time
```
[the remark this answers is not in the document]
cwli24: @patrickhousley It's the unix epoch time when the page or document was created, and `time` here is the high res time since then

cwli24 (approved):
Pending tests passing

src/features/jserrors/aggregate/index.js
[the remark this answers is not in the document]
patrickhousley: Oh the `firstSeenAt` needs to be an actual timestamp instead of the time since the page started loading.

tests/specs/js-errors/error-payload.e2e.js
metal-messiah: The way the error module works is that it aggregates many errors into a single object with a count.  This is just checking that of the 20 that threw, the timestamp should match the 1st only.

As for checking other errors containing the property, if thats a concern we should just test them all

metal-messiah: Thanks i think i see what you were meaning about the browser_stack_hash.  Will put in a more eloquent solution

PR #576, opened 2023-06-08
feat: Remove img, jsonp, and xhrGet methods

patrickhousley:
Unused agent network methods have been removed. All data transmitted by the agent will now default to using XMLHttpRequest POST calls except for data transmitted during a visibility change or page unload. Visibility changes and page unloads will use sendBeacon with a fallback to fetch with keepalive. For browsers that do not support sendBeacon, the final data transmission will use XMLHttpRequest POST.
---

### Overview

Removing the unused `sendData` methods `jsonp`, `img', and `xhrGet`. All data will now be transmitted using `XMLHttpRequest` via `POST`. For browsers that support `sendBeacon`, the final harvest will fallback to `fetch` with `keepalive: true`.

All tests for `harvest.js` and `harvest-scheduler.js` modules have been converted to Jest and WDIO. In doing so, I have created many auto-mock files in `__mock__` directories. Using this style of mocking modules will let us manage the mock in a central location for each module and eventually enabled auto mocking globally for our jest unit tests.

Minor cleanup work was done with the `util/global-scope.js` and `browser-version/*.js` modules. Since these modules exported constants that were determined at runtime, I moved them into the `common/constants` folder into a single module. This will reduce import/require statements in the final bundle.

The pull request checks workflow has been updated to add the size check comment to PRs. The babel and ts configs have been updated so the test files and mocks do not get included in our `/dist/` directory.

### Related Issue(s)

- [NEWRELIC-8494](https://issues.newrelic.com/browse/NEWRELIC-8494)

Functional test conversions:

- [NEWRELIC-9027](https://issues.newrelic.com/browse/NEWRELIC-9027) - functional/disable-harvest.test.js
- [NEWRELIC-9035](https://issues.newrelic.com/browse/NEWRELIC-9035) - functional/err/harvest.test.js
- [NEWRELIC-9043](https://issues.newrelic.com/browse/NEWRELIC-9043) - functional/final-harvest.test.js
- [NEWRELIC-9046](https://issues.newrelic.com/browse/NEWRELIC-9046) - functional/harvest.test.js
- [NEWRELIC-9079](https://issues.newrelic.com/browse/NEWRELIC-9079) - functional/spa/harvest.test.js
- [NEWRELIC-9095](https://issues.newrelic.com/browse/NEWRELIC-9095) - functional/stn/harvest.test.js
- [NEWRELIC-9106](https://issues.newrelic.com/browse/NEWRELIC-9106) - functional/xhr/harvest.test.js

Browser test conversions:

- [NEWRELIC-9252](https://issues.newrelic.com/browse/NEWRELIC-9252) - browser/harvest-scheduler.browser.js
- [NEWRELIC-9118](https://issues.newrelic.com/browse/NEWRELIC-9118) - browser/harvest.browser.js

### Testing

Unit and integration tests have been updated.


src/common/util/submit-data.js
```diff
@@ -78,15 +34,24 @@ submitData.xhr = function xhr ({ url, body = null, sync, method = 'POST', header
...
+ * @param {string=} args.body - The Stringified body.
+ * @param {string=} [args.method=POST] - The XHR method to use.
+ * @param {{key: string, value: string}[]} [args.headers] - The headers to attach.
+ * @returns {XMLHttpRequest}
  */
-submitData.img = function img ({ url }) {
-  var element = new Image()
-  element.src = url
-  return element
+export function fetchKeepAlive ({ url, body = null, method = 'POST', headers = [{ key: 'content-type', value: 'text/plain' }] }) {
```
metal-messiah, line 45: Does this cause any errors in FF or just not honor the "keep-alive" part?

patrickhousley: I will have to check but was that an issue the last time we tried fetch with keepalive?

metal-messiah: I don't believe we've ever actually utilized fetch keepalive in the agent, so i'm not sure what firefox foes there.  I'm assuming it just ignores that property and treats it like a standard fetch

patrickhousley: Tested in Chrome 65, edge 13, firefox 55...they all still make the fetch call and seem to ignore the keepalive part.

src/common/browser-version/ios-version.js
```diff
@@ -1,11 +0,0 @@
-export const isiOS = /(iPad|iPhone|iPod)/g.test(navigator.userAgent)
-
-/* Feature detection to get our version(s). */
-
-// Shared Web Workers introduced in iOS 16.0+ and n/a in 15.6-
-export const iOS_below16 = isiOS && Boolean(typeof SharedWorker === 'undefined')
-/*
-  ^ It was discovered in Safari 14 (https://bugs.webkit.org/show_bug.cgi?id=225305) that the buffered flag in PerformanceObserver
-  did not work. This affects our onFCP metric in particular since web-vitals uses that flag to retrieve paint timing entries.
-  This was fixed in v16+.
```
cwli24, line 10: This comment disappeared but should be retained in runtime constants.

src/features/page_view_timing/aggregate/index.js
```diff
@@ -196,7 +196,7 @@ export class Aggregate extends AggregateBase {
   onHarvestFinished (result) {
     if (result.retry && this.timingsSent.length > 0) {
       for (var i = 0; i < this.timingsSent.length; i++) {
-        this.timings.push(this.timingsSent[i])
+        this.timings.unshift(this.timingsSent[i])
       }
```
cwli24, line 200: `this.timings` was filled in chronological order aka per call to `addTiming`, and that's the order in which it should be resent out, right?
If so, the unshift 1-by-1 actually reverses the order of the sent timings. This can do what Trace has done, simply:
````javascript
this.timings = this.timingsSent.concat(this.timings)

Comment (cwli24) [reply src/features/page_view_timing/aggregate/index.js:200]: Or is `this.timings` wrong too? Should it be `[3,2,1]` rather than `[1,2,3]` ?

File: src/common/harvest/harvest.js [hunk 4]
```diff
@@ -51,87 +39,88 @@ export class Harvest extends SharedContext {
...
    * @param {NetworkSendSpec} spec Specification for sending data
    */
-  sendX (spec) {
-    const { endpoint, opts } = spec
-    var submitMethod = getSubmitMethod(endpoint, opts)
-    if (!submitMethod) return false
-    var options = {
-      retry: submitMethod.method === submitData.xhr
+  sendX (spec = {}) {
+    const submitMethod = submitData.getSubmitMethod({ isFinalHarvest: spec.opts?.unload })
```
metal-messiah, line 46: Will this always be valid and defined now? (old behavior had a case where submitMethod was being checked for existence)

src/common/constants/runtime.js
```diff
@@ -0,0 +1,63 @@
...
+export const ffVersion = (() => {
+  const match = navigator.userAgent.match(/Firefox[/\s](\d+\.\d+)/)
+  if (Array.isArray(match) && match.length >= 2) {
+    return +match[1]
+  }
+
+  return 0
+})()
+
+export const supportsSendBeacon = !!navigator.sendBeacon
```
cwli24, line 63: [no action needed]

FYI, I'm going to create a separate ticket to merge sendBeacon -> fetch keepalive, since they're effectively doing the same thing, so don't go too crazy with sendBeacon specific tests.

cwli24: NEWRELIC-9384

src/common/harvest/harvest.js
patrickhousley: Yea, `getSubmitMethod` will now always return a submit method.

src/common/harvest/harvest.js
```diff
@@ -51,87 +39,88 @@ export class Harvest extends SharedContext {
...
+      if (endpoint === 'events') {
+        body = body.e
       } else {
-        fullUrl = fullUrl + encodeObj(payload.body, agentRuntime.maxBytes)
+        body = stringify(body)
       }
-    } else body = payload.body
+    }
+
+    if (!body || body.length === 0 || body === '{}' || body === '[]') {
```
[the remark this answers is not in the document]
patrickhousley: In my testing, I have not seen this issue any more since this line was added and the call to `cleanPayload` on line 87.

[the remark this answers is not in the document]
patrickhousley: This should be fully covered by the jest unit tests.

src/common/constants/runtime.js
patrickhousley: The only thing that might keep us from doing that is the fact that firefox does not support the `keepalive` flag on fetch.

src/features/page_view_timing/aggregate/index.js
patrickhousley: Good catch @cwli24 the loop actually isn't needed at all.

src/common/constants/runtime.js
metal-messiah: We've already gone down this road and we can not utilize fetch keep alive for ie or ff.

src/common/util/submit-data.js
```diff
@@ -78,15 +55,24 @@ submitData.xhr = function xhr ({ url, body = null, sync, method = 'POST', header
...
+  return fetch(url, {
+    method,
+    headers: headers.reduce((aggregator, header) => {
+      aggregator.push([header.key, header.value])
```
metal-messiah, line 70: Does `headers` work with an array of arrays here?  I've always just seen it as a flat object of k:v defs so just double checking.

patrickhousley: Yup: https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers

metal-messiah (approved):
I think this covers all the topics mentioned and I have not been able to find any faults from the migration of the previous implementation.  This still likely represents a change that affects the entire agent, so I think we should consider deploying this as a standalone shipment, but i think it looks G2G. approved!

patrickhousley:
> To clarify since it wasn't explicitly stated elsewhere, this will remove the ability for IE11 (those still on it) to send out data on page unload events, yes?

Not entirely. For IE11, the img method was used before and now XHR will be used. Both methods are a best effort and we cannot guarantee that either will actually result in the data making it out of the browser before the browser kills the JS process.

bjfield (approved):

PR #578, opened 2023-06-08
chore: Convert functional/obfuscate.test.js to WDIO (NEWRELIC-9056)

bjfield:
Replaces JIL functional test obfuscate.test.js with a new WDIO test file.
---

### Overview

- Replaces JIL functional test obfuscate.test.js with a new WDIO file.
- Moves a helper function that's now used only once out of the helper file into the remaining test file where it's used.
- Modifies the test server so the config object passed to the command API from the test handle connector is first serialized, then deserialized, so RegExp objects are preserved and not converted to empty objects.

### Related Issue(s)

[NEWRELIC-9056](https://issues.newrelic.com/browse/NEWRELIC-9056)

### Testing

- `npm run wdio -- --retry=false tests/specs/obfuscate.e2e.js`
- `./tools/jil/bin/cli.js tests/worker/all-uncategorized.test.js` (because helper function was relocated)


tools/browser-matcher/common-matchers.mjs
```diff
@@ -22,3 +22,14 @@ export const supportsFetch = new SpecMatcher()
...
+// Some browsers support basic fetch API, but not all supporting functions.
+// E.g. arrayBuffer on ios@10 generates an error when used with FormData instance.
+// MDN shows this function as not supported: https://developer.mozilla.org/en-US/docs/Web/API/Body/arrayBuffer
+export const supportsFetchExtended = new SpecMatcher()
+  .include('safari>=11.1') // MDN says no support (11.1 currently latest), but 11.1 is accounted for in the tests
+  .include('chrome>=42')
+  .include('edge>=14')
+  .include('firefox>=40')
+  .include('ios>=11.3')
+  .include('android')
```
bjfield, line 35: Unsure of the relevance and history of the `fetchExt` (supportsFetchExtended) matcher. Seems like it's probably not needed given how old the browsers are (ios>=11.3, safari>=11.1), and we could just use the supportsFetch matcher now.

patrickhousley: I think this depends on how far back we want to support the "all browsers" testing.

tests/specs/obfuscate.e2e.js
```diff
@@ -0,0 +1,92 @@
...
+
+    await Promise.all([
+      ajaxPromise,
+      errorsPromise,
+      insPromise,
+      resourcePromise,
+      spaPromise,
+      timingsPromise,
+      rumPromise
+    ]).then(([
```
patrickhousley, line 58: Can we get rid of the promise chains please and just use `async..await`?

tools/browser-matcher/common-matchers.mjs
bjfield: The same question of relevance applies to the old versions in our supportsFetch matcher, I suppose. I'll leave it for now for clarity around potential future old browser edge cases.

tests/specs/obfuscate.e2e.js
```diff
@@ -0,0 +1,92 @@
...
+  withBrowsersMatching(supportsFetchExtended)('should apply to all payloads', async () => {
+    const spaPromise = browser.testHandle.expectEvents()
+    const ajaxPromise = browser.testHandle.expectAjaxEvents()
+    const timingsPromise = browser.testHandle.expectTimings()
+    const errorsPromise = browser.testHandle.expectErrors()
+    const insPromise = browser.testHandle.expectIns()
+    const resourcePromise = browser.testHandle.expectResources()
+    const rumPromise = browser.testHandle.expectRum()
+
+    const config = {
```
patrickhousley, line 13: Should probably move this variable declaration outside the test, above the describe. You can also probably remove all the properties except the `init.obfuscate` properties. The test server handle will merge the properties with the defaults in `tools/wdio/plugins/testing-server/default-asset-query.mjs`.

patrickhousley (approved):

metal-messiah (approved):

PR #581, opened 2023-06-13
chore: Convert api.test.js from JIL to WDIO - NEWRELIC-9026

bjfield:
Replaces the tests/functional/api.test.js file with tests/specs/api.e2e.js.
---

### Overview

- Adds matchers for notIE and consistentTimingData
- Removes old JIL test file
- Fleshes out the rest of api.e2e.js

### Related Issue(s)

[NEWRELIC-9026](https://issues.newrelic.com/browse/NEWRELIC-9026)

### Testing

`npm run wdio -- --retry=false tests/specs/api.e2e.js`


tests/specs/api.e2e.js
```diff
@@ -64,7 +67,7 @@ describe('newrelic api', () => {
     })
   })
 
-  describe('noticeError api', () => {
+  describe('noticeError()', () => {
```
patrickhousley, line 70: Keep it consistent ;)

tests/specs/api.e2e.js
```diff
@@ -99,4 +102,205 @@ describe('newrelic api', () => {
...
+        browser.testHandle.expectRum(),
+        browser.url(await browser.testHandle.assetURL('api/finished.html', {
+          init: {
+            page_view_timing: {
+              enabled: false
+            },
+            ins: {
+              harvestTimeSeconds: 2
+            }
+          }
```
patrickhousley, line 119: I would check any config objects like this...if they are necessary I mean. The default feature harvest time is 5 seconds and that seems pretty stable for everyone and fast enough.

tests/specs/api.e2e.js
```diff
@@ -99,4 +102,205 @@ describe('newrelic api', () => {
...
+        }))
+          .then(() => browser.waitForAgentLoad())
+      ])
+
+      const pageActions = insResult.request.body.ins
+      expect(pageActions).toBeDefined()
+      expect(pageActions.length).toEqual(1) // exactly 1 PageAction was submitted
+      expect(pageActions[0].actionName).toEqual('finished') // PageAction has actionName = finished
```
patrickhousley, line 127: I have found expecting on a specific indexed event to be flakey. Could just do a quick search for the event to see if it was included in the request body.

```suggestion
      const finishedAction = Array.from(insResult.request.body.ins || [])
        .find(e => e.actionName === 'finished')
      expect(finishedAction).toBeDefined()
```

tests/specs/api.e2e.js
```diff
@@ -99,4 +102,205 @@ describe('newrelic api', () => {
...
+      const oneHundredX = 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
+      const twoHundredCharacterString = ninetyNineY + oneHundredX + 'q'
+      const ri = {
+        one: '201',
+        three: twoHundredCharacterString
+      }
+      ri[twoHundredCharacterString] = '2'
+      const queryRi = JSON.parse(errorsResult.request.query.ri)
+      expect(twoHundredCharacterString.length).toBe(200)
+      expect(queryRi).toEqual(ri)
```
patrickhousley, line 206: This test is a bit confusing. First, line 205 is unnecessary. It is testing a variable constructed in the test. Second, could use [`expect.toMatch`](https://jestjs.io/docs/expect#tomatchregexp--string) instead of constructing the string. That would remove lines 196-203 ish.

```suggestion
      const queryRi = JSON.parse(errorsResult.request.query.ri)
      expect(queryRi.one).toEqual('201')
      expect(queryRi.three).toMatch(/y{99}x{100}q/)
      expect(Object.keys(queryRi)).toContain(expect.stringMatching(/y{99}x{100}q/))
```

tests/specs/api.e2e.js
```diff
@@ -99,4 +102,205 @@ describe('newrelic api', () => {
...
+          scriptString: `
+            newrelic.setUserId(456)
+            newrelic.setUserId({'foo':'bar'})
+            newrelic.noticeError('fake1')
+            `
+        }))
+          .then(() => browser.waitForAgentLoad())
+      ])
+
+      const firstErrorsResult = await browser.testHandle.expectErrors(3000)
```
patrickhousley, line 276: Should probably move this into the `Promise.all` above. Possible edge case here where the page loads / errors harvest before the test runner can fully execute this line of code. Would also remove the timeout value and remove the harvestTimeSeconds above.

tests/specs/api.e2e.js
```diff
@@ -99,4 +102,205 @@ describe('newrelic api', () => {
...
+      expect(firstErrorsResult.request.body.err[0]).toHaveProperty('custom')
+      expect(firstErrorsResult.request.body.err[0].custom[ERRORS_INBOX_UID]).not.toBeDefined() // Invalid data type (non-string) does not set user id
+
+      await browser.execute(function () {
+        newrelic.setUserId('user123')
+        newrelic.setUserId()
+        newrelic.noticeError('fake2')
+      })
+
+      const secondErrorsResult = await browser.testHandle.expectErrors(3000)
```
patrickhousley, line 287: Same here. Use Promise all.

tools/browser-matcher/common-matchers.mjs
```diff
@@ -22,3 +22,21 @@ export const supportsFetch = new SpecMatcher()
   .include('firefox>=40')
   .include('ios>=11')
   .include('android')
+
+// Firefox <= 19 has inconsistent timing data.
```
patrickhousley, line 26: I think we need to settle on a minimum browser version that we simply will never test. I am pretty sure you can't test FF 19 with W3C protocol.

tests/specs/api.e2e.js
bjfield: What kind of flakey have you seen with expecting specific indexed events? Like the number of items in `.ins` might vary? Whoever wrote this test originally noted that they were expecting exactly one PageAction, which sounds like a stable expectation worth testing for.

tests/specs/api.e2e.js
bjfield: Elegant!

tests/specs/api.e2e.js
patrickhousley: Yea if there is expected to only be one item in `ins`, that makes sense.

tools/browser-matcher/common-matchers.mjs
bjfield: I don't know what a sensible minimum is, but a 10 year old Firefox probably falls well below it. I'll remove this matcher.

tests/specs/api.e2e.js
bjfield: Do you mean you'd prefer to leave these as `... api` instead of `()`?

patrickhousley: No, you missed one.

patrickhousley: Maybe not. I might not be reading the diff correctly.

patrickhousley (approved):

metal-messiah (approved):

PR #582, opened 2023-06-13
chore: Remove redundant JIL test file metrics.test.js - NEWRELIC-9049

bjfield:
Since metrics.e2e.js already covers the same tests, the JIL test file can be removed.
---

### Overview

I verified that `metrics.e2e.js `covers the same scenarios as the removed JIL file.

### Related Issue(s)

[NEWRELIC-9049](https://issues.newrelic.com/browse/NEWRELIC-9049)

### Testing

`npm run wdio -- --retry=false tests/specs/metrics.e2e.js`


bjfield:
> This file was left in place as a performance comparison of sorts between JIL and WDIO.

The WDIO version has a couple additional tests now. Has it outlived its usefulness as a performance comparison?

If we keep it around I’d like to add a comment in the file about the purpose.

patrickhousley:
> The WDIO version has a couple additional tests now. Has it outlived its usefulness as a performance comparison?

Probably yea. Hell, this is git so we can always recover it.

patrickhousley (approved):

metal-messiah (approved):

PR #583, opened 2023-06-13
feat: Add error mode to session trace - NEWRELIC-8666 NEWRELIC-9283

cwli24:
Session-enabled traces now have an "error mode" capability in which the feature lies dormant until an exception is thrown. Upon the error, the past 30 seconds of data is retained and sent, and the feature becomes fully activated again.
---
### Overview

Following https://github.com/newrelic/newrelic-browser-agent/pull/545, Trace can now run in error mode, which makes it silent until such time an error occurs and it starts harvesting the 30 seconds prior and in full thereafter. (Similar to experimental Replay.)

Some timer nodes are also removed (evaluated as ok) to reduce noise in reported data.

### Related Issue(s)
NEWRELIC-8666
NEWRELIC-9283

### Testing

All workflows from #545 should remain unaffected.
In addition, these should now hold true as well:

- Granted that Replay is **on**, _on a new session page_:
  - if RUM `stn:1`, Trace will run fully and normally, harvesting everything regardless of Replay mode.
  - if `stn:0`, Trace mode will mirror Replay mode (abiding by truth table): off to off, full to full, error to error.
- When Replay starts in **Error** mode:
  - it doesn't immediately start harvesting per usual.
  - it switches to FULL capture and starts harvesting upon an error being thrown (and recorded by JS error feature).
    - the first harvest should be much like the normal initial harvest -- e.g., gets `ptid`
    - data was not dropped or missed
      - e.g., error thrown before 30 seconds of page life should yield everything like load timings that are normally sent
    - data only pertains to the past 30 seconds of page life from present
      - e.g., error thrown after 30 seconds of page load event (roughly) should exclude load timings
      - note: nodes are included as long as their _end_ time falls within the 30 sec window, i.e. their duration overlaps
    - if error gets thrown _before_ page load (that is, before Trace has started), then it starts with FULL capture, rather than in silent ERROR mode.
- When Replay aborts aka `REPLAY_ABORTED` event is emitted,
  - if Trace is in Full mode, it sends the pending nodes and permanently stops.
  - if Trace is in Error (or Off) mode, nothing happens -- it's already not harvesting.
- When user navigates away or hides page, ST final harvest is **not** send while in error mode.
- When Trace (also) starts in Error mode on a page, opening a new tab-same page will also start Trace in Error mode.
  - if Trace transitions to Full mode (e.g., from error being thrown) or to Off (e.g., from directly setting the `runtime.session.state.sessionTraceMode = 0`) on the new tab, switching back to the starting page causes its mode to change respectively as well.


cwli24:
@bjfield new title longer than 70 chars

metal-messiah (approved):

patrickhousley (approved):

PR #592, opened 2023-06-21
chore: Disable jspdf tests for iOS

patrickhousley:
Disabling jspdf tests for iOS.
---

### Overview

iOS via Appium does not like the canvas code in the html2canvas third-party script.

### Related Issue(s)

N/A

### Testing

N/A


metal-messiah:
I'm confused on this one. Didn't we have this test working in JIL in iOS? (Just not >=16)

patrickhousley:
> I'm confused on this one. Didn't we have this test working in JIL in iOS? (Just not >=16)

I have spent a lot of time looking into this issue. You are right, the original JIL test passes but the WDIO test seems to not do anything.

JIL: https://app.saucelabs.com/tests/07691a507e324dd3b840786878a178c7
WDIO: https://app.saucelabs.com/tests/c92821cb75154097bda1f8752ee3f414

Notice in the JIL result, the body updates with the text `PDF Generated`. I updated the asset file to append that text after jsPDF was done running. It never gets appended in WDIO. I have tried also used `browser.execute` and attaching the function to the window. There is just something about how jsPDF works that breaks wdio/appium/xcuitest/w3c.

bjfield (approved):

metal-messiah (approved):

PR #593, opened 2023-06-21
chore: Fix inconsistent IE issues with err/attributes.e2e.js

bjfield:
Intermittently WDIO was failing in IE to capture the ID of the selector used to trigger clicks.
---

### Overview

The error was `Error: Malformed type for "elementId" parameter of command elementClick` and it affected a number of the tests intermittently. The new logic mostly uses browser.execute to interact directly with the scripts and elements.

Some of the tests didn't actually need to use a click event because they weren't testing a SPA click interaction.

### Related Issue(s)

N/A

### Testing

`npm run wdio -- --retry=false tests/specs/err/attributes.e2e.js -P -b ie@11`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9687,9 +9687,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001505",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001505.tgz",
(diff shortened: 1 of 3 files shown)


patrickhousley:
I am seriously questioning if we should try to avoid running code with `.click()` and just use `execute` to run a script attached on the window within the asset.

patrickhousley (approved):
[Tests](https://github.com/newrelic/newrelic-browser-agent/actions/runs/5350565926/jobs/9703328542#step:8:191) passing.

metal-messiah (approved):

PR #594, opened 2023-06-22
fix: Fix illegal invocation error on final harvest

patrickhousley:
An issue from 1.234.0 has been addressed in which an illegal invocation error could occur during a final harvest as a result of calling `addPageAction` and passing `window.location` as an argument before a navigation.
---

### Overview

Reverting a change from #521. Calling `addPageAction` and passing `window.location` as an argument before a navigation would result in calling `.toString` on the global `window.location`. Doing so during a page unload causes an `Illegal Invocation` error due to the global `window.location.toString()` losing it's context.

### Related Issue(s)

https://issues.newrelic.com/browse/NEWRELIC-9370

### Testing

Test added to ensure this issue does not happen again.

### DIFF
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -244,13 +244,9 @@ export class Harvest extends SharedContext {
       }
       return Object.entries(input || {})
         .reduce((accumulator, [key, value]) => {
-          if (
-            value !== null &&
-            value !== undefined && (
-              (typeof value === 'number') ||
+          if ((typeof value === 'number') ||
               (typeof value === 'string' && value.length > 0) ||
-              (Object.entries(value).length > 0)
-            )
+              (typeof value === 'object' && Object.keys(value || {}).length > 0)
           ) {
             accumulator[key] = value
           }
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -244,7 +244,14 @@ export class Harvest extends SharedContext {
       }
       return Object.entries(input || {})
         .reduce((accumulator, [key, value]) => {
-          if (value !== null && value !== undefined && value.length) {
+          if (
+            value !== null &&
+            value !== undefined && (
+              (typeof value === 'number') ||
+              (typeof value === 'string' && value.length > 0) ||
+              (Object.entries(value).length > 0)
+            )
+          ) {
             accumulator[key] = value
           }
 
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -59,7 +59,7 @@ export class Harvest extends SharedContext {
   send (spec = {}) {
     const caller = this.obfuscator.shouldObfuscate() ? this.obfuscateAndSend.bind(this) : this._send.bind(this)
 
-    return caller({ ...spec, payload: this.cleanPayload(spec.payload) })
+    return caller(spec)
   }
 
   /**
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -244,7 +244,7 @@ export class Harvest extends SharedContext {
       }
       return Object.entries(input || {})
         .reduce((accumulator, [key, value]) => {
-          if (value !== null && value !== undefined && value.toString()?.length) {
+          if (value !== null && value !== undefined && value.length) {
             accumulator[key] = value
           }
 
--- a/src/common/harvest/harvest.test.js
+++ b/src/common/harvest/harvest.test.js
@@ -304,9 +304,7 @@ describe('_send', () => {
 
   test('should set body to events when endpoint is events', () => {
     spec.endpoint = 'events'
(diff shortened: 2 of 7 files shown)


src/common/harvest/harvest.test.js
```diff
@@ -304,9 +304,7 @@ describe('_send', () => {
 
   test('should set body to events when endpoint is events', () => {
     spec.endpoint = 'events'
-    spec.payload.body.e = {
-      [faker.datatype.uuid()]: faker.lorem.sentence()
-    }
+    spec.payload.body.e = faker.lorem.sentence()
```
patrickhousley, line 307: This test was the reason for the original change. However, `body.e` will always be a string so I changed the test to pass.

tests/assets/api/addPageAction-unload.html
```diff
@@ -0,0 +1,24 @@
...
+</head>
+
+<body>
+  <div id="content">
+    This page is used to call `addPageAction` immediately followed by a navigation.
+  </div>
+  <a id="btn" href="#btn">Click Me</a>
+  <script>
+    window.triggerPageActionNavigation = function () {
+      newrelic.addPageAction('pageaction', window.location)
```
metal-messiah, line 18: We should probably consider enforcing that objects passed to pageAction API be shallow k:v obj of [string]: string|number|boolean|null|undefined

metal-messiah (commented):
Excellent job diagnosing this

tests/specs/ins/harvesting.e2e.js
```diff
@@ -0,0 +1,22 @@
+describe('ins harvesting', () => {
+  it('NEWRELIC-9370: should not throw an exception when calling addPageAction with window.location before navigating', async () => {
+    const testUrl = await browser.testHandle.assetURL('api/addPageAction-unload.html')
+    await browser.url(testUrl)
+      .then(() => browser.waitForAgentLoad())
+
+    const [pageActionsHarvest] = await Promise.all([
+      browser.testHandle.expectIns(),
+      browser.testHandle.expectErrors(10000, true),
```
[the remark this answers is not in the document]
patrickhousley: It's an added precaution. If you revert the code change, this will actually capture the invocation error keying you into whats actually happening.

tests/assets/api/addPageAction-unload.html
patrickhousley: I agree. Something we can take into consideration when redesigning harvest.

patrickhousley: I put in a placeholder ticket for now: https://issues.newrelic.com/browse/NEWRELIC-9438

tests/specs/ins/harvesting.e2e.js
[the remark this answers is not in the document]
metal-messiah: Not needed for this PR but we need a good system for handling internal agent errors so they don't pollute customer streams as much. Possibly need to utilize ierr more as well as consume it more downstream

metal-messiah (approved):

src/common/harvest/harvest.js
cwli24, line 1: There's an extra `cleanPayload` call on L62 of this file btw, which is redundant.

src/common/harvest/harvest.js
```diff
@@ -244,7 +244,14 @@ export class Harvest extends SharedContext {
...
         .reduce((accumulator, [key, value]) => {
-          if (value !== null && value !== undefined && value.toString()?.length) {
+          if (
+            value !== null &&
+            value !== undefined && (
+              (typeof value === 'number') ||
+              (typeof value === 'string' && value.length > 0) ||
+              (Object.entries(value).length > 0)
+            )
+          ) {
```
cwli24, line 254:
```suggestion
          if ((typeof value === 'number') ||
              (typeof value === 'string' && value.length > 0) ||
              (typeof value === 'object' && Object.keys(value).length > 0)) {
```

patrickhousley: `typeof null === 'object'` ;)

src/common/harvest/harvest.js
patrickhousley: removed 👍

src/common/harvest/harvest.js
metal-messiah: more of a philosophical question here... but **should** payloads that are nullish even get to the harvest?  Seems like it would make more sense to filter those out at the point of creation (the feature).

patrickhousley: if features have nothing to harvest, they shouldn't send anything to the harvest module

metal-messiah: i mean yes, but this also handles the case where payloads are valid but some fields in the payload have empty strings or null or whatever.  But should that be handled in the feature?

cwli24 (approved):

Issue #599, opened 2023-06-23
UnhandledRejection error while loading the agent 1.235.0


ilario-pierbattista:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

Starting from yesterday evening (2023-06-22 17:00 UTC), I'm getting the following error while loading the agent in browsers with strict privacy settings:

> UnhandledRejection
Non-Error promise rejection captured with value: undefined

It's similar to #392. In the browser console I see chunks loading failures, and then the error `Uncaught (in promise)`.

The agent is injected dinamically in HTML by the PHP agent.

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

## Steps to Reproduce

Use a browser with strict privacy settings. It will block the requests to `https://js-agent.newrelic.com`.

[NOTE]: # ( Please be as specific as possible. )

## Expected Behavior

Even if the browser blocks the agent download I expect to get only warnings in console. Rejections should be handled.

[NOTE]: # ( Tell us what you expected to happen. )

## Relevant Logs / Console output

Sample warnings:
```
New Relic: Downloading runtime APIs failed... 
New Relic: A problem occurred when starting up session manager. This page will not start or extend any session.
  ChunkLoadError: Loading chunk 860 failed.
  (error: https://js-agent.newrelic.com/860.e997904f-1.235.0.min.js)
New Relic: Downloading and initializing page_view_event failed...
  ChunkLoadError: Loading chunk 412 failed.
  (error: https://js-agent.newrelic.com/lazy-feature-loader.0ba331d7-1.235.0.min.js)
[...] and so on
```

Error:
```
Uncaught (in promise) undefined [A:6:32604](https://website/page)
    a https://website/page:6
    InterpretGeneratorResume self-hosted:1359
    AsyncFunctionThrow self-hosted:760
    (Asinc.: async)
    importAggregator https://website/page:6
    nrWrapper https://website/page:6
    (Asinc.: EventListener.handleEvent)
    nrWrapper https://website/page:6
    s https://website/page:6
    o https://website/page:6
    importAggregator https://website/page:6
    F https://website/page:6
    start https://website/page:6
    forEach self-hosted:203
    start https://website/page:6
    <anonima> https://website/page:6
    <anonima> https://website/page:6
    <anonima> https://website/page:6
    <anonima> https://website/page:6
```

[...]

PR #603, opened 2023-06-23
fix: Handle chunk load promise error introduced in 1.235.0 - NR-129244

cwli24:
Starting in 1.235.0, loading the agent in browsers with strict privacy settings resulted in an uncaught error in a promise. This error has been handled, and chunk load failures again result only in a warning.
---

### Overview

Eliminate promise rejection.

### Related Issue(s)

NR-129244 // #599 

### Testing

Verify that when one of the feature's aggregate import is blocked & fails, no uncaught promise is thrown.

### DIFF
--- a/src/features/session_replay/replay-mode.js
+++ b/src/features/session_replay/replay-mode.js
@@ -15,8 +15,8 @@ export async function getSessionReplayMode (agentId) {
     const newrelic = gosNREUM()
     // Should be enabled by configuration and using an agent build that includes it (via checking that the instrument class was initialized).
     if (getConfigurationValue(agentId, 'session_replay.enabled') && typeof newrelic.initializedAgents[agentId].features.session_replay === 'object') {
-      const success = await newrelic.initializedAgents[agentId].features.session_replay.onAggregateImported // if Replay could not initialize, this throws a (uncaught) rejection
-      if (success) return await sharedChannel.sessionReplayInitialized // wait for replay to determine which mode it's after running its sampling logic
+      const srInitialized = await newrelic.initializedAgents[agentId].features.session_replay.onAggregateImported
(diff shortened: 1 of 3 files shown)


src/features/session_replay/replay-mode.js
```diff
@@ -15,8 +15,8 @@ export async function getSessionReplayMode (agentId) {
     const newrelic = gosNREUM()
     // Should be enabled by configuration and using an agent build that includes it (via checking that the instrument class was initialized).
     if (getConfigurationValue(agentId, 'session_replay.enabled') && typeof newrelic.initializedAgents[agentId].features.session_replay === 'object') {
-      await newrelic.initializedAgents[agentId].features.session_replay.onAggregateImported // if Replay could not initialize, this throws a (uncaught) rejection
-      return await sharedChannel.sessionReplayInitialized // wait for replay to determine which mode it's after running its sampling logic
+      const srInitialized = await newrelic.initializedAgents[agentId].features.session_replay.onAggregateImported
+      if (srInitialized) return await sharedChannel.sessionReplayInitialized // wait for replay to determine which mode it's after running its sampling logic
```
patrickhousley, line 19: What happens when `srInitialized` is `false`? This should probably have a default return.

cwli24: It bleeds down into the `return MODE.OFF` like the other 2 checks

patrickhousley: Not if an exception is not thrown. Right.

patrickhousley: NVM i can't read.

patrickhousley (approved):

bjfield (approved):

PR #609, opened 2023-07-05
feat: Stop reporting ajax events going to same beacon

cwli24:
Network requests made to the same hostname as the agent's reporting beacon endpoint are no longer captured, by default. As an example, outgoing XHRs to the `bam.nr-data.net` URL where analytic data is typically sent would not reflectively record itself.
---

### Overview

Primary purpose is to make "internal URL" reporting dynamic, so that when the beacon url is changed each time, the `deny_list` setting doesn't also have to change to retain the block. In other words, it's no longer necessary to include `bam-cell.nr-data.net`, etc. in the deny_list to prevent agent from reporting its own generated traffic by simply toggling the flag.

This also helps facilitate users who don't care about the agent's own reporting from showing on their NR1 dashboard and from being billed for such data.

### Related Issue(s)

NR-99149, NR-136392
Bug: NR-140471

### Testing

Verify that ajax events are not sent for network requests going to any domain that `info.beacon` or `info.errorBeacon` is also set to, when `init.ajax.block_internal = true` (default), but they then are when the flag is flipped to be `false`.

Also check that whether the initial `init.ajax.deny_list` is undefined or a preset list doesn't affect this behavior.


src/features/ajax/aggregate/index.js
```diff
@@ -19,15 +19,23 @@ export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
   constructor (agentIdentifier, aggregator) {
     super(agentIdentifier, aggregator, FEATURE_NAME)
+    const agentInit = getConfiguration(agentIdentifier)
+    const allAjaxIsEnabled = agentInit.ajax.enabled !== false
+
+    register('xhr', storeXhr, this.featureName, this.ee)
+    if (!allAjaxIsEnabled) return // feature will only collect timeslice metrics & ajax trace nodes if it's not fully enabled
```
patrickhousley, line 26: Was this change for session trace or replay?

patrickhousley: Should the return be after the denyList set?

tools/testing-server/test-handle.js
```diff
@@ -265,7 +265,8 @@ module.exports = class TestHandle {
           loader: 'full',
           config: {
             licenseKey: this.#testId
-          }
+          },
+          init: { ajax: { block_internal: false } }
```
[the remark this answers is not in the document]
patrickhousley: I think this is causing the test failure in WDIO:

```
2023-07-10T17:50:05.5808462Z [chrome 114.0.5735.91 windows #0-32] Running: chrome (v114.0.5735.91) on windows
2023-07-10T17:50:05.5809066Z [chrome 114.0.5735.91 windows #0-32] Session ID: e48715854c0c4d68b3fea71fa2e431f9
2023-07-10T17:50:05.5809523Z [chrome 114.0.5735.91 windows #0-32]
2023-07-10T17:50:05.5810045Z [chrome 114.0.5735.91 windows #0-32] » /tests/specs/xhr/deny-list.e2e.js
2023-07-10T17:50:05.5810597Z [chrome 114.0.5735.91 windows #0-32] Ajax events to beacon endpoint
2023-07-10T17:50:05.5811305Z [chrome 114.0.5735.91 windows #0-32]    ✖ not captured when blocked
2023-07-10T17:50:05.5811843Z [chrome 114.0.5735.91 windows #0-32]    ✖ not captured when blocked
2023-07-10T17:50:05.5812370Z [chrome 114.0.5735.91 windows #0-32]    ✖ not captured when blocked
2023-07-10T17:50:05.5812979Z [chrome 114.0.5735.91 windows #0-32]    ✖ not captured when blocked
2023-07-10T17:50:05.5813565Z [chrome 114.0.5735.91 windows #0-32]    ✓ captured when allowed
2023-07-10T17:50:05.5813993Z [chrome 114.0.5735.91 windows #0-32]
2023-07-10T17:50:05.5814451Z [chrome 114.0.5735.91 windows #0-32] 1 passing (52.9s)
2023-07-10T17:50:05.5814893Z [chrome 114.0.5735.91 windows #0-32] 1 failing
```

src/features/ajax/aggregate/index.js
cwli24: Need to add test that disabled ajax still harvest metrics & ajax trace nodes for sure

cwli24: Oh so race between deny list being set and `storeXhr` doesn't matter because storeXhr would still early return on init config (not subject to race) and not execute any code dependent on deny list.

cwli24:
https://github.com/newrelic/newrelic-browser-agent/actions/runs/5478559213?pr=609

src/features/ajax/aggregate/index.js
```diff
@@ -19,15 +19,23 @@ export class Aggregate extends AggregateBase {
...
+
+    const agentInfo = getInfo(agentIdentifier)
+    let denyList = agentInit.ajax.block_internal ? (agentInit.ajax.deny_list || []).concat(agentInfo.beacon, agentInfo.errorBeacon) : agentInit.ajax.deny_list
+    setDenyList(denyList)
```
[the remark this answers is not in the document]
metal-messiah: Agree

src/features/ajax/aggregate/index.js
metal-messiah: Is this going to break the drain behavior since this can return before invoking drain()

cwli24: ^Oh yeah, thanks forgot about that drain line at the bottom. Yeah need to drain before early exit fs.

cwli24: Verified that if ajax enabled config is false, it won't be imported at all (since at least v1220) hence no need to test for timeslice & ajax nodes when it's disabled.

metal-messiah:
Lets merge with main and run all wdio tests, but otherwise looks good

metal-messiah (approved):

tests/specs/xhr/deny-list.e2e.js
```diff
@@ -1,6 +1,6 @@
 describe('Ajax events to beacon endpoint', () => {
   it('not captured when blocked', async () => {
-    let url = await browser.testHandle.assetURL('spa/ajax-deny-list.html') // requests going to beacon domain URL not collected by default
+    let url = await browser.testHandle.assetURL('spa/ajax-deny-list.html', { init: { ajax: { block_internal: true } } })
```
metal-messiah, line 3: Does this mean that the default behavior (block internal) is not working?

metal-messiah: nvm, looks like this is just affecting internal testing default behavior

metal-messiah (approved):

PR #610, opened 2023-07-06
fix: Fix potential property of undefined errors

cwli24:
Patch a couple of areas that can yield reading property of undefined or null errors in the RUM call and jsonp wrapper.
---

### Overview

See summary

### Related Issue(s)

NR-122092
NR-138554

### Testing

All regression

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9687,9 +9687,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001509",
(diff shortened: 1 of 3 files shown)


patrickhousley (approved):
The initialized features issue may be a race condition between the page load event and the JS fully instantiating all the feature instrument classes.

For the jsonp issue, we probably need to discuss our continued support for jsonp. The callback regex is not going to always work since the callback query parameter name could be anything.

PR #614, opened 2023-07-13
feat: reduce jserrors wrapping and remove onerror use

patrickhousley:
Replaces the use of the `onerror` global function with `addEventListener` to capture errors that go unhandled by customer code. This does not change how unhandled promise rejects are captured. Additional wrappings for events, timers, and animation fames have been removed in favor of capturing unhandled errors using a global event listener. 
---

### Overview

This PR removes wrappings for events, timers, and animation frames from the jserrors feature. Those wrappings are not specifically necessary unless another feature (spa) needs and performs the wrapping.

Errors are now primarily captured using a global event handler for `error` and `unhandledrejection` events. These events are captured and, if necessary, converted to Error-like objects that can be processed by the aggregator. This removes the use of the global `onerror` function and removes the need for customer's to not override our `onerror` function. 

The jserrors feature retains a listener on the event emitter for `fn-err` events. In the event the spa feature is loaded, this handler will capture the error earlier than the global error handlers allowing for the spa feature to properly decorate the error with interaction data.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-138375

### Testing

All the test cases should still pass with this change.


src/features/jserrors/instrument/index.js
```diff
@@ -5,64 +5,59 @@
...
+      if (!this.abortHandler) return
+      handle('ierr', [this.#castError(error), now(), true], undefined, FEATURE_NAMES.jserrors, this.ee)
     })
 
-    // Replace global error handler with our own.
-    this.origOnerror = globalScope.onerror
-    globalScope.onerror = this.onerrorHandler.bind(this)
+    globalScope.addEventListener('unhandledrejection', (promiseRejectionEvent) => {
+      if (!this.abortHandler) return
+      if (this.#seenErrors.has(promiseRejectionEvent.reason)) {
```
metal-messiah, line 44: Does I didn't think fn-err fired in the promise wrapper.  Is this validated?

metal-messiah: if yes and it ends up getting sent as an error in the fn-err handler, its not getting marked as a promise exception

patrickhousley: Verified the promise wrapper does not use the `fn-err` event so unhandled promise rejections will get sent as such.
image

patrickhousley: I guess technically the `seenErrors` bit here is not necessary. I will remove it.

patrickhousley:
[Tests](https://github.com/newrelic/newrelic-browser-agent/actions/runs/5546582910) pass with expected safari 15 errors.

metal-messiah (approved):

src/features/jserrors/instrument/index.js
```diff
@@ -5,64 +5,55 @@
 
 import { handle } from '../../../common/event-emitter/handle'
 import { now } from '../../../common/timing/now'
-import { getOrSet } from '../../../common/util/get-or-set'
-import { wrapRaf, wrapTimer, wrapEvents, wrapXhr } from '../../../common/wrap'
-import './debug'
 import { InstrumentBase } from '../../utils/instrument-base'
-import { FEATURE_NAME, NR_ERR_PROP } from '../constants'
+import { FEATURE_NAME } from '../constants'
```
cwli24, line 9: Can you remove NR_ERR_PROP from the constants file too?

src/features/jserrors/instrument/index.js
```diff
@@ -71,67 +62,55 @@ export class Instrument extends InstrumentBase {
...
   }
 
+  #castError (error) {
+    if (error instanceof Error) {
+      return error
+    }
+
+    if (typeof error.message !== 'undefined') {
+      return new UncaughtError(error.message, error.filename, error.lineno, error.colno)
+    }
```
cwli24, line 76: Is there a case when this is called with an error that's not an Error instance yet has a message?
I see `fn-err` and all `internal-error` events emitted with an actual error object from try-catch.

src/features/jserrors/instrument/index.js
```diff
@@ -71,67 +62,55 @@ export class Instrument extends InstrumentBase {
...
-
-/**
- * Attempts to cast an unhandledPromiseRejection reason (reject(...)) to an Error object
- * @param {*} reason - The reason property from an unhandled promise rejection
- * @returns {Error} - An Error object with the message as the casted reason
- */
-function castReasonToError (reason) {
-  let prefix = 'Unhandled Promise Rejection: '
-  if (reason instanceof Error) {
+    if (typeof promiseRejectionEvent.reason === 'undefined') return new Error(prefix)
```
cwli24, line 96: Should keep in mind that when we create a new Error, the stack trace is going to point back to this code (us), so users will likely first come to us saying there's some error being thrown in our code. 

That said, if they call `reject()` in promise without an arg, should this be more descriptive than just the prefix string?

src/features/jserrors/instrument/index.js
patrickhousley: `throw` can be passed any value.

src/features/jserrors/instrument/uncaught-error.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Represents an uncaught non Error type error. This class does
+ * not extend the Error class to prevent an invalid stack trace
+ * from being created. Use this class to cast thrown errors that
+ * do not use the Error class (strings, etc) to an object.
+ */
+export class UncaughtError {
+  constructor (message, filename, lineno, colno) {
+    this.name = 'UncaughtError'
```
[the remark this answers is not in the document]
patrickhousley: What name would you suggest?

src/features/jserrors/instrument/index.js
patrickhousley: This code was not changed, just moved. That said, I also questioned why we were creating new Error instances here for the same reason. I thought about changing this to use the `UncaughtError` class.

PR #615, opened 2023-07-13
chore: update pr workflow

patrickhousley:
Workflow updates to provide faster feedback on pull requests, cleanup nightly jobs, and make re-usable workflows for jil testing.
---

### Overview

Workflow updates:

- Pull Request Checks
  - Jobs not needing secrets have been moved to be ran on the `pull_request` event. The remaining jobs will run on manual dispatch.
  - Updated the `WDIO Coverage` and `WDIO` jobs so they run in parallel. Keeping the wdio coverage execution in a separate job so it runs fast and isolated.
  - Added jil unit and functional tests to run after all wdio tests have finished.
- JIL: All Browsers
  - New reusable workflow that calls `JIL: Single Browser` for each browser and collection (unit, functional).
- JIL: Single Browser
  - New reusable workflow that runs a single collection of jil tests for a specified browser.
- Main Push
  - Updated workflow to prevent it from running multiple times simultaneously. If multiple PRs are merged back to back to `main`, only the workflow start for the last merge will finish executing. All others will be cancelled.
- Nightly
  - Added jil smoke and polyfill jobs to this workflow and deleted the old workflow files.
- WDIO: All browsers
  - Fixed the ref inputs not passing through when using workflow call

### Related Issue(s)

https://issues.newrelic.com/browse/NR-137546

### Testing

N/A


metal-messiah (approved):
Nice job with this! -- Approved, just don't forget to update the PR description

PR #617, opened 2023-07-14
fix: prevent invalid error stack traces

patrickhousley:
Removing the instantiation of the Error class in the jserrors feature for instances where a non-Error value is thrown from customer's code. In such cases, an Error instance would be created and the stack trace would point back to the agent code as the source. This causes confusion since it makes it seem like the agent code is throwing the error when that is not the case. Instead, the internal UncaughtError class will be used and no stack trace will exist.
---

### Overview

Addressing PR comments from https://github.com/newrelic/newrelic-browser-agent/pull/614.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-138375

### DIFF
--- a/src/features/jserrors/constants.js
+++ b/src/features/jserrors/constants.js
@@ -1,4 +1,3 @@
 import { FEATURE_NAMES } from '../../loaders/features/features'
 
 export const FEATURE_NAME = FEATURE_NAMES.jserrors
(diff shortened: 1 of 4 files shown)


cwli24 (commented):
Some of the tests need re-adjusting

patrickhousley:
Verified the Safari 15 test failures were due to known SL Safari15 issues.

PR #621, opened 2023-07-18
fix: Fix serialization issue with array data

metal-messiah:
Update the harvester to account for array payloads, which in niche cases could convert array payloads to indexed objects.
---

### Overview
This PR updates the cleanPayload mechanics to account for array payloads.

### Related Issue(s)
[NR-143331
](https://issues.newrelic.com/browse/NR-143331)

### Testing
Session replay wdio tests + harvest.test.js should continue to pass


src/common/harvest/harvest.js
```diff
@@ -239,9 +239,8 @@ export class Harvest extends SharedContext {
    */
   cleanPayload (payload = {}) {
     const clean = (input) => {
-      if ((typeof Uint8Array !== 'undefined' && input instanceof Uint8Array) || typeof input === 'string') {
-        return input.length > 0 ? input : null
-      }
+      if ((typeof Uint8Array !== 'undefined' && input instanceof Uint8Array) || Array.isArray(input)) return input
+      if (typeof input === 'string') return input.length > 0 ? input : null
```
patrickhousley, line 243: Was there a case where `input.length > 0` was not working for the other data types? Just wondering why string got split to its own line.

metal-messiah: I forget at this point but I think it was because returning null for the array if 0 length was breaking something. We could look deeper

patrickhousley: This is fine. If `input` is an empty array, it would return `null` which gets stringified to `'null'` in the `_send` method bypassing the empty check on line 120.

patrickhousley (approved):

metal-messiah:
Circle back and diagnose IE11 issues on unit tests (JIL)

PR #624, opened 2023-07-21
fix: Capture potential missing tail data for session trace

cwli24:
Send buffered session trace data in certain cases when it's being discarded at the tail end of a captured session.
---

### Overview

In the old session trace behavior, it had discarded or not sent remaining session trace nodes when it hit the max duration life of the feature, even though the nodes were added before the max duration time. Likewise, on EoL of the page, session trace would not run final harvest if the buffered nodes is less than the required minimum of 30, which causes that valid data to be amiss. This PR patches both of those issues.

### Related Issue(s)

NR-138555

### Testing

Workflows listed in bug ticket's AC list should hold true, namely:

- trace sends final harvest of < 30 nodes when in standalone
- if there are any additional nodes buffered between the max duration point of time and its immediately preceding interval harvest, those are also sent, not discarded


src/features/session_trace/aggregate/index.js
```diff
@@ -145,6 +148,14 @@ export class Aggregate extends AggregateBase {
     }
     /* --- EoS --- */
 
+    setTimeout(() => {
```
patrickhousley, line 151: Hmm...couldn't we do this in the prepareHarvest method instead of creating a new timeout?

cwli24: Initially tried but it became really complicated because of how many factors are at play here. Like you'd have to consider the "final" or remaining nodes harvest and that `#prepareHarvest` is triggered on regular interval harvests and `subscribeToEOL`. It becomes a lot simpler to separate this logic from having to think of when and how `#prepareHarvest` is called.

src/features/session_trace/aggregate/index.js
```diff
@@ -189,16 +193,16 @@ export class Aggregate extends AggregateBase {
   }
 
   #prepareHarvest (options) {
-    /* Standalone refers to the legacy version of ST before the idea of 'session' or the Replay feature existed.
-      It has a different behavior on returning a payload for harvest than when used in tandem with either of those concepts. */
     if (this.isStandalone) {
-      if (now() > MAX_TRACE_DURATION) { // been collecting for over the longest duration we should run for, empty trace object so ST has nothing to send
-        this.#scheduler.stopTimer()
-        this.trace = {}
+      if (this.ptid && now() >= MAX_TRACE_DURATION) {
```
cwli24, line 197: (In practice this isn't an actual issue, but if the purpose is to last send buffered data, `ptid` isn't really required. Though edge & theoretical case, `#prepareHarvest` does apply to the first `runHarvest` call too, wherein `ptid` is undefined and subsequently fetched.)

patrickhousley:
I verified the safari tests that failed in github actions are passing locally. They appear to just be flakey.

PR #626, opened 2023-07-25
[WIP] Basic spa poc

metal-messiah:
---


src/common/aggregate/aggregator.js
```diff
@@ -82,12 +84,36 @@ export class Aggregator extends SharedContext {
...
       if (results[type].length) hasData = true
       delete this.aggregatedData[type]
     }
     return hasData ? results : null
   }
+
+  toArray (type, obj) {
+    if (typeof obj !== 'object') return []
+
+    return mapOwn(obj, (key, value) => this.getValue(key, value, type))
```
patrickhousley, line 98: The `key` parameter is never used and I would like to continue removing uses of `mapOwn`

```suggestion
    return Object.values(obj).map(value => this.getValue(value, type))
```

src/common/aggregate/aggregator.js
```diff
@@ -82,12 +84,36 @@ export class Aggregator extends SharedContext {
...
+  }
+
+  getValue (key, value, type) {
```
patrickhousley, line 101:
```suggestion
  getValue (value, type) {
```

src/features/basic_spa/instrument/index.js
```diff
@@ -0,0 +1,60 @@
...
+    const eventsEE = wrapEvents(this.ee)
+
+    const trackURLChange = (args) => handle('newURL', [now(), '' + window.location, args?.type], undefined, this.featureName, this.ee)
+
+    historyEE.on('pushState-end', trackURLChange)
+    historyEE.on('replaceState-end', trackURLChange)
+
+    windowAddEventListener('hashchange', trackURLChange, true, this.removeOnAbort?.signal)
+    windowAddEventListener('load', trackURLChange, true, this.removeOnAbort?.signal)
+    windowAddEventListener('popstate', trackURLChange, true, this.removeOnAbort?.signal)
```
patrickhousley, line 30: Do we need both the history wrappers on lined 25-26 and this event listener?

metal-messiah: These are ports over from the original spa feature, needs eval as to what is actually needed there

src/common/timing/time-to-interactive.js
```diff
@@ -0,0 +1,32 @@
...
+  #longTasks = 0
+  constructor ({ startTimestamp, buffered = false }) {
+    this.#startTimestamp = startTimestamp
+    this.#ltUnsub = longTask.subscribe(({ name, value, attrs }) => {
+      this.#latestLT = attrs.ltStart + value
+      this.#longTasks++
+      this.#setLtTimer()
+    }, buffered)
+
+    return new Promise((resolve, reject) => {
```
patrickhousley, line 19: Don't use return in a constructor. I'm honestly surprised eslint didn't complain about this.

src/features/basic_spa/aggregate/index.js
```diff
@@ -0,0 +1,92 @@
...
+    this.interactionInProgress = new Interaction(this.agentIdentifier)
+    this.interactionInProgress.on('finished', this.completeInteraction.bind(this))
+    this.interactionInProgress.on('cancelled', this.cancelInteraction.bind(this))
+    if (trigger) this.interactionInProgress.trigger = trigger
+    if (category) this.interactionInProgress.category = CATEGORY.ROUTE_CHANGE
+    if (startedAt) this.interactionInProgress.start = startedAt
+  }
+
+  cancelInteraction () {
+    this.interactionInProgress = null
```
patrickhousley, line 77: I think there is a potential here for a memory leak. If an interaction is cancelled, it needs to cleanup its `TimeToInteractive` subscription.

src/common/timing/time-to-interactive.js
metal-messiah: something feels gross about an empty constructor with the same logic in a class method but fine :)

patrickhousley: You could use a static factory method and a private constructor.

patrickhousley: Oh yea, thats a TS thing. lol .... still could use a factory method.

patrickhousley: Looking through the code, it looks like you only use the `.then` method and don't actually care that it's a promise.....technically you could just have a public then method.

src/common/aggregate/aggregator.js
```diff
@@ -82,12 +84,36 @@ export class Aggregator extends SharedContext {
...
+    // err type can be decorated with an interaction ID if present
+    if (type !== 'err') return value
+    const shouldHold = this.checkInteractions(value) // add an interaction ID if exists
+    if (shouldHold) return
+    return value
+  }
+
+  checkInteractions (value) {
+    const spaFeature = getFeatureState({ agentIdentifier: this.sharedContext.agentIdentifier, featureName: FEATURE_NAMES.basicSpa })
+    const { shouldHold, interactions } = spaFeature?.hasInteraction?.({ timestamp: value?.metrics?.time?.t }) || {}
```
cwli24, line 111: Where is this `shouldHold`? the .hasInteraction only ever returns an object with just a single interactions property.

src/features/spa/aggregate/index.js
```diff
@@ -720,6 +720,8 @@ export class Aggregate extends AggregateBase {
       }
       baseEE.emit('interactionSaved', [interaction])
       state.interactionsToHarvest.push(interaction)
+
```
cwli24, line 723: Reminder to delete

src/features/page_view_event/aggregate/index.js
```diff
@@ -109,6 +109,7 @@ export class Aggregate extends AggregateBase {
           activateFeatures(JSON.parse(responseText), this.agentIdentifier)
           this.drain()
         } catch (err) {
+          console.error(err)
```
cwli24, line 112: Reminder to delete

src/features/utils/lazy-feature-loader.js
```diff
@@ -32,6 +32,8 @@ export function lazyFeatureLoader (featureName, featurePart) {
         return import(/* webpackChunkName: "session_trace-aggregate" */ '../session_trace/aggregate')
       case FEATURE_NAMES.spa:
```
cwli24, line 33: Likewise can take old spa out

src/features/basic_spa/aggregate/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+  completeInteraction (ixn) {
+    if (!ixn) return
+    this.interactionsToHarvest.push(ixn)
+    this.interactionInProgress = null
+    this.scheduler.scheduleHarvest(0.1)
+  }
+
+  hasInteraction ({ timestamp }) {
+    if (!timestamp) return { interaction: undefined }
```
cwli24, line 88: I think this is meant to return `interactions`? Can you align function name & property singular or plural?

src/features/ajax/aggregate/index.js
```diff
@@ -119,11 +106,19 @@ export class Aggregate extends AggregateBase {
...
-        var interactionId = this.spaNode.interaction.id
-        spaAjaxEvents[interactionId] = spaAjaxEvents[interactionId] || []
-        spaAjaxEvents[interactionId].push(event)
+      const spaFeature = getFeatureState({ agentIdentifier, featureName: FEATURE_NAMES.basicSpa })
+      const {
+        interactions
+      } = spaFeature?.hasInteraction?.({ timestamp: event.startTime }) || {}
+
+      // if the ajax happened inside an ixn window (found a match), add it to the ixn
+      if (interactions?.length) {
```
cwli24, line 115: What if there is more than 1 interaction that's returned? How can it make sense that there's more than 1 interaction at a time and that this ajax is a child of every one (multiple)?

cwli24 (commented):
So while this is less problem prone than the existing spa, it could use a big sweeping.

Overall, a lack of jsdoc, documentation behind the process, and unclear semantics make this hard to trace through. For example, on one line, I'm reading about "interactions" -> "ixn" to have it suddenly use a function that's within the `BelNode` class -- is an interaction simply a bel-serializer node or is it made up of nodes? It's confusing and unclear.
The way things are passed around is also reminiscent of a top issue with writing/reading/debugging javascript code with dynamic types: I'm getting lost on what a function should pass back and both the type and available properties of objects.

The other part of the confusion-complication is the way things are structured. For example, the belnode is responsible for `cancel()`-ing itself. Something that should be a simple data structure is sharing the logic/responsibility of its handler.

src/common/timing/now.js
```diff
@@ -5,5 +5,5 @@
 
 // This is our own layer around performance.now. It's not strictly necessary, but we keep it in case of future mod-ing of the value for refactor purpose.
 export function now () {
-  return Math.round(performance.now())
+  return Math.floor(performance.now())
```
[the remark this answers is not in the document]
metal-messiah: Screenshot 2023-09-15 at 9 46 18 AM

src/common/timing/time-to-interactive.js
```diff
@@ -0,0 +1,38 @@
...
+  cancel () {
+    clearTimeout(this.#ltTimer)
+    this.#ltUnsub()
+    // this.#rejector()
+  }
+
+  #setLtTimer () {
+    clearTimeout(this.#ltTimer)
+    this.#ltTimer = setTimeout(() => {
+      this.#resolver({ value: Math.round(Math.max(this.#startTimestamp, this.#latestLT)), attrs: { startTimestamp: this.#startTimestamp, longTasks: this.#longTasks } })
```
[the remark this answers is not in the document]
metal-messiah: essentially to accommodate for either being started on the fly, or representing a past time, ie from the page load origin time

src/features/ajax/aggregate/index.js
metal-messiah: it could be a child of both a page load interaction and a click interaction if both are happening simultaneously while the ajax occurs under this model, since its timestamp-based alone.

PR #631, opened 2023-08-01
fix: prevent dollar symbol only variable names

patrickhousley:
Prevent the build process of the agent from creating minified variable names that only contain the dollar symbol. If such a variable is immediately followed by an ampersand, as could be the case in a minified logic block, it could be picked up by  string replacement logic applied to outgoing HTML within their server.
---

### Overview

Prevent the build process of the agent from creating minified variable names that only contain the dollar symbol. If such a variable is immediately followed by an ampersand, as could be the case in a minified logic block, it could be picked up by  string replacement logic applied to outgoing HTML within their server.

In the one case we saw, it appears the customer is either performing their own injection logic or performing some custom string replacement or interpolation on the HTML after the APM agent has injected the browser agent loader.

If this needs to be expanded upon, a more robust solution would be to override the `nth_identifier` setting in the terser plugin settings:

```
nth_identifier: {
  get: (numberOfCharacters) => {
    const chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'.split('')
    let base = 52 // The first character of a variable name must be alpha
    let result = ''
    numberOfCharacters++
    do {
      numberOfCharacters--
      result += chars[numberOfCharacters % base]
      numberOfCharacters = Math.floor(numberOfCharacters / base)
      base = 62
    } while (numberOfCharacters > 0)
    return result
  }
}
```

### Related Issue(s)

https://issues.newrelic.com/browse/NR-147504

### Testing

Modify the webpack config to minify the local build and use the test server to verify.

### DIFF
--- a/tools/scripts/post-build-cleanup.js
+++ b/tools/scripts/post-build-cleanup.js
@@ -25,6 +25,15 @@ function prependSemicolon (fileName, text) {
     return fs.promises.readFile(`${buildDir}/${fileName}`, 'utf-8')
   }))
 
+  files.forEach((contents, i) => {
+    if (builtFileNames[i].includes('-loader') && builtFileNames[i].endsWith('.js')) {
+      const matches = contents.match(/\$&/)
+      if (Array.isArray(matches) && matches.length > 0) {
+        throw new Error(`Loader file ${builtFileNames[i]} contains a character sequence that could break injection due to string replacement: ${JSON.stringify(matches)}`)
+      }
+    }
+  })
+
   let prepended = 0
   await Promise.all(files
     .map((f, i) => {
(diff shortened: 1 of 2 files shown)


tools/scripts/post-build-cleanup.js
```diff
@@ -25,6 +25,15 @@ function prependSemicolon (fileName, text) {
     return fs.promises.readFile(`${buildDir}/${fileName}`, 'utf-8')
   }))
 
+  files.forEach((contents, i) => {
+    if (builtFileNames[i].includes('-loader') && builtFileNames[i].endsWith('.js')) {
+      const matches = contents.match(/\$&/)
```
metal-messiah, line 30: Good start and we can fortify this selector later for more problematic statements too

metal-messiah (approved):

PR #635, opened 2023-08-08
feat: Centralize web vitals timings

metal-messiah:
Centralize web vitals timings to be accessible from any part of the agent.
---

### Overview
This PR utilizes a singleton export of the vitals timings with a pub/sub methodology to share identical data across the agent.  This will create better consistency and replace areas of the agent leaning on differing methodologies.

### Related Issue(s)
NR-101330

### Testing
Existing timing tests should continue to pass.
New jest tests have been added for each vital file
New jest test has been made to `replace` some of the web-vitals `browser` tests that became broken by changing this behavior


src/common/vitals/cumulative-layout-shift.js
```diff
@@ -0,0 +1,23 @@
...
+import { VitalMetric } from './vital-metric'
+
+export const cls = new VitalMetric(VITAL_NAMES.CUMULATIVE_LAYOUT_SHIFT, undefined, (x) => x)
+let resolver
+export const waitForCumulativeLayoutShift = (buffered = false) => {
+  return new Promise((resolve, reject) => {
+    /* Cumulative Layout Shift - We don't have to limit this callback since cls is stored as a state and only sent as attribute on other timings.
+      reportAllChanges ensures our tracked cls has the most recent rolling value to attach to 'unload' and 'pagehide'. */
+    if (buffered && cls.isValid) resolve(cls)
+    else resolver = resolve
```
patrickhousley, line 12: Is it possible that this line could be called multiple times resulting in `resolver` getting re-assigned and leaving callers of `waitForCumulativeLayoutShift` waiting on a promise that will never resolve?

metal-messiah: good point there

src/common/vitals/vital-metric.js
```diff
@@ -0,0 +1,35 @@
...
+    this.attrs = {}
+    if (typeof roundingMethod === 'function') this.#roundingMethod = roundingMethod
+    if (value !== undefined) this.#value = value
+  }
+
+  get value () { return this.#roundingMethod(this.#value) }
+  set value (v) { this.#value = v }
+
+  get isValid () { return this.value >= 0 }
+}
```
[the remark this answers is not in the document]
patrickhousley: Could add a `once` option:

```
subscribe (callback, options) {
  let cb = callback
  if (options.once) {
    cb = (...args) => {
      this.#subscribers.remove(cb)
      callback(...args)
    }
  }
  this.#subscribers.add(cb)
  return () => {
    this.#subscribers.remove(cb)
  }
}
```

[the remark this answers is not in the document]
metal-messiah: I like this idea. Could remove the promise behavior as well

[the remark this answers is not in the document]
metal-messiah: changed to more closely match this approach

src/common/vitals/vital-metric.js
```diff
@@ -0,0 +1,51 @@
+export class VitalMetric {
+  #values = []
+  #roundingMethod = Math.floor
+  #subscribers = new Set()
+
+  constructor (name, roundingMethod) {
+    this.name = name
+    this.attrs = {}
```
patrickhousley, line 8: Make these fields private. They should not be allowed to be changed outside this class. (I am assuming here)

src/features/page_view_event/aggregate/index.js
```diff
@@ -104,20 +103,18 @@ export class Aggregate extends AggregateBase {
       }
     }
 
-    try { // PVTiming sends these too, albeit using web-vitals and slightly different; it's unknown why they're duplicated, but PVT should be the truth
-      var entries = globalScope.performance.getEntriesByType('paint')
-      entries.forEach(function (entry) {
-        if (!entry.startTime || entry.startTime <= 0) return
+    try {
+      queryParameters.fp = String(firstPaint.value.current)
+      queryParameters.fcp = String(firstContentfulPaint.value.current)
```
patrickhousley, line 108: I don't think the wrapping in `String()` is necessary. If these are query parameters, they will always be sent as a string.

src/common/vitals/index.js
```diff
@@ -0,0 +1,8 @@
+import { cumulativeLayoutShift } from './cumulative-layout-shift'
+import { firstContentfulPaint } from './first-contentful-paint'
+import { firstInputDelay } from './first-input-delay'
+import { firstPaint } from './first-paint'
+import { interactionToNextPaint } from './interaction-to-next-paint'
+import { largestContentfulPaint } from './largest-contentful-paint'
+
+export { cumulativeLayoutShift, firstContentfulPaint, firstInputDelay, firstPaint, interactionToNextPaint, largestContentfulPaint }
```
patrickhousley, line 8: Could this whole file be written as `export * from 'blah'`

src/features/page_view_timing/aggregate/index.js
```diff
@@ -27,86 +24,18 @@ export class Aggregate extends AggregateBase {
...
-          })
-        }
-      } catch (e) {}
-    } else {
-      onFCP(({ name, value }) => {
-        if (pageStartedHidden || this.alreadySent.has(name)) return
-        this.alreadySent.add(name)
-        this.addTiming(name.toLowerCase(), value)
-      })
+    const handleVitalMetric = ({ name, current: value, attrs }) => {
```
patrickhousley, line 28: Maybe make this a class private method.

patrickhousley (commented):
I think this is a lot cleaner and better structured from having an object that gets imported, wrote to, and read from willy nilly in the code base. I do wonder if we should not go ahead and move the other timings that are just in the page view timing aggregator: `endCurrentSession` and `recordPageUnload`.

src/common/vitals/vital-metric.js
```diff
@@ -0,0 +1,51 @@
...
+      try {
+        cb(this.value)
+      } catch (e) {
+        // ignore errors
+      }
+    })
+  }
+
+  get isValid () {
+    return this.value.current >= 0
```
patrickhousley, line 33: This should probably be `this.#values.length > 0`. Just looking to the comments down around the `cls` metric in the page view timing aggregator.

src/common/vitals/index.js
metal-messiah: Yes but this guards from adding stuff later to that file that may not be meant to be barrel'd up. 

I'm good either way tho

patrickhousley: True. Personally I hate barrels because I've been on the shit end of figuring out why bundles where so big and including stuff. I am curious if it matters in the size department of our outputs. Thats really all I care about here.

metal-messiah: Yeah true. Let's just skip the barrel and be explicit

patrickhousley:
Looks good. Need jest tests for the new `src/common/vitals` files.

metal-messiah:
https://github.com/newrelic/newrelic-browser-agent/actions/runs/5863677585

src/common/constants/runtime.js
```diff
@@ -69,3 +71,5 @@ export const ffVersion = (() => {
 export const isIE = Boolean(isBrowserScope && window.document.documentMode) // deprecated property that only works in IE
 
 export const supportsSendBeacon = !!navigator.sendBeacon
+
+export const offset = Math.floor(globalScope?.performance?.timeOrigin || globalScope?.performance?.timing?.navigationStart || Date.now())
```
cwli24, line 75: Do you want to just change the runtime obj to use this too? This is being duplicated rn

src/common/vitals/cumulative-layout-shift.js
```diff
@@ -0,0 +1,13 @@
+import { onCLS } from 'web-vitals'
+import { VITAL_NAMES } from './constants'
+import { VitalMetric } from './vital-metric'
+import { isBrowserScope } from '../constants/runtime'
+
+export const cumulativeLayoutShift = new VitalMetric(VITAL_NAMES.CUMULATIVE_LAYOUT_SHIFT, (x) => x)
+
+if (isBrowserScope) {
+  onCLS(({ value, entries }) => {
+    if (value === cumulativeLayoutShift.value.current) return
```
cwli24, line 10: If the purpose is to _not_ have adjacent (reported) values be the same, note that the value in `.value.current` has been potentially rounded, so this comparison may return a false negative.

src/common/vitals/vital-metric.js
cwli24, line 1: I know this is really late to bring up, but I don't really see a big value to having this "onion layer" interface inserted between web-vitals and our usage code. The `#values` and `entries` arrays continues to balloon over lifetime and just consumes memory. This also makes working (us/others) with the code a bit harder in that `on___` is well documented and easy to use via its readme docs or webdev pages, but now it's replaced by an intermediary that requires reading or modifying this class and/or individual timing modules.
(Treat this as a side thread)

src/common/vitals/vital-metric.js
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor (name, roundingMethod) {
+    this.name = name
+    this.attrs = {}
+    if (typeof roundingMethod === 'function') this.#roundingMethod = roundingMethod
+  }
+
+  update ({ value, entries, attrs, addConnectionAttributes = false }) {
+    this.#values.push(this.#roundingMethod(value))
+    this.entries = entries
```
cwli24, line 15: The old values are kept, then so should the old entries. The new `entries` may not contain the same subset of records as the existing `entries`, so `this.entries` needs to be appended or merged each update too?

src/features/page_view_timing/aggregate/index.js
cwli24, line 1: All the `alreadySent` that ensures timings are sent once per hard page load as previously decided got removed. This prevented web-vital internal reporting and reset from diluting existing customer dashboards that don't account for soft/cached page loads. Has this been re-discussed & approved by UI or Cheryl?

src/common/constants/runtime.js
metal-messiah: yes, because it is nice to be able to access without an identifier.  But offset is used in many places so we should do this separately.  Adding a ticket

src/features/page_view_timing/aggregate/index.js
```diff
@@ -180,8 +102,8 @@ export class Aggregate extends AggregateBase {
     Mitigation: We've set initial CLS to null so that it's omitted from timings like 'pageHide' in that edge case. It should only be included if onCLS callback was executed at least once.
     Future: onCLS value changes should be reported directly & CLS separated into its own timing node so it's not beholden to 'pageHide' firing. It'd also be possible to report the real final CLS.
     */
-    if (this.cls !== null) {
-      attrs.cls = this.cls
+    if (cumulativeLayoutShift.value.current >= 0) {
```
cwli24, line 105: The default `value.current` in which CLS has never been set yet is now `undefined` via `#value[-1]` of an empty array. This should check `!== undefined` then rather than 0.

src/common/constants/runtime.js
metal-messiah: https://issues.newrelic.com/browse/NR-151155

src/common/vitals/first-contentful-paint.test.js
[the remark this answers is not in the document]
metal-messiah: They are each testing that if the reporting mechanism (usually `on<whatever>` from web-vitals, but also other mech's as fallbacks) emit a value, that the `VitalMetric` is updated.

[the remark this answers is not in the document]
cwli24: I mean to say that the actual reporting heuristics aren't being tested--they can't be as these are jest tests and the on___ are mocked to become trivial number-repeaters. And so, as you say, really just "`VitalMetric` is updated" is being tested, and not the process of reporting itself for each timing.

src/features/page_view_timing/aggregate/index.js
metal-messiah: it hasn't been removed.  All the vitals that checked for `alreadySent` now just check internally to see if they are already `valid` (which means they already emitted to the subscribers).  If they are, they do not emit again.  There are also test cases validating this in all the vitals that behave this way.

cwli24: You're right; I missed the `isValid` - disregard this.

src/common/vitals/first-contentful-paint.test.js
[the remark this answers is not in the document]
cwli24: Is there a test that checks multiple consumers subscribed to a timing get the same data? Don't see it in vital-metric.test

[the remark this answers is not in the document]
metal-messiah: can add that for sure

src/common/vitals/cumulative-layout-shift.js
metal-messiah: b6bf3b1af9b4eb9c8788f9e59cfa0855e49499db

src/common/vitals/vital-metric.js
metal-messiah: f6bb821e7788e7027484256438a5ab086de38cce

src/common/vitals/cumulative-layout-shift.js
```diff
@@ -7,7 +7,7 @@ export const cumulativeLayoutShift = new VitalMetric(VITAL_NAMES.CUMULATIVE_LAYO
 
 if (isBrowserScope) {
   onCLS(({ value, entries }) => {
-    if (value === cumulativeLayoutShift.value.current) return
+    if (cumulativeLayoutShift.roundingMethod(value) === cumulativeLayoutShift.value.current) return
```
cwli24, line 10: Storing the unrounded value would be better because current: 83.3 and new: 83.4 would make this a false positive under flooring.

cwli24: (Vital numbers have a bigger granularity so it's easy to lose accuracy, unlike performance domhighrestimestamp numbers.)

cwli24: I think you also flatten the new `entries` as an array in other commit, but if you want to be able to tell which set of entries is responsible for yielding which value point, they would have to be appended as-is (2-d array).

metal-messiah: Guess it depends on your definition. Checking that the rounded values are equal is important because it's the rounded value that gets reported (and we don't want to re-emit it)

patrickhousley (approved):

src/common/vitals/vital-metric.js
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor (name, roundingMethod) {
+    this.name = name
+    this.attrs = {}
+    this.roundingMethod = typeof roundingMethod === 'function' ? roundingMethod : Math.floor
+  }
+
+  update ({ value, entries = [], attrs = {}, addConnectionAttributes = false }) {
+    this.#values.push(this.roundingMethod(value))
+    this.entries.push(...entries)
```
cwli24, line 14: Might've mentioned this already but you may want to keep the entries segregated like `.push(entries)` so value is kept 1:1 with each **set**. Spreading entries make them indistinguishable.

src/common/vitals/vital-metric.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  get isValid () {
+    return this.#values.length > 0 && this.value.current >= 0
+  }
+
+  subscribe (callback) {
+    this.#subscribers.add(callback)
+    if (this.isValid) callback(this.value)
```
cwli24, line 42: If there's already data (prev calls to `.update`), you'd want to cb with _each_ of those value-entries.

metal-messiah: not following what you mean here, `this.value` is a getter which returns previous, current, and entries (all the values seen). Is that what you mean?
```javascript
get value () {
    return {
      previous: this.#values[this.#values.length - 2],
      current: this.#values[this.#values.length - 1],
      name: this.name,
      attrs: this.attrs,
      entries: this.entries
    }
  }
```

https://github.com/newrelic/newrelic-browser-agent/pull/635/files/7b056dd4bb8c7c1b9a78ecb9bc015c78d57fa2e1#diff-51f5c7465d0ec05605d24c9d75d2580ee046be3b14756d95033e85c7ceeed67cR26-R34

cwli24: Well suppose `this.#values = [1, 5, 2, 4]` when `subscribe` is called. It runs for the current value but none of the pre-existing. This differs from what observers with buffered would do. In my eyes, `subscribe` being for only the most current "updates" and future updates but not past updates is not a common pattern; parallel to the EE not executing the `.on(___,cb)` with the most recent emitted same-type event & args. 

I think to sum it up, this could be confirmed to be either _only_ forward-looking (like ee.on) or both forward and **all** backwards looking (like PO.observer buffered)

src/common/vitals/vital-metric.js
cwli24: In the future, we may have callbacks provided to `subscribe` that cares about a set of PerformanceEntry's for a/each particular value derived from such set. I can think of such use case from some questions posed by Cheryl or UI so far in recent history.

src/common/vitals/vital-metric.js
metal-messiah: this was somewhat patterned off of rxjs's behaviorSubject, which will emit the latest value when subscribed to https://stackblitz.com/edit/rxjs-behaviorsubject-simpleexample?devtoolsheight=100&file=index.ts . Ours just has extra metadata attached as well on that emission.

if that's confusing, we can always iterate on this pattern tho

metal-messiah: Not really though, because it also reports the entries, which contains the full history of the emissions

metal-messiah: Just to be clear -- Basically it always reports the full history (entries), but the returned object also has shortcut methods the get the latest value or previous value

src/common/vitals/vital-metric.js
metal-messiah: Let's address that in the future if needed and not treat it as a blocker since that would require basically a full rewrite of the PR for something that doesn't exist in the agent today

cwli24: But all that means is just `this.entries.push(entries)` and you've separated them?

src/common/vitals/vital-metric.js
cwli24: If a cwv emits (valueA derived from setA entries) and (valueB derived from setB entries) and (valueC from setC), it doesn't make sense you can just report valueC and [...setA, ...setB, ...setC] (can't even differentiate the line between the sets) and expect the receiver to reconstruct valueA & valueB.

src/common/vitals/vital-metric.js
cwli24: Let's defer this as entries are not really used or sent to NR atm, but the history of values still need addressing.

src/common/vitals/vital-metric.js
metal-messiah: changed to include all updates as history objects, see: https://github.com/newrelic/newrelic-browser-agent/pull/635/commits/39fc1e91a53329c4c2609d030e4f31e0f64cc762

PR #636, opened 2023-08-08
chore: multi-agent single page

patrickhousley:
This modifies the internals of the agent wrappings to support chaining those wrappings when more then one agent bundle is loaded on the same page.
---

### Overview

This modifies the internals of the agent wrappings to support chaining those wrappings when more then one agent bundle is loaded on the same page. When wrapper methods are called within the same bindle, the globals will only be wrapped once to maintain how the agent works today. However, if more than one agent bundle, for example two different agent versions, are loaded on the same page, this changes the agent so that the second agent will still wrap the already wrapped globals. This in effect allows both agents to instrument the browser runtime and harvest it's own data.

This PR also includes workflow changes:
- Several of the scripts from `tools/scripts` have now been turned into composite github actions. These can still be used on local machines by CDing into the action and running `npm i && node index.js`. This was done to isolate the script and it's dependencies from the agent code with the eventual goal being to remove unnecessary dependencies from the agent and resolve false positive dependabot findings.
- `.github/actions/build-ab` is used to build the A/B script for the different environments and does so by combining the "current" and "dev" script from the CDN. For the dev and staging environments, it will also include "experiments".
- `.github/actions/fastly-purge` is used to purge one or more asset paths within fastly.
- `.github/actions/fastly-verify` is used to "fetch" one or more assets from the fastly cdn to verify the asset is available.
- `.github/actions/internal-ab` is used to call other composite actions to build and publish the internal A/B scripts and flush the CDN cache.
- `.github/actions/s3-upload` is used to upload all files within a given directory to a given S3 bucket and sub-path.
- The publish experiments and dev workflows have been update to make use of the new composite actions but still fundamentally function the same and have the same results as the old versions of those workflows.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-117113

### Testing

This should go through the full battery of tests. None of the tests were touched since these changes should not change how the agent function normally for a customer in any way.


metal-messiah:
I think something is going on with the build size action.  Its drastically larger for all builds

patrickhousley:
I ran the script manually and am getting the same size results. I plan on working on the diff script in another PR so maybe I will find something then.

metal-messiah (approved):
Set up looks good.  When processes/tests are prepped, approved for merge

PR #637, opened 2023-08-11
feat: Reducing the agent chunk count

patrickhousley:
Reducing the number JavaScript files the agent loads asynchronously down to a single file. This means customer's users will now only need to load one additional JavaScript file for the New Relic Browser Agent to operate. This additional JavaScript file will continue to only be loaded upon the document "load" event. This change not only reduces the number of additional requests the agent must make but also reduces the amount of code that must be loaded by eliminating duplication.
---

### Overview

Header pretty much says it all. Below are the code size statistics and the CDN cost statistics are attached in the JIRA.

#### Current Loader Statistics

|Loader|# of Assets|Raw Size|GZip Size|
|---|---|---|---|
|lite|8|106 KB|29 KB|
|pro|12|172 KB|46 KB|
|spa|13|205 KB|53 KB|
|experimental|15|381 KB|110 KB|

#### New Loader Statistics

|Loader|# of Assets|Raw Size|GZip Size|
|---|---|---|--|
|lite|1|74 KB|24 KB|
|pro|1|111 KB|35 KB|
|spa|1|135 KB|41 KB|
|experimental|3|315 KB|99 KB|

> Sizes also include the loader.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-149736

### Testing

Load the 1mb test page to verify the async chunk is still being loaded on page load. Otherwise, the existing wdio and jil tests should continue passing.

### DIFF
--- a/tools/webpack/configs/standard.mjs
+++ b/tools/webpack/configs/standard.mjs
@@ -1,4 +1,5 @@
 import path from 'path'
+import webpack from 'webpack'
 import { merge } from 'webpack-merge'
 import commonConfig from './common.mjs'
 import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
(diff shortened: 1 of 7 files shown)


patrickhousley:
The size check workflow is going to be broken on this and future PRs. I will fix that workflow after this has been merged so we have files in the CDN to test the workflow updates with.

tools/webpack/configs/standard.mjs
```diff
@@ -14,57 +15,122 @@ import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
...
+      entry: {
+        'nr-loader-rum': path.join(env.paths.src, 'cdn/lite.js'),
+        'nr-loader-rum.min': path.join(env.paths.src, 'cdn/lite.js')
+      },
+      plugins: [
+        new webpack.IgnorePlugin({
+          checkResource: (resource, context) => {
+            if (context.match(/features\/utils/) && resource.indexOf('aggregate') > -1) {
+              // Only allow page_view_event, page_view_timing, and metrics features
+              return !resource.match(/page_view_event\/aggregate|page_view_timing\/aggregate|metrics\/aggregate/)
```
metal-messiah, line 30: Did we determine if maintaining this list is actually necessary?  Or will it get tree-shaken out anyway?

patrickhousley: I did not try but I can.

patrickhousley: Yea we will have to maintain this list. Without this plugin, all the features get included in the async chunk for the lite and pro agent.

metal-messiah (approved):

PR #648, opened 2023-08-17
feat: Add best effort to detect Ajax events before instantiation

metal-messiah:
Add a "best effort" at detecting ajax events that may have happened on the page before the agent was instantiated by utilizing the ResourceTiming browser API.
---

### Overview
This PR adds a "best effort" at detecting ajax events that may have happened on the page before the agent was instantiated by utilizing the ResourceTiming browser API. This can create mostly-complete ajax event data.  It will run once, at the instantiation of the AJAX feature, before wrappings and handlers have been set up, to avoid data duplication.

### Related Issue(s)
https://issues.newrelic.com/browse/NR-151536

### Testing
A new e2e was added to validate this behavior.


src/features/ajax/instrument/index.js
```diff
@@ -34,6 +34,29 @@ export class Instrument extends InstrumentBase {
...
+    // because it was deferred from loading in some way -- e.g. 'deferred' script loading tags or other lazy-loading techniques
+    //
+    // it is "imperfect" because we cannot capture the following with the API vs wrapping the events directly:
+    // * requestBodySize - (txSize -- request body size)
+    // * method - request type (GET, POST, etc)
+    // * callbackDuration - (cbTime -- sum of resulting callback time)
+    try {
+      const initiators = { xmlhttprequest: 'xhr', fetch: 'fetch', beacon: 'beacon' }
+      globalScope?.performance?.getEntriesByType('resource').forEach(resource => {
+        if (Object.keys(initiators).includes(resource.initiatorType) && resource.responseStatus !== 0) {
```
cwli24, line 49:
```suggestion
        if (resource.initiatorType in initiators && resource.responseStatus !== 0) {
```
^is a much faster operation btw

patrickhousley (approved):

src/features/ajax/instrument/index.js
metal-messiah: Screenshot 2023-08-23 at 10 34 55 PM

Much* is subjective i guess when we are talking about 3 keys -- it's negligible, but we can go with this

PR #652, opened 2023-08-22
chore: Remove allow_bfcache flag

cwli24:
---

### Overview

PR cleans up unused (old) code associated with the bf cache support change in the past. This flag has been on for multiple agent versions with no report of issues, hence by consensus is due to be removed at this time.

### Related Issue(s)

NR-137640

### Testing

Code deletion only. No change in default behavior of the agent.

### DIFF
--- a/src/common/config/state/init.js
+++ b/src/common/config/state/init.js
@@ -8,7 +8,6 @@ const model = () => {
     maskInputOptions: { password: true }
   }
   return {
-    allow_bfcache: true, // *cli - temporary feature flag for BFCache work
     privacy: { cookies_enabled: true }, // *cli - per discussion, default should be true
     ajax: { deny_list: undefined, block_internal: true, enabled: true, harvestTimeSeconds: 10 },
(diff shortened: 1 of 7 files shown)


src/common/config/state/init.js
```diff
@@ -8,7 +8,6 @@ const model = () => {
     maskInputOptions: { password: true }
   }
   return {
-    allow_bfcache: true, // *cli - temporary feature flag for BFCache work
```
metal-messiah, line 11: does the agent fail or warn or anything negative if a customer is setting this value via a previous version implementation and we remove it from the config model?

cwli24: The set fn just ignores it via the configurable.js

metal-messiah:
We should consider making this a feat just so that it makes it into the release notes for visibility

patrickhousley (approved):
Approved but this should be changed to a `feat` so customers are aware in case they were using this feature flag for some reason.

PR #656, opened 2023-08-23
feat: Add mode to enable agent to not harvest until user consent

metal-messiah:
Add ability for customers to wait to start harvesting. This behavior will utilize a configuration setting in tandem with an API call to trigger features to begin harvesting.
---

### Overview
This PR adds the ability to both delay the agent from harvesting, and trigger the agent to begin harvesting only after customer input.  

possible configurations:
* init.<feature_name>.autoStart = false
    - features with false flag will not harvest until api is called
    - others will harvest immediately

Possible implementations:
* customer calls `newrelic.start()` with empty arguments
    - Starts harvest for all features
* customer calls `newrelic.start(feature_name)`
    - Starts harvest for single feature matching start call string arg
* customer calls `newrelic.start([feature_name, feature_name, etc...])`
    - Starts harvest for all features matching start call array arg

### Related Issue(s)
https://issues.newrelic.com/browse/NR-152577
https://issues.newrelic.com/browse/NR-152568
https://issues.newrelic.com/browse/NR-137050

### Testing
New tests have been added to validate this behavior


src/common/config/state/init.js
```diff
@@ -23,6 +23,18 @@ const model = () => {
...
+      jserrors: true,
+      metrics: true,
+      page_action: true,
+      page_view_event: true,
+      page_view_timing: true,
+      session_replay: true,
+      session_trace: true,
+      spa: true
+    },
+
```
patrickhousley, line 37:
```suggestion
```

src/common/drain/drain.js
```diff
@@ -42,9 +42,10 @@ function curateRegistry (agentIdentifier) {
  * @param {string} agentIdentifier - A unique 16 character ID corresponding to an instantiated agent.
  * @param {string} featureName - A named group into which the feature's buffered events are bucketed.
  */
-export function drain (agentIdentifier = '', featureName = 'feature') {
+export function drain (agentIdentifier = '', featureName = 'feature', force) {
   curateRegistry(agentIdentifier)
 
+  if (force) registry[agentIdentifier].delete(featureName)
```
patrickhousley, line 48: Do we ever use the `force` flag?

metal-messiah: all the features use it if the "auto" property is false

src/features/ajax/aggregate/index.js
```diff
@@ -128,6 +127,7 @@ export class Aggregate extends AggregateBase {
     }
 
     function prepareHarvest (options) {
+      console.log('ajax harvest!')
```
patrickhousley, line 130: Missed console log.

metal-messiah, line 130: remove console log

src/features/page_action/aggregate/index.js
```diff
@@ -32,17 +31,15 @@ export class Aggregate extends AggregateBase {
...
-    register('block-ins', () => {
-      this.blocked = true
-      scheduler.stopTimer(true)
-    }, this.featureName, this.ee)
+    this.waitForFlags(['ins']).then(([enabled]) => {
+      if (enabled) {
+        var scheduler = new HarvestScheduler('ins', { onFinished: (...args) => this.onHarvestFinished(...args) }, this)
+        scheduler.harvest.on('ins', (...args) => this.onHarvestStarted(...args))
+        scheduler.startTimer(this.harvestTimeSeconds, 0)
+      }
```
metal-messiah, line 39: and else block that sets `blocked = true`

src/features/session_replay/aggregate/index.js
```diff
@@ -121,16 +120,26 @@ export class Aggregate extends AggregateBase {
...
 
-      this.waitForFlags(['sr']).then(([flagOn]) => this.initializeRecording(
+      Promise.all([
+        this.waitForFlags(['sr']),
+        this.waitForOptIn()
+      ]).then(([[flagOn]]) => this.initializeRecording(
         flagOn,
         Math.random() < getConfigurationValue(this.agentIdentifier, 'session_replay.errorSampleRate'),
         Math.random() < getConfigurationValue(this.agentIdentifier, 'session_replay.sampleRate')
       )).then(() => sharedChannel.onReplayReady(this.mode)) // notify watchers that replay started with the mode
```
patrickhousley, line 130: Is there any chance this could result in an unhandled promise rejection?

src/features/session_replay/aggregate/index.js
```diff
@@ -121,16 +120,26 @@ export class Aggregate extends AggregateBase {
...
 
-      drain(this.agentIdentifier, this.featureName)
+      this.drain()
     }
   }
 
+  waitForOptIn () {
+    return new Promise((resolve, reject) => {
```
patrickhousley, line 137: Looks like `reject` is unused here.

src/features/session_replay/aggregate/index.js
```diff
@@ -121,16 +120,26 @@ export class Aggregate extends AggregateBase {
         }
       }, this.featureName, this.ee)
 
-      this.waitForFlags(['sr']).then(([flagOn]) => this.initializeRecording(
+      Promise.all([
+        this.waitForFlags(['sr']),
+        this.waitForOptIn()
+      ]).then(([[flagOn]]) => this.initializeRecording(
```
metal-messiah, line 126: I think this is no longer necessary with the opt in being tied to the importAggregator() call now, should remove

src/features/session_replay/aggregate/index.js
```diff
@@ -121,16 +120,26 @@ export class Aggregate extends AggregateBase {
...
     }
   }
 
+  waitForOptIn () {
+    return new Promise((resolve, reject) => {
+      if (!getConfigurationValue(this.agentIdentifier, 'session_replay.requireOptIn')) resolve()
+      registerHandler('sr-opt-in', () => resolve(), this.featureName, this.ee)
+    })
+  }
+
```
metal-messiah, line 142: remove

src/features/utils/instrument-base.test.js
```diff
@@ -10,6 +10,18 @@ import { warn } from '../../common/util/console'
...
+  jserrors: true,
+  metrics: true,
+  page_action: true,
+  page_view_event: true,
+  page_view_timing: true,
+  session_replay: true,
+  session_trace: true,
+  spa: true
+}
+
```
patrickhousley, line 24:
```suggestion
```

tests/specs/session-replay/rrweb-configuration.e2e.js
```diff
@@ -30,6 +30,27 @@ describe.withBrowsersMatching(notIE)('RRWeb Configuration', () => {
...
+      await browser.execute(function () {
+        newrelic.sessionReplay().optIn()
+      })
+      wasInitialized = await browser.execute(function () {
+        return Object.values(newrelic.initializedAgents)[0].features.session_replay.featAggregate.initialized
+      })
+      expect(wasInitialized).toEqual(true)
+    })
+  })
+
```
metal-messiah, line 53: remove

src/features/session_replay/aggregate/index.js
metal-messiah: yes potentially, but this was missed after cloning from the other SR opt in branch and this is no longer needed. will remove

src/features/session_replay/aggregate/index.js
metal-messiah: this was missed after cloning from the other SR opt in branch and this is no longer needed. will remove

metal-messiah:
Theres some remnants from the clone of the other SR opt in POC that need to be removed.

patrickhousley:
Might want to run `npm run cdn:webpack -- --env mode=prod --profile --json=compilation-stats.json` and load the stats json file in https://webpack.github.io/analyse or https://chrisbateman.github.io/webpack-visualizer/ to see why the lite agent loader grew so much. Maybe it's nothing, maybe it's fine, just would double check.

tests/specs/session-replay/rrweb-configuration.e2e.js
metal-messiah: changed to test SR functionality with new method

metal-messiah:
> Might want to run `npm run cdn:webpack -- --env mode=prod --profile --json=compilation-stats.json` and load the stats json file in https://webpack.github.io/analyse or https://chrisbateman.github.io/webpack-visualizer/ to see why the lite agent loader grew so much. Maybe it's nothing, maybe it's fine, just would double check.

After PR comment fixes and removing unused code -- when building prod locally, it builds to 27.91 kb, a 0.12kb increase.  Will follow up more if needed

metal-messiah:
https://github.com/newrelic/newrelic-browser-agent/actions/runs/5969345072

src/common/config/state/init.js
```diff
@@ -23,6 +23,17 @@ const model = () => {
...
+      ajax: true,
+      jserrors: true,
+      metrics: true,
+      page_action: true,
+      page_view_event: true,
+      page_view_timing: true,
+      session_replay: true,
+      session_trace: true,
+      spa: true
+    },
```
metal-messiah, line 36: still seeking feedback on how to structure this part

metal-messiah: leading candidate:
`<feature_name>.autoStart = boolean`

src/common/util/feature-flags.js
```diff
@@ -15,26 +14,25 @@ const bucketMap = {
   sr: [FEATURE_NAMES.sessionReplay, FEATURE_NAMES.sessionTrace]
 }
 
+const sentIds = new Set()
```
patrickhousley, line 17: I would like us to get to a point where we do not have module level variables like this. It is what caused this headache of not being able to instantiate more than one micro agent from the same bundle.

tests/assets/instrumented-manual.html
```diff
@@ -0,0 +1,36 @@
...
+      NREUM.init.auto = {
+              ajax: false,
+              jserrors: false,
+              metrics: false,
+              page_action: false,
+              page_view_event: false,
+              page_view_timing: false,
+              session_trace: false,
+              spa: false
+            }
```
patrickhousley, line 20: This is fine, I just needed a place to add this comment 😉 . The file `tools/wdio/plugins/testing-server/default-asset-query.mjs` needs to be updated with the defaults please.

src/common/util/feature-flags.js
metal-messiah: i agree, for the most part.  Module-level vars are fine if they are namespaced per agentID, which `sentIds` will be.  Still work to be done to land on a common pattern throughout the agent, but the most common pattern today is the agentID namespace in a stateful var like that. Bigger chore to rip that behavior out of the agent completely, but could be done

patrickhousley: Yea I think this is a job for another day and we should settle on a pattern. One which I think I would like is to have the Agent instance be the top level scope for everything else but that has implications for some of the "global" stuff we do. Like you said, a conversation for another day.

src/loaders/api/api.js
```diff
@@ -122,6 +122,22 @@ export function setAPI (agentIdentifier, forceDrain) {
     return appendJsAttribute('application.version', value, 'setApplicationVersion', false)
   }
 
+  apiInterface.start = (features, cb) => {
```
patrickhousley, line 125: Don't forget the angler update for the SM calling this API should generate.

metal-messiah: good call

src/common/config/state/init.js
cwli24, line 1: Could you update PR description please? heh

src/features/page_action/aggregate/index.js
```diff
@@ -32,17 +31,17 @@ export class Aggregate extends AggregateBase {
...
-    this.ee.on(`drain-${this.featureName}`, () => { if (!this.blocked) scheduler.startTimer(this.harvestTimeSeconds, 0) })
-
-    // if rum response determines that customer lacks entitlements for ins endpoint, block it
-    register('block-ins', () => {
-      this.blocked = true
-      scheduler.stopTimer(true)
-    }, this.featureName, this.ee)
+    this.waitForFlags(['ins']).then(([enabled]) => {
+      if (enabled) {
+        var scheduler = new HarvestScheduler('ins', { onFinished: (...args) => this.onHarvestFinished(...args) }, this)
```
cwli24, line 36: Minor convention styling
```suggestion
        const scheduler = new HarvestScheduler('ins', { onFinished: (...args) => this.onHarvestFinished(...args) }, this)
```

tools/testing-server/constants.js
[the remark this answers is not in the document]
cwli24: Also `allow_bfcache: true,` got removed in main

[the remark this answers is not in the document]
metal-messiah: putting it into the testing-server allows `npm run test-server` to also have the base configs supplied to it, instead of only during active test injections via wdio.

The reason to keep `loader: spa` in wdio instead of moving that too, is that the default loader for `JIL` is `pro`, so this retains that behavior.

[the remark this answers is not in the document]
cwli24: Got it. So before wdio had its own configs, jil had its own config, now they share some configs but also have their own?

src/features/utils/instrument-base.js
```diff
@@ -45,7 +45,12 @@ export class InstrumentBase extends FeatureBase {
     */
     this.onAggregateImported = undefined
 
-    if (auto) registerDrain(agentIdentifier, featureName)
+    /** used in conjunction with newrelic.start() to defer harvesting in features */
+    this.requiresOptIn = (getConfigurationValue(this.agentIdentifier, `${this.featureName}.autoStart`) === false || getConfigurationValue(this.agentIdentifier, 'auto') === false)
```
cwli24, line 49: there's no `auto` field in the init model

src/features/utils/instrument-base.js
```diff
@@ -56,6 +61,20 @@ export class InstrumentBase extends FeatureBase {
...
+
+    if (this.requiresOptIn && !this.optedIn) {
+      // this feature requires an opt in...
+      // wait for API to be called
+      this.ee.on(`${this.featureName}-opt-in`, () => {
+        // register the feature to drain only once the API has been called, it will drain when importAggregator finishes for all the features
+        // called by the api in that cycle
+        registerDrain(this.agentIdentifier, this.featureName)
+        this.optedIn = true
+        this.importAggregator()
```
cwli24, line 73: The way this is done makes it non-obvious for any edge case bugs, such as similar to the interaction-timer issue with too many redundant handlers over time.

I recommend getting rid of `this.optedIn` (less unneeded logic vars the better) and creating a separate function rather than recursing itself. If you want to doubly ensure the code path of `this.importAggregator()` only runs once, you can either make the ee handler a singly function or tbh `featAggregate` already does that job, though you could also just set `this.auto = false` for extra assurance.

src/loaders/api/api.js
cwli24: Does `agent-base` already need to include these fn for typing?

src/loaders/api/api.js
```diff
@@ -122,6 +122,22 @@ export function setAPI (agentIdentifier, forceDrain) {
...
+    try {
+      const featNames = Object.values(FEATURE_NAMES)
+      if (!features) features = featNames
+      features = Array.isArray(features) && features.length ? features : [features]
+      if (features.some(f => !featNames.includes(f))) return warn(`Invalid feature name supplied. Acceptable feature names are: ${featNames}`)
+
+      if (!features.includes(FEATURE_NAMES.pageViewEvent)) features.push(FEATURE_NAMES.pageViewEvent)
```
cwli24, line 132:
```suggestion
      if (!features) features = featNames
      else {
        features = Array.isArray(features) && features.length ? features : [features]
        if (features.some(f => !featNames.includes(f))) return warn(`Invalid feature name supplied. Acceptable feature names are: ${featNames}`)
  
        if (!features.includes(FEATURE_NAMES.pageViewEvent)) features.push(FEATURE_NAMES.pageViewEvent)
      }
```

src/loaders/api/api.js
metal-messiah: yes, will need to update that as well

metal-messiah: updated typings, need to do angler PR still

src/features/utils/instrument-base.js
metal-messiah: if youre worried about extra logic, all of that is equally as complicated as just staging another call to importAggregator, and there is no way for it to trigger ee.on more than once, so thats not really an issue here...

but i do like the this.auto = false route so i will modify to use that

src/loaders/api/api.js
metal-messiah: https://source.datanerd.us/agents/angler/pull/481

src/loaders/agent-base.js
```diff
@@ -92,4 +92,13 @@ export class AgentBase {
   addRelease (name, id) {
     warn('Call to agent api addRelease failed. The agent is not currently initialized.')
   }
+
+  /**
+   * Starts a set of agent features if not running in "autoStart" mode
+   * {@link https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/start/}
+   * @param {string|string[]|undefined} name The feature name(s) to start.  If no name(s) are passed, all features will be started
+   */
+  start (featureNames) {
```
patrickhousley, line 101: This should be moved to the `Agent` class since this API method does not apply to the micro agent.

metal-messiah: The `start` api has now been retrofitted into the micro-agent as well.

patrickhousley (approved):

PR #659, opened 2023-08-24
feat: Origin of agent webpack chunks now changeable

cwli24:
Allow the agent to operate through a proxy. The agent needs to be able to fetch its source code and send analytics to a different domain, and this solves the former. NPM is not affected.
---

### Overview

In support of NR-152264 / Proxy Support, a configuration has been exposed that will change where the agent chunk is loaded from. In context, our default domain `js-agent.newrelic.com` is on many ad blocker list today which prevent us from collecting data from end users. The use case here is to circumvent that block by using a proxy to fetch agent assets. (The beacon can also be changed so that analytics data flow throw another--or the same--proxy server.)
```javascript
init: {
  proxy: {
    beacon,
    assets
  }
}
```

Note that this applies to agent pulled from the CDN (installation methods: copy+paste or APM-injected). Agent installed via NPM should not be affected by this change, as it's irrelevant.

### Related Issue(s)

NR-149705

### Testing

Existing CDN / NPM agent tests ensure new code does not break default functionality. New e2e has been added that tests the chunk source switching behavior.

UPDATE: this has been merged with #673 .


tools/webpack/env.mjs
```diff
@@ -71,10 +71,11 @@ export default async (env) => {
...
 
-  process.env.PATH_VERSION = PATH_VERSION
-  process.env.SUBVERSION = SUBVERSION
-  process.env.PUBLIC_PATH = PUBLIC_PATH
-  process.env.VERSION = VERSION
+  // -- These are likely not being used(?)
+  // process.env.PATH_VERSION = PATH_VERSION
+  // process.env.SUBVERSION = SUBVERSION
+  // process.env.PUBLIC_PATH = PUBLIC_PATH
+  // process.env.VERSION = VERSION
```
metal-messiah, line 78: lets just remove this if it isnt being used

src/common/config/state/info.js
```diff
@@ -6,6 +6,7 @@ const model = {
   beacon: nrDefaults.beacon,
   errorBeacon: nrDefaults.errorBeacon,
   // others must be populated by user
+  assetsPath: undefined,
```
cwli24, line 9: Move this to the `init` block

src/common/url/check-url.js
```diff
@@ -0,0 +1,12 @@
+export function isValidAssetUrl (string) {
+  /* Note: This API is not available in IE11 and throws in Safari v14- when base is undefined.
+    Hence, this will always return false for those environments.
+  */
+  try {
+    let url = new URL(string)
```
cwli24, line 6: try swapping this out to `parseUrl`

src/common/url/check-url.js
cwli24, line 1: Accept string with or w/o scheme, but ignore it anyways and enforce https.
Make sure port goes through to final setting.

src/common/url/check-url.js
cwli24: Doesn't work. This function tries to ensure customer provides a domain:port aka origin; however, `parseUrl` interprets relative paths with the current location as base.

src/common/config/state/init.js
```diff
@@ -9,6 +9,7 @@ const model = () => {
   }
   return {
     allow_bfcache: true, // *cli - temporary feature flag for BFCache work
+    assetsPath: undefined,
```
metal-messiah, line 12: can we just add a little comment to this for future sanity
```suggestion
    assetsPath: undefined, // if a value is set here by a customer, it will be used to overwrite the webpack asset path used to fetch assets
```

src/common/url/check-url.js
```diff
@@ -1,12 +1,27 @@
...
+    properUrl = new URL(string)
+    properUrl.protocol = 'https:' // enforce HTTPS over HTTP
   } catch (_) {
-    return false
+    return ''
   }
-  if (!string.endsWith('/')) return false // webpack concats base verbatim so this is a stringent req
-  return true
+  // We want a url with scheme, domain, and port minus username|password => origin; webpack concats this verbatim, so an ending slash is required
+  return properUrl.origin + '/'
```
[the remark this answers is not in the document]
cwli24: Keeping in sync with beacon, does that support path too? 👀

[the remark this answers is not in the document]
metal-messiah: I think technically it would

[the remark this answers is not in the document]
patrickhousley: Technically it would and I would like to keep the beacon and asset in sync with what we support.

src/common/url/check-url.js
patrickhousley, line 1: This function could be name more generically and should be applied to the beacon properties as well since those could be changed by the customer. Ultimately, I feel we need to standardize on what we expect in those parameters like it must contain `https` unless `ssl === false`, it must contain a domain, and can contain a port and path. Build a parser around those requirements and run it on the properties in the `configure.js` file. Update the harvest file to take out the reliance on the ssl setting and let the configure file handle constructing the correct base url for harvesting.

metal-messiah: Could also use the getter setter pattern in init.js that session replay is using

cwli24: `ssl` doesn't have a UI toggle (maybe it did?) so not even sure when it's used. It's always defaulted to `undefined` when APM installed, and anything aside from explicitly setting it to `false` would result in `https` in beacon. 

Here's my point:

1. For old users, if they are still on http, they will retain beacon without scheme and `ssl = false` and http is used.
2. For both old & new users,
    a. if they don't specify a scheme (and ssl is not false), https is forcibly used
    b. if they do specify a scheme, they must use https

We should just cut http out.

patrickhousley: The `ssl` flag was most likely introduced for integration testing only. It was probably never meant to be used by customers.

metal-messiah:
I think we need to regroup on the config values and implementation of this.  We will likely actually need a new property for both values (something like init.proxy.beacon, init.proxy.assets) instead of overwriting beacon/errorBeacon.  This will be necessary for compatibility/sanity with customers reverting back to the standard beacon values in the DB (once those have been changed to a custom value, theres no easy way to get the initial value back for the customer).  

I also can't help but feel in combination with #673 that we may be over-engineering all these url checks.  The combo of those 2 PRs is growing the lite build by more than 5%, for something that should be relatively benign (just adding a property and overwriting the beacon and asset values if set) and unused by the large majority of customers.  Validating the url could likely be accomplished by good docs to follow and less by hard-coding checks and parsing all over the agent so we should circle back on this as well.

cwli24:
> I think we need to regroup on the config values and implementation of this. We will likely actually need a new property for both values (something like init.proxy.beacon, init.proxy.assets) instead of overwriting beacon/errorBeacon. This will be necessary for compatibility/sanity with customers reverting back to the standard beacon values in the DB (once those have been changed to a custom value, theres no easy way to get the initial value back for the customer).
> 
> I also can't help but feel in combination with #673 that we may be over-engineering all these url checks. The combo of those 2 PRs is growing the lite build by more than 5%, for something that should be relatively benign (just adding a property and overwriting the beacon and asset values if set) and unused by the large majority of customers. Validating the url could likely be accomplished by good docs to follow and less by hard-coding checks and parsing all over the agent so we should circle back on this as well.

Reworked: validations removed and the input moved to a new object in the init config.

src/loaders/configure/configure.js
```diff
@@ -28,13 +36,21 @@ export function configure (agentIdentifier, opts = {}, loaderType, forceDrain) {
   const updatedInit = getConfiguration(agentIdentifier)
   runtime.denyList = [
     ...(updatedInit.ajax?.deny_list || []),
-    ...(updatedInit.ajax?.block_internal
-      ? [
-          info.beacon,
-          info.errorBeacon
-        ]
-      : [])
+    ...(updatedInit.ajax?.block_internal ? [info.beacon, info.errorBeacon] : [])
```
metal-messiah, line 39: Does this need to include the proxyBeacon as well, since that would ingest the data if set

metal-messiah: nvm i see it below

src/loaders/api/apiAsync.js
```diff
@@ -66,7 +64,8 @@ export function setAPI (agentIdentifier) {
     const agentInfo = getInfo(agentIdentifier)
     if (!agentInfo.beacon) return
 
-    var url = scheme + '://' + agentInfo.beacon + '/1/' + agentInfo.licenseKey
+    const scheme = (getConfigurationValue(agentIdentifier, 'ssl') === false) ? 'http' : 'https'
+    let url = `${scheme}://${agentInfo.beacon}/1/${agentInfo.licenseKey}`
```
patrickhousley, line 68: Does this need to support the proxy configuration?

src/loaders/configure/configure.js
```diff
@@ -28,13 +36,21 @@ export function configure (agentIdentifier, opts = {}, loaderType, forceDrain) {
...
   ]
+  if (!alreadySetOnce) {
+    alreadySetOnce = true
+    if (init.proxy?.assets) {
+      redefinePublicPath(init.proxy.assets + '/') // much like the info.beacon & init.proxy.beacon, this input should not end in a slash, but one is needed for webpack concat
+      handle(SUPPORTABILITY_METRIC_CHANNEL, ['Config/AssetsUrl/Changed'], undefined, FEATURE_NAMES.metrics, agentEE)
+      runtime.denyList.push(init.proxy.assets)
```
patrickhousley, line 46: This should probably also be dependent upon the `block_internal` setting.

src/loaders/configure/configure.js
```diff
@@ -28,13 +36,21 @@ export function configure (agentIdentifier, opts = {}, loaderType, forceDrain) {
...
+    }
+    if (init.proxy?.beacon) {
+      handle(SUPPORTABILITY_METRIC_CHANNEL, ['Config/BeaconUrl/Changed'], undefined, FEATURE_NAMES.metrics, agentEE)
+      runtime.denyList.push(init.proxy.beacon)
```
patrickhousley, line 50: Same here.

tools/testing-server/plugins/agent-injector/loader-transform.js
```diff
@@ -20,13 +20,14 @@ async function getLoaderContent (request, reply, testServer) {
...
-  if (!loaderFileStats.isFile()) {
+  let file
+  try {
+    file = (await fs.promises.readFile(loaderFilePath)).toString()
+  } catch (_) {
     throw new Error(`Could not find loader file ${loaderFilePath}`)
   }
-
-  return (await fs.promises.readFile(loaderFilePath)).toString()
+  return file
```
patrickhousley, line 30: Is this to keep the server from crashing if the requested loader file is not found?

tools/webpack/configs/common.mjs
```diff
@@ -59,9 +59,12 @@ export default (env, asyncChunkName) => {
         namespace: `NRBA-${env.VERSION}.${env.SUBVERSION}`,
         filename: '[file].map[query]',
         moduleFilenameTemplate: 'nr-browser-agent://[namespace]/[resource-path]?[loaders]',
-        publicPath: env.PUBLIC_PATH,
+        publicPath: env.PUBLIC_PATH, // -- leaving source maps on CDN only as they're not strictly necessary
         append: env.SUBVERSION === 'PROD' ? false : '//# sourceMappingURL=[url]'
       }),
+      new webpack.DefinePlugin({ // passing default asset path as string to agent
+        WEBP_CONST_PUBLIC_PATH: JSON.stringify(env.PUBLIC_PATH)
+      }),
```
patrickhousley, line 67: In my local testing, we don't actually need these changes in the webpack config and we can reduce the public-path.js file down to just setting the global `__webpack_public_path__`. This way, webpack is set with the default public path and it only gets overridden if the function is called to set it.

src/loaders/configure/public-path.js
```diff
@@ -0,0 +1,9 @@
+// Set the default CDN or remote for fetching the assets; NPM shouldn't change this var.
+if (typeof WEBP_CONST_PUBLIC_PATH !== 'undefined') {
+  __webpack_public_path__ = WEBP_CONST_PUBLIC_PATH // eslint-disable-line
+}
+
```
patrickhousley, line 5: Based on my comment in the webpack config, this code can actually be removed.

```suggestion
```

tools/testing-server/plugins/agent-injector/loader-transform.js
cwli24: Yes, I tweaked this a bit to show an actual full error message during a troubleshooting session because it actually threw an unhandled exception before on bad cases.

src/loaders/api/apiAsync.js
patrickhousley: 0.017% > 0% Personally, I don't understand the use-case for this API but it is being used. If those customer's using this API try to use proxy, it's not going to work like it should.

cwli24: Updated to mirror harvest

tools/webpack/configs/common.mjs
cwli24: Proxy e2e still pass after reverting change to this file 👍 . Thanks for finding out

tests/specs/proxy.e2e.js
```diff
@@ -0,0 +1,56 @@
+import { notIE, notSafari, notIOS } from '../../tools/browser-matcher/common-matchers.mjs'
+
+describe('Using proxy servers -', () => {
```
cwli24, line 3: Move `notIE` up to here

patrickhousley (approved):

PR #660, opened 2023-08-25
chore: Internal entitlements check

metal-messiah:
This PR adds another segment to the a-b script, which will hit NG and get user entitlements.  These entitlements will be used to help determine if the SR feature should be lazy-loaded using the `start()` api.
---

### Overview
This PR adds another segment to the a-b script, which will hit NG and get user entitlements.  These entitlements will be used to help determine if the SR feature should be lazy-loaded using the `start()` api.

### Related Issue(s)
NR-154857


.github/actions/build-ab/index.js
```diff
@@ -133,3 +134,68 @@ if (['dev', 'staging'].includes(args.environment)) {
...
 }
+
+// add entitlements NG script shim
+(async () => {
+  if (environment === 'dev') return
+  const entitlementsEndpoints = {
+    dev: '',
+    staging: 'https://staging-one.newrelic.com',
+    prod: 'https://one.newrelic.com',
+    'eu-prod': 'https://one.eu.newrelic.com'
```
patrickhousley, line 145: Are these not the same domains for the UI? Couldn't we just use a relative URL in the XHR?

.github/actions/build-ab/index.js
```diff
@@ -133,3 +134,68 @@ if (['dev', 'staging'].includes(args.environment)) {
     }
   }
 }
+
+// add entitlements NG script shim
+(async () => {
```
patrickhousley, line 139: You do not need to wrap this in an `iife`. For github actions, you can assume a global `async` is present.

.github/actions/build-ab/index.js
```diff
@@ -133,3 +134,68 @@ if (['dev', 'staging'].includes(args.environment)) {
...
+  if (environment === 'dev') return
+  const entitlementsEndpoints = {
+    dev: '',
+    staging: 'https://staging-one.newrelic.com',
+    prod: 'https://one.newrelic.com',
+    'eu-prod': 'https://one.eu.newrelic.com'
+  }
+  const entitlementsNG = `
+  try {
```
patrickhousley, line 148: So I was considering using a templating library before when I was converting the original scripts to a github action. I decided against it because there was no code that needed to be injected. Since that is changing, maybe we should consider EJS or Handlebars.

.github/actions/build-ab/index.js
metal-messiah: yeah i guess thats true.  we can go to a relative url.  Theres the off chance that the API and the UI segregate at some point but i think that we can cross that bridge if and when that ever happens.

.github/actions/build-ab/index.js
metal-messiah: cool didnt know that

.github/actions/build-ab/index.js
patrickhousley: Yea that would be cool. I have no preference to what we use. Just something easy!!!

.github/actions/build-ab/index.js
```diff
@@ -6,43 +6,11 @@ import { S3Client, ListObjectsCommand } from '@aws-sdk/client-s3'
...
-        'bam-cell.nr-data.net'
-      ]
-    },
-    session_replay: {
-      enabled: true,
-      sampleRate: 0.5,
-      errorSampleRate: 1
-    }
-  },
+const template = Handlebars.compile(await fs.promises.readFile('./template.js', 'utf-8'))
```
patrickhousley, line 11: If this works, good. I just want to put here that I, personally, would prefer an absolute path to the file being read. It gives me warm and fuzzies that something stupid about the CI/CD environment isn't going to break it.

.github/actions/build-ab/template.js
```diff
@@ -0,0 +1,87 @@
...
+      applicationID: {{{args.appId}}}
+    },
+    info: {
+      beacon: 'staging-bam.nr-data.net',
+      errorBeacon: 'staging-bam.nr-data.net',
+      licenseKey: {{{args.licenseKey}}},
+      applicationID: {{{args.appId}}},
+      sa: 1
+    }
+  }
```
patrickhousley, line 36: Formatting. This object is indented one too many.

patrickhousley (approved):
This looks good and my comments are more just nitpicks.

metal-messiah:
Do not merge until #656 is complete

PR #664, opened 2023-08-28
chore: actions share node modules

patrickhousley:
---

### Overview

Changed the actions so they share node modules and a package.json. This should resolve the dependency not found error in the shared scripts.


cwli24 (approved):
3rd time's the charm? 🤞

PR #665, opened 2023-08-28
chore: adding runtime sm

patrickhousley:
---

### Overview

This adds three new supportability metrics that are based on the detected runtime environment. This will defer from loader type in cases where the customer is building their own agent from source or npm.

### Related Issue(s)

N/A


metal-messiah (approved):
Approved but don't merge until tests pass

PR #668, opened 2023-08-29
fix: Fixing issue with leaking event listeners

### LINKED ISSUE #556: InteractionTimer leaks event listeners / debounces
**Note:** I hope this is the correct repository. I found my way here via searching the minified sources found in the browser.

## Description

InteractionTimer leaks event listeners / debounces. Or that is my best guess, you should look into it deeper.

If you leave a tab open for a long time (a few days) this will gradually slow down all click and keydown events.

For example, I left a tab open for a week, and click events took 6.2 seconds, and during that the most time was spent in a newrelic script doing hundreds of thousands of repeated `setTimeout` and `clearTimeout` calls.

I did another test on a completely different page, and in four days it slowed click events to 300ms, with the same symptoms.

Searching for the code in the performance trace points to this file https://github.com/newrelic/newrelic-browser-agent/blob/1ee675d232fb9a233c36ad0a2b998739fbaa1b8b/src/common/timer/interaction-timer.js

image

image

## Steps to Reproduce

1. Have a page with newrelic monitoring and a JavaScript click event handler.
2. Leave the page open for a long time, for example multiple days, depending on the device.
3. (I didn't interact with the page much, but I'm not sure if some interaction is required.)
4. Start a performance trace in the browser and trigger the JavaScript click event handler. Or just wait long enough that it slows down to multiple seconds.

## Expected Behavior

Monitoring does not noticeably affect website performance.

This line:

image

Points to this:

image

## Your Environment

Chromium 113.0.5672.63

Also reproduced on Chrome on Android.

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

patrickhousley:
Fixing an issue where session timeout would result in rewrapping global event handlers. For instances where the site would be open for long periods of time, such as kiosks, this would eventually result in a degradation in event handler performance.
---

### Overview

When our session entity timed out, it would re-initialize and re-wrap event handlers. For pages that stay open for long periods of time, this would eventually result in a noticeable performance degradation for event handlers on the page. 

### Related Issue(s)

https://issues.newrelic.com/browse/NR-137712
Closes #556

### Testing

To reproduce this issue, or test the fix, the easiest way is to change the default session timeout values to 30s for the inactivity timer and 60s for the session timer. If you are reproducing the issue using what is currently in main, add console logs to each of the session entity methods to show when those methods are called. What you will find is that those method calls increase with each session timeout. With the fix in place, the number of method calls for a session reset stays static.


src/common/session/session-entity.js
```diff
@@ -42,31 +42,37 @@ export class SessionEntity {
...
+
+    // Only setup EE and wrap events once per instance
+    this.ee = ee.get(agentIdentifier)
+    wrapEvents(this.ee)
+
+    this.setup(opts)
+  }
+
+  setup ({ value = generateRandomHexString(16), expiresMs = DEFAULT_EXPIRES_MS, inactiveMs = DEFAULT_INACTIVE_MS }) {
+    console.log('Session.setup')
```
metal-messiah, line 65: Console

metal-messiah (approved):
Looks good minus the console log, nice find

src/common/session/session-entity.js
patrickhousley: Shit I thought I got all of them lol thanks

cwli24 (commented):
This is just a question, but did you find that our wrap-events does **not** check that it's already wrapped? I ask because I thought wrap-function checks the `flag` and doesn't re-wrap.

patrickhousley:
Good catch @cwli24 I did miss the underlying issue. So the actual problem is in the InteractionTimer where we add a listener to the events EE. When the session times out, we create a new InteractionTimer but the old one never gets fully destroyed because this listener is never destroyed. In fact, not only is this a memory leak, but it also hangs onto the session entity reference and, when an event happens like a click, the session entity is refreshed multiple times, once for every listener created due to the session expiring.

image

src/common/session/constants.js
```diff
@@ -1,3 +1,5 @@
 export const PREFIX = 'NRBA'
-export const DEFAULT_EXPIRES_MS = 14400000
-export const DEFAULT_INACTIVE_MS = 1800000
+// export const DEFAULT_EXPIRES_MS = 14400000
+// export const DEFAULT_INACTIVE_MS = 1800000
```
metal-messiah, line 3: These just need to be set back to the real values

patrickhousley:
jil tests passed [here](https://github.com/newrelic/newrelic-browser-agent/actions/runs/6017644279)

PR #669, opened 2023-08-29
feat: Removing old supportability metrics

patrickhousley:
Removing supportability metrics that are no longer needed including polyfill detection, bfcache usage, body and query bytes sent count, and custom attributes size.
---

### Overview

Removing the below supportability metrics:
- Polyfill detection
- bfcache restores
- body and query bytes sent for each endpoint
- js attributes size

Updating framework detection for appexp to detect:
- nextjs - requires react detection since this is ssr for react
- nuxtjs - requires vue detection since this is ssr for vue
- angular universal - requires angular detection since this is ssr for angular
- svelte - new framework detection for us
- sveltelit - requires svelte detection since this is ssr for svelte
- preact - new framework detection for us
- preact ssr - requires preact detection since this is ssr for preact
- qwik - new framework detection for us - this one is an interesting html first framework
- electron - not really a framework but im curious if anyone is using our agent in their electron apps

### Related Issue(s)

https://issues.newrelic.com/browse/NR-94597
https://source.datanerd.us/agents/angler/pull/480


metal-messiah:
This will likely need to be a `feat` since it affects agent harvests and customer billing (in a good way)

tests/specs/metrics.e2e.js
```diff
@@ -357,15 +200,15 @@ function loaderTypeSupportabilityMetric (loaderType) {
 
     const supportabilityMetrics = unloadSupportMetricsResults.request.body.sm
     expect(supportabilityMetrics).toEqual(expect.arrayContaining([{
-      params: { name: expect.stringContaining('Generic/Version/') },
+      params: { name: expect.stringContaining('Generic/DistMethod/') },
       stats: { c: 1 }
     }]))
     expect(supportabilityMetrics).toEqual(expect.arrayContaining([{
-      params: { name: expect.stringContaining('Generic/DistMethod/') },
+      params: { name: `Generic/LoaderType/${loaderTypesMapped[loaderType]}/Detected` },
```
metal-messiah, line 207: Does this test account for the worker loader type?

patrickhousley: No. It would take more work to support worker here and I am not sure that work is worth it right now.

patrickhousley: No. It would take more work to support worker here and I am not sure that work is worth it right now.

cwli24:
> This will likely need to be a `feat` since it affects agent harvests and customer billing (in a good way)

Are they charged for metrics though? I thought they don't--only for data stored under their own account.

src/features/metrics/aggregate/framework-detection.js
cwli24, line 1: So we know, are these reliable in the manner they're captured? (We don't exactly have every framework setup in test suite)

patrickhousley: I feel like there is a possibility that we may have some false positives/negatives. Like you said, we don't test all the frameworks. I found sample sites for each framework and tested them manually. I don't think we should add integration tests for every one of these frameworks, and/or every version of said framework.

cwli24 (approved):
Just would like to see page restore come back, but otherwise this looks good

Issue #666, opened 2023-08-28
JSErrors on MicroAgent don't report the events


maac4422:
## Description

I'm using MicroAgent but I don't receive the JSErros on the new relic, I tested it forcing a throw exception in my app, and it doesn't reflect that event.

NOTE: I tried the same with Agent and the event is reported fine.

## Steps to Reproduce

```
import { MicroAgent } from '@newrelic/browser-agent/loaders/micro-agent'
// @ts-ignore
import { JSErrors } from '@newrelic/browser-agent/features/jserrors'
// @ts-ignore
import { Ajax } from '@newrelic/browser-agent/features/ajax'
// @ts-ignore
import { Metrics } from '@newrelic/browser-agent/features/metrics'
// @ts-ignore
import { SessionTrace } from '@newrelic/browser-agent/features/session_trace'

const newRelicOptions = {
  init: {distributed_tracing: {enabled: true}, privacy: {cookies_enabled: true}, ajax: {deny_list: [""]}},
  info: {beacon: "", errorBeacon: "", licenseKey: NEW_RELIC_LICENSE_KEY, applicationID: NEW_RELIC_AGENT_ID, sa: 1},
  loader_config: {accountID: NEW_RELIC_ACCOUNT_ID, trustKey: NEW_RELIC_ACCOUNT_ID, agentID: NEW_RELIC_AGENT_ID, licenseKey: NEW_RELIC_LICENSE_KEY, applicationID: NEW_RELIC_AGENT_ID},
}

function createNewRelicLogger() {
  let agent: MicroAgent | null = null

  if (!agent) {
    agent = new MicroAgent({
      ...newRelicOptions,
      features:[
        JSErrors,
        Ajax,
        Metrics,
        SessionTrace
      ]
    })
  }

  function error(error: string | Error, customAttributes?: object | undefined) {
    agent?.noticeError(error, customAttributes)
  }

  function start(){
    agent?.start()
  }

  return {
    error,
    start
  }
}

const newRelicLogger = createNewRelicLogger()

export default newRelicLogger
```

## Expected Behavior

The throw exception report correct on the new relic dashboard

## Your Environment

- React with Vite, and Typescript.

- Browser: Google Chrome

metal-messiah:
Hey 👋 @maac4422 

[...]

PR #670, opened 2023-08-31
fix: Issues with types resolution

patrickhousley:
Fixing issues with typescript throwing errors when trying to resolve the typings for individual features with `"moduleResolution": "node"` set in the `tsconfig.json`. `"moduleResolution": "nodenext"` and `"moduleResolution": "node16"` will enable typescript use of the `exports` field of our `package.json` but could also break other libraries and forces developers to change all local imports to include file extensions.
---

### Overview

Our `package.json` contains `exports` for types but typescript only uses the `exports` field when `"moduleResolution": "nodenext"` is set in the `tsconfig.json`. This setting has other ramifications that may make it impossible for customers to use. To resolve the issue for customers that cannot use `"moduleResolution": "nodenext"`, the `typesVersions` settings have been updated to include type declarations for all the items exported in the `exports` settings.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-155273
closes: #666

### Testing

I added some ts module resolution testing in the vite test wrapper.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9108,9 +9108,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001524",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001524.tgz",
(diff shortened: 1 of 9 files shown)


cwli24 (approved):
Confirmed that the tsc with types-resolver config has no problems with typescript version 4.0(.8), 4.6(.4), or 4.9(.5).

For future reference: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#exports-is-prioritized-over-typesversions

PR #673, opened 2023-09-01
chore: Add support metrics on proxy url change and validate beacon url

cwli24:
---

### Overview

Building on #659 , this include the `beacon` and `errorBeacon` properties of `info` into URL validation when they are set (changed) different from the default NR endpoints. Also, SM is emitted upon either of those or the `init.assetsPath` property successfully changing to detect proxy usage.

### Related Issue(s)

NR-154245

### Testing

New test cases in proxy.e2e.js.
However, browser implementation plus test vendor limitation prevents IE11 and Safari from being auto tested.
- Safari >14 manually verified to work the same as in each of the test cases for other browsers.
- IE11 (and Safari <=14) do not support proxying of neither beacon nor asset chunks but has been manually verified that:
  - passing the public default of `info: { beacon|errorBeacon: 'bam.nr-data.net' }` configuration does not change existing behavior.
  - passing `beacon` or `errorBeacon` as any other value, including being `undefined` in the `info` object results in fallback to the public default URL of `https://bam.nr-data.net/` and does not break functionality of agent.
  - setting the `assetsPath` also does nothing but does not break functionality either.

^ These could use more testing scrutiny.


src/common/url/check-url.js
patrickhousley, line 1: Would like a better name for this file so it is not confused for something that could be used anywhere else in the agent.

src/common/url/check-url.js
```diff
@@ -0,0 +1,37 @@
...
+import { warn } from '../util/console'
+import { isIE } from '../constants/runtime'
+
+/**
+ * Checks input string to see if it's acceptable as agent's public path for loading source code.
+ * @param {string} string - supposed url string
+ * @param {boolean} [allowUnsecure] - when scheme isn't included in url string, this allows http: to be prepended if true
+ * @returns Absolute url if 'string' is accepted. Empty string if it's denied.
+ */
+export function validateServerUrl (string, allowUnsecure = false, localTest = false) {
```
patrickhousley, line 11: We need a better way of altering this code for local testing rather than a parameter. We need a way to allow insecure for our testing purposes but remove support for it in our production builds.

src/common/url/check-url.js
```diff
@@ -0,0 +1,37 @@
...
+  if (!string || typeof string !== 'string') return ''
```
patrickhousley, line 12: What if we returned undefined instead of a string?

cwli24:
Merged into #659 as one PR after re-work.

PR #675, opened 2023-09-03
chore: fixed issues in the CI pipeline

metal-messiah:
---


.github/actions/shared-utils/fetch-retry.js
```diff
@@ -16,12 +16,16 @@ export async function fetchRetry (url, options) {
         retry: undefined
       })
 
-      if (result.ok) {
+      status = result?.status,
```
patrickhousley, line 19: This actually doesn't make sense to have in a while loop any more since it is a recursive fn call.

patrickhousley (approved):
Thanks for fixing all these issues.

PR #677, opened 2023-09-05
feat: Add SR to pro and spa builds (only runs for LP customers)

metal-messiah:
Add the SR feature to Pro and Pro+Spa builds of the agent. The session replay library shipping with this version of the browser agent is in *limited preview* and is not turned on by default. To use the feature, users will need to be part of the limited preview customer group and configure their browser application settings in the UI.
---

### Overview
This PR exposes the SR feature to the pro and spa builds of the agent.

### Related Issue(s)
https://issues.newrelic.com/browse/NR-156695

### Testing
Tests should be unaffected


patrickhousley (approved):
Just a couple of QQ. Should we add the two additional bundles created for SR to the PR size check output? Should we add additional details to our library size doc page regarding the two additional bundles required for SR?

PR #702, opened 2023-09-13
fix: Final harvest does not happen when initial RUM call fails

cwli24:
Fixes the final harvest behavior in cases when the RUM request fails. The initial request is meant to act as a gateway for the agent to start reporting data, and its failure should result in the agent not sending anything. As of previous version, the metrics and page view timing features do not follow this rule, and that has been corrected.
---

### Overview

This cuts down on disassociated data from user page "sessions" wherein their agent never sent any RUM or interval analytics yet does sent disjoint metrics and timings that cannot be tied to anything else of said page session.

### Related Issue(s)

NR-138551

### Testing

Tests added but can also manually verify that following a failed RUM request, nothing else gets send either on interval or on refresh (sendBeacon).

### DIFF
--- a/src/features/metrics/aggregate/index.js
+++ b/src/features/metrics/aggregate/index.js
@@ -32,6 +32,7 @@ export class Aggregate extends AggregateBase {
     // *cli, Mar 23 - Per NR-94597, this feature should only harvest ONCE at the (potential) EoL time of the page.
     scheduler = new HarvestScheduler('jserrors', { onUnload: () => this.unload() }, this)
     scheduler.harvest.on('jserrors', () => ({ body: this.aggregator.take(['cm', 'sm']) }))
+    this.ee.on(`drain-${this.featureName}`, () => { scheduler.started = true }) // this is needed to ensure EoL is "on" and sent
 
     this.drain()
   }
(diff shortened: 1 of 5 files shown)


patrickhousley (approved):

src/features/metrics/aggregate/index.js
metal-messiah, line 1: This brings up a good point, but IMO it'd probably be simpler and more reliable if this feature just waited for the `err` flag before setting up the harvester and/or running the `checks()`, like the other features do, and we remove the whole block-err behavior.

cwli24: It could but it wouldn't fix it for PVT as that doesn't depend on any flag.

metal-messiah: Where in this PR is the problem being addressed for PVT?

cwli24: PVT `startTimer` is tied to the drain event, but it's scheduler initializing is not. It's not just PVT as this requires all features to have "started" before being allowed to final harvest.

metal-messiah (approved):

PR #703, opened 2023-09-14
chore: always upload code coverage

patrickhousley:
---

### Overview

Even if wdio testing failes, when coverage is enabled, coverage will always be uploaded.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-157916

### DIFF
--- a/.github/workflows/wdio-single-browser.yml
+++ b/.github/workflows/wdio-single-browser.yml
@@ -88,12 +88,12 @@ jobs:
             ${{ inputs.additional-flags || '' }}
       - name: Find pull request
         id: pull-request-target
-        if: ${{ inputs.coverage }}
+        if: ${{ !cancelled() && inputs.coverage }}
         uses: ./.github/actions/find-pull-request
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/wdio-single-browser.yml
```diff
@@ -88,12 +88,12 @@ jobs:
             ${{ inputs.additional-flags || '' }}
       - name: Find pull request
         id: pull-request-target
-        if: ${{ inputs.coverage }}
+        if: ${{ always() && inputs.coverage }}
```
metal-messiah, line 91: Do we want this to run even if the action is cancelled?

patrickhousley: I think if you cancel a workflow it cancels all jobs and steps...right?

patrickhousley: Just read the warning in the GHA docs. I will change this real quick.

metal-messiah (approved):

PR #706, opened 2023-09-14
feat: Removing hash from chunk asset name

patrickhousley:
Removing the hash from the chunk asset file name. This will allow internal and external automation to predict the file name of each asset chunk as long as the loader type and version number is known.
---

### Overview

Removing the hash from the chunk asset file name. This will allow internal and external automation to predict the file name of each asset chunk as long as the loader type and version number is known.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-160947

### DIFF
--- a/tools/webpack/configs/common.mjs
+++ b/tools/webpack/configs/common.mjs
@@ -47,7 +47,7 @@ export default (env, asyncChunkName) => {
 
         return env.SUBVERSION === 'PROD' ? `[name]${env.PATH_VERSION}.js` : '[name].js'
       },
-      chunkFilename: env.SUBVERSION === 'PROD' ? `[name].[chunkhash:8]${env.PATH_VERSION}.min.js` : `[name]${env.PATH_VERSION}.min.js`,
+      chunkFilename: `[name]${env.PATH_VERSION}.min.js`,
       path: env.paths.build,
       publicPath: env.PUBLIC_PATH,
       clean: false,


tools/webpack/configs/common.mjs
```diff
@@ -47,7 +47,7 @@ export default (env, asyncChunkName) => {
 
         return env.SUBVERSION === 'PROD' ? `[name]${env.PATH_VERSION}.js` : '[name].js'
       },
-      chunkFilename: env.SUBVERSION === 'PROD' ? `[name].[chunkhash:8]${env.PATH_VERSION}.min.js` : `[name]${env.PATH_VERSION}.min.js`,
+      chunkFilename: `[name]${env.PATH_VERSION}.min.js`,
```
metal-messiah, line 50: Just want to double check here,  is there any chance that our internal deployments, which often share the same version number, will all try to point at the same async chunk since there is no hash differentiation?

ie does 1.240.0-experiment-1 behind the scenes still just point at nr-spa-1.240.0, same as regular 1.240.0?

patrickhousley: No. Experiments are contained in a sub-directory of the S3 bucket `/experiments/<branch-name>`. We only ever have one version of the agent in the `/dev/` and `/test/` sub-directories.

metal-messiah (approved):

PR #707, opened 2023-09-14
chore: Make SR firstTimestamp more reliable

metal-messiah:
Set the first timestamp earlier and more often to try to prevent undefined timestamps from being processed
---

### Overview
There were a small number of Session Replay events being observed with an `undefined` firstTimestamp value.  This is an effort to reduce the likelihood of that.

### Related Issue(s)
NR-160973


src/features/session_replay/aggregate/index.js
```diff
@@ -262,6 +262,8 @@ export class Aggregate extends AggregateBase {
       warn('Recording library was never imported')
       return this.abort()
     }
+    this.clearTimestamps()
+    this.setTimestamps({ timestamp: Date.now() })
```
patrickhousley, line 266: Maybe this should use the time we capture when the agent first loads.

patrickhousley (approved):

src/features/session_replay/aggregate/index.js
```diff
@@ -213,6 +216,8 @@ export class Aggregate extends AggregateBase {
   getHarvestContents () {
     const agentRuntime = getRuntime(this.agentIdentifier)
     const info = getInfo(this.agentIdentifier)
+    const firstTimestamp = this.timestamp.event.first || this.timestamp.cycle.first
+    const lastTimestamp = this.timestamp.event.last || this.timestamp.cycle.last
```
cwli24, line 220: Edge case where either the `event.first` or `event.last` exists but not the other? Then calculation would be using one of `event` and one of `cycle`. Is that guaranteed to be impossible?

cwli24: Not an issue

PR #712, opened 2023-09-15
chore: New experiments publishing process

patrickhousley:
---

### Overview

New experiments and A/B script publishing process:
- Experiments workflow will now ask which environment to publish the experiment to. Only that environment's A/B scripts will be rebuilt after the experiment is published into a sub-directory of `/experiments/{{environment}}/` in the S3 bucket.
  - Experiments can still be published to both `dev` and `staging` but it must be done by running the workflow twice, once for each environment.
- The publish-release and publish-dev workflow will no longer pass different values for the currently released and next versions of the loader.
- The publish-release workflow will no longer gate the publishing of the `current` loader on the publishing of the A/B environment scripts.
- Loader version loading order has been changed:
  - The latest released loader version will be the first loader to execute on internal environments and will report data to the internal NR1 account.
  - The latest loader built from the main branch will be the second loader to execute on internal environments and will report to internal Browser account.
  - Experiments will be the third loader(s) to execute on internal environments and will report to internal Browser account.
  - A final postamble script will be loaded for any post loader code that we need to run in the environment after the loader has started but before the page load event fires.
- The latest released, latest main, experiments, and postamble scripts all load independently to ensure a bug in one script does not keep the other loaders from running.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-160203

### DIFF
--- a/.github/actions/build-ab/index.js
+++ b/.github/actions/build-ab/index.js
@@ -1,6 +1,7 @@
 import fs from 'fs'
 import path from 'path'
 import url from 'url'
+import process from 'process'
 import { AssumeRoleCommand, STSClient } from '@aws-sdk/client-sts'
 import { S3Client, ListObjectsCommand } from '@aws-sdk/client-s3'
 import { v4 as uuidv4 } from 'uuid'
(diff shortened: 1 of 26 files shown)


.github/actions/build-ab/index.js
```diff
@@ -9,84 +10,131 @@ import { fetchRetry } from '../shared-utils/fetch-retry.js'
...
-  path.resolve(__dirname, '../../../temp'),
-  `${args.environment}.js`
-)
-await fs.promises.mkdir(path.dirname(outputFile), { recursive: true })
-if (fs.existsSync(outputFile)) {
-  await fs.promises.rm(outputFile)
+// 1. Create the released environment script
+const releasedScriptRequest = await fetchRetry(`${args.released}?_nocache=${uuidv4()}`, { retry: 3 })
+
+if (!releasedScriptRequest.ok() || typeof releasedScriptRequest.text !== 'function') {
```
metal-messiah, line 22: if fetchRetry doesn't return the fetch promise (which i think is what happens if it fails to load 3 times), .ok() will not be a function and that will probably be the error that throws

something like `.ok is not a function`

.github/actions/build-ab/index.js
```diff
@@ -9,84 +10,131 @@ import { fetchRetry } from '../shared-utils/fetch-retry.js'
...
-  while (isTruncated) {
-    const { IsTruncated, NextMarker, CommonPrefixes } = await s3Client.send(listCommand)
-    if (CommonPrefixes) CommonPrefixes.forEach(prefix => experimentsList.add(prefix.Prefix))
-
-    if (IsTruncated) {
-      listCommand.input.Marker = NextMarker
-    } else {
-      isTruncated = false
-    }
+  if (!latestScriptRequest.ok() || typeof latestScriptRequest.text !== 'function') {
```
metal-messiah, line 42: same here

.github/actions/build-ab/index.js
```diff
@@ -9,84 +10,131 @@ import { fetchRetry } from '../shared-utils/fetch-retry.js'
...
+    latestScriptOutput,
+    releasedScriptTemplate({
+      args, latestScript
+    }),
+    { encoding: 'utf-8' }
+  )
+  fileList.push(latestScriptOutput)
+
+  // 3. Creating experiments script
+  if (!args.role || !args.bucket || !args.region) {
```
metal-messiah, line 59: Can you please add a check/message for the env vars that AWS uses as well?

patrickhousley: There is a check in the action script. If the env variables are not set, it just skips the experiments. If I add it here, it will throw an error. We only ever need to connect to S3 when building for dev and staging environments.

.github/actions/build-ab/index.js
patrickhousley: Fixed

.github/actions/build-ab/index.js
patrickhousley: Fixed

.github/actions/build-ab/index.js
```diff
@@ -9,84 +10,131 @@ import { fetchRetry } from '../shared-utils/fetch-retry.js'
...
+    console.warn('Skipping experiments; AWS role, bucket, and region must be defined.')
+  } else if (!process.env.AWS_ACCESS_KEY_ID || !process.env.AWS_SECRET_ACCESS_KEY) {
+    console.warn('Skipping experiments; AWS credentials missing in environment variables.')
   } else {
```
metal-messiah, line 63: Do we want this check to run if its in prod or eu?

.github/actions/build-ab/templates/postamble.js
```diff
@@ -97,4 +45,4 @@ try {
   if (!!newrelic && !!newrelic.noticeError) newrelic.noticeError(err)
 }
 
-if (!!newrelic && !!newrelic.setApplicationVersion && '{{{env}}}' === 'staging') newrelic.setApplicationVersion( '' + Math.floor(Math.random() * 10) + '.' + Math.floor(Math.random() * 10) + '.' + Math.floor(Math.random() * 10) )
+if (!!newrelic && !!newrelic.setApplicationVersion && '{{{args.environment}}}' === 'staging') newrelic.setApplicationVersion( '' + Math.floor(Math.random() * 10) + '.' + Math.floor(Math.random() * 10) + '.' + Math.floor(Math.random() * 10) )
```
metal-messiah, line 48: Should we have it set the NREUM global back to the original values here?

metal-messiah: yep

PR #713, opened 2023-09-16
chore: SR - handle meta nodes more gracefully

metal-messiah:
handle meta nodes more gracefully
---

### Overview
A new discovery found that the replayer actually needs BOTH the meta node, and subsequent snapshot node in order to render properly.  This is an effort to keep those events together.

### Testing
Existing tests should continue to pass.  A new test will be added to check for meta flag

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -64,6 +64,8 @@ export class Aggregate extends AggregateBase {
      * -- When visibility changes from "hidden" -> "visible", it must capture a full snapshot for the replay to work correctly across tabs
     */
     this.hasSnapshot = false
+    /** Payload metadata -- Should indicate that the payload being sent has a meta node. The meta node should always precede a snapshot node. */
+    this.hasMeta = false
     /** Payload metadata -- Should indicate that the payload being sent contains an error.  Used for query/filter purposes in UI */
     this.hasError = false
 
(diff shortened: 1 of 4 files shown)


src/features/session_replay/aggregate/index.js
```diff
@@ -64,6 +64,8 @@ export class Aggregate extends AggregateBase {
      * -- When visibility changes from "hidden" -> "visible", it must capture a full snapshot for the replay to work correctly across tabs
     */
     this.hasSnapshot = false
+    /** Payload metadata -- Should indicate that the payload being sent has a meta node. The meta node should always precede a snapshot node. */
```
patrickhousley, line 67: Can you elaborate on the last sentence in this comment? Is "node" in reference the the network request or part of the object we send to the network.

metal-messiah: "Node" is such a non-descriptive word 😆  What I mean by this is, every time RRWeb emits an "event" (node as I called it above), it flows through our `store()` method. In practice, RRWeb always emits a "meta" node, then a "snapshot" node, but there is a chance that those two could be arbitrarily broken apart by the harvest scheduler, which could have some weird consequences (mostly viewport sizing issues) if the downstream API grabs the snapshot node that has no meta node. 

 This effort is to shove the "meta" node back into the payload if it detects a snapshot without a meta

metal-messiah: theres a (small, but real) chance it could split based on the interval timing (if the interval triggers a harvest after it got a meta node, but before it got a snapshot node), or based on the size (though super small here since the meta nodes are tiny)

patrickhousley: Thanks!

patrickhousley (approved):

PR #717, opened 2023-09-18
chore: fix endless metrics call

metal-messiah:
Fix an issue where metrics unloading would create an endless loop
---

### Related Issue(s)
Fixes endless loop related to #702


cwli24 (approved):
Manually verified in huddle

PR #721, opened 2023-09-19
fix: Validation of SR configurations

metal-messiah:
Update the block selector, mask selector, and mask input setters to handle empty strings
---

### Overview
block_selector should always return [data-nr-block] as well as any customer-supplied configs in a comma delimited string.  This failed when `''` was passed as an argument, as the agent would append `,''` to the string, breaking the agent.

other configs could also benefit from being validated, so those were added here as well.

### Related Issue(s)
NR-162325

### Testing
Tests have been added.  Can be confirmed manually by supplying `''` as the block_selector argument to init.session_replay in a test-server page and inspecting the `initalizedAgents` config property.

### DIFF
--- a/src/common/config/state/init.js
+++ b/src/common/config/state/init.js
@@ -1,9 +1,12 @@
+import { isValidSelector } from '../../dom/query-selector'
 import { DEFAULT_EXPIRES_MS, DEFAULT_INACTIVE_MS } from '../../session/constants'
+import { warn } from '../../util/console'
 import { gosNREUMInitializedAgents } from '../../window/nreum'
 import { getModeledObject } from './configurable'
 
 const model = () => {
   const hiddenState = {
+    mask_selector: '*',
     block_selector: '[data-nr-block]',
     mask_input_options: { password: true }
   }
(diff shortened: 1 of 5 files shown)


src/common/config/state/init.js
```diff
@@ -55,14 +55,14 @@ const model = () => {
         return hiddenState.block_selector
       },
       set block_selector (val) {
-        hiddenState.block_selector += `,${val}`
+        if (val) hiddenState.block_selector += `,${val}`
```
patrickhousley, line 58: Need to be a little more diligent here I think.

```
const queryCheck = (s) => document.createDocumentFragment().querySelector(s)
const isSelectorValid = (selector) => {
  try { queryCheck(selector) } catch { return false }
  return true
}
...
const hiddenState = {
  block_selector: new Set(['[data-nr-block]']),
  mask_input_options: { password: true }
}
...
get block_selector () {
  return Array.from(hiddenState.block_selector)
}
set block_selector (val) {
  if (typeof val !== 'string') return
  val.split(',').forEach(selector => {
    if (isSelectorValid(selector)) hiddenState.block_selector.add(selector)
    else warn(`An invalid session_replay.block_selector was provided: ${selectpr}`)
  })
}
```

tests/specs/session-replay/rrweb-configuration.e2e.js
```diff
@@ -236,6 +236,19 @@ describe.withBrowsersMatching(notIE)('RRWeb Configuration', () => {
 
       expect(JSON.stringify(body).includes('testing')).toBeFalsy()
     })
+
+    it('block_selector: should not extend on empty string argument', async () => {
```
patrickhousley, line 240: This is IMHO something that we should move to a unit test.

src/common/config/state/init.js
metal-messiah: normally i'm not a big proponent of agent-side validation but this makes sense since it is critical to blocking sensitive data :+1:

metal-messiah: probably should apply this validSelector check to other selectors as well, like mask-text selector

patrickhousley: The other issue, or reason to do this, would be that an error is thrown to the console if an invalid selector is provided.

tests/specs/session-replay/rrweb-configuration.e2e.js
metal-messiah: converted to a jest test

patrickhousley (approved):

PR #726, opened 2023-09-20
chore: add SM when SR cant send due to size

metal-messiah:
Add a SM when the SR feature cannot send due to DOM size
---

### Related Issue(s)
https://issues.newrelic.com/browse/NR-163032


metal-messiah:
Needs Angler PR as follow up

patrickhousley:
I could argue this should be a feature since SR has already been put into the release notes as a feature and this could help customers understand why they may not see SRs. Also, maybe this should be an error so it shows up on the customer's error dashboard? Not sure there though.

patrickhousley (approved):

PR #727, opened 2023-09-21
chore: SR validation - only warn if not empty string

metal-messiah:
Only warn if the block config is not an empty string, which is the default supplied by the API if not set
---

### DIFF
--- a/src/common/config/state/init.js
+++ b/src/common/config/state/init.js
@@ -82,7 +82,7 @@ const model = () => {
       },
       set block_selector (val) {
         if (isValidSelector(val)) hiddenState.block_selector += `,${val}`
-        else warn('An invalid session_replay.block_selector was provided and will not be used', val)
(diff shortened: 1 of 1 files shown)


src/common/config/state/init.js
```diff
@@ -82,7 +82,7 @@ const model = () => {
       },
       set block_selector (val) {
         if (isValidSelector(val)) hiddenState.block_selector += `,${val}`
-        else warn('An invalid session_replay.block_selector was provided and will not be used', val)
+        else if (val !== '') warn('An invalid session_replay.block_selector was provided and will not be used', val)
```
patrickhousley, line 85: Still throws a warning for `null` but probably a super edge case.

metal-messiah: I think the group consensus was to not "allow" a null value for block selector anyway so this might be ok

PR #731, opened 2023-09-23
chore: Add more control over the polyfill builds

metal-messiah:
Add more control over the polyfill builds by explicitly building the agent instance
---

### Overview
This is necessary to exclude Session Replay entirely from the polyfill build.  This should be evaluated if this is the desired behavior.

### DIFF
--- a/src/cdn/polyfills/spa.js
+++ b/src/cdn/polyfills/spa.js
@@ -4,4 +4,27 @@
  */
 
 import '../polyfills.js'
-import '../spa'
+import { Agent } from '../loaders/agent'
+
+import { Instrument as InstrumentPageViewEvent } from '../../features/page_view_event/instrument'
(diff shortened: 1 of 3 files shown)


src/cdn/polyfills/spa.js
```diff
@@ -4,4 +4,27 @@
...
+    InstrumentXhr,
+    InstrumentPageViewEvent,
+    InstrumentPageViewTiming,
+    InstrumentSessionTrace,
+    InstrumentMetrics,
+    InstrumentPageAction,
+    InstrumentErrors,
+    InstrumentSpa
+  ],
+  loaderType: 'spa'
```
metal-messiah, line 29: Could even change this loaderType now to indicate its the polyfill build for SM purposes

metal-messiah: @patrickhousley https://source.datanerd.us/agents/angler/pull/490

PR #734, opened 2023-09-25
feat: Add messaging about Session Replay abort behavior

metal-messiah:
Add messaging to signal why a replay aborted, which could happen for a number of reasons.
---

### Overview
This PR adds messaging to signal why a replay aborted

### Testing
Tests should be unaffected


src/features/session_replay/aggregate/index.js
```diff
@@ -253,7 +253,7 @@ export class Aggregate extends AggregateBase {
   onHarvestFinished (result) {
     // The mutual decision for now is to stop recording and clear buffers if ingest is experiencing 429 rate limiting
     if (result.status === 429) {
-      this.abort()
+      this.abort('429')
```
patrickhousley, line 256:
```suggestion
      this.abort('429: Too many requests')
```

patrickhousley (approved):

src/features/session_replay/aggregate/index.js
patrickhousley: image

PR #740, opened 2023-09-27
fix: Set Session Replay first chunk flags more reliably

metal-messiah:
Add logic to set firstChunk flag and replay mode on new replays more reliably.
---

### Overview
This PR increases reliability of SR payload decoration in two ways:
* it adds logic to decorate firstChunk flags better while running in error mode
* it adds logic to ensure replay mode syncs across tab changes

### Related Issue(s)
https://newrelic.slack.com/archives/C04QZSS6XPV/p1695774682575609?thread_ts=1695766285.826099&cid=C04QZSS6XPV

### Testing
Existing tests should continue to pass

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -100,7 +100,11 @@ export class Aggregate extends AggregateBase {
       this.ee.on(SESSION_EVENTS.PAUSE, () => { this.stopRecording() })
       // The SessionEntity class can emit a message indicating the session was resumed (visibility change). This feature must start running again (if already running) if that occurs.
       this.ee.on(SESSION_EVENTS.RESUME, () => {
+        // if the mode changed on a different tab, it needs to update this instance to match
+        const { session } = getRuntime(this.agentIdentifier)
+        this.mode = session.state.sessionReplay
         if (!this.initialized || this.mode === MODE.OFF) return
+        if (this.mode === MODE.ERROR) this.isFirstChunk = true
         this.startRecording()
       })
 
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -59,8 +59,6 @@ export class Aggregate extends AggregateBase {
     /** can shut off efforts to compress the data */
     this.shouldCompress = true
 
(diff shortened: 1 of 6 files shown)


src/features/session_replay/aggregate/index.js
```diff
@@ -100,7 +100,11 @@ export class Aggregate extends AggregateBase {
       this.ee.on(SESSION_EVENTS.PAUSE, () => { this.stopRecording() })
       // The SessionEntity class can emit a message indicating the session was resumed (visibility change). This feature must start running again (if already running) if that occurs.
       this.ee.on(SESSION_EVENTS.RESUME, () => {
+        // if the mode changed on a different tab, it needs to update this instance to match
+        const { session } = getRuntime(this.agentIdentifier)
+        this.mode = session.state.sessionReplay
         if (!this.initialized || this.mode === MODE.OFF) return
+        if (this.mode === MODE.ERROR) this.isFirstChunk = true
```
cwli24, line 107: `isFirstChunk` is a property of the outgoing payload that marks that payload as the first from the page's agent. This logic should belong with the sending of payload which isn't happening within this callback. 

Also, `prepareHarvest` has this line already:
`this.isFirstChunk = !!session.isNew || this.mode === MODE.ERROR`. I'm confused about this flow

cwli24: Oops, I read that wrong; that line isn't in `prepareHarvest` but it really ought to be

metal-messiah: `!!session.isNew` is true forever until that page is closed, so that wont work there, since that would make every payload be `isFirstChunk: true`.  We only reference the isNew part when the agent instance starts up.  Since subsequent tabs will get isNew: false, and only the first tab will get isNew: true

metal-messiah: ```outgoing payload that marks that payload as the first from the page's agent```

Just to be clear, isFirstPayload needs to mark the first payload of a session (including other tabs)

metal-messiah: this is resolved by using the session entity for state keeping

cwli24 (approved):

PR #741, opened 2023-09-27
fix: Tap session entity into storage api for changes across tabs

metal-messiah:
Tap session entity into storage event listener for stateful changes across tabs.
---

### Overview
There are certain cases where the agent needs to update when changes happen in backgrounded tabs. Utilizing the storage API allows for updates from other tabs.

### Testing
A test has been added to ensure that a backgrounded tab could terminate an existing session replay

### DIFF
--- a/src/common/session/session-entity.js
+++ b/src/common/session/session-entity.js
@@ -11,6 +11,7 @@ import { getModeledObject } from '../config/state/configurable'
 import { handle } from '../event-emitter/handle'
 import { SUPPORTABILITY_METRIC_CHANNEL } from '../../features/metrics/constants'
 import { FEATURE_NAMES } from '../../loaders/features/features'
+import { windowAddEventListener } from '../event-listener/event-listener-opts'
 
 export const MODE = {
   OFF: 0,
@@ -31,7 +32,8 @@ const model = {
 export const SESSION_EVENTS = {
   PAUSE: 'session-pause',
   RESET: 'session-reset',
-  RESUME: 'session-resume'
+  RESUME: 'session-resume',
+  CROSS_TAB_UPDATE: 'session-cross-tab-update'
 }
 
 export class SessionEntity {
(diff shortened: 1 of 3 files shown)


src/features/session_replay/aggregate/index.js
```diff
@@ -122,8 +134,10 @@ export class Aggregate extends AggregateBase {
           this.mode = MODE.FULL
           // if the error was noticed AFTER the recorder was already imported....
           if (recorder && this.initialized) {
-            this.stopRecording()
-            this.startRecording()
+            if (globalScope?.document.visibilityState === 'visible') {
```
cwli24, line 137: There are a couple other things going on in this handler that doesn't make sense when page hidden. Is this awaiting feedback from Steve?

src/common/session/session-entity.js
```diff
@@ -31,7 +32,8 @@ const model = {
 export const SESSION_EVENTS = {
   PAUSE: 'session-pause',
   RESET: 'session-reset',
-  RESUME: 'session-resume'
+  RESUME: 'session-resume',
+  CROSS_TAB_UPDATE: 'session-cross-tab-update'
```
cwli24, line 36: Can this just be called `UPDATE`

src/features/session_replay/aggregate/index.js
metal-messiah: We are awaiting Steve on some things yes but lets talk about the other things that should be short circuited if not visible

src/common/session/session-entity.js
cwli24: `SESSION_EVENTS.UPDATE` / `session-update` should only ever happen when the NRBA session got updated, as a subset of storage updates

src/features/session_replay/aggregate/index.js
```diff
@@ -104,6 +104,18 @@ export class Aggregate extends AggregateBase {
         this.startRecording()
       })
 
+      this.ee.on(SESSION_EVENTS.CROSS_TAB_UPDATE, (data) => {
+        if (!this.initialized) return
+        if (this.mode !== MODE.OFF && data.sessionReplay === MODE.OFF) this.abort()
```
cwli24, line 109: abort missing reason

cwli24:
Do existing tests suffice to test for each of these transition behavior in the session update handler?

src/common/session/session-entity.js
metal-messiah: If we do this we should re-arrange some of the code.  Theres some places that are setting the mode directly,  If we tap into the session entity update, we should just have that be the one place that gets set, and then the handler updates the SR state (i think).  Theres some chicken vs egg stuff happening by tapping into a generic update event if we dont restructure some stuff

metal-messiah: I think we might benefit from adding a "type" or something to the emitted event so we can tell the difference between same-page session updates and cross-tab updates

cwli24 (approved):

patrickhousley (approved):
Tested locally and looks good. I still think we should probably move the constants from the session entity class to the constants file.

PR #744, opened 2023-10-02
feat: Remove deprecated API inlineHit

patrickhousley:
Removing the browser agent API inlineHit. This API was deprecated in 2014 and is no longer supported or used by newer versions of the browser agent.
---

### Overview

Removing the deprecated inlineHit api.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-166453


cwli24 (approved):
🟢 if we got PM go-ahead already.

PR #750, opened 2023-10-03
feat: Do not report ajax nodes in session traces if in deny list

metal-messiah:
Prevent ajax nodes derived from calls found in the ajax deny list from populating Session Trace payloads
---

### Overview
I an effort to reduce COGS and costs for customers before GA'ing ST, this PR prevents ajax nodes derived from calls in the ajax deny list from populating Session Trace payloads

### Related Issue(s)
https://issues.newrelic.com/browse/NR-167035

### Testing
Tests have been adjusted to account for STN no longer having NR ajax nodes by default


cwli24:
Can you revise title? It's confusing since `block_internal` default is true.

tests/browser/stn/stn.browser.js
```diff
@@ -69,71 +69,82 @@ function runTests () {
     let fiVal = 30
     let fidVal = 8
 
+    let xhr = new XMLHttpRequest()
+    xhr.open('GET', 'https://api.publicapis.org/entries')
+    xhr.send()
+    xhr.addEventListener('load', () => setTimeout(runTests, 1000))
+
```
cwli24, line 76: I just migrated this file to a jest test in a PR branch, and actual XHR requests don't work well there so let's rethink this

src/features/ajax/aggregate/index.js
```diff
@@ -96,7 +94,7 @@ export class Aggregate extends AggregateBase {
           handle(SUPPORTABILITY_METRIC_CHANNEL, ['Ajax/Events/Excluded/App'], undefined, FEATURE_NAMES.metrics, ee)
         }
         return
-      }
+      } else handle('bstXhrAgg', ['xhr', hash, params, metrics], undefined, FEATURE_NAMES.sessionTrace, ee)
```
cwli24, line 97: Should make this a separate if-check based off `block_internal` value, as stated, rather than relying on the deny-list which may include non-internal URLs .

metal-messiah: I guess what we are really after is that the denylist impacts the STN nodes.  if we are denying AJAX events we should deny STN as well, since this is pre-monetization

metal-messiah: I will revise the desc to better match that

cwli24: IIRC the last time we attempted to move these into behind the deny-list, it affected ingest and was reverted. Has John flip the light to green?

metal-messiah: Session traces are not currently billable -- we are currently just eating any costs associated with storing stn data -- so if anything this will actually reduce COGS. There's no better time to remove data than before it's actually billable.

cwli24 (approved):

PR #758, opened 2023-10-04
fix: Isolate browser interaction node

patrickhousley:
Improving context isolation for events the SPA feature tracks to allow for multiple agent bundles to live on the same page. This is only intended for internal New Relic usage at this time and should not affect customers.
---

### Overview

When running multiple agent bundles on the same page, we get two initial page load interactions but only one interaction for all soft navigations. This was caused by the SPA feature not scoping the scope object it was adding to the events. Adding scoping now results in both agent bundles sending their own interactions for soft navigations.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-152262

### DIFF
--- a/src/features/spa/aggregate/index.js
+++ b/src/features/spa/aggregate/index.js
@@ -20,6 +20,7 @@ import { FEATURE_NAMES } from '../../../loaders/features/features'
 import { AggregateBase } from '../../utils/aggregate-base'
 import { firstContentfulPaint } from '../../../common/vitals/first-contentful-paint'
 import { firstPaint } from '../../../common/vitals/first-paint'
+import { bundleId } from '../../../common/ids/bundle-id'
 
 const {
   FEATURE_NAME, INTERACTION_EVENTS, MAX_TIMER_BUDGET, FN_START, FN_END, CB_START, INTERACTION_API, REMAINING,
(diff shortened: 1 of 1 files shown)


src/features/spa/aggregate/index.js
```diff
@@ -166,7 +167,7 @@ export class Aggregate extends AggregateBase {
     register(FN_START, function (args, eventSource) {
       var ev = args[0]
       var evName = ev.type
-      var eventNode = ev.__nrNode
+      var eventNode = ev[`__nrNode:${bundleId}`]
```
metal-messiah, line 170: Is there any reason to keep the `__nrNode:` piece to this?  Kinda strange syntax

metal-messiah:
since this is potentially changing production feature data we'd be better off making this a fix or feat for visibility (imo)

metal-messiah (approved):
Approved pending comments

src/features/spa/aggregate/index.js
patrickhousley: Not really but honestly I think we need to revisit this and the IDs used in function wrapper, etc. This only supports multiple bundles and does not work for multiple agents of the same bundle. I also wonder if there is a better way to maintain context data like this without modifying the object that is being passed around.

PR #764, opened 2023-10-06
feat: Detect GraphQL operation names and types in AJAX calls

metal-messiah:
Detect GraphQL operation names and types in AJAX calls and decorate AJAX events with that information.
---

### Overview
This PR adds the ability for the agent to parse AJAX request bodies in search of the operation name and type for GQL calls. This should support singular GQL calls as well as batched GQL array bodies.

### Related Issue(s)
https://issues.newrelic.com/browse/NR-166525

### Testing
New jest and e2e tests have been added to validate both standalone and spa ajax calls.


patrickhousley:
Does apollo use xhr or fetch

metal-messiah:
> Does apollo use xhr or fetch

Apollo client uses fetch

src/features/spa/aggregate/serializer.js
```diff
@@ -128,6 +127,11 @@ export class Serializer extends SharedContext {
             nullable(node.dt && node.dt.traceId, addString, true) +
             nullable(node.dt && node.dt.timestamp, numeric, false)
           )
+
+          // add params.gql here
+          var ajaxAttrParts = addCustomAttributes({ ...(params.gql || {}) }, addString)
+          children = children.concat(ajaxAttrParts)
+          attrCount = ajaxAttrParts.length
```
cwli24, line 134: I don't _think_ ajax nodes rn can have children, do you know if this is true and/or this requires consumer parsing?

src/features/ajax/instrument/index.js
```diff
@@ -93,7 +93,7 @@ function subscribeToEvents (agentIdentifier, ee, handler, dt) {
     ctx.xhrGuids = {}
     ctx.lastSize = null
     ctx.loadCaptureCalled = false
-    ctx.params = this.params || {}
+    ctx.params = this.params
```
cwli24, line 96: Is `this.params` never undefined? Or does it not matter if `ctx.params` isnt' an object?

src/features/ajax/instrument/index.js
```diff
@@ -157,6 +157,8 @@ function subscribeToEvents (agentIdentifier, ee, handler, dt) {
 
     this.startTime = now()
 
+    this.body = data
```
cwli24, line 160: This is not needed & can be removed if this PR isn't affecting XHR (which it's not)

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,65 @@
...
+  const contents = parseGQLContents(gql)
+  if (!contents) return
+  if (typeof contents !== 'object' || !contents.query) return
+
+  const operationName = contents.operationName || 'anonymous'
+  const operationType = contents.query.split(' ')[0]
+  return {
+    operationName, // the operation name of the indiv query
+    operationType, // query, mutation, or subscription,
+    operationFramework: 'GraphQL'
```
cwli24, line 22: Is L22 here really needed?

src/features/ajax/aggregate/index.js
```diff
@@ -119,6 +120,9 @@ export class Aggregate extends AggregateBase {
         event.spanTimestamp = xhrContext.dt.timestamp
       }
 
+      // parsed from the AJAX body, looking for operationName param & parsing query for operationType
+      event.gql = params.gql = parseBatchGQL(bodyString)
```
cwli24, line 124:
```suggestion
      event.gql = parseBatchGQL(bodyString)
```

src/features/spa/aggregate/serializer.js
metal-messiah: Ajax nodes use children to report custom attributes today.  Will validate via `experiment` publish in NR1 once the test suite issues have been resolved

src/features/ajax/instrument/index.js
metal-messiah: I think this was an accidental removal.  Will add back in

src/features/ajax/aggregate/index.js
metal-messiah: params.gql can be id'd in the SPA feature, since it references the object context

src/features/ajax/instrument/index.js
metal-messiah: it is... the body context is evaluated in the aggregator via parseGQL() call

metal-messiah: this is necessary for cases where the browser has been extended with a fetch polyfill, which uses xhr under the hood

src/features/ajax/aggregate/gql.js
metal-messiah: i think if we some how centralize this behavior to be extraneous to just graphql (maybe do something similar for lightning framework for instance) it will be helpful to know the framework

metal-messiah: but for this implementation, i think it is not needed since its ignored in the batch processor.  Can remove if desired, but if theres ever a case we use the single parser it will be absent again

src/features/ajax/instrument/index.js
metal-messiah: good catch thanks, yes it does

metal-messiah: added

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,84 @@
...
+ * @property {string} operationFramework Framework responsible for the operation
+ */
+
+/**
+ * @param {string|Object} gql The GraphQL object body sent to a GQL server
+ * @returns {GQLMetadata}
+ */
+export function parseGQL (gql) {
+  if (!gql) return
+  const contents = gql
```
patrickhousley, line 14: Is this an unnecessary variable creation?

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,84 @@
...
+  if (typeof contents !== 'object' || !contents.query || typeof contents.query !== 'string') return
```
patrickhousley, line 15: Can simplify this.

```suggestion
  if (typeof contents !== 'object' || typeof contents.query !== 'string') return
```

src/features/ajax/aggregate/index.js
```diff
@@ -119,6 +120,9 @@ export class Aggregate extends AggregateBase {
         event.spanTimestamp = xhrContext.dt.timestamp
       }
 
+      // parsed from the AJAX body, looking for operationName param & parsing query for operationType
+      event.gql = params.gql = parseBatchGQL(parseGQLContents(this.body)) || parseGQL(parseGQLQueryString(this?.parsedOrigin?.search))
```
patrickhousley, line 124: Could we simplify this by having a single function we call passing the body and search string and assuming the gql module will take care of figuring out whether to use body or search string and whether batching is being used?

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,84 @@
...
+ * @property {string} operationName Name of the operation
+ * @property {string} operationType Type of the operation
+ * @property {string} operationFramework Framework responsible for the operation
+ */
+
+/**
+ * @param {string|Object} gql The GraphQL object body sent to a GQL server
+ * @returns {GQLMetadata}
+ */
+export function parseGQL (gql) {
```
patrickhousley, line 12: Just a general though here but it would be nice if this module had a single export that took in the necessary parameters and returned the metadata. The rest of the function should not be exported and can be tested through the public interface.

src/features/ajax/aggregate/gql.js
metal-messiah: little faster to early return here if the string is empty, but little consequence either way :+1:

src/features/ajax/aggregate/gql.js
metal-messiah: yes, will clean up ty

src/features/ajax/aggregate/index.js
metal-messiah: yeah, good idea

metal-messiah:
self-note: we need to safeguard against bodies that are encoded or compressed or not JSON

src/common/util/type-check.test.js
```diff
@@ -0,0 +1,13 @@
+import { isTrueObject } from './type-check'
+
+test('isTrueObject', () => {
+  expect(isTrueObject([])).toEqual(false)
+  expect(isTrueObject(1)).toEqual(false)
+  expect(isTrueObject('string')).toEqual(false)
+  expect(isTrueObject(new Blob([]))).toEqual(false)
```
patrickhousley, line 7:
```suggestion
  expect(isTrueObject(new Blob([]))).toEqual(false)
  expect(isTrueObject(new Date())).toEqual(false)
  expect(isTrueObject(null)).toEqual(false)
  expect(isTrueObject(undefined)).toEqual(false)
  expect(isTrueObject(function () {})).toEqual(false)
  expect(isTrueObject(/./)).toEqual(false)
  expect(isTrueObject(Math)).toEqual(false)
  expect(isTrueObject(window.location)).toEqual(false)
  expect(isTrueObject(Object.create(null))).toEqual(false)
```

src/common/util/type-check.js
```diff
@@ -0,0 +1,3 @@
+export function isTrueObject (obj) {
```
patrickhousley, line 1:
```suggestion
/**
 * Tests a passed object to see if it is a pure object or not. All non-primatives in JS
 * are technically objects and would pass a `typeof` check.
 * @param {*} obj Input object to be tested
export function isPureObject (obj) {
```

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,97 @@
...
+ * @property {string} operationType Type of the operation
+ * @property {string} operationFramework Framework responsible for the operation
+ */
+
+/**
+ *
+ * @param {object|string} body Ajax request body
+ * @param {string} query Ajax request query param string
+ * @returns {GQLMetadata | undefined}
+ */
```
patrickhousley, line 15:
```suggestion
/**
 * Parses and returns the graphql metadata from a network request. If the network
 * request is not a graphql call, undefined will be returned.
 * @param {object|string} body Ajax request body
 * @param {string} query Ajax request query param string
 * @returns {GQLMetadata | undefined}
 */
```

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,97 @@
...
+
+    opNames.push(operation.operationName)
+    opTypes.push(operation.operationType)
+  }
+
+  if (!opTypes.length) return
+  return {
+    operationName: opNames.join(','), // the operation name of the indiv query -- joined by ',' for batched results
+    operationType: opTypes.join(','), // query, mutation, or subscription -- joined by ',' for batched results
+    operationFramework: 'GraphQL'
```
patrickhousley, line 61: Do we need an additional property here or a different framework name to indicate it was a batched gql call?

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,97 @@
...
+      return
+    }
+  } else contents = gqlContents
+  if (!isTrueObject(contents) && !Array.isArray(contents)) return
+  let isValid = false
+  if (Array.isArray(contents)) isValid = contents.some(x => validateGQLObject(x))
+  else isValid = validateGQLObject(contents)
+  if (!isValid) return
+
+  return contents
```
patrickhousley, line 82:
```suggestion
  let contents
  
  if (!gqlContents || (typeof gqlContents !== 'string' && typeof gqlContents !== 'object')) return
  else if (typeof gqlContents === 'string') {
    try {
      contents = JSON.parse(gqlContents)
    } catch {
      // must be a JSON object
      return
    }
  } else contents = gqlContents
  
  if (!isTrueObject(contents) && !Array.isArray(contents)) return
  
  let isValid = false
  if (Array.isArray(contents)) isValid = contents.some(x => validateGQLObject(x))
  else isValid = validateGQLObject(contents)
  
  if (!isValid) return
  return contents
```

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function parseGQLQueryString (gqlQueryString) {
+  try {
+    if (!gqlQueryString || typeof gqlQueryString !== 'string') return
+    const params = new URLSearchParams(gqlQueryString)
+    return parseGQLContents(Object.fromEntries(params))
+  } catch (err) {
+    // do nothing for now?
```
patrickhousley, line 91:
```suggestion
    // Parsing failed, return undefined
    return
```

src/features/ajax/aggregate/gql.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param {object|string} body Ajax request body
+ * @param {string} query Ajax request query param string
+ * @returns {GQLMetadata | undefined}
+ */
+export function parseGQL ({ body, query } = {}) {
+  if (!body && !query) return
+  const gqlBody = parseBatchGQL(parseGQLContents(body))
+  if (gqlBody) return gqlBody
+  const gqlQuery = parseSingleGQL(parseGQLQueryString(query))
+  if (gqlQuery) return gqlQuery
```
patrickhousley, line 21: You could probably reduce the overall `try..catch` blocks in this module by just putting the body of this function in a `try..catch`. If any error is thrown from any of the other functions, parsing should be considered failed and undefined should be returned. It is also the top level function of this module and provides the most protection from errors in the future.

src/features/ajax/aggregate/gql.js
metal-messiah: I think we could add that later if necessary.  Unclear if that will be useful and could increase ingest arbitrarily (same could be said for framework tbh, but rn that seems more "needed" as it is the only thing that indicates what the data is representing)

patrickhousley (approved):

PR #765, opened 2023-10-06
fix: Add first harvest of session flags to RUM and Trace

cwli24:
Mark the first RUM call and Session Trace harvest of any given session so that it becomes queryable in New Relic.
---

### Overview

Flag in query param of respective request indicating it is the first outgoing payload of a given session.

### Related Issue(s)

NR-166515
NR-166516

### Testing

New test cases added


src/common/session/session-entity.js
```diff
@@ -140,7 +141,7 @@ export class SessionEntity {
 
     // The fact that the session is "new" or pre-existing is used in some places in the agent.  Session Replay and Trace
     // can use this info to inform whether to trust a new sampling decision vs continue a previous tracking effort.
-    if (this.isNew === undefined) this.isNew = !Object.keys(initialRead).length
+    this.isNew = !Object.keys(initialRead).length
```
metal-messiah, line 144: This is going to break `isNew` since it uses the undefined check in tandem with line 229 in the `reset()` call

src/common/session/session-entity.js
```diff
@@ -25,9 +25,10 @@ const model = {
   inactiveAt: 0,
   expiresAt: 0,
   updatedAt: Date.now(),
-  sessionReplay: MODE.OFF,
+  sessionReplayMode: MODE.OFF,
   sessionReplaySentFirstChunk: false,
   sessionTraceMode: MODE.OFF,
+  traceHarvestStarted: false,
```
metal-messiah, line 31: Is this necessary since you will be able to tell that a session `isNew` per page?  only report firstHarvest flag if isNew?

cwli24: Falls into the same idea that the SR has with the reason why SentFirstChunk was added -- the real first trace payload can come from any tab theoretically.

This isn't an issue with RUM (flag) however, bc that is unconditionally always sent at load of a new tab, basing off `isNew` makes sense there.

src/common/session/session-entity.js
cwli24: I removed this bc I think it's extraneous precisely because `reset` will erase the `isNew` property for the `session` manager. Whenever `setup` is ran, `isNew` should be undefined anyways _and_ `isNew` should be set anew in tandem with `initialRead` state.

src/common/session/session-entity.js
metal-messiah: Just want to double check here -- Is there a race condition though were both tabs could read from session state as `traceHarvestStarted: false` before it has updated, theoretically sending 2 payloads with 

`!this.agentRuntime.session.state.traceHarvestStarted`

cwli24: Yeah, so the new `SESSION_EVENTS.UPDATE` just needs to be listened on, will add into stn

cwli24: Wait, the session-entity listener to storage updates already manages this. That will update the session.state object.

src/features/page_view_event/aggregate/index.js
```diff
@@ -69,6 +69,8 @@ export class Aggregate extends AggregateBase {
       at: info.atts
     }
 
+    if (agentRuntime.session) queryParameters.fsh = Number(agentRuntime.session.isNew) // "first session harvest" aka RUM request or PageView event of a session
```
metal-messiah, line 72: Everything looks good, just one quick philosophical question here,  Do we want to always send the `fsh` (like shown here), or should we just only send it if its truthy?

metal-messiah:
Approved pending re-run of PR check tests before merging

metal-messiah (approved):

src/features/page_view_event/aggregate/index.js
cwli24: If it's not sent at all, there's no session tracking running; distinct from sending '0' ig. That might matter if you're trying to find statistics.

Also does downstream need to be aware of this going as `fsh` before merge? Does it matter what it is as long as they know the exact string?

metal-messiah: They have not done the work. It will just have no effect (dropped) until they implement the property on their end

PR #766, opened 2023-10-06
feat: Isolate spa event observation context

patrickhousley:
Introducing a new agent instance private observation context store and implementing it for the isolation of observation contexts for SPA events. This makes it possible to run multiple agents from the same code bundle without interference causing one agent to not report observation data. The observation context also removes the need to alter the event objects. The private observation context store will be extended into other parts of the agent over time.
---

### Overview

Creating a new ObservationContext class that utilizes a WeakMap to allow agent instances to associate context data with observed scopes like objects, callback functions, promises, etc.

The first POC use-case included here is a rewrite of #758 using the observation context instance instead of modifying the event object directly.

The ticket has been updated with screenshots show the WeakMap properly allowing garbage collection and browser interactions still being reported correctly.

### Related Issue(s)

https://issues.newrelic.com/browse/NR-168327
https://issues.newrelic.com/browse/NR-168326

### DIFF
--- a/src/common/context/observation-context.js
+++ b/src/common/context/observation-context.js
@@ -1,0 +1,20 @@
+import { gosNREUM } from '../window/nreum'
+
+export class ObservationContext {
+  static getObservationContextByAgentIdentifier (agentIdentifier) {
+    const nr = gosNREUM()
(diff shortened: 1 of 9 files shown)


patrickhousley:
Superseded by #903

PR #768, opened 2023-10-09
feat: Add network info to all page view timing events

cwli24:
Add client network information to all PageViewTiming events in New Relic. The attributes currently only exist on LCP and FI event types and will now be reported with all others such as load and pagehide events.
---

### Overview

FR to add network attributes to other pvt nodes. It'll be shown now via NRQL and may be picked up by UI in the future to highlight the at-the-time network conditions when certain metrics are captured.

### Related Issue(s)

NR-168402

### Testing

Modified existing tests.

### DIFF
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -94,6 +94,7 @@ export class Aggregate extends AggregateBase {
 
   addTiming (name, value, attrs) {
     attrs = attrs || {}
+    addConnectionAttributes(attrs)
 
     // If cls was set to another value by `onCLS`, then it's supported and is attached onto any timing but is omitted until such time.
     /*
@@ -174,4 +175,14 @@ export class Aggregate extends AggregateBase {
 
     return payload
   }
+}
+
+function addConnectionAttributes (obj) {
+  var connection = navigator.connection || navigator.mozConnection || navigator.webkitConnection // to date, both window & worker shares the same support for connection
+  if (!connection) return
+
+  if (connection.type) obj['net-type'] = connection.type
+  if (connection.effectiveType) obj['net-etype'] = connection.effectiveType
+  if (connection.rtt) obj['net-rtt'] = connection.rtt
+  if (connection.downlink) obj['net-dlink'] = connection.downlink
 }
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -94,7 +94,7 @@ export class Aggregate extends AggregateBase {
 
   addTiming (name, value, attrs) {
     attrs = attrs || {}
-    addConnectionAttributes(attrs)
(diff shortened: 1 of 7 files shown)


src/features/page_view_timing/aggregate/index.js
```diff
@@ -94,6 +94,7 @@ export class Aggregate extends AggregateBase {
 
   addTiming (name, value, attrs) {
     attrs = attrs || {}
+    addConnectionAttributes(attrs)
```
metal-messiah, line 97: By moving this out of the vital wrapper, any objects pulled from the buffer will get the network information at the time of calling, not the time they were observed. 

Before:
A metric flowed through the vitalmetric class --> it was stored with network info at the time of flowing through --> saved in buffer --> future subscribers would get that value, which would have the network attributes at the time of saving.  This is likely an edge case, but i'm not sure its the best way to ensure accuracy.  Thoughts there?

Now:
By moving this out of the vital wrapper, any objects pulled from the buffer will get the network information at the time of calling, not the time they were observed. 

Before:
A metric flowed through the vitalmetric class --> it is stored without any network metrics --> saved in buffer --> future subscribers would get that value, which would have no network attributes at the time of saving, instead getting the network attributes at the time of retrieval.

cwli24: One of those both-sides debates, but imo a) the vital metric class is trying to do too much if it's also tracking network conditions and b) that's more of a job for the overall pvt (aggregate) process. Up until of late, `addTiming` functions as the central point when buffered data is processed into event units--still is--and is called directly by other non-vital timing events,  so it made the most sense to tack on network info there.
Your case specifically targets a subscriber that wants that info with past vitals (or other timings in general) which doesn't exist just yet, but if and when we get to that, the solution should be generalized to apply to all timings. If that's an issue, we need to rethink how #handleVitalmetric is calling addTiming but also having a separate system/pathway of storing another version of the same data if that makes sense. E.g., maybe addTiming should return it's final processed output and if the vital metrics class needs to store something, it should store that instead.

cwli24: We can add it in vital metrics and check in `addTiming` if it exists and add it again if it doesn't (for non-vital timings) but splitting it up that way causes confusion.

metal-messiah: I think it would just be simplest to add it to the vital timing at the time it was recorded.  If the consumers of the metric use or dont use it is inconsequential, because at the end of the day we just want to record it when it happens and not later, even if that gap is small or an edge case -- simply to save ourselves from ourselves later if the consumption of those singletons' subscriptions ever gets deferred in some way.

metal-messiah (approved):
Approved pending a comment about the async connection attributes behavior

src/features/page_view_timing/aggregate/index.js
cwli24, line 97:
```suggestion
    addConnectionAttributes(attrs) // network conditions may differ from the actual for VitalMetrics when they were captured
```

PR #769, opened 2023-10-10
chore: support development code blocks in build:

patrickhousley:
---

### Overview

The build has been updated to support us adding code blocks that are meant for local or development builds of the agent.

```
// develblock:start
console.log('This code will be removed in production builds and npm package builds')
// develblock:end
```

### Related Issue(s)

This is a possible solution for https://issues.newrelic.com/browse/NR-169078

### Testing

Copy the above code block into any source file and run an NPM build or production webpack build. The output files should not contain the block of code. Likewise, running a webpack local or dev build should retain the code block.


metal-messiah (commented):
Theres some challenges here we need to think through:
* we dont test the prod build
* we do test the dev build (which would have develblocks)

The potential problem:
This concept has the potential to break the prod build if implemented incorrectly (ie, devel block is added for some code, but not for the consumer(s) of that code).  This is compounded by the fact that the prod build is not tested, meaning broken code could ship and we wouldnt know it until its being consumed somewhere (NR1 or worse).

What is the plan to ensure we can protect ourselves from ourselves before merging a PR with this in place?

PR #776, opened 2023-10-18
feat: Allow boolean values in setCustomAttribute

metal-messiah:
Allow boolean values to be supplied to newrelic.setCustomAttribute() API calls.
---

### Overview
Old versions of the agent allowed boolean values for setCustomAttribute, which did not match our documentation.  An effort was conducted some time ago to align the agent with the docs, and booleans were forcibly denied.  This is causing some migration issues for some customers, and there is no clear reason to exclude booleans from this API.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-171312

### Testing
Tests have been updated to account for the change

### DIFF
--- a/src/loaders/api/api.component-test.js
+++ b/src/loaders/api/api.component-test.js
@@ -274,12 +274,12 @@ describe('setAPI', () => {
       expect(console.warn).toHaveBeenCalledWith(expect.stringContaining('Name must be a string type'))
     })
 
-    test.each([undefined, {}, []])('should return early and warn when value is not a string, number, or null (%s)', (value) => {
+    test.each([undefined, {}, [], Symbol])('should return early and warn when value is not a string, number, or null (%s)', (value) => {
       const args = [faker.datatype.uuid(), value]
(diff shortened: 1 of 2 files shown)


src/loaders/api/api.component-test.js
```diff
@@ -274,12 +274,12 @@ describe('setAPI', () => {
       expect(console.warn).toHaveBeenCalledWith(expect.stringContaining('Name must be a string type'))
     })
 
-    test.each([undefined, {}, []])('should return early and warn when value is not a string, number, or null (%s)', (value) => {
+    test.each([undefined, {}, [], Symbol])('should return early and warn when value is not a string, number, or null (%s)', (value) => {
```
patrickhousley, line 277:
```suggestion
    test.each([undefined, {}, [], Symbol('foobar')])('should return early and warn when value is not a string, number, or null (%s)', (value) => {
```

patrickhousley (approved):

metal-messiah:
Docs Site PR --> https://github.com/newrelic/docs-website/pull/14908/files

PR #785, opened 2023-10-23
chore: Expose base agent instance under initializedAgents property

cwli24:
---

### Overview

This PR:

- tidies up multiple places calling `gosNREUMInitializedAgents` with different properties into a singular call that assigns the new agent instance onto the newrelic.initializedAgents global object.
- introduces the assignment and modification of new or existing properties on the base agent instance itself and passing that reference directly to other functions.

### Related Issue(s)

Derived during revisions on #626 

### Testing

Checks everything still works or configures correctly, and that the `newrelic.initializedAgents` interface has the same layout (e.g., `newrelic.initializedAgents[<id>].api` still holds all the api associated with that agent) and superset of information as the existing interface:
-NEW-
image

-OLD-
image

set___ calls should keep the respective property under the respective agent (identifier) updated on the global.


src/loaders/agent.js
```diff
@@ -46,7 +46,9 @@ export class Agent extends AgentBase {
     // Future work is being planned to evaluate removing this behavior from the backend, but for now we must ensure this call is made
     this.desiredFeatures.add(PageViewEvent)
 
-    Object.assign(this, configure(this.agentIdentifier, options, options.loaderType || 'agent'))
+    gosNREUMInitializedAgents(this.agentIdentifier, this) // append this agent onto the global NREUM.initializedAgents
+    // ^Currently, this needs to happen before configure() since it will add the agent to the global that's looked up by set-config functions inside there.
```
metal-messiah, line 50: should we just put line 49 inside at the top of configure to eliminate the cross-dependency?  this will already exist in there now

src/loaders/configure/configure.js
```diff
@@ -43,13 +46,13 @@ export function configure (agentIdentifier, opts = {}, loaderType, forceDrain) {
...
-  gosNREUMInitializedAgents(agentIdentifier, api, 'api')
-  gosNREUMInitializedAgents(agentIdentifier, exposed, 'exposed')
-  addToNREUM('activatedFeatures', activatedFeatures)
+  setRuntime(agent.agentIdentifier, runtime)
 
-  return api
+  if (!alreadySetOnce) {
+    setTopLevelCallers() // no need to set global APIs on newrelic obj more than once
+    addToNREUM('activatedFeatures', activatedFeatures)
+  }
```
metal-messiah, line 54: nitpick but can you just combine this with the other alreadySetOnce if statement on 37?

src/loaders/micro-agent.js
```diff
@@ -91,8 +92,7 @@ export class MicroAgent extends AgentBase {
           }
         })
       })
-      gosNREUMInitializedAgents(this.agentIdentifier, this.features, 'features')
-      return this
+      return true
```
metal-messiah, line 95: not sure if this actually matters or not, but before it was returning the class instance instead of a boolean

src/loaders/agent.js
cwli24: So the real problem in this case really is just that the first set of set___ calls done by configure won't add (update) those configurations on the class instance itself since they won't find anything on the global yet, should this line go at the bottom of the constructor.
It wouldn't be an issue of course if those calls were just passed the instance argument to which they can just update the property on the instance in the grand scheme of things.

But yes to your comment, this can just be moved up top.

src/loaders/micro-agent.js
cwli24: In theory, not unless someone is actually using the ret value as more than a truth check from `microAgent.run()` -- this just makes more sense since the error case returns `false`. But I also have some failing tests rn so hold that thought!

patrickhousley: I agree with @cwli24 here. The user would already have reference to the agent instance for the purposes of calling `start` so it doesn't make sense that this function returns the agent instance.

src/common/config/state/init.js
```diff
@@ -114,7 +114,8 @@ export function getConfiguration (id) {
 export function setConfiguration (id, obj) {
   if (!id) throw new Error(missingAgentIdError)
   _cache[id] = getModeledObject(obj, model())
-  gosNREUMInitializedAgents(id, _cache[id], 'config')
+  const agentInst = gosNREUMInitializedAgents(id)
+  if (agentInst) agentInst.init = _cache[id]
```
metal-messiah, line 118: same here

src/common/config/state/runtime.js
```diff
@@ -35,5 +35,6 @@ export function getRuntime (id) {
 export function setRuntime (id, obj) {
   if (!id) throw new Error('All runtime objects require an agent identifier!')
   _cache[id] = getModeledObject(obj, model)
-  gosNREUMInitializedAgents(id, _cache[id], 'runtime')
+  const agentInst = gosNREUMInitializedAgents(id)
+  if (agentInst) agentInst.runtime = _cache[id]
```
metal-messiah, line 39: same

src/common/config/state/info.js
```diff
@@ -45,5 +45,6 @@ export function getInfo (id) {
 export function setInfo (id, obj) {
   if (!id) throw new Error('All info objects require an agent identifier!')
   _cache[id] = getModeledObject(obj, model)
-  gosNREUMInitializedAgents(id, _cache[id], 'info')
+  const agentInst = gosNREUMInitializedAgents(id)
+  if (agentInst) agentInst.info = _cache[id]
```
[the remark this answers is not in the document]
cwli24: Assuming it exists on the global is risky & could break; this ties back to https://github.com/newrelic/newrelic-browser-agent/pull/785#discussion_r1370360058 because it wouldn't exist yet on agent startup if not for moving the assignment up to the top. So it's not a known issue now, but some other thing could remove the reference, so it's better to just have it guarded. This'll go away when the _cache is replaced with just modifying `this.info` directly.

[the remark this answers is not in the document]
metal-messiah: I guess the designation I am calling out is that we shouldn't call it "get or set" if it doesn't actually set if not getting.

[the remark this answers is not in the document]
cwli24: I suppose (though that's outlined in the jsdoc). Still, I don't think this is worth the effort to split & update, keeping in mind this'll go away anyways some time later.

metal-messiah (approved):

PR #794, opened 2023-10-27
chore: add webview testing

patrickhousley:
---

### Overview

Adding webview testing. The testing is very basic and just ensures the agent loads and send data.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-164543


.github/workflows/wdio-all-browsers.yml
```diff
@@ -85,3 +85,19 @@ jobs:
       browser-target: ie@11
       additional-flags: -P
     secrets: inherit
+
+  ios-webview:
+    uses: ./.github/workflows/wdio-single-browser.yml
```
metal-messiah, line 90: I think we need to alter the single browser yml to provide a better name since the env can now be altered

(by this i mean there are now two jobs identically named android@*, ios@* -- same with the coverage test, etc)

patrickhousley: Good callout. Fixed in latest push.

PR #796, opened 2023-10-30
fix: Extending Promise with ES6 class

patrickhousley:
Extending the global Promise using ES6 classes after the agent has loaded results in the extended class properties not being present on the created promise instance. This can break third-party libraries and customer code that relies on extending the Promise class with additional properties and methods. Those properties will now be properly copied to the wrapped Promise instance the agent creates.
---

### Overview

Updating the promise wrapper to check for additional instance level properties during construction. If found, the property descriptor for those properties will be applied to the new wrapped promise instance that is created and returned by the agent.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-155491


patrickhousley:
Closing. This breaks jspdf and possibly other libraries that es6 extends Promise.

PR #798, opened 2023-11-01
fix: Wrapped promise allows ES6 sub-classing

patrickhousley:
Sub-classing the global Promise using ES6 classes after the agent has loaded results in the extended class properties not being present on the created promise instance. The wrapper has been moved to use ES6 sub-classing allowing further sub-classing.
---

### Overview

Rewrote the promise wrapper to use ES6 sub-classing of the global Promise.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-155491


patrickhousley:
Closing for now. We can revisit this if we need to after the spa refactor.

PR #800, opened 2023-11-02
feat: Add basic support for deferring the browser agent loader script

metal-messiah:
Add basic support for deferring the browser agent loader script, including exposing a DOM-level event for external use to indicate the agent is successfully running.
---

### Overview
This PR does the following:
* adds an event that fires when the feature-flags module is triggered to indicate the agent is running
* Adds a basic test for `defer`, `async`, and dynamic injection methods of the loader
* Updates the test server to allow for injecting the loader as `default`, `defer`, `async`, and `injection`

### Related Issue(s)
NR-173949
NR-173951
NR-175756

SPIKE --> https://newrelic.atlassian.net/l/cp/3YhWixQz

### Testing
A new test has been added that loads the agent with various script loading tags and tests for the basic events to be reported.


src/common/util/feature-flags.js
```diff
@@ -43,6 +43,9 @@ export function activateFeatures (flags, agentIdentifier) {
     }
   })
   sentIds.add(agentIdentifier)
+
+  // let any window level subscribers know that the agent is running
+  dispatchGlobalEvent({ loaded: true, agentIdentifier })
```
patrickhousley, line 48: What about sending the agent instance instead of the identifier? Thinking about the fact that this will be like a public API and difficult to change later.

tools/testing-server/plugins/agent-injector/loader-transform.js
```diff
@@ -30,6 +22,35 @@ async function getLoaderContent (request, reply, testServer) {
...
+      return `<script type="text/javascript">
+        window.addEventListener('load', function(){
+        let script = document.createElement('script');
+        script.src = "${loaderFilePath}";
+        document.body.append(script);
+        })
+      </script>`
+    default:
+      return `<script type="text/javascript">${await getLoaderContent(loaderFilePath)}</script>`
+  }
```
patrickhousley, line 51: Need to support `nonce` here from latest code in main.

tools/testing-server/plugins/agent-injector/loader-transform.js
```diff
@@ -42,12 +63,13 @@ module.exports = function (request, reply, testServer) {
...
       if (chunkString.indexOf('{loader}') > -1) {
-        const replacement = await getLoaderContent(request, reply, testServer)
+        const loaderFilePath = getLoaderFilePath(request, testServer, !!request.query?.script)
+        const loaderScript = await getLoaderScript(request.query?.script, loaderFilePath)
         done(
           null,
           chunkString.replace(
             '{loader}',
-            `<script type="text/javascript">${sslShim}${replacement}</script>`
+            `<script type="text/javascript">${sslShim}</script>${loaderScript}`
```
patrickhousley, line 72: This `script` should also have a `nonce`

src/common/util/feature-flags.js
metal-messiah: will modify this to only emit the loaded: true portion until bigger architecture changes are made to enable the agent instance comment

tools/testing-server/plugins/agent-injector/loader-transform.js
```diff
@@ -30,6 +22,35 @@ async function getLoaderContent (request, reply, testServer) {
...
+}
+
+async function getLoaderScript (scriptType, loaderFilePath, nonce) {
+  switch (scriptType) {
+    case 'defer':
+      return `<script src="${loaderFilePath}" defer></script>`
+    case 'async':
+      return `<script src="${loaderFilePath}" async></script>`
+    case 'injection':
+      return `<script type="text/javascript">
```
patrickhousley, line 42: All of these script tags need the `nonce`.

patrickhousley (approved):
Approved pending the one comment

PR #803, opened 2023-11-03
feat: Session Replay API

metal-messiah:
Add two new APIs for controlling session replay behaviors -- newrelic.recordReplay() and newrelic.pauseReplay().
---

### Overview
This PR adds two new APIs that can give more control over session replay behaviors:
* newrelic.recordReplay()
    - This API will be ignored if the Session Replay entitlement was not seen
    - This API will cause a replay to start recording if not started
    - This API will cause a replay to resume recording if previously started but not actively recording
    - This API will be ignored if a replay is already in progress
    - This API will shift the "mode" from "off" or "error" to "full" mode, and replays will continue running across pages until stopped by the pauseReplay() API or the session ending.
    - This API will start a session trace if not already in progress. The related session trace will not be stopped by "pauseReplay"
* newrelic.pauseReplay()
    - This API will cause a replay to pause recording if actively recording
    - This API will be ignored if a replay is not already in progress
    - This API will shift the "mode" from "full" or "error" to "off" mode, and replays will not continue running on any subsequent page until resumed using the recordReplay() API.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-167584

### Testing
New tests have been added to validate these behaviors.


src/features/session_replay/aggregate/index.js
```diff
@@ -152,46 +155,67 @@ export class Aggregate extends AggregateBase {
...
    * @param {boolean} errorSample - the true/false state of the error sampling decision
    * @param {boolean} fullSample - the true/false state of the full sampling decision
+   * @param {boolean} ignoreSession - whether to force the method to ignore the session state and use just the sample flags
    * @returns {void}
    */
-  async initializeRecording (entitlements, errorSample, fullSample) {
+  async initializeRecording (errorSample, fullSample, ignoreSession) {
     this.initialized = true
-    if (!entitlements) return
+    if (!this.entitled) return
```
metal-messiah, line 218: Add check to see if it's already recording (from the api) and early return

src/features/session_replay/aggregate/index.js
```diff
@@ -113,6 +113,9 @@ export class Aggregate extends AggregateBase {
     /** Hold on to the last meta node, so that it can be re-inserted if the meta and snapshot nodes are broken up due to harvesting */
     this.lastMeta = undefined
 
+    /** set by BCS response */
+    this.entitled = false
+
     const shouldSetup = (
       getConfigurationValue(agentIdentifier, 'privacy.cookies_enabled') === true &&
       getConfigurationValue(agentIdentifier, 'session_trace.enabled') === true
```
patrickhousley, line 121: Should this be `session_replay.enabled`?

metal-messiah: This is a modest attempt to not run sr if trace is not going to run since every replay needs a trace

src/features/session_replay/aggregate/index.js
```diff
@@ -152,46 +155,67 @@ export class Aggregate extends AggregateBase {
...
-        flagOn,
-        (Math.random() * 100) < getConfigurationValue(this.agentIdentifier, 'session_replay.error_sampling_rate'),
-        (Math.random() * 100) < getConfigurationValue(this.agentIdentifier, 'session_replay.sampling_rate')
-      )).then(() => sharedChannel.onReplayReady(this.mode)) // notify watchers that replay started with the mode
+      this.waitForFlags(['sr']).then(([flagOn]) => {
+        this.entitled = flagOn
+        this.initializeRecording(
+          (Math.random() * 100) < getConfigurationValue(this.agentIdentifier, 'session_replay.error_sampling_rate'),
+          (Math.random() * 100) < getConfigurationValue(this.agentIdentifier, 'session_replay.sampling_rate')
+        )
```
patrickhousley, line 188: Is it possible that this code could execute after a customer has called the API to start a recording? Any risk with this if they do?

src/features/session_replay/aggregate/index.js
```diff
@@ -433,6 +457,13 @@ export class Aggregate extends AggregateBase {
     return this.estimateCompression(this.payloadBytesEstimation + newBytes) + 1000
   }
 
+  actuallyStop (forceHarvest) {
```
patrickhousley, line 460:
```suggestion
  /**
   * Forces the agent into OFF mode so that changing tabs or navigating
   * does not restart the recording. This is used when the customer calls
   * the stopRecording API.
   */
  forceStop (forceHarvest) {
```

Also curious why the contents of this function where not just in the event handler above. Maybe just to keep the event handlers clean and short, which is good.

src/features/session_replay/aggregate/index.js
```diff
@@ -152,46 +155,67 @@ export class Aggregate extends AggregateBase {
...
+        if (this.blocked || !this.entitled) return
+        // if it isnt already (fully) initialized... initialize it
+        if (!recorder) this.initializeRecording(false, true, true)
+        // its been initialized and imported the recorder but its not recording (mode === off || error)
+        else if (this.mode !== MODE.FULL) this.switchToFull()
+        // if it gets all the way to here, that means a full session is already recording... do nothing
+      }, this.featureName, this.ee)
+
+      registerHandler('pauseReplay', () => {
+        this.actuallyStop(this.mode !== MODE.ERROR)
```
patrickhousley, line 169: Shouldn't we force harvest if the mode is error? Confusing.

tests/specs/session-replay/mode.e2e.js
```diff
@@ -107,11 +160,54 @@ describe.withBrowsersMatching(notIE)('Session Replay Sample Mode Validation', ()
...
+      initialized: true,
+      events: expect.any(Array),
+      mode: 0
+    }))
+
+    await Promise.all([
+      browser.execute(function () {
+        newrelic.noticeError(new Error('test'))
+      }), browser.pause(1000)
+    ])
```
patrickhousley, line 207: Maybe add two more test cases:
1. Call the stop API in a script right below the loader. How does calling the api before the page load event affect the session replay state?
2. Call the start API in a script right below the loader. How does calling the api before the page load event affect the session replay state?

src/features/session_replay/aggregate/index.js
metal-messiah: if the mode is error and we force a harvest -- it will take the 30s buffer we've been storing up and harvest it -- which we should only actually want to get a data harvest if running in full mode

tools/testing-server/plugins/test-handle/index.js
```diff
@@ -22,7 +22,8 @@ module.exports = fp(async function (fastify, testServer) {
     }
 
     if (request.query.nonce) {
-      reply.header('content-security-policy', `script-src 'nonce-${request.query.nonce}'`)
+      // Must have 'unsafe-eval' in case build was ran with coverage enabled
+      reply.header('content-security-policy', `script-src 'nonce-${request.query.nonce}' 'unsafe-eval'`)
```
metal-messiah, line 26: 👍

src/features/session_replay/aggregate/index.js
metal-messiah: there is the chance this gets called after the api has executed, but there is a short-circuit in `initializeRecording` that prevents it from rerunning if already recording

patrickhousley (approved):

PR #805, opened 2023-11-06
feat: Enable SRI and nonce attributes for async chunks

patrickhousley:
Additional JavaScript resources loaded by the agent will now include a SHA512 integrity hash and will respect the presence of a nonce token on scripts loaded on the page. This gets the agent closer to being able to run within an environment with a more secure content security policy.
---

### Overview

Additional JavaScript resources loaded by the agent will now include a SHA512 integrity hash and will respect the presence of a nonce token on scripts loaded on the page. This gets the agent closer to being able to run within an environment with a more secure content security policy.

The integrity hash for chunks is calculated during the build and included in the loader. The nonce token code is backwards compatible to IE11 by checking for any scripts that have a nonce and using it.

This does not address the loader injected into the page. Until the loader includes an integrity hash and all APM agents injecting the browser agent support injecting the nonce attribute, the agent will still not be able to run without our current CSP guidelines.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-118087


src/loaders/configure/nonce.js
```diff
@@ -0,0 +1,22 @@
...
+    const scriptTags = document.querySelectorAll('script')
+    for (let i = 0; i < scriptTags.length; i++) {
+      const scriptTag = scriptTags[i]
+      if (scriptTag.nonce) {
+        return scriptTag.nonce
+      }
+    }
+  } catch (ex) {
+    // Swallow error and proceed like nonce is not defined
+  }
```
metal-messiah, line 19: This assumes that the first script is the loader, are we good with that assumption?  Would we be better off just not trying to support IE11?

metal-messiah (approved):
approved with comment

src/loaders/configure/nonce.js
patrickhousley: For IE11, we are assuming that any script containing a `nonce` will work because that script would otherwise fail if that `nonce` was not in the CSP. I feel like this is a safe assumption.

PR #808, opened 2023-11-09
feat: Soft Navigations trial

cwli24:
Introducing a lighter and simplified version of the existing spa feature that tracks soft navigations. This new feature still sends `BrowserInteraction` events, but the duration and finishing of those interactions have been redefined. As a replacement, it ought to reduce complexity and integration issues around the existing spa and exhibit better and faster performance.
---

### Overview

Soft navigations feature is disabled unless the `soft-nav` feature flag is passed into the `init` configuration, in which case soft nav will _replace_ the old spa on the page. This feature captures "interactions" based on the following sequences in **exact chronological** order:

1. `click`, `keydown`, or `submit` event on the page occurs
  a. an interaction can also be started through the `newrelic.interaction()` api
3. the location (url) of the page is altered
4. there's a dom change noticed by `MutationObserver`

No ensuing single or chained callbacks, network requests, async timers, etc. are tracked, and they do not affect the length of the interaction, unlike the old spa method. The `createTracer` spa api will have no effect on soft nav, though the other spa apis remain the same.

### Related Issue(s)

Building on #626 
Angler: https://source.datanerd.us/agents/angler/pull/529

### Testing

Jest & wdio tests for feature added based on [this google doc](https://docs.google.com/document/d/15hjIGA9R8eYbnb-Xu2lfttT_a-CHHAgIYfxMYo1NDyg/edit)


src/features/page_view_event/aggregate/index.js
```diff
@@ -111,6 +111,7 @@ export class Aggregate extends AggregateBase {
           activateFeatures(JSON.parse(responseText), this.agentIdentifier)
           this.drain()
         } catch (err) {
+          console.error(err)
```
metal-messiah, line 114: console clean up

src/features/soft_navigations/aggregate/index.js
```diff
@@ -0,0 +1,163 @@
...
+      this.initialPageLoadInteraction.finish(loadEventTime)
+      this.interactionsToHarvest.push(this.initialPageLoadInteraction)
+      this.initialPageLoadInteraction = null
+      this.scheduler.startTimer(harvestTimeSeconds, 0.1) // give buffered (ajax & jserror) events some time to settle before sending the initial page load
+    })
+
+    this.interactionInProgress = null // aside from the "page load" interaction, there can only ever be 1 ongoing at a time
+
+    this.blocked = false
+    registerHandler('block-spa', () => { // if rum response determines that customer lacks entitlements for spa endpoint, this feature shouldn't harvest
```
metal-messiah, line 41: This should probably be implemented in the opposite way, where it waits for the 'spa' flag and then initializes.  Could be a race between harvesting the initial ixn and the block-spa flag

src/features/page_view_event/aggregate/index.js
cwli24: This was actually on purpose to aid troubleshooting why RUM fails

tests/assets/spa/xhr-before-load.html
cwli24, line 1: This is a remnant from POC PR -- should be removed(?)

tests/assets/spa/xhr.html
cwli24, line 1: Also this one, should prob revert this.

src/features/page_view_event/aggregate/index.js
metal-messiah: If that's the case it would be better to put the `err` object as the second argument into the `warn(...)` call on 116 to keep it tidy

tests/assets/spa/xhr-before-load.html
metal-messiah: yes can be removed as it was just used for manually testing the POC

tests/assets/spa/xhr.html
metal-messiah: Same here, was just used for manually testing the POC

src/features/jserrors/aggregate/index.js
```diff
@@ -197,98 +197,94 @@ export class Aggregate extends AggregateBase {
...
-    handle('errorAgg', msg, undefined, FEATURE_NAMES.sessionTrace, this.ee)
-    handle('errorAgg', msg, undefined, FEATURE_NAMES.spa, this.ee)
-    handle('errorAgg', msg, undefined, FEATURE_NAMES.sessionReplay, this.ee)
-
+    // Trace sends the error in its payload, and both trace & replay simply listens for any error to occur.
+    const jsErrorEvent = [type, bucketHash, params, newMetrics, customAttributes]
+    handle('errorAgg', jsErrorEvent, undefined, FEATURE_NAMES.sessionTrace, this.ee)
+    handle('errorAgg', jsErrorEvent, undefined, FEATURE_NAMES.sessionReplay, this.ee)
     // still send EE events for other features such as above, but stop this one from aggregating internal data
     if (this.blocked) return
```
metal-messiah, line 205: Do we need to worry about spa (old or new) still having access to the error event if the error feature is blocked?  This change adds new behavior that stops the jserror event from flowing on if blocked

src/common/window/nreum.js
```diff
@@ -61,7 +61,8 @@ export function gosNREUMOriginals () {
       EV: globalScope.Event,
       PR: globalScope.Promise,
       MO: globalScope.MutationObserver, // this'll be undefined if not in a web window
-      FETCH: globalScope.fetch
+      FETCH: globalScope.fetch,
+      RAF: globalScope.requestAnimationFrame?.bind(globalScope) // this'll be undefined if not in a web window
```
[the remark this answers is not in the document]
cwli24: Oh I thought jserror was still using it; seems like we have dead code rn. Will do: remove wrap-raf module

src/features/jserrors/aggregate/index.js
```diff
@@ -197,98 +197,94 @@ export class Aggregate extends AggregateBase {
...
-      if (params._interactionNodeId) {
-        params.parentNodeId = params._interactionNodeId.toString()
-        delete params._interactionNodeId
-      }
-
-      var hash = item[1] + interaction.root.attrs.id
-      var jsAttributesHash = stringHashCode(stringify(customParams))
-      var aggregateHash = hash + ':' + jsAttributesHash
+      delete params._softNavAttributes // cleanup temp properties from synchronous evaluation; this is harmless when async from soft nav (properties DNE)
+      delete params._softNavFinished
```
metal-messiah, line 235: is this stuff duplicative to the `#storeJserrorForHarvest` arguments of `softNavCustomAttrs = {}`?

src/features/jserrors/aggregate/index.js
```diff
@@ -197,98 +197,94 @@ export class Aggregate extends AggregateBase {
...
       delete params._interactionNodeId
 
-      var hash = item[1]
-      var jsAttributesHash = stringHashCode(stringify(customParams))
+      var hash = wasSaved ? item[1] + interaction.root.attrs.id : item[1]
+      var jsAttributesHash = stringHashCode(stringify(allCustomAttrs))
       var aggregateHash = hash + ':' + jsAttributesHash
 
-      this.aggregator.store(item[0], aggregateHash, item[2], item[3], customParams)
+      this.aggregator.store(item[0], aggregateHash, params, item[3], allCustomAttrs)
```
metal-messiah, line 273: since you added a method for `storeJsErrorForHarvest`, doesn't this seem duplicative now?

src/features/soft_navigations/aggregate/ajax-node.js
```diff
@@ -0,0 +1,57 @@
...
+      addString(this.domain),
+      addString(this.path),
+      numeric(this.txSize),
+      numeric(this.rxSize),
+      this.requestedWith,
+      addString(this.nodeId),
+      nullable(this.spanId, addString, true) + nullable(this.traceId, addString, true) + nullable(this.spanTimestamp, numeric)
+    ]
+    let allAttachedNodes = []
+    if (typeof this.gql === 'object') allAttachedNodes = addCustomAttributes(this.gql, addString)
```
metal-messiah, line 48: Don't ajax nodes under old spa get other custom attributes besides just GQL?  Or is that handled somewhere else?

src/features/soft_navigations/aggregate/ajax-node.js
```diff
@@ -0,0 +1,57 @@
...
+    this.children.forEach(node => allAttachedNodes.push(node.serialize())) // no children is expected under ajax nodes at this time
```
metal-messiah, line 49: just to re-state what i said above, i do think there are custom attr children nodes that old spa adds for custom attrs

src/features/soft_navigations/aggregate/index.js
```diff
@@ -0,0 +1,254 @@
...
+import { AjaxNode } from './ajax-node'
+import { InitialPageLoadInteraction } from './initial-page-load-interaction'
+import { Interaction } from './interaction'
+
+export class Aggregate extends AggregateBase {
+  static featureName = FEATURE_NAME
+  constructor (agentIdentifier, aggregator, { domObserver }) {
+    super(agentIdentifier, aggregator, FEATURE_NAME)
+
+    const harvestTimeSeconds = getConfigurationValue(agentIdentifier, 'spa.harvestTimeSeconds') || 10
```
metal-messiah, line 20: this should use the soft_nav init item you added if they are to be separated

src/features/soft_navigations/aggregate/index.js
```diff
@@ -0,0 +1,254 @@
...
+      this.initialPageLoadInteraction = null
+      // Report metric on the initial page load time
+      handle(SUPPORTABILITY_METRIC_CHANNEL, ['SoftNav/Interaction/InitialPageLoad/Duration/Ms', Math.round(loadEventTime)], undefined, FEATURE_NAMES.metrics, this.ee)
+    })
+
+    this.latestRouteSetByApi = null
+    this.interactionInProgress = null // aside from the "page load" interaction, there can only ever be 1 ongoing at a time
+
+    this.blocked = false
+    this.waitForFlags(['spa']).then(([spaOn]) => {
```
metal-messiah, line 47: Doesnt have to happen as part of this initial PR but we should consider if we can gate all the event generation flows from happening if the spa flag is falsy

src/features/soft_navigations/aggregate/interaction.js
```diff
@@ -0,0 +1,159 @@
...
+    this.eventSubscription = new Map([
+      ['finished', []],
+      ['cancelled', []]
+    ])
+    this.forceSave = this.forceIgnore = false
+    if (this.trigger === API_TRIGGER_NAME) this.createdByApi = true
+  }
+
+  updateDom (timestamp) {
+    this.domTimestamp = (timestamp || performance.now()) // default timestamp should be precise for accurate isActiveDuring calculations
```
metal-messiah, line 47: should performance.now() in these fallbacks use the `now()` implementation in the agent which also utilizes Math.floor() to normalize all the values?

src/features/soft_navigations/aggregate/interaction.js
```diff
@@ -0,0 +1,159 @@
...
+    this.historyTimestamp = (timestamp || performance.now())
+  }
+
+  seenHistoryAndDomChange () {
+    return this.historyTimestamp > 0 && this.domTimestamp > this.historyTimestamp // URL must change before DOM does
+  }
+
+  on (event, cb) {
+    if (!this.eventSubscription.has(event)) throw new Error('Cannot subscribe to non pre-defined events.')
+    if (typeof cb !== 'function') throw new Error('Must supply function as callback.')
```
metal-messiah, line 61: Just wanna double check that we want to throw errors here instead of warns()

src/features/soft_navigations/constants.js
```diff
@@ -0,0 +1,29 @@
+import { FEATURE_NAMES } from '../../loaders/features/features'
+
+export const INTERACTION_TRIGGERS = [
+  'click', // e.g. user clicks link or the page back/forward buttons
+  'keydown', // e.g. user presses left and right arrow key to switch between displayed photo gallery
+  'submit' // e.g. user clicks submit butotn or presses enter while editing a form field
+]
```
metal-messiah, line 7: We should probably evaluate if this covers the cases we want (including touch events, scroll events, etc)

src/features/soft_navigations/instrument/index.js
```diff
@@ -0,0 +1,66 @@
...
+      oncePerFrame = true
+      originals.RAF(() => { // waiting for next frame to time when any visuals are supposedly updated
+        handle('newDom', [now()], undefined, this.featureName, this.ee)
+        oncePerFrame = false
+      })
+    })
+
+    const processUserInteraction = debounce((event) => {
+      handle('newUIEvent', [event], undefined, this.featureName, this.ee)
+      domObserver.observe(document.documentElement || document.body, { attributes: true, childList: true, subtree: true, characterData: true })
```
metal-messiah, line 48: is the domObserver.observe call supposed to be called every time the debounce cb happens?

src/loaders/agent.js
```diff
@@ -48,17 +47,18 @@ export class Agent extends AgentBase {
     // Future work is being planned to evaluate removing this behavior from the backend, but for now we must ensure this call is made
     this.desiredFeatures.add(PageViewEvent)
 
+    this.runSoftNavOverSpa = [...this.desiredFeatures].some(instr => instr.featureName === FEATURE_NAMES.softNav)
```
metal-messiah, line 50: i think this looks like it still needs to be hooked up to the feature_flags mechanism

src/features/soft_navigations/instrument/index.js
metal-messiah: Also should this only be watching for changes to the body? (documentElement will give the whole tree)

src/features/soft_navigations/instrument/index.js
```diff
@@ -0,0 +1,66 @@
...
+    historyEE.on('pushState-end', trackURLChange)
+    historyEE.on('replaceState-end', trackURLChange)
+
+    try {
+      this.removeOnAbort = new AbortController()
+    } catch (e) {}
+    const trackURLChangeEvent = (evt) => handle('newURL', [evt.timeStamp, '' + window.location], undefined, this.featureName, this.ee)
+    windowAddEventListener('popstate', trackURLChangeEvent, true, this.removeOnAbort?.signal)
+
+    let oncePerFrame = false // attempt to reduce dom noice since the observer runs very frequently with below options
```
metal-messiah, line 36: running more frequently than expected also might be a consequence of calling .observe multiple times (every time it debounces)??

src/features/jserrors/aggregate/index.js
cwli24: No, the params are sent as is AFAICT(ell). These need to be scrubbed so they're not sent under the errors.

src/features/jserrors/aggregate/index.js
cwli24: storeJsErrorForHarvest (soft nav) has somewhat different flow from on onInteractionDone (old spa), and while they certainly share similarities, I think it'd be too messy to combine them. The intention is to delete onInteractionDone when old spa gets removed. I made a comment for that, work?

src/features/soft_navigations/aggregate/ajax-node.js
cwli24: In `serializeInteraction` for serializer.js, for type 2 (ajax) nodes, only that is tacked onto `children` and no other attributes. You may be thinking of bel.7;2 payloads handled directly by the ajax feature itself, in which case it prepends `jsAttributes` as well.

src/features/soft_navigations/aggregate/ajax-node.js
cwli24: Addressed

src/features/soft_navigations/aggregate/index.js
cwli24: You mean don't buffer stuff if spa is 0? We can. Make the instrument listeners check our abort signal, and pass that fn to aggregate to abort() here. Would be easier if we had a "parent" reference back to the instrument file, but can just pass it in like how domObserver is passed in, for now.
Do you want to proceed with this?

src/features/soft_navigations/aggregate/interaction.js
cwli24: So I ran into a problem here that required finer accuracy. The UI event timeStamp can be something like `100.123`. If this event updateDom at `100.456` and that gets rounded down to `100`, then you end up with an interaction with start and end timestamps of `100.123` and `100` (before it's harvested and rounded down in serialization), which causes problems.
Sure, we can round down the event timeStamp too so that it's `100` and `100` during runtime. However, consider an ajax request that then fires at `100.789`. If you round this down and assign it to the interaction, that's incorrect. If you don't round this down, then an ajax request that fires at `100.234` which should be attributed to the interaction cannot seek it.

src/features/soft_navigations/aggregate/interaction.js
cwli24: I think so, yeah. This wouldn't be a user related error so warning them only confuses them. `on` is an internal fn. It'd be a dev error, and we can detect this much faster/earlier this way. Moral of the story: don't do `interaction.on('whateveryouwant', ()=>{...})` without adding 'whateveryouwant' to the `eventSubscription` map.

src/features/soft_navigations/constants.js
cwli24: AFAICF(ind), these are all that Google said they're looking at. This area could use feedback from trial users on, me thinks.

src/features/soft_navigations/instrument/index.js
cwli24: It's disconnected on ixn timeouts and also on UI ixn finish. I wanted to limit the callback execution whenever it's not necessary (no ixn in progress, so no point to be observing dom) to reduce performance hit.

Just discovered that `.end()` api doesn't flow thru disconnect though, thanks for bringing that to my attn. I'm going to move the disconnect() into `setClosureHandlers` which sets up cleanup code on either started interactions.

src/loaders/agent.js
cwli24: The final output for `runSoftNavOverSpa` is actually completed on the next line inside configure of configure.js, L48. The information I needed to make the decision is fragmented into different scopes, hence I had to split this up. Didn't want to change `configure` and pass desiredFeatures or a separate temp flag in (this line can be interpreted as the temp flag anyways).

src/features/soft_navigations/instrument/index.js
cwli24: The cb is executed per change or group of changes and there's no guarantee that all of those records are batched together per frame or per any other set interval. Calling observe on the same observer doesn't do that either.

> If you call observe() on a node that's already being observed by the same MutationObserver, all existing observers are automatically removed from all targets being observed before the new observer is activated.
> If the same MutationObserver is not already in use on the target, then the existing observers are left alone and the new one is added.

src/features/soft_navigations/aggregate/ajax-node.js
metal-messiah: does it inherit the custom attributes for ajax children from the parent browser ixn then?

src/features/soft_navigations/aggregate/index.js
metal-messiah: yes but lets separate that work to a new ticket item, ty

src/features/soft_navigations/aggregate/ajax-node.js
cwli24: Which is to say that the [staging app on NR1](https://one.newrelic.com/nr1-core/ajax-requests/home/Mjk0MjgxMXxCUk9XU0VSfEFQUExJQ0FUSU9OfDQwNzUwNDQ3NQ?duration=1800000&filters=%28domain%20IN%20%28%27BROWSER%27%2C%20%27BROWSER%27%29%20AND%20type%20IN%20%28%27APPLICATION%27%2C%20%27MICRO_FRONTEND%27%29%29&state=bfdad770-a66b-849a-b4e7-830133a13da1) as an example may be overcounting.

src/features/soft_navigations/aggregate/index.js
cwli24: done, NR-217495

src/features/soft_navigations/aggregate/ajax-node.js
metal-messiah: Thanks, please note this as a follow up item to investigate further in the GQL CDD

tests/browser/xhr/ajax-events.browser.js
metal-messiah, line 1: FYI this file seems to have a lot of issues in the JIL test suite across all the browsers

src/features/soft_navigations/instrument/index.js
cwli24: changed to <body> tag instead of the whole <html>. Performance of observe/disconnect vs. perma-observe started in slack thread.

tests/assets/spa/errors/discarded-interaction.html
```diff
@@ -19,7 +19,7 @@
     document.addEventListener('click', function () {
       setTimeout(function foo () {
         (function bar () {
-          fetch('/json')
+          if (typeof fetch !== 'undefined') fetch('/json') // because IE sucks
```
patrickhousley, line 22: Why not just use XHR here?

metal-messiah (approved):

tests/assets/spa/errors/discarded-interaction.html
cwli24: Valid, I think I did it bc fetch is easier with 1 line instead of 4 or 5. 😆

EduardDopler:
Any plans to pull only either the old or the new implementation into the browser? Currently, both are part of the spa build. Hence the significant increase in size.

patrickhousley:
> Any plans to pull only either the old or the new implementation into the browser? Currently, both are part of the spa build. Hence the significant increase in size.

It is not possible in our current deployment mechanics. Our plans are to eventually remove the old SPA feature completely which will reduce the bundle size. In the mean time, the NPM package allows customers to build their own customized agent with the features they care about.

PR #809, opened 2023-11-09
feat: Expose library versions used to capture session replay data

metal-messiah:
Expose library versions used to capture session replay data, to help sync visualizations in the UI Replayer components across version migrations.
---

### Overview
This PR exposes the rrweb lib version used to capture the reply data to a public namespace to be used for UI syncing purposes.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-178304

### Testing
Tests have been updated to account for the change

### DIFF
--- a/src/loaders/configure/nonce.js
+++ b/src/loaders/configure/nonce.js
@@ -2,7 +2,7 @@
 
 __webpack_require__.nc = (() => {
   try {
-    if (document.currentScript && document.currentScript.nonce) {
+    if (document && document.currentScript && document.currentScript.nonce) {
       return document.currentScript.nonce
     }
   } catch (ex) {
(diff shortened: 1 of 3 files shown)


src/loaders/configure/nonce.js
```diff
@@ -2,7 +2,7 @@
 
 __webpack_require__.nc = (() => {
   try {
-    if (document.currentScript && document.currentScript.nonce) {
+    if (document && document.currentScript && document.currentScript.nonce) {
       return document.currentScript.nonce
     }
   } catch (ex) {
```
cwli24, line 8: Can't this whole try-catch just be the undefined-err proof: `return typeof document !== 'undefined' && document.currentScript?.nonce` ?

metal-messiah: Wasnt sure if Webpack transpiles these native overrides, since this was the way Patrick wrote this method.  The optional chaining would break IE if not.

plus your example would return `false` instead of `undefined` for cases where document is not defined which i'm also not sure what webpack would do with when adding the nonce to scripts

cwli24 (approved):

src/loaders/configure/nonce.js
metal-messiah: shared in slack

PR #810, opened 2023-11-10
feat: Add relative timestamps to Session Replay payloads

metal-messiah:
Add relative timestamps to Session Replay payloads to allow for server-side clock skew adjustments and improve the API pagination experience.
---

### Overview
This PR adds the `rst` param, which is seen on all other browser agent payloads, as well as a relative representation of the first payload node to allow for server-side clock skew adjustments if the server time and the rst time differ drastically.

### Related Issue(s)
NR-177602

### Testing
Tests have been updated to account for the extra params

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -25,6 +25,7 @@ import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
 import { handle } from '../../../common/event-emitter/handle'
 import { FEATURE_NAMES } from '../../../loaders/features/features'
 import { RRWEB_VERSION } from '../../../common/constants/env'
+import { now } from '../../../common/timing/now'
 
 export const AVG_COMPRESSION = 0.12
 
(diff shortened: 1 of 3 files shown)


src/features/session_replay/aggregate/index.js
```diff
@@ -301,12 +305,14 @@ export class Aggregate extends AggregateBase {
...
           'replay.lastTimestamp': lastTimestamp,
           'replay.durationMs': lastTimestamp - firstTimestamp,
           'replay.nodes': this.events.length,
           'session.durationMs': agentRuntime.session.getDuration(),
           agentVersion: agentRuntime.version,
           session: agentRuntime.session.state.value,
+          rst: relativeNow,
           hasMeta: this.hasMeta,
           hasSnapshot: this.hasSnapshot,
           hasError: this.hasError,
```
cwli24, line 318: Late to ask but should this share the same or common base as the default query in harvest.js used by all the other feats?

metal-messiah: unfortunately no, since the base query is formatted differently than the blob endpoint requires. see `browser_monitoring_key` etc

metal-messiah: specifically this area
Screenshot 2023-11-10 at 10 43 20 AM

cwli24 (approved):
There's also a `globalScope.performance.now()` you could change into `now()` if you're so inclined in `setTimestamp()`.

PR #821, opened 2023-11-19
feat: Migrate Session Traces to Use Blob Consumer with Feature Flags

metal-messiah:
Create migration path for Session Traces to use the Blob Consumer, which aligns Session Traces and Session Replays and allows for better ingest scaling.
---

### Overview
This PR creates a new Agg for Session Traces which uses the Blob Consumer as the ingest pipeline.  There are many key principals to work through when evaluating this PR:
* The new agg will honor BCS and WILL NOT care about what the state of SR is.  BCS is changing to be self-sufficient and we can honor the flags that are provided to us without so much cross-contamination of the feature sets. The new module comes with a set of new assumptions and the tests of that module match those assumptions.  Everything is gated by feature_flag, including how tests operate.  One of those assumptions is that BCS will report sampling to `STS` as 0,1,2 (off, full, error), and `ST` will denote entitlements. This will also apply to session replay in the form of `SRS` and `SR` respectively.
* The new agg will report to the blob consumer using the same method as SR. It will report under the eventType of `BrowserSessionChunk`.
* The new agg will generate its own PTID and will not rely on a response from ingest to do so.  To replace existing behaviors in the backend, the `firstSessionHarvest` flag is used.
* Many new tests have been made, and several more have been edited to align with some testing structure changes.  The bulk of this PR is testing file/architecture changes.  The key concept was to take existing ST and SR tests and utilize their concepts to make matching ST-blob tests.
* A change to drain/feature-flags was made.  This should be highly beneficial but warrants a thorough look-through since drain is a highly-consequential action in the agent code.

### Related Issue(s)

[ST SPEC](https://source.datanerd.us/agents/rum-specs/blob/main/browser/SessionTrace-Blob.md)
[SR SPEC](https://source.datanerd.us/agents/rum-specs/blob/main/browser/SessionReplay-Blob.md)
https://new-relic.atlassian.net/browse/NR-57395 - `xhrWrappable` now fully removed

[Feature](https://new-relic.atlassian.net/browse/NR-165202)

### Testing
Dozens of new E2E tests were created, and many more tests had small changes to align on a common testing architecture change.


src/common/util/feature-flags.js
```diff
@@ -24,21 +23,17 @@ export function activateFeatures (flags, agentIdentifier) {
...
-    if (bucketMap[flag]) {
-      bucketMap[flag].forEach(feat => {
-        if (!num) handle('block-' + flag, [], undefined, feat, sharedEE)
-        else handle('feat-' + flag, [], undefined, feat, sharedEE)
-        handle('rumresp-' + flag, [Boolean(num)], undefined, feat, sharedEE) // this is a duplicate of feat-/block- but makes awaiting for 1 event easier than 2
-      })
-    } else if (num) handle('feat-' + flag, [], undefined, undefined, sharedEE) // not sure what other flags are overlooked, but there's a test for ones not in the map --
+    if (!num) sharedEE.emit('block-' + flag, [])
+    else sharedEE.emit('feat-' + flag, [])
+    sharedEE.emit('rumresp-' + flag, [num]) // this is a duplicate of feat-/block- but makes awaiting for 1 event easier than 2
```
metal-messiah, line 28: using handle here creates an unneeded race condition between emitting to features to enable them and being able to drain.  All features should already be present on the page before this emits. This is the pattern that was used in <=v1216 for many years, and is simpler to rely on.  This allows us to no longer need the middleware buffer in ST for the re-write.

src/features/session_replay/aggregate/index.js
```diff
@@ -184,15 +184,14 @@ export class Aggregate extends AggregateBase {
...
+        this.entitled = !!flagOn
         this.initializeRecording(
           (Math.random() * 100) < getConfigurationValue(this.agentIdentifier, 'session_replay.error_sampling_rate'),
           (Math.random() * 100) < getConfigurationValue(this.agentIdentifier, 'session_replay.sampling_rate')
         )
       }).then(() => sharedChannel.onReplayReady(this.mode)) // notify watchers that replay started with the mode
-
-      this.drain()
     }
+    this.drain()
```
metal-messiah, line 194: This was implemented because the feature would not drain if `shouldSetup` was not true, which could affect other features from draining

cwli24:
reminder; handler-cache.js can be deleted -- dead code

src/features/session_replay/instrument/index.js
```diff
@@ -44,10 +54,14 @@ export class Instrument extends InstrumentBase {
   }
 
   async #startRecording (mode) {
+    if (this.featAggregate) {
```
metal-messiah, line 57: Also need a check in here to see if recorder exists, and if recorder is recording --> early return

src/features/session_replay/instrument/index.js
```diff
@@ -44,10 +54,15 @@ export class Instrument extends InstrumentBase {
...
+      if (this.featAggregate.entitled) this.featAggregate.initializeRecording(mode, true)
+      /** dont need to reimport the recorder so just return if the featAgg exists */
+      return
+    }
+    if (this.recorder?.recording) return
```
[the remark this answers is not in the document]
cwli24: Here's some case flows:
Ex. A

1. api recordReplay called -> import Recorder, fetching
2. preload evaluated -> quits because a startup already in process -> queues agg import
3. Recorder imported & starts recording
4. agg imports

Ex. B

1. preload evaluated -> import Recorder, fetching
2. api recordReplay called -> quits because a startup already in process
3. Recorder imported & starts recording -> queues agg import
4. agg imports

Similar scenario for if preload evaluated to false, just that recordReplay will do the startup.

src/features/session_replay/instrument/index.js
```diff
@@ -44,10 +54,15 @@ export class Instrument extends InstrumentBase {
   }
 
   async #startRecording (mode) {
+    if (this.featAggregate) {
+      if (this.featAggregate.entitled) this.featAggregate.initializeRecording(mode, true)
```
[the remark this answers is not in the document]
cwli24: #apiRestartRecording should:

1. check if the agg is imported.
  a. if so, check if it's currently recording and if so, done
  b. if not recording, start recording if entitled to do so, done. If not entitled, done.
2. check if a startup is progress. if so, done. if not, start it up. ( aka just call `#startRecording`)

It's close to what you have here, except for the bit about checking for recording in-progress, but easier to read & analyze if preload start doesn't deal with step 1 logic.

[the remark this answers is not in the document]
cwli24: One more thing here. `mode` has to be a class instance variable that's changeable because of async. Consider if preload startup with session-inherited ERROR mode, but then api recordReplay is called. A startup is already in progress, but that startup will end up recording using ERROR instead of FULL as the wanted behavior in this case. Therefore, the recording should use whatever is the latest mode at the time.
I.e. `#apiRestartRecording` will set the local mode to be FULL, and _then_ quit to let the in-progress startup finish.

src/features/session_replay/aggregate/index.js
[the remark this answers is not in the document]
cwli24: Nvm, I think you're supposed to **delete** that instead since L107 is replacing it.

src/features/session_replay/aggregate/index.js
```diff
@@ -112,23 +102,34 @@ export class Aggregate extends AggregateBase {
...
+
     const { error_sampling_rate, sampling_rate, autoStart, block_selector, mask_text_selector, mask_all_inputs, inline_stylesheet, inline_images, collect_fonts } = getConfigurationValue(this.agentIdentifier, 'session_replay')
 
-    this.waitForFlags(['sr']).then(([flagOn]) => {
-      this.entitled = flagOn
+    this.waitForFlags(['srs', 'sr']).then(([srMode, entitled]) => {
+      // waitForFlags now returns the raw value (0, 1) of the flag.
+      // Eventually BCS will report this as (0, 1, 2) for the appropriate mode, and report entitlements separately.
+      // This will need to be fixed to honor that when that change is made. Returning the raw flag value enables that work.
+      this.entitled = !!entitled
```
cwli24, line 122: Just saying
```suggestion
      this.entitled = Boolean(entitled)
```

src/features/session_replay/aggregate/index.js
```diff
@@ -112,23 +102,34 @@ export class Aggregate extends AggregateBase {
...
       if (!this.entitled && this.recorder?.recording) {
-        this.abort(ABORT_REASONS.ENTITLEMENTS)
         handle(SUPPORTABILITY_METRIC_CHANNEL, ['SessionReplay/EnabledNotEntitled/Detected'], undefined, FEATURE_NAMES.metrics, this.ee)
+        this.recorder?.abort(ABORT_REASONS.ENTITLEMENTS)
```
cwli24, line 125: There's no `.abort` in recorder. Did you mean `this.abort`?

src/features/session_replay/aggregate/index.js
```diff
@@ -112,23 +102,34 @@ export class Aggregate extends AggregateBase {
...
-      this.initializeRecording(
-        (Math.random() * 100) < error_sampling_rate,
-        (Math.random() * 100) < sampling_rate
-      )
+      this.initializeRecording(srMode)
     }).then(() => {
-      if (this.mode === MODE.OFF) args?.recorder?.stopRecording() // stop any conservative preload recording launched by instrument
-      sharedChannel.onReplayReady(this.mode) // notify watchers that replay started with the mode
-    })
+      if (this.mode === MODE.OFF) this.recorder?.stopRecording() // stop any conservative preload recording launched by instrument
```
cwli24, line 130: These lines are very roundabout. Instead of having it split up, can `initializeRecording` host L124-125 if not entitled and then return?

src/features/session_replay/aggregate/index.js
```diff
@@ -180,12 +182,9 @@ export class Aggregate extends AggregateBase {
...
       // The session is new... determine the mode the new session should start in
-      if (fullSample) this.mode = MODE.FULL // full mode has precedence over error mode
-      else if (errorSample) this.mode = MODE.ERROR
-      // If neither are selected, then don't record (early return)
-      else {
-        return
-      }
+      this.mode = srMode
+      // If off, then don't record (early return)
+      if (this.mode === MODE.OFF) return
```
cwli24, line 187:
```suggestion
      // If off, then don't record (early return)
      if ((this.mode = srMode) === MODE.OFF) return
```

cwli24 (changes requested):

src/common/util/feature-flags.js
```diff
@@ -24,21 +23,17 @@ export function activateFeatures (flags, agentIdentifier) {
...
+    activatedFeatures[flag] = num
   })
 
   // Let the features waiting on their respective flags know that RUM response was received and that any missing flags are interpreted as bad entitlement / "off".
   // Hence, those features will not be hanging forever if their flags aren't included in the response.
   Object.keys(bucketMap).forEach(flag => {
     if (activatedFeatures[flag] === undefined) {
-      bucketMap[flag]?.forEach(feat => handle('rumresp-' + flag, [false], undefined, feat, sharedEE))
+      bucketMap[flag]?.forEach(feat => sharedEE.emit('rumresp-' + flag, [false]))
       activatedFeatures[flag] = false
```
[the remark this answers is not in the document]
metal-messiah: ty

src/common/util/feature-flags.test.js
```diff
@@ -47,14 +47,14 @@ test('emits the right events when feature flag = 1', () => {
...
   activateFeatures(flags, agentIdentifier)
 
-  const sharedEE = jest.mocked(eventEmitterModule.ee.get).mock.results[0].value
+  const sharedEE = eventEmitterModule.ee.get(agentIdentifier).emit
 
-  // each flag gets emitted to each of its mapped features, and a feat- AND a rumresp- for every emit, so (1+2+1+1+2+2)*2 = 16
-  expect(handleModule.handle).toHaveBeenCalledTimes(16)
-  expect(handleModule.handle).toHaveBeenNthCalledWith(1, 'feat-stn', [], undefined, FEATURE_NAMES.sessionTrace, sharedEE)
-  expect(handleModule.handle).toHaveBeenLastCalledWith('rumresp-sr', [true], undefined, FEATURE_NAMES.sessionTrace, sharedEE)
+  // each flag gets emitted to each of its mapped features, and a feat- AND a rumresp- for every emit, so (1+1+1+2)*2 = 14
```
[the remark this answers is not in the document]
metal-messiah: there's no longer a reason to map through the `bucketMap` for the emits since they do not respect a buffer channel, which dropped this to 10 calls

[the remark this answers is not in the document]
metal-messiah:
```suggestion
```

src/features/jserrors/aggregate/index.js
```diff
@@ -51,17 +51,14 @@ export class Aggregate extends AggregateBase {
     const scheduler = new HarvestScheduler('jserrors', { onFinished: (...args) => this.onHarvestFinished(...args) }, this)
     scheduler.harvest.on('jserrors', (...args) => this.onHarvestStarted(...args))
 
-    // Don't start harvesting until "drain" for this feat has been called (which currently requires RUM response).
-    this.ee.on(`drain-${this.featureName}`, () => {
-      if (!this.blocked) scheduler.startTimer(harvestTimeSeconds) // and only if ingest will accept jserror payloads
+    this.waitForFlags(['err']).then(([isOn]) => {
+      if (isOn) scheduler.startTimer(harvestTimeSeconds)
```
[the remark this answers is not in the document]
metal-messiah: There may need to be some other changes to accommodate that, if anything further down the constructor references the scheduler, but yeah this makes sense in general

src/features/session_replay/aggregate/index.js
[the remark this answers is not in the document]
metal-messiah: ty

src/features/page_view_event/aggregate/index.js
```diff
@@ -128,8 +128,8 @@ export class Aggregate extends AggregateBase {
           activateFeatures(flags, this.agentIdentifier)
           this.drain()
         } catch (err) {
-          this.ee.abort()
-          warn('RUM call failed. Agent shutting down.', err)
+          this.ee.aborted = true
+          warn('RUM call failed. Agent shutting down.')
```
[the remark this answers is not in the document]
metal-messiah: 🤷

src/features/session_replay/instrument/index.js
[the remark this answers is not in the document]
metal-messiah: ```The context here is you don't know if replay was stopped/paused/unstarted to begin with. So this shouldn't initializeRecording potentially again```

we should talk through this to get everything aligned, because initializeRecording checks for `.blocked` which would be the only case where we would not want to start recording again.

src/features/session_replay/aggregate/index.js
metal-messiah: webpack transpiles that to `!!` anyway since its shorter, just fyi

src/features/session_replay/aggregate/index.js
metal-messiah: probably a merge conflict remnant, ty

src/features/session_replay/aggregate/index.js
metal-messiah: potato potahto, but sure

src/features/session_replay/aggregate/index.js
metal-messiah: pointless nesting and harder to read tbh

cwli24:
```
/*
 * Copyright 2023 New Relic Corporation. All rights reserved.
 */
/**
 * @file Primes the Session Replay feature for lazy loading.
 *
 * NOTE: This code is under development and dormant. It will not download to instrumented pages or record any data.
 * It is not production ready, and is not intended to be imported or implemented in any build of the browser agent until
 * functionality is validated and a full user experience is curated.
 */
import { DEFAULT_KEY, MODE, PREFIX } from '../../../common/session/constants'
import { InstrumentBase } from '../../utils/instrument-base'
import { FEATURE_NAME } from '../constants'
import { hasReplayPrerequisite, isPreloadAllowed } from '../shared/utils'

export class Instrument extends InstrumentBase {

  #mode
  constructor (agentIdentifier, aggregator, auto = true) {
    super(agentIdentifier, aggregator, FEATURE_NAME, auto)
    let session
    try {
      session = JSON.parse(localStorage.getItem(`${PREFIX}_${DEFAULT_KEY}`))
    } catch (err) { }

    if (hasReplayPrerequisite(agentIdentifier)) {
      this.ee.on('recordReplay', () => this.#apiStartOrRestartReplay())
    }

    if (this.#canPreloadRecorder(session)) {
      this.#mode = session?.sessionReplayMode
      this.#preloadStartRecording()
    } else {
      this.importAggregator()
    }
  }

  // At this point wherein session state exists already but we haven't init SessionEntity aka verify timers.
  #canPreloadRecorder (session) {
    if (!session) { // this might be a new session if entity initializes: conservatively start recording if first-time config allows
      // Note: users with SR enabled, as well as these other configs enabled by-default, will be penalized by the recorder overhead EVEN IF they don't actually have or get
      // entitlement or sampling decision, or otherwise intentionally opted-in for the feature.
      return isPreloadAllowed(this.agentIdentifier)
    } else if (session.sessionReplayMode === MODE.FULL || session.sessionReplayMode === MODE.ERROR) {
      return true // existing sessions get to continue recording, regardless of this page's configs or if it has expired (conservatively)
    } else { // SR mode was OFF but may potentially be turned on if session resets and configs allows the new session to have replay...
      return isPreloadAllowed(this.agentIdentifier)
    }
  }

  #alreadyStarted = false
  /**
   * This func is use for early pre-load recording prior to replay feature (agg) being loaded onto the page. It should only setup once, including if already called and in-progress.
   */
  async #preloadStartRecording () {
    if (this.#alreadyStarted) return
    this.#alreadyStarted = true

    try {
      const { Recorder } = (await import(/* webpackChunkName: "recorder" */'../shared/recorder'))

      this.recorder ??= new Recorder({ mode: this.#mode, agentIdentifier: this.agentIdentifier })
      this.recorder.startRecording()
      this.abortHandler = this.recorder.stopRecording
    } catch (e) {}
    this.importAggregator({ recorder: this.recorder })
  }

  /**
   * Called whenever startReplay API is used. That could occur any time, pre or post load.
   */
  #apiStartOrRestartReplay () {
    if (this.featAggregate) { // post-load; there's possibly already an ongoing recording
      if (!this.featAggregate.recorder?.recording && this.featAggregate.entitled) this.featAggregate.initializeRecording(MODE.FULL, true)
      // if already recording or not entitled, don't start (again); don't importAgg again either
    } else { // pre-load
      this.#mode = MODE.FULL
      this.#preloadStartRecording()
      // There's a race here wherein either:
      // a. Recorder has not been initialized, and we've set the enforced mode, so we're good, or;
      // b. Record has been initialized, possibly with the "wrong" mode, so we have to correct that + restart.
      if (this.recorder && this.recorder.parent.mode !== MODE.FULL) {
        this.recorder.parent.mode = MODE.FULL
        this.recorder.stopRecording()
        this.recorder.startRecording()
        this.abortHandler = this.recorder.stopRecording
      }
    }
  }
}
```

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
+        browser_monitoring_key: this.agentInfo.licenseKey,
+        type: 'BrowserSessionChunk',
+        app_id: this.agentInfo.applicationID,
+        protocol_version: '0',
+        attributes: encodeObj({
+          // this section of attributes must be controllable and stay below the query param padding limit -- see QUERY_PARAM_PADDING
+          // if not, data could be lost to truncation at time of sending, potentially breaking parsing / API behavior in NR1
+          // trace payload metadata
+          'trace.firstTimestamp': this.agentRuntime.offset + earliestTimeStamp,
+          'trace.lastTimestamp': this.agentRuntime.offset + latestTimeStamp,
```
cwli24, line 171: Can reduce length by sending the relative time and letting DEM do the calculation bc unix time is very long.

cwli24: Should also reduce some of these long names. Like `trace.` prefix is not necessary, and `start` under a trace chunk should be understandable to services working on it. In general, we've been very lax lately on name sizing--which is more of an issue for query params than payload bodies; idk if that's a good or bad thing as transfer sizes is both a cost to us and extra bill to the customer(?). I imagine that was the bigger reason things were packed or abbreviated in "the old days" (vs. any old browser concerns with payload size).

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
+          'trace.nodes': stns.length,
+          'trace.originTimestamp': this.agentRuntime.offset,
+          // other payload metadata
+          agentVersion: this.agentRuntime.version,
+          ...(firstSessionHarvest && { firstSessionHarvest }),
+          ...(hasReplay && { hasReplay }),
+          ptid: `${this.agentRuntime.ptid}`,
+          rst: now(),
```
cwli24, line 179: Do we really still need rst with those 3 other timestamps?

src/features/utils/instrument-base.js
```diff
@@ -102,7 +102,7 @@ export class InstrumentBase extends FeatureBase {
           return
         }
         const { lazyFeatureLoader } = await import(/* webpackChunkName: "lazy-feature-loader" */ './lazy-feature-loader')
-        const { Aggregate } = await lazyFeatureLoader(this.featureName, 'aggregate')
+        const { Aggregate } = await lazyFeatureLoader(this.featureName, 'aggregate', getConfigurationValue(this.agentIdentifier, 'feature_flags'))
```
cwli24, line 105: Remnant? Unused & not sure what for.

src/features/utils/lazy-feature-loader.js
```diff
@@ -11,7 +11,7 @@ import { FEATURE_NAMES } from '../../loaders/features/features'
  * @param featurePart Name of the feature part to load; should be either instrument or aggregate
  * @returns {Promise<InstrumentBase|FeatureBase|null>}
  */
-export function lazyFeatureLoader (featureName, featurePart) {
+export function lazyFeatureLoader (featureName, featurePart, featureFlags = []) {
```
cwli24, line 14: See comment on instrument-base.js

src/loaders/agent-base.js
cwli24, line 1: Also a remnant of something. No observationContext usage.

src/loaders/agent.js
```diff
@@ -85,8 +85,8 @@ export class Agent extends AgentBase {
       delete newrelic.initializedAgents[this.agentIdentifier]?.features // GC mem used internally by features
       delete this.sharedAggregator
       // Keep the initialized agent object with its configs for troubleshooting purposes.
-      newrelic.ee?.abort() // set flag and clear global backlog
-      delete newrelic.ee?.get(this.agentIdentifier) // clear this agent's own backlog too
+      const thisEE = newrelic.ee.get(this.agentIdentifier)
+      thisEE.aborted = true // set flag and clear backlog
```
cwli24, line 89: Should also still delete the agent obj itself under the global EE, no?

src/loaders/api/api.js
```diff
@@ -198,7 +198,7 @@ export function setAPI (agentIdentifier, forceDrain, runSoftNavOverSpa = false)
       drain(agentIdentifier, 'api')
     }).catch(() => {
       warn('Downloading runtime APIs failed...')
-      drain(agentIdentifier, 'api', true)
+      instanceEE.aborted = true
```
cwli24, line 201: Just pointing out that we're aborting the entire agent because the async apis (finished, setErrorHandler, addToTrace, addRelease) failed to import. You sure?

src/loaders/configure/configure.js
```diff
@@ -52,6 +52,7 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
     ...(updatedInit.ajax.deny_list || []),
     ...(updatedInit.ajax.block_internal ? internalTrafficList : [])
   ]
+  runtime.ptid = agent.agentIdentifier
```
cwli24, line 55: Does `ptid` on the RUM or any earlier payloads not mess with anything?

cwli24 (changes requested):
Nicely done though

tools/wdio/plugins/custom-commands.mjs
```diff
@@ -96,18 +96,21 @@ export default class CustomCommands {
...
      */
-    browser.addCommand('enableSessionReplay', async function () {
+    browser.addCommand('enableSessionReplay', async function (sampling_rate = 100, error_sampling_rate = 100) {
+      const stMode = Math.random() * 100 < sampling_rate ? 1 : Math.random() * 100 < error_sampling_rate ? 2 : 0
       await browser.testHandle.scheduleReply('bamServer', {
         test: testRumRequest,
         permanent: true,
         body: JSON.stringify({
           stn: 1,
+          sts: 1,
```
cwli24, line 106: Assuming bcs does it's stuff correctly, is sts supposed to be uncorrelated to srs? This wouldn't have simply been the case with our side logic with how much conditional branching was happening. I just want to make sure our test server & tests capture real behavior.

src/features/metrics/aggregate/index.js
cwli24, line 1: Also down on L134, you will always have a `agentRuntime.ptid` now since that's set in configure. Will need to change the condition.

src/common/harvest/harvest-scheduler.js
```diff
@@ -102,7 +102,7 @@ export class HarvestScheduler extends SharedContext {
       if (!submitMethod) return false
 
       const retry = !opts?.unload && submitMethod === submitData.xhr
-      payload = this.opts.getPayload({ retry, opts })
+      payload = this.opts.getPayload({ retry, ...opts })
```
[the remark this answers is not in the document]
metal-messiah: fixed

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
+  constructor (agentIdentifier, aggregator) {
     super(agentIdentifier, aggregator, FEATURE_NAME)
     this.agentRuntime = getRuntime(agentIdentifier)
+    this.agentInfo = getInfo(agentIdentifier)
 
     // Very unlikely, but in case the existing XMLHttpRequest.prototype object on the page couldn't be wrapped.
-    if (!this.agentRuntime.xhrWrappable) return
+    if (!this.agentRuntime.xhrWrappable) {
+      deregisterDrain(this.agentIdentifier, this.featureName)
+      return
```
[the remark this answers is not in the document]
metal-messiah: Screenshot 2024-03-24 at 12 43 48 PM
Haha! you tell me.  I'll just rip this out

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
 
-    function trivial (node) {
-      const limit = 4
-      return !!(node && typeof node.e === 'number' && typeof node.s === 'number' && (node.e - node.s) < limit)
+  /** Switch from "off" or "error" to full mode (if entitled) */
+  switchToFull () {
+    if (this.mode === MODE.FULL || !this.entitled) return
+    const prevMode = this.mode
+    this.mode = MODE.FULL
+    if (prevMode === MODE.OFF && this.initialized) return this.initialize(this.mode, this.entitled, true)
```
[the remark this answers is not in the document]
metal-messiah: I can try to steamline the logic in this block a little better

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -0,0 +1,283 @@
...
+    this.parent = parent
+  }
+
+  /** Central function called by all the other store__ & addToTrace API to append a trace node. */
+  storeSTN (stn) {
+    if (this.parent.blocked) return
+    if (this.nodeCount >= this.parent.maxNodesPerHarvest) { // limit the amount of pending data awaiting next harvest
+      if (this.parent.agentRuntime.session.state.sessionTraceMode !== MODE.ERROR) {
+        // trigger a harvest early
+        this.parent.scheduler.runHarvest({ needResponse: true })
```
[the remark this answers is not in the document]
metal-messiah: in the run-through I did mention the harvest early change instead of dropping data. Dropping data is not something we should do (period) if we are now monetizing ST, there's literally no good reason to do so.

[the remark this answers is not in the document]
metal-messiah: `I don't like this as we have no idea what the status of the scheduler or trace aggregate is to be executing an initial harvest--or any harvest`

I'm not following the logic here.  It checks if it's in error mode in this block.  If the feature starts at any time in full mode, it immediately harvests... meaning error mode doesnt force a harvest, and full mode wouldve already harvested the "initial" harvest (which has no real meaning at the moment by the way)... This also has precedent as the SR feature does the same thing.

If it makes you feel better we can add a check for the harvester's status, or backlog the data in some way, but i think the risk here is pretty low considering the alternative you are proposing is to drop the data altogether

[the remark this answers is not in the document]
metal-messiah: I added a check for harvesting as well as a temp bucket to hold any 'blocked' data while harvesting

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
-      e: metrics.time,
-      o: params.message,
-      t: params.stackHash
-    }
-    this.storeSTN(node)
+    /** Only start actually harvesting if running in full mode at init time */
+    if (this.mode === MODE.FULL) this.startHarvesting()
+    this.agentRuntime.session.write({ sessionTraceMode: this.mode })
+    /** drain and kick off the registerHandlers to start processing any buffered data */
+    if (!this.drained) this.drain()
```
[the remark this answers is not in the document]
metal-messiah: was added because initialize could techically be called more than once.  I will just move it to a new spot to be safer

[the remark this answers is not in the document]
metal-messiah: actually initialize should only be called once, can remove

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
-        n: key,
-        s: val,
-        e: val,
-        o: 'document',
-        t: 'timing'
+      // The SessionEntity can have updates (locally or across tabs for SR mode changes), (across tabs for ST mode changes).
+      // Those updates should be sync'd here to ensure this page also honors the mode after initialization
+      this.ee.on(SESSION_EVENTS.UPDATE, (eventType, sessionState) => {
+        // this will only have an effect if ST is NOT already in full mode
+        if (this.mode !== MODE.FULL && (sessionState.sessionReplayMode === MODE.FULL || sessionState.sessionTraceMode === MODE.FULL)) this.switchToFull()
```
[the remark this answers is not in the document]
metal-messiah: this is specifically aimed at integrating to the SR API being triggered (e.g. sending SR into full mode)

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
-  storeEvent (currentEvent, target, start, end) {
-    if (this.shouldIgnoreEvent(currentEvent, target)) return
 
-    const evt = {
-      n: this.evtName(currentEvent.type),
-      s: start,
-      e: end,
-      t: 'event'
-    }
+    /** ST/SR sampling flow in BCS - https://drive.google.com/file/d/19hwt2oft-8Hh4RrjpLqEXfpP_9wYBLcq/view?usp=sharing */
```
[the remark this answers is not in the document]
metal-messiah: this is a DEM chart, you need to open it with the included draw.io to get full res. I don't agree on this one, I find it helpful to document flow patterns, esp. for nuanced ones like this, and sometimes those things can be gated by login, with no impact to customers or risk for us.

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
-    }
-
-    if (this.isStandalone && now() >= MAX_TRACE_DURATION) {
+  /** Called by the harvest scheduler at harvest time to retrieve the payload.  This will only actually return a payload if running in full mode */
+  prepareHarvest (options = {}) {
+    if (this.traceStorage.nodeCount <= REQ_THRESHOLD_TO_SEND && !(options.isFinalHarvest || options.forceNoRetry) && this.everHarvested) {
+      // Only harvest when more than some threshold of nodes are pending, after the very first harvest, with the exception of the last outgoing harvest.
       return
     }
+    if (this.mode === MODE.OFF && this.traceStorage.nodeCount === 0) return
```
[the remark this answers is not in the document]
metal-messiah: i'm not sure we should be forcing a harvest to run here when aborting...  a viz change will have already triggered a harvest, and a session expiration will have done so as well, so the other cases that could cause abort() to call are likely those that wouldnt need a final harvest, but instead just shut it down

src/features/session_trace/aggregate/index.js
metal-messiah: there is no payload size concern here.  The blob consumer does not do any translating of data (ie magic) so it's important to provide the properties with a certain level of explicitness.  Not only this, but we are aligning to a cross team spec on these matters.  Accordingly, the timestamp values are not going to be evaluated by the backend for performance reasons, and the API needs raw searchable values for timestamps for querying and paging purposes.  I would suggest you take some time to read the rum specs, timestamp specs, and blob consumer specs if you have concerns here.

src/features/session_trace/aggregate/index.js
metal-messiah: thanks, missed removing this

src/loaders/configure/configure.js
metal-messiah: This is actually the reason the request was made for us to calculate it, so that PVE's could be tied to a trace by ptid

tools/wdio/plugins/custom-commands.mjs
metal-messiah: they will be inherently correlated, but for our purposes we are to treat them independently

src/features/metrics/aggregate/index.js
metal-messiah: no longer needed

src/features/session_trace/instrument/index.js
```diff
@@ -18,7 +20,11 @@ export class Instrument extends InstrumentBase {
   static featureName = FEATURE_NAME
   constructor (agentIdentifier, aggregator, auto = true) {
     super(agentIdentifier, aggregator, FEATURE_NAME, auto)
-    if (!isBrowserScope) return // session traces not supported outside web env
+    const canTrackSession = isBrowserScope && getConfigurationValue(agentIdentifier, 'privacy.cookies_enabled') === true
```
cwli24, line 23: This is the exact check done by existing `enableSessionTracking`. We shouldn't duplicate it if there's really no reason to; can you use the existing fn?

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,220 @@
...
-    const node = {
-      n: 'Ajax',
-      s: metrics.time,
-      e: metrics.time + metrics.duration,
-      o: params.status + ' ' + params.method + ': ' + params.host + params.pathname,
-      t: 'ajax'
-    }
-    this.storeSTN(node)
+  /** This module does not auto harvest by default -- it needs to be kicked off.  Once this method is called, it will then harvest on an interval */
+  startHarvesting () {
```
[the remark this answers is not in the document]
cwli24: 2nd item still needs to be done

src/features/session_trace/aggregate/trace/storage.js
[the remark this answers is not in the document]
cwli24: `agentRuntime.session.state.sessionTraceMode` is something that can exist as long as session entity started up.
It doesn't give a clear indicator that the `scheduler` is ready, has been ready, or has gone through ready-shutdown process, etc.
In addition, again this is conflating what this class & function is meant to do. So I still think if nodes are supposed to be sent when buffered is full and trace is in FULL mode, that should be done in the aggregate/index.js, who also has the full visibility scope of scheduler.

src/features/session_trace/aggregate/index.js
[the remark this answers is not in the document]
cwli24: nvm, sounds like the abort check will be on the feature's `blocked` flag instead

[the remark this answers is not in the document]
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/821/commits/2dd66e863bcf9364750f6188eb04344743cdef68

src/features/session_trace/aggregate/index.js
```diff
@@ -1,538 +1,195 @@
...
-    }
-
-    if (this.trace[stn.n]) this.trace[stn.n].push(stn)
-    else this.trace[stn.n] = [stn]
-
-    this.nodeCount++
+  /** This module does not auto harvest by default -- it needs to be kicked off.  Once this method is called, it will then harvest on an interval */
+  startHarvesting () {
+    if (this.scheduler.started || this.scheduler.aborted || this.blocked) return
+    this.scheduler.runHarvest({ needResponse: true })
```
cwli24, line 104: {needResponse: true} is not needed anymore; can take out here & take it out of the trace.js

cwli24: `this.scheduler.aborted` is also extraneous since that's only set by stoptime(true)

cwli24:
To do: check for proper merge after #1008 pulled in

patrickhousley (approved):

PR #829, opened 2023-12-01
chore: fix dry run automation

patrickhousley:
---

### Overview

Fixes the issue with the dry run flag not working and an issue with the internal release script not using the correct parameters based on selected environment.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-185891

### DIFF
--- a/.github/workflows/publish-internal.yml
+++ b/.github/workflows/publish-internal.yml
@@ -36,16 +36,43 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: lts/*
+      - name: Prepare target environment
+        id: target-environment
+        run: |
+          if [[ "${{ inputs.nr_environment }}" == "prod" ]]; then
+            echo "nrba_app_id=${{ secrets.INTERNAL_PRODUCTION_APPLICATION_ID }}" >> $GITHUB_OUTPUT
+            echo "nrba_license_key=${{ secrets.INTERNAL_LICENSE_KEY }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-internal.yml
```diff
@@ -36,16 +36,43 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: lts/*
+      - name: Prepare target environment
+        id: target-environment
+        run: |
+          if [[ "${{ inputs.nr_environment }}" == "prod" ]]; then
+            echo "nrba_app_id=${{ secrets.INTERNAL_PRODUCTION_APPLICATION_ID }}" >> $GITHUB_OUTPUT
+            echo "nrba_license_key=${{ secrets.INTERNAL_LICENSE_KEY }}" >> $GITHUB_OUTPUT
```
metal-messiah, line 44: With the emphasis on security lately, is there a risk here that GH could fail to obfuscate the secret data in some cases? https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow
Would it be safer to just abstract this out one layer and have ind. actions call to `internal-ab` and supply the correct inputs?

patrickhousley: You can look at https://github.com/newrelic/newrelic-browser-agent/actions/runs/7065468372/job/19235524467 to see the GHA automatically obfuscates any secrets values.

metal-messiah (approved):

PR #830, opened 2023-12-04
fix: Fix API Warning Messages

metal-messiah:
Fix warning messages given by API calls made before the agent has initialized
---

### Overview
The Agent gave an incorrect message for `setUserId` when called before the agent was initialized.  This change aligns all API calls to give the same generic message.

### Related Issue(s)
NR-186348

### Testing
All tests should be unaffected

### DIFF
--- a/src/loaders/agent-base.js
+++ b/src/loaders/agent-base.js
@@ -2,11 +2,12 @@
 
 import { warn } from '../common/util/console'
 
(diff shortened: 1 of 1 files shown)


patrickhousley (approved):
Would rather use a private class function instead of a function outside the class but either is fine.

PR #832, opened 2023-12-06
feat: Session Replay Dynamic Loading

metal-messiah:
Allow Session Replay modules to load earlier in the page lifecycle if a replay is ongoing
---

### Overview
This PR enables earlier loading of SR recording libs by doing the following:
1. Isolating all the recording-specific logic to a shared module
  - Abstracting the "meta data" to the event cycle itself instead of the top level class (things like hasError, hasMeta, etc.)
  - Creating a helper class to keep the cycle state easier to manage
2. Checking the session state at load time to see if a recording is in progress (state 1 or 2)
 - Reads the local storage directly bypassing the need to set up the session entity
3. Async Importing the new shared recording module in the SR instrumentation file if deemed allowed
4. Passing the recorder along to the aggregate file if early imported
5. using the passed recorder in the agg file if exists, if not, lazy loading the shared recorder and initializing it after page load

Changes to note:
* Constants have been moved
* Tests have been slightly modified to support the new constants structure and new method for getting events from the recorder
* A problem was noted during this exercise: 
  - The agg tried to "prevent" the harvester from harvesting if it estimated that the payload was too large. This was unreliable to rely on solely, so an extra step was added. 
  - If the agg estimates that the payload is too large, now it will still try to compress it and re-check the payload size before aborting completely.
  - This required changing some assumptions in certain shut-down tests
* It became evident that under our restrictions of a 1mb snapshot payload, that inline-images was never going to be reliable.  The code supporting that (src code, tests) was ripped out.
* SRI is unreliable and causes issues with SR chunks.  It was commented out until a more reliable solution is reached
### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-186806

### Testing
* Existing tests have been modified to support the new architecture, but the test conditions should remain intact and pass
* New test has been added to check if preloaded recorder captures and reports SR data


src/features/session_replay/instrument/index.js
```diff
@@ -9,13 +9,29 @@
...
 export class Instrument extends InstrumentBase {
   static featureName = FEATURE_NAME
   constructor (agentIdentifier, aggregator, auto = true) {
     super(agentIdentifier, aggregator, FEATURE_NAME, auto)
-    this.importAggregator()
+    let session
+    try {
+      session = JSON.parse(localStorage.getItem('NRBA_SESSION'))
+      if (session.sessionReplayMode !== MODE.OFF) {
+        ;(async () => {
```
patrickhousley, line 24: Would maybe move this to a private class function.

patrickhousley (approved):
A couple of minor comments but otherwise looks good.

PR #833, opened 2023-12-08
chore: prevent s3-upload of existing files

patrickhousley:
---

### Overview

Updated the s3-upload action to first check if any of the files being uploaded have a long-term cache policy and already exist in the s3 bucket. If a file is found matching these two criteria, the action will fail and prevent the uploading of the files.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-185887

### DIFF
--- a/.github/actions/s3-upload/index.js
+++ b/.github/actions/s3-upload/index.js
@@ -2,7 +2,7 @@ import fs from 'fs'
 import path from 'path'
 import * as core from '@actions/core'
 import { AssumeRoleCommand, STSClient } from '@aws-sdk/client-sts'
-import { S3Client } from '@aws-sdk/client-s3'
+import { S3Client, HeadObjectCommand, NotFound } from '@aws-sdk/client-s3'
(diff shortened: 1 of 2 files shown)


.github/actions/s3-upload/index.js
```diff
@@ -53,6 +47,42 @@ const uploads = await Promise.all(filesToUpload
       params.Key = `${dirPath}/${params.Key}`
     }
 
+    return params
+  })
+
+const existingFiles = await Promise.all(uploadCommandParams
+  .filter(params => params.CacheControl.indexOf('max-age=31536000') > 0)
```
metal-messiah, line 54: This looks good but just want to double check -- Is this sufficient to select the files that should be rewritten ( explicit loaders, chunks) but not those that need to allow it (.x loaders, current)

patrickhousley: The files that are not cached for one year are the same ones that should be overwritten with each release AND files that are not in the root of the S3 bucket.

PR #834, opened 2023-12-11
feat: Add a flag to note that a payload occurred during a replay

metal-messiah:
Added a flag to mark harvests as having occurred during an active session replay. This will aid cross-feature querying in NR1.
---

### Overview
This PR adds the `hr` flag to note that a payload "has replay".  This was already implemented in Session Traces so this extends that pattern to all payloads. This is a very simple implementation, only reading from the session entity at harvest time to determine the replay mode.

### Related Issue(s)
NR-189641

### Testing
Tests should continue to pass

### DIFF
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -103,7 +103,7 @@ export class Harvest extends SharedContext {
     if (customUrl) url = customUrl
     if (raw) url = `${protocol}://${perceviedBeacon}/${endpoint}`
 
-    const baseParams = !raw && includeBaseParams ? this.baseQueryString() : ''
+    const baseParams = !raw && includeBaseParams ? this.baseQueryString(qs) : ''
     let payloadParams = encodeObj(qs, agentRuntime.maxBytes)
     if (!submitMethod) {
       submitMethod = submitData.getSubmitMethod({ isFinalHarvest: opts.unload })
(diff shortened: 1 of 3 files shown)


src/common/harvest/harvest.js
```diff
@@ -180,7 +180,8 @@ export class Harvest extends SharedContext {
       '&ck=0', // ck param DEPRECATED - still expected by backend
       '&s=' + (runtime.session?.state.value || '0'), // the 0 id encaps all untrackable and default traffic
       encodeParam('ref', ref),
-      encodeParam('ptid', (runtime.ptid ? '' + runtime.ptid : ''))
+      encodeParam('ptid', (runtime.ptid ? '' + runtime.ptid : '')),
+      encodeParam('hr', (runtime?.session?.state.sessionReplayMode === 1 ? '1' : '0'), qs) // hasReplay
```
patrickhousley, line 184: So only if we are in full session replay?

patrickhousley (approved):
Be sure to update specs repo

src/common/harvest/harvest.js
metal-messiah: yep

metal-messiah:
https://source.datanerd.us/agents/rum-specs/pull/7

PR #835, opened 2023-12-11
fix: disabling sri integration

patrickhousley:
A bug has been found with the agent build and SRI integration. The SRI usage within the agent is being disabled to prevent issues with session replay limited preview.
---

### Overview

Disabled the SRI plugin and test for now. Also removed the custom plugins and converted our chunking to standard webpack configuration.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-189746


metal-messiah (approved):

patrickhousley:
This may not need to be merged if we merge #837

PR #837, opened 2023-12-13
fix: loader missing sub-resource integrity hashes

patrickhousley:
Sub-resource integrity hashes were not being properly generated and included in the loader for async chunks related to session replay. This would cause a warning to appear in the browser console but would not block those chunks from loading. This would have only affected customers taking part in the session replay limited preview.
---

### Overview

This problem was caused by a couple of issues:
1. The webpack plugin that we relied on to generate the SRI hashes was not picking up the recorder and compressor chunks so no hash was being generated.
2. We had post build scripts that modified the files after the webpack build finished. While this issue was not seen before, it would have been if we did not move those processes internal to webpack after fixing the SRI hash generation.

Change Highlights:
- Moved all the post build scripts into individual webpack plugins.
- Wrote a new SRI plugin that correctly generates and injects hashes for all async chunks.
  - While this plugin does not have nearly as much functionality as the original, it does work for our needs and I doubt we will ever need the additional functionality around HMR and the HTML plugin.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-189746

### Testing

The SRI integration test has been updated to force session replay and verify all chunks were loaded with integrity hashes.


tests/specs/csp.e2e.js
```diff
@@ -40,17 +43,30 @@ describe('Content Security Policy', () => {
...
         .then(() => browser.waitForAgentLoad())
     ])
 
+    await browser.waitUntil(
+      () => browser.execute(function () {
+        return window.chunkIntegratyValues.length === 3
+      })
+    )
     const foundIntegrityValues = await browser.execute(function () {
       return window.chunkIntegratyValues
```
cwli24, line 66: Typo: `chunkIntegrityValues` if you care -- would have to update the html var too.

tests/specs/csp.e2e.js
```diff
@@ -40,17 +43,30 @@ describe('Content Security Policy', () => {
...
     })
 
-    expect(foundIntegrityValues.length).toEqual(1)
+    expect(foundIntegrityValues.length).toEqual(3)
```
cwli24, line 69: This expect becomes redundant because execution will only pass the previous lines of code if the number of values hits 3.

tools/webpack/plugins/loader-apm-check.mjs
```diff
@@ -0,0 +1,42 @@
...
+    compiler.hooks.thisCompilation.tap(this.#pluginName, (compilation) => {
+      compilation.hooks.processAssets.tap(
+        {
+          name: this.#pluginName,
+          stage: compilation.compiler.webpack.Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_INLINE + 1,
+          additionalAssets: true
+        },
+        (assets) => {
+          Object.entries(assets)
+            .filter(([assetKey]) => assetKey.indexOf('-loader') > -1 && assetKey.endsWith('.js'))
```
cwli24, line 23: Both this & non-ascii: only the `.min.js` that APM uses or does it matter/not matter?

cwli24 (approved):
Built & confirmed each plugins is doing its work.

PR #844, opened 2023-12-16
feat: Expose cross-domain stylesheets to Session Replay

metal-messiah:
Improve inline stylesheet capture mechanisms by exposing cross-domain stylesheets to Session Replay recorder
---

### Overview
This PR exposes cross-domain stylesheets by evaluating and re-injecting them to the DOM if deemed unable to parse via JS.  Stylesheets that are loaded as cross-domain without the crossOrigin: Anonymous tag will be blocked from Javascript parsing. This workaround allows the stylesheet to load for js parsing (and does not cause a re-render since the same contents were already evaluated by the browser)

### Related Issue(s)
https://newrelic.slack.com/archives/C04QZSS6XPV/p1702610835131929

### Testing
New test condition was added to evaluate this behavior


cwli24:
Won't do

PR #849, opened 2023-12-22
chore: converting tests

patrickhousley:
---

### Overview

Converting JIL tests to jest/wdio.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-138412
https://new-relic.atlassian.net/browse/NR-138424
https://new-relic.atlassian.net/browse/NR-138540
https://new-relic.atlassian.net/browse/NR-138522
https://new-relic.atlassian.net/browse/NR-138520
https://new-relic.atlassian.net/browse/NR-138427


tools/wdio/config/base.conf.mjs
```diff
@@ -38,7 +38,7 @@ export default function config () {
     framework: 'mocha',
     mochaOpts: {
       ui: 'bdd',
-      timeout: args.timeout,
+      timeout: 120000,
```
patrickhousley, line 41: This change was done so that our timeout CLI parameter only affects the agent request expect timeouts. Before this change, passing `-t 120000` would cause the expects and sauce browser session to timeout at 120 seconds resulting in subsequent tests failing because the sauce browser session was terminated.

patrickhousley: Our default test timeout is 85s. I increased the session timeout to 120s so that a single test timing out does not cause the whole session/suite to timeout.

metal-messiah (approved):

src/common/event-emitter/contextual-ee.component-test.js
```diff
@@ -142,6 +142,42 @@ describe('event-emitter buffer', () => {
...
+          expect.anything()
+        ])
+      ]
+    }))
+    expect(ee.isBuffering(eventType)).toEqual(true)
+
+    drain('globalEE')
+    ee.buffer([eventType])
+    ee.emit(eventType, eventArgs)
+    ee.emit(eventType, eventArgs)
```
cwli24, line 172: You got a duplicate line here btw

cwli24, line 172: Dup'd line here btw
```suggestion
```

patrickhousley: That was done on purpose.

PR #862, opened 2024-01-19
fix: Bubble Up API Methods to Top-Level Instance For NPM

metal-messiah:
Fix an issue where the NPM wrapping of the agent would not bubble the APIs up to the top-level of the instance, causing invalid warnings to log.
---

### Overview
This PR makes sure that the top level instance calls the internal APIs directly instead of relying on object assignment.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-216981

### Testing
New tests have been added to validate that NPM builds have access to the top level api methods

### DIFF
--- a/src/loaders/agent-base.js
+++ b/src/loaders/agent-base.js
@@ -5,11 +5,8 @@ import { warn } from '../common/util/console'
 export class AgentBase {
   /** Tries to execute the api and enerates a generic warning message with the api name injected if unsuccessful */
   #try (method, ...args) {
-    try {
-      this.api[method](...args)
-    } catch (err) {
-      warn(`Call to agent api ${method} failed. The agent is not currently initialized.`)
-    }
+    if (typeof this.api?.[method] !== 'function') warn(`Call to agent api ${method} failed. The agent is not currently initialized.`)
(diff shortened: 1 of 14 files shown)


src/loaders/agent-base.js
```diff
@@ -3,9 +3,13 @@
...
 export class AgentBase {
-  /** Generates a generic warning message with the api name injected */
-  #warnMessage (api) {
-    return `Call to agent api ${api} failed. The agent is not currently initialized.`
+  /** Tries to execute the api and enerates a generic warning message with the api name injected if unsuccessful */
+  #try (method, ...args) {
+    try {
+      this.api[method](...args)
+    } catch (err) {
+      warn(`Call to agent api ${method} failed. The agent is not currently initialized.`)
```
patrickhousley, line 11: The try catch is assuming any exception is indicative of the agent not being fully initialized. What happens when one of the API methods throws an exception for a different reason? It may be better to change this to specifically test for the existence of the method?

metal-messiah: good call, will change

patrickhousley (approved):

PR #868, opened 2024-01-26
fix: Fix deferred Session Replay payloads

metal-messiah:
Force agent to send preloaded session replay payloads if present at load time.
---

### Overview
This PR forces the agent to harvest at agg load time if there are preload payloads.  This alleviates an issue where the payload would race with standard payloads at unload time.

### Testing
New tests have been added and old tests have been modified to account for earlier harvests


patrickhousley (approved):

metal-messiah:
firefox is flaky on some STN tests in GHA.  running locally cannot reproduce.
Screenshot 2024-01-26 at 10 35 46 PM

PR #870, opened 2024-01-27
fix: Pass unload options to simultaneous harvests in Session Replay

metal-messiah:
Pass unload options to simultaneous harvests in Session Replay to allow all calls to utilize sendBeacon when unloading
---

### Overview
This should alleviate an issue where the page unloads before the agg has successfully sent preloaded payloads and creates a race condition between xhr and sendBeacon.

### Testing
All tests should continue to pass

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -221,7 +221,7 @@ export class Aggregate extends AggregateBase {
     }
   }
 
-  prepareHarvest () {
+  prepareHarvest ({ opts } = {}) {
     if (!this.recorder) return
     const recorderEvents = this.recorder.getEvents()
     // get the event type and use that to trigger another harvest if needed
@@ -252,7 +252,7 @@ export class Aggregate extends AggregateBase {
     const { session } = getRuntime(this.agentIdentifier)
     if (!session.state.sessionReplaySentFirstChunk) this.syncWithSessionManager({ sessionReplaySentFirstChunk: true })
     this.recorder.clearBuffer()
-    if (recorderEvents.type === 'preloaded') this.scheduler.runHarvest()
+    if (recorderEvents.type === 'preloaded') this.scheduler.runHarvest(opts)
     return [payload]
   }
 
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -28,8 +28,6 @@ import { MODE, SESSION_EVENTS, SESSION_EVENT_TYPES } from '../../../common/sessi
 import { stringify } from '../../../common/util/stringify'
 import { stylesheetEvaluator } from '../shared/stylesheet-evaluator'
 
-let gzipper, u8
-
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
   // pass the recorder into the aggregator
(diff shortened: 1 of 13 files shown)


metal-messiah:
Firefox is being flaky with an unchanged stn test.  Passes locally
image

src/features/session_replay/aggregate/index.js
```diff
@@ -221,7 +221,7 @@ export class Aggregate extends AggregateBase {
     }
   }
 
-  prepareHarvest () {
+  prepareHarvest ({ opts } = {}) {
```
cwli24, line 224: stylistic tweak, but in accordance with the other 3 features that pass opt into their `prepareHarvest` fn, can this be changed to
```suggestion
  prepareHarvest (options) {
```
and have the schedule pass in `({ ...opts, retry})`

cwli24: Or better yet, just straight pass `{ unload: true}` into the runHarvest call. There's two separate opts-passing system between feature and the runHarvest (which includes the `scheduleHarvest` used on retries) cycle, and this may overwrite something?

src/features/session_replay/aggregate/index.js
```diff
@@ -252,7 +252,7 @@ export class Aggregate extends AggregateBase {
     const { session } = getRuntime(this.agentIdentifier)
     if (!session.state.sessionReplaySentFirstChunk) this.syncWithSessionManager({ sessionReplaySentFirstChunk: true })
     this.recorder.clearBuffer()
-    if (recorderEvents.type === 'preloaded') this.scheduler.runHarvest()
+    if (recorderEvents.type === 'preloaded') this.scheduler.runHarvest(opts)
```
cwli24, line 255: It still seems inefficiently redundant to me that this runs unload harvest nested within the pagehide unload harvest callback, followed by vischange unload harvest callback so that prepareHarvest is being called 3 times here.

cwli24: I also thought we landed on sending 'preloaded' nodes immediately after feature-ready state in reg XHR instead.

metal-messiah: > It still seems inefficiently redundant to me that this runs unload harvest nested within the pagehide unload harvest callback, followed by vischange unload harvest callback so that prepareHarvest is being called 3 times here.

This is not happening, the problem that is happening is that unload is triggering prepareHarvest, and prepareHarvest is triggering other subsequent harvests (line 255) without the unload params so they are trying to be sent via xhr instead of beacon if that occurs.

src/features/session_replay/aggregate/index.js
metal-messiah: it isnt always going to be unload: true

metal-messiah: > There's two separate opts-passing system between feature and the runHarvest (which includes the scheduleHarvest used on retries) cycle, and this may overwrite something?

The opts being held are the same opts that were originally passed into runHarvest, which just means the opts will be preserved for other triggered runHarvests within the same loop

src/features/session_replay/aggregate/index.js
cwli24: That's also as I understand it, but why is prepareHarvest calling into itself for 1, and 2, unloading is already being handled by the scheduler module itself, so while features can dictate modifications to payload if it's an unload, it's redundant to have features do the same job the scheduler does

metal-messiah: I'm not sure i'm following what you are proposing.  Can you dictate more clearly what strategy you are advocating for (specifically for handing preload payloads, of which there could be variable number of and of which could be excessively large due to snapshotting)?  There is currently no mechanism in the harvester to call for a feature to prep a payload multiple times.  which seems less of a duty for the harvest scheduler and more on the feature to manage that relationship.  There are a lot of places that our features trigger the scheduler to harvest immediately.  this just uses one more, and in doing so provides the rules that it did not use to.

patrickhousley (approved):

PR #877, opened 2024-02-06
fix: Fix adblock memory leak

Joe-Degler:
Fixed memory leaks caused by request resolve failure (e.G. due to adblocker)
---

### Overview

On our website ( playbook.com ), some users were noticing severe performance degredation when uploading big files. Memory usage grew, and cancelling of the upload did not resolve the issue, eventually causing in a crash of the upload or the website itself.

After investigation, we were able to pin down the issue to the New Relic Browser Agent. However, the issue only occured when adblock (in my case: uBlock Origin) was enabled.

We noticed that two URLs specifically were being blocked: `https://js-agent.newrelic.com/nr-spa-1.245.0.min.js` and `https://bam.nr-data.net/`. 

(Note: Issue occured also with newest version of Browser Agent)

The issue itself is not actually related to uploading - this only made the symptoms much more obvious, due to the size of chunked ArrayBuffers. Even when using the website normally, a slow upwards creep of memory heap size happens.

The main problem is the `buffer` / `backlog` of the ee (event emitter?) holding on to every raw event, but in case of failure to load additional assets or submitting them it's never getting flushed. This results in the backlog growing indefinitely, slowly consuming the users memory. Upload makes this particularly noticable because the buffer (in this case, I think it was `spa` specifically) holds on to every ArrayBuffer reference.

We were able to work around the issue with some monkeypatching - it would be great to get these issues fixed upstream, instead.

We found 3 specific places that resulted in this problem:

---
**Problem #1: `send()` method of `harvest.js` is not using the appropriate event listener for failing requests**

[Link to Change](https://github.com/newrelic/newrelic-browser-agent/pull/877/commits/77accde396872122ae014a6da54c9cd1f5b1db30)

The harvester attempts to submit data to `https://bam.nr-data.net/`.

The `load` [event](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/load_event) is not fired if the XHR is aborted client-side. This causes the callback to never be fired - in turn, never calling `cbCallback` (even with a 400+), which in turn again, never causes the abort [here](https://github.com/newrelic/newrelic-browser-agent/pull/877/files#diff-f006e010fc9058844e6c293ce7f7248a2c743172b1d28bdb11eba9ae236d2627R114) to be called.

Replacing this event with the `loadend` [event](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/loadend_event) seems appropriate. It fires even on an error, but returns an "failed" XHR object with status code 0. 

I've adjusted the `sent` boolean to return `false` if the request has not been sent (eg, didn't leave the client). **I'm not 100% sure on the implications here, and this change is irrelevant for the fix itself.**

With this change, the callback is now successfully invoked for any successful, failed, or errored request.

---
**Problem #2: The `abort()` function of the `contextual-ee.js` replaces the backlog object**

[Link to change](https://github.com/newrelic/newrelic-browser-agent/pull/877/commits/7dec931bd0fe05817223e10a0b3d932286c8b075)

When `abort()` is called, it replaces `globalInstance.backlog` with a fresh new `{}` object. However, all "sub-features" (or emitters, I guess?) hold a direct reference to the original `ee.backlog` object. As they reference this object directly, two things happen:
- Even when the main emitter is aborted, in some circumstances `force` was true, causing the emitter to continue trying to emit
- Because the sub-emitter is using a direct reference to `backlog` (which has been stored on initialization), it accesses the "old" (already discarded by the main emitter) object indefinitely. On this object, the feature / group still exists, so it passes the `if (bufferGroup)` check and continues pushing events to the buffer. This also causes it to never be garbage collected, as references still exist, obviously.

Instead of replacing the main emitter with an empty object, instead iterate over its keys and delete all the existing groups. This clears the buffer, as expected.

---
**Problem #3: Failure to import additional code does not drain buffers and does not unset handlers**

[Link to change](https://github.com/newrelic/newrelic-browser-agent/pull/877/commits/020c6f6e25ab696878e47115366e37a6908cc507)

There's two distinct places I could find that try to initiate an additional import of code from `https://js-agent.newrelic.com/` to import additional features. In both of these, if they fail, the corresponding buffers are not drained adequately, and keep emitting to the buffer. 
The first case is in `api.js`, which, if it fails, only raises a warning that `Downloading runtime APIs failed`. No effort is made in draining and de-registering the `api` buffer/feature.

The second case is in `instrument-base.js`, in which it does start a drain in the exception handler for each drain that failed, but the drain is immediately short-circuiting because the feature is already registered in the registry for the agent. 

I went with a "crude" fix here, still aligning with what I think is the spirit of the architecture, in the form of an `force` parameter, which forces the method to drain the group immediately. 

---

All three of these fixes were necessary to banish the memory leak and make things work properly. They're separated neatly in their own commits.

### Related Issue(s)

[I found this unresolved issue from a year ago.](https://forum.newrelic.com/s/hubtopic/aAX8W0000008fH8WAI/memory-leak-caused-when-browseragent-not-initialized-propertly#:~:text=Memory%20leak%20caused%20when%20BrowserAgent%20not%20initialized%20propertly,-4&text=There%20seems%20to%20be%20an,is%20causing%20a%20memory%20leak.&text=After%20some%20investigation%2C%20we%20found,associated%20with%20the%20loaded%20page.)

### Testing

Replicating this is suprisingly simple. Install `uBlock origin`, or, alternatively, block both urls locally in your environments. 
If you're testing locally, `uBlock Origin` will only block the egress traffic to `https://bam.nr-data.net/`, but allow loading more data through `localhost:63342/build/nr-spa.min.js`. Depending on what you're testing, you may have to block the import URL or not.

Create a simple index.html file similar to the github instructions, importing `nr-loader-spa.min.js`. 

Evaluate `newrelic.ee.backlog`  - you'll observe that it's filled with data. Depending on the activities you do on the website, you'll also see that the backlog keeps increasing. 

If you've *only* blocked the data egress sink:

When applying the commit that fixes the harvest callback: 
you'll notice that `newrelic.ee.aborted == true`, and that `newrelic.ee.backlog == {}`, but if you check out `newrelic.initializedAgents[<foobar>].features[<barfoo>].ee.backlog`, you'll see it's still referencing the old buffer. 

Additionally, applying the commit that fixes abort:
You'll notice that now, the ee of the features of the initializedAgents are also empty, as the original `newrelic.ee.backlog` now got emptied, as opposed to replaced with an empty object.

If you've also blocked importing additional code from `nr-spa.min.js`:
Load the page, observe how `newrelic.ee.backlog` is filled with data. that may keep growing and isn't flushed.

Apply the commit that fixes the drain on import error.

Observe how `newrelic.ee.backlog` is now empty, and does not grow anymore.

### DIFF
--- a/src/common/drain/drain.js
+++ b/src/common/drain/drain.js
@@ -41,13 +41,14 @@ function curateRegistry (agentIdentifier) {
  * its own named group explicitly, when ready.
  * @param {string} agentIdentifier - A unique 16 character ID corresponding to an instantiated agent.
  * @param {string} featureName - A named group into which the feature's buffered events are bucketed.
+ * @param {boolean} force - Whether to force the drain to occur immediately, bypassing the registry and staging logic.
  */
-export function drain (agentIdentifier = '', featureName = 'feature') {
+export function drain (agentIdentifier = '', featureName = 'feature', force = false) {
   curateRegistry(agentIdentifier)
   // If the feature for the specified agent is not in the registry, that means the instrument file was bypassed.
   // This could happen in tests, or loaders that directly import the aggregator. In these cases it is safe to
   // drain the feature group immediately rather than waiting to drain all at once.
-  if (!agentIdentifier || !registry[agentIdentifier].get(featureName)) return drainGroup(featureName)
+  if (!agentIdentifier || !registry[agentIdentifier].get(featureName) || force) return drainGroup(featureName)
 
   // When `drain` is called, this feature is ready to drain (staged).
   registry[agentIdentifier].get(featureName).staged = true
--- a/src/common/drain/drain.test.js
+++ b/src/common/drain/drain.test.js
@@ -46,6 +46,15 @@ describe('drain', () => {
     expect(emitSpy).toHaveBeenCalledWith('drain-pumbaa', expect.anything())
   })
 
+  test('immediately drains when force is true', () => {
+    registerDrain('abcd', 'page_view_timing')
+    registerDrain('abcd', 'page_view_event')
+
+    let emitSpy = jest.spyOn(ee.get('abcd'), 'emit')
+    drain('abcd', 'page_view_event', true)
+    expect(emitSpy).toHaveBeenNthCalledWith(1, 'drain-page_view_event', expect.anything())
+  })
+
   test('defaults to "feature" group when not provided one', () => {
     let emitSpy = jest.spyOn(ee.get('abcd'), 'emit')
     drain('abcd')
--- a/src/common/event-emitter/contextual-ee.component-test.js
+++ b/src/common/event-emitter/contextual-ee.component-test.js
@@ -143,6 +143,19 @@ describe('event-emitter buffer', () => {
     expect(ee.isBuffering(eventType)).toEqual(false)
   })
 
+  test('it should empty the backlog on abort as opposed to replacing it', async () => {
+    const { ee } = await import('./contextual-ee')
+
+    const backlog = {
+      api: ['foo', 'bar', 'baz']
+    }
+    ee.backlog = backlog
+    ee.abort()
+
+    expect(ee.backlog).toBe(backlog)
+    expect(ee.backlog).toEqual({})
+  })
+
   test('it should not buffer after drain', async () => {
     const { ee } = await import('./contextual-ee')
     const { drain } = await import('../drain/drain')
--- a/src/common/event-emitter/contextual-ee.js
+++ b/src/common/event-emitter/contextual-ee.js
@@ -144,5 +144,11 @@ function ee (old, debugId) {
 
 function abort () {
   globalInstance.aborted = true
-  globalInstance.backlog = {}
+  // The global backlog can be referenced directly by other emitters,
+  // so we need to delete its contents as opposed to replacing it.
+  // Otherwise, these references to the old backlog would still exist
+  // and the keys will not be garbage collected.
+  Object.keys(globalInstance.backlog).forEach(key => {
+    delete globalInstance.backlog[key]
+  })
 }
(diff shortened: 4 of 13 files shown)


patrickhousley:
Thank you so much @Joe-Degler I am taking a look at this now.

patrickhousley:
All unit and jil tests for all browsers are passing locally. One test failed in wdio but I think it's just flakiness.

```
[Safari iOS #16-28] Running: Safari on iOS
[Safari iOS #16-28] Session ID: 303df308a0a74ba59381d108bd8d0a83
[Safari iOS #16-28]
[Safari iOS #16-28] » /tests/specs/session-replay/ingest.e2e.js
[Safari iOS #16-28] Session Replay Ingest Behavior
[Safari iOS #16-28]    ✖ Should empty event buffer when sending (3 retries)
[Safari iOS #16-28]    ✓ Should stop recording if 429 response (1 retries)
[Safari iOS #16-28]
[Safari iOS #16-28] 1 passing (3m 22.9s)
[Safari iOS #16-28] 1 failing
[Safari iOS #16-28]
[Safari iOS #16-28] 1) Session Replay Ingest Behavior Should empty event buffer when sending
[Safari iOS #16-28] expect(received).toBeGreaterThan(expected)

Expected: > 0
Received:   0
[Safari iOS #16-28] Error: expect(received).toBeGreaterThan(expected)
[Safari iOS #16-28]
[Safari iOS #16-28] Expected: > 0
[Safari iOS #16-28] Received:   0
[Safari iOS #16-28]     at Context.<anonymous> (/Users/phousley/code/newrelic-browser-agent/tests/specs/session-replay/ingest.e2e.js:18:43)
[Safari iOS #16-28]     at processTicksAndRejections (node:internal/process/task_queues:95:5)
[Safari iOS #16-28]

Spec Files:	 1075 passed, 3 retries, 1 failed, 24 skipped, 1100 total (100% completed) in 01:01:47
```

I will run through the manual tests and work on building out e2e tests for the three specific issues tomorrow.

patrickhousley (approved):

Joe-Degler:
No Problem! Do you need me to do anything else?

PR #884, opened 2024-02-09
fix: Add safe logic to snapshots

metal-messiah:
Safe-guarded a race condition where an asynchronous call to fix a stylesheet could trigger the agent to try to force a snapshot after it has already stopped recording due to a visibility change.
---

### Overview
The agent makes an async call to attempt to fix non-inlinable stylesheets, but this operation is asynchronous.  If anything triggers the agent to stop recording before the async job resolves, it will errantly attempt to take a snapshot with no active recording.  This behavior safe-guards that in two ways with a status check and a try/catch.

### Related Issue(s)
https://newrelic.slack.com/archives/C04QZSS6XPV/p1707505446130859

### DIFF
--- a/src/features/session_replay/shared/recorder.js
+++ b/src/features/session_replay/shared/recorder.js
@@ -166,7 +166,12 @@ export class Recorder {
 
   /** force the recording lib to take a full DOM snapshot.  This needs to occur in certain cases, like visibility changes */
   takeFullSnapshot () {
-    recorder.takeFullSnapshot()
+    try {
+      if (!this.recording) return
+      recorder.takeFullSnapshot()
+    } catch (err) {
+      // in the off chance we think we are recording, but rrweb does not, rrweb's lib will throw an error.  This catch is just a precaution
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/features/session_replay/shared/recorder.js
```diff
@@ -166,7 +166,12 @@ export class Recorder {
 
   /** force the recording lib to take a full DOM snapshot.  This needs to occur in certain cases, like visibility changes */
   takeFullSnapshot () {
-    recorder.takeFullSnapshot()
+    try {
+      if (!this.recording) return
+      recorder.takeFullSnapshot()
```
patrickhousley, line 171: Should we abort the network call as well?

patrickhousley (approved):

src/features/session_replay/shared/recorder.js
metal-messiah: By the time we get to this the network call should have resolved so we would need to abort elsewhere and TBH i dont think its worth the extra complexity to manage

Issue #895, opened 2024-02-15
Typescript Newrelic Browser agent don't have noticeError method


master96:
[NOTE]: # ( ^^ Wrong types for Browser Agent ^^ )

## Description

[NOTE]: # ( When BrowserAgent is initialized it exposes noticeError method but when object is logged there is no such method available. It's actually inside api key of browser agent )

NoticeError function is missing from BrowserAgent it's available inside api key but not typed.

## Steps to Reproduce

I have following code 

```javascript
import { BrowserAgent } from '@newrelic/browser-agent/loaders/browser-agent';

export let newrelicWebClient: BrowserAgent;
  const options = {
    init: {
      distributed_tracing: { enabled: true },
      privacy: { cookies_enabled: true },
      ajax: { deny_list: ['*****'] },
    }, // NREUM.init
    info: {
      beacon: '*****',
      errorBeacon: '*****',
      licenseKey: '*****',
      applicationID: '*****',
      sa: 1,
    }, // NREUM.info
    loader_config: {
      accountID: '*****',
      trustKey: '*****',
      agentID: '*****',
      licenseKey: '*****',
      applicationID: '*****',
    }, // NREUM.loader_config
  };

  // The agent loader code executes immediately on instantiation.
  newrelicWebClient = new BrowserAgent(options);
  newrelicWebClient?.setApplicationVersion('*****' || null);


and i am calling it like

console.log(newrelicWebClient);
 newrelicWebClient?.noticeError(e,{some_key:'value'})
 
```
in console log i get 

![image](https://github.com/newrelic/newrelic-browser-agent/assets/12187986/817e254d-59ae-4a60-a8e9-6e597e8833eb)

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )

* ex: Browser name and version - Chrome
* ex: Operating System and version - Mac
* Browser agent version involved (e.g., v1.229.0) - "@newrelic/browser-agent": "1.251.0",
* APM language and version, if applicable (e.g., Dotnet 10.2.0), and auto|manual

## Additional context

[TIP]:  # ( Add any other context about the problem here. )

patrickhousley:
Hi @master96 and thanks for reaching out. Are you getting a specific error? The method is there, you just have to drill down the inheritance chain.

<img width="908" alt="image" src="https://github.com/newrelic/newrelic-browser-agent/assets/3984979/55299063-8293-418e-9d27-ef8291d1d6c1">

[...]

PR #896, opened 2024-02-16
chore: add tests to check for exposed api methods

patrickhousley:
---

### Overview

This just adds some tests to verify we are exposing the api methods correctly on the npm agent. We are not going to type the `api` property on the agent for now.

### Related Issue(s)

closes https://github.com/newrelic/newrelic-browser-agent/issues/895

### DIFF
--- a/tests/specs/api.e2e.js
+++ b/tests/specs/api.e2e.js
@@ -1,8 +1,38 @@
 import { notIE } from '../../tools/browser-matcher/common-matchers.mjs'
+import { apiMethods, asyncApiMethods } from '../../src/loaders/api/api-methods'
 
 describe('newrelic api', () => {
   afterEach(async () => {
     await browser.destroyAgentSession()
+  })
+
+  it('should expose api methods', async () => {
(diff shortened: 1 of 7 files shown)


tests/specs/api.e2e.js
```diff
@@ -1,10 +1,40 @@
...
+    const agentInstanceApiMethods = await browser.execute(function () {
+      function getAllPropertyNames (obj) {
+        let result = new Set()
+        while (obj) {
+          Object.getOwnPropertyNames(obj).forEach(p => result.add(p))
+          obj = Object.getPrototypeOf(obj)
+        }
+        return [...result]
+      }
+      return getAllPropertyNames(Object.values(newrelic.initializedAgents)[0])
```
cwli24, line 25: Would be less dependent on knowing the inheritance, and if that ever change, to just do an `expect` loop through `typeof newrelic.initializedAgents[0][method] === 'function'`, no?

patrickhousley: Could do that but the code grabs all properties of the object traversing the inheritance chain. Either would work IMHO.

metal-messiah (approved):

PR #898, opened 2024-02-22
feat: Apply deny list to ajax metrics

Wrhector:
The ajax deny list has historically only applied to the collection of ajax events and not ajax metrics. With this feature, the deny list will also be applied to the collection and reporting of ajax timeslice metric data. If an ajax request matches an entry in the deny list, both events and metrics will be blocked from being collected in the agent. This feature will be behind a feature flag to assess impact before final complete rollout. Thanks to community member [@Wrhector](https://github.com/Wrhector) for contributing this feature.
---

### Overview

Updating the ajax feature to apply the deny list to the collection of ajax metrics if the feature flag is enabled. If the feature flag is not present, ajax metrics should continue to be collected.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-239921


patrickhousley (changes requested):
Blocking this PR while we discuss the implication internally.

patrickhousley:
Running test suite:
https://github.com/newrelic/newrelic-browser-agent/actions/runs/8205283383
https://github.com/newrelic/newrelic-browser-agent/actions/runs/8205299927

patrickhousley (approved):
All tests passing.

PR #899, opened 2024-02-23
fix: Page load after session timeouts don't start new session

cwli24:
When a session expires or becomes inactive while the user does not have the page open and they re-open the page, that is now treated as a new session.
---

### Overview

We had an issue where off-page expiration or timeouts of session resulted in the next on-page load to initialize a session that isn't new (i.e. `isNew` false). This is not intended and has been fixed.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-230914

### Testing

New jest test cases added to check for correct program flow and `isNew` value rather than with entire agent. This suffices to get the behavior we want.

### DIFF
--- a/src/common/session/session-entity.component-test.js
+++ b/src/common/session/session-entity.component-test.js
@@ -236,25 +236,25 @@ describe('reset()', () => {
 })
 
 describe('isNew', () => {
-    test('isNew is true after the session resets by timers | true -> true', async () => {
+    test('is true after the session resets by timers | true -> true', async () => {
       const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
       expect(sessionInstance.isNew).toEqual(true)
       sessionInstance.reset() // imitate what expireMs and inactiveMs does on activation
       expect(sessionInstance.isNew).toEqual(true)
     })
-    test('isNew is true after the session resets by timers | false -> true', () => {
+    test('is true after the session resets by timers | false -> true', () => {
       storage.set(`${PREFIX}_${key}`, { ...model, value, expiresAt: Date.now() + 100000, inactiveAt: Date.now() + 100000, updatedAt: Date.now() })
       const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
       expect(sessionInstance.isNew).toEqual(false)
       sessionInstance.reset()
       expect(sessionInstance.isNew).toEqual(true)
     })
-    test('isNew is true if reset happens on initialization after time outs off-view | expiresAt', () => {
+    test('is true if reset happens on initialization after time outs off-view | expiresAt', () => {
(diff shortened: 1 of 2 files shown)


src/common/session/session-entity.component-test.js
```diff
@@ -235,6 +235,18 @@ describe('reset()', () => {
...
+  expect(sessionInstance.isNew).toEqual(true)
+  sessionInstance.reset() // imitate what expireMs and inactiveMs does on activation
+  expect(sessionInstance.isNew).toEqual(true)
+})
+test('isNew is true if reset happens on initialization after time outs off-view', () => {
+  storage.set(`${PREFIX}_${key}`, { ...model, value, expiresAt: Date.now() - 1, inactiveAt: Date.now() - 1, updatedAt: Date.now() })
+  const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
+  expect(sessionInstance.isNew).toEqual(true)
+})
+
```
metal-messiah, line 249:
```suggestion
describe('isNew', () => {
  test('isNew is true after the session resets by timers', async () => {
    const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
    expect(sessionInstance.isNew).toEqual(true)
    sessionInstance.reset() // imitate what expireMs and inactiveMs does on activation
    expect(sessionInstance.isNew).toEqual(true)
  })
  test('isNew is true if reset happens on initialization after time outs off-view | expiresAt', () => {
    storage.set(`${PREFIX}_${key}`, { ...model, value, expiresAt: Date.now() - 1, inactiveAt: Date.now() + 100000, updatedAt: Date.now() })
    const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
    expect(sessionInstance.isNew).toEqual(true)
  })
  test('isNew is true if reset happens on initialization after time outs off-view | inactiveAt', () => {
    storage.set(`${PREFIX}_${key}`, { ...model, value, expiresAt: Date.now() + 100000, inactiveAt: Date.now() - 1, updatedAt: Date.now() })
    const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
    expect(sessionInstance.isNew).toEqual(true)
  })
  test('isNew is true if reset happens on initialization after time outs off-view | invalid data', () => {
    const storage = new LocalMemory({ [`${PREFIX}_${key}`]: { invalid_fields: true } })
    const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
    expect(sessionInstance.isNew).toEqual(true)
  })
})
```

metal-messiah: nice work on this.  the above just splits the test apart to ensure that both expires and inactive timestamps (and invalid data in general) still lead to a positive isNew value

metal-messiah (approved):

PR #903, opened 2024-02-28
feat: Isolate observation context per agent

patrickhousley:
Initial implementation of an isolated observation context per agent running on the same page. This reduces the modification of instances of globals like events and instead each agent will track it's own observation context internally.
---

### Overview

This implements the very minimum of an isolated observation context with backwards compatibility for features that have not been migrated yet. This is being done as part of the timestamp work to make it easier to apply NR time to events and to observe the RUM call without having to make too many modifications to the harvester or hard code URLs.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-222414


src/common/context/observation-context-manager.js
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Returns the observation context tied to the supplied construct. If there has been
+   * no observation construct created, an empty object is created and stored as the current
+   * context.
+   * @param key {unknown} The construct being observed such as an XHR instance
+   * @return {EventContext} An object of key:value pairs to track as
+   * part of the observation
+   */
+  getCreateContext (key) {
```
patrickhousley, line 33: The event emitter is still responsible for creating the observation context. I am not sure that is the right place for it and it's more just an easy central place to do it.

metal-messiah:
We should probably sync on what this is fixing to make sure we are on the same page since it increases the agent build size

src/loaders/agent-base.js
```diff
@@ -1,12 +1,26 @@
...
  * @typedef {import('./api/interaction-types').InteractionInstance} InteractionInstance
  */
 
 export class AgentBase {
+  agentIdentifier
+  observationContext = new ObservationContextManager()
+
+  constructor (agentIdentifier = generateRandomHexString(16)) {
+    this.agentIdentifier = agentIdentifier
+    this.eventEmitter = ee.get(agentIdentifier)
```
metal-messiah, line 18: Is there a purpose for storing this on the class?  It doesn't look like its referenced by anything.  If the answer is yes, we should consider if we should call it `this.ee` to match the pattern found in the features

tests/specs/api.e2e.js
```diff
@@ -15,12 +15,14 @@ describe('newrelic api', () => {
...
+        var result = new Set()
         while (obj) {
-          Object.getOwnPropertyNames(obj).forEach(p => result.add(p))
+          Object.getOwnPropertyNames(obj).forEach(function (p) {
+            return result.add(p)
+          })
           obj = Object.getPrototypeOf(obj)
         }
-        return [...result]
+        return Array.from(result)
```
metal-messiah, line 25: ICR Do we have a polyfill for this already for ie11?

patrickhousley: Yes.

src/loaders/agent-base.js
patrickhousley: It is not referenced per se but it continues the pattern that I believe we are, or should be, moving toward where `Agent` is the top-level class that composes all other class instances / data that makes up a running browser agent. Technically, it is not referenced anywhere right now and I could remove the assignment.

As for the name, I would argue that we should not perpetuate the use of acronyms and short names like `ee`. Yes, technically we all know what it means but that is most likely because we know the code base. I would like us to start using more descriptive variable names. The name does not make a difference in the final bundle size but it can help make the code easier to read and reason about.

metal-messiah (approved):

src/loaders/agent-base.js
patrickhousley: Fair point. I will update here to remove the attachment to the class instance for now until that larger effort is undertaken.

PR #911, opened 2024-03-05
feat: Calculate New Relic time in the agent

patrickhousley:
Migrating the process of time-stamping harvested data to the agent. The agent will utilize the date and time reported by the ingest pipeline during the RUM call as well as browser network and timing stats to derive a server time offset that can be applied to harvested data as event occur. This will improve the ability to visualize the harvested data on a timeline within New Relic UI.
---

### Overview

Adding a new time keeper class to the agent that will utilize the RUM network call and the resource performance APIs to calculate the server time offset. This can then be used by features and other aspects of the agent to correct an absolute timestamp from browser time to server time and convert a time value relative to page origin to a timestamp that accounts for the server time offset.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-222414


src/common/timing/time-keeper.js
```diff
@@ -0,0 +1,90 @@
...
+    if (!responseDateHeader) {
+      throw new Error('Missing date header on rum response.')
+    }
+
+    const resourceEntries = globalScope.performance.getEntriesByName(rumRequestUrl, 'resource')
+    if (!Array.isArray((resourceEntries)) || resourceEntries.length === 0) {
+      throw new Error('Missing rum request performance entry.')
+    }
+
+    let medianRumOffset = 0
```
cwli24, line 50: Nitpick: can we change this var to be something bit more intuitive, like `estOneWayNetworkTime` ?

src/common/timing/time-keeper.js
```diff
@@ -0,0 +1,90 @@
...
+    this.#serverOffset = 0
+    if (typeof resourceEntries[0].responseStart === 'number' && resourceEntries[0].responseStart !== 0) {
+      // Cors is enabled and we can make a more accurate calculation of NR server time
```
cwli24, line 53: FMI, in what cases does rum service not provide cors header allowing any domain, or at least the [Timing-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Timing-Allow-Origin)?

src/common/timing/time-keeper.js
```diff
@@ -0,0 +1,90 @@
...
+      medianRumOffset = (resourceEntries[0].responseStart - resourceEntries[0].requestStart) / 2
+      this.#serverOffset = Math.floor(resourceEntries[0].requestStart + medianRumOffset)
+    } else {
+      // Cors is disabled or erred, we need to use a less accurate calculation
+      medianRumOffset = (resourceEntries[0].responseEnd - resourceEntries[0].fetchStart) / 2
+      this.#serverOffset = Math.floor(resourceEntries[0].fetchStart + medianRumOffset)
+    }
+
+    // This is the NR server time
+    const serverTimestamp = Date.parse(responseDateHeader) - medianRumOffset
```
cwli24, line 63: The "(estimated) current server time" at this point should be the date stamped by the server **+** the time it took that to be relayed back aka return network time (`medianRumOffset` / `estOneWayNetworkTime`).

src/common/timing/time-keeper.js
```diff
@@ -0,0 +1,90 @@
...
+    // Corrected page origin time
+    this.#correctedOriginTime = Math.floor(serverTimestamp - this.#serverOffset)
```
cwli24, line 65: I'm not understanding `serverOffset` or this calculation. So that I'm clear, an event that occur at some later time will have a relative DOMHighResTimeStamp that's based on client `timeOrigin`, different from `serverTimestamp` aka "server timeOrigin".
We want to map that old DOMHighResTimeStamp to a new DOMHighResTimeStamp that's relative to `serverTimestamp` at this particular point in time, right?
If so, why not capture the DOMHighResTimeStamp right here at this point in time, and _subtract_ it from the event's later DOMHighResTimeStamp -- let's call this `newPerfNow`? Then the "server time" at which the client side event occurred would simply be `serverTimestamp` + `newPerfNow`.

src/common/timing/time-keeper.js
patrickhousley: I feel like the name given is more descriptive. It is the median calculated time the RUM call took offset from the current page origin time.

src/common/timing/time-keeper.js
patrickhousley: The RUM call did not have the Timing-Allow-Origin header or the necessary CORS Expose Header header. I had to add those in a separate PR. That said, we should never actually hit the alternate code as there is no browser we support that does not support this branch of code. If a customer is using a proxy and alters the headers, thats on them. So, I could remove the alternate code path or leave it as a "just in case".

src/common/timing/time-keeper.js
patrickhousley: [Here](https://whimsical.com/calculated-nr-time-5ihwHhG6VhcTGV9Czi6BfU) is the whimsical that details out the calculations a little more. Let me know if that and the above answers are still unclear.

src/common/timing/time-keeper.js
cwli24: Oh, so this is est server _before_ RUM?

src/common/timing/time-keeper.js
cwli24: image
`serverTimestamp` doesn't seem to exist in the whimsical; it uses the dateHeader parsed directly without subtracting the rumOffset. This in part threw me off because in the final equation, `medianRumOffset` is subtracted twice.

src/features/utils/feature-base.js
```diff
@@ -16,5 +17,7 @@ export class FeatureBase {
      * @type {boolean}
      */
     this.blocked = false
+
+    this.timeKeeper = TimeKeeper.getTimeKeeperByAgentIdentifier(this.agentIdentifier)
```
metal-messiah, line 21: Just to clarify -- It looks like the plan for the future will be for features to access the timekeeper from their class instance props?

metal-messiah (approved):
General concept looks good, g2g when tests are done

src/features/utils/feature-base.js
patrickhousley: Correct. Putting the timekeeper here reduces the amount of code in each feature to go and grab the agent timekeeper and makes it easier to use the timekeeper.

PR #913, opened 2024-03-08
feat: Add rum-response metadata handler

metal-messiah:
The initial connect call can now supply browser app metadata.  This adds a handler to store and use that metadata for future harvests.
---

### Overview
This PR introduces a handler for the agent metadata response from the RUM call.  This will initially contain a real entityGuid for the browser app, regardless if it was APM injected or standalone.  The entityGuid implementations will be introduced in subsequent PRs, as it will be applied to SR, ST and Logs payloads in the future. 

### Related Issue(s)
NR-215487

### Testing
A new test has been added to account for the rum-supplied app id.


src/common/config/state/runtime.js
```diff
@@ -18,6 +18,7 @@ const model = {
   origin: '' + globalScope.location,
   ptid: undefined,
   releaseIds: {},
+  rumCallMetadata: {},
```
patrickhousley, line 21: Nitpick

```suggestion
  agentMetadata: {},
```

patrickhousley (approved):

PR #918, opened 2024-03-18
feat: Update agent to use new relic server time

patrickhousley:
Update the agent to use New Relic server time as the timestamp for harvested page action and error data.
---

### Overview

This updates all the harvested data to use the calculated NR server time as the the basis for timestamping harvested data. See the [CDD](https://newrelic.atlassian.net/wiki/spaces/INST/pages/3180429574/Normalize+Browser+Agent+Timestamps) for spec changes.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-222430
https://new-relic.atlassian.net/browse/NR-222419
https://new-relic.atlassian.net/browse/NR-222426

### DIFF
--- a/src/common/context/shared-context.js
+++ b/src/common/context/shared-context.js
@@ -1,4 +1,5 @@
 import { warn } from '../util/console'
+import { TimeKeeper } from '../timing/time-keeper'
 
 const model = {
   agentIdentifier: '',
@@ -17,5 +18,7 @@ export class SharedContext {
     } catch (err) {
       warn('An error occured while setting SharedContext', err)
     }
+
+    this.timeKeeper = TimeKeeper.getTimeKeeperByAgentIdentifier(this.sharedContext.agentIdentifier)
   }
 }
--- a/src/common/context/shared-context.js
+++ b/src/common/context/shared-context.js
@@ -1,9 +1,9 @@
 import { warn } from '../util/console'
-import { TimeKeeper } from '../timing/time-keeper'
 
 const model = {
   agentIdentifier: '',
-  ee: undefined
+  ee: undefined,
+  timeKeeper: undefined
 }
 
 export class SharedContext {
@@ -16,9 +16,7 @@ export class SharedContext {
         if (Object.keys(model).includes(key)) this.sharedContext[key] = value
       })
     } catch (err) {
-      warn('An error occured while setting SharedContext', err)
+      warn('An error occurred while setting SharedContext', err)
     }
-
-    this.timeKeeper = TimeKeeper.getTimeKeeperByAgentIdentifier(this.sharedContext.agentIdentifier)
   }
 }
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -179,7 +179,7 @@ export class Harvest extends SharedContext {
       encodeParam('v', VERSION),
       transactionNameParam(info),
       encodeParam('ct', runtime.customTransaction),
-      '&rst=' + this.timeKeeper.now(),
+      '&rst=' + this.sharedContext.timeKeeper.now(),
       '&ck=0', // ck param DEPRECATED - still expected by backend
       '&s=' + (runtime.session?.state.value || '0'), // the 0 id encaps all untrackable and default traffic
       encodeParam('ref', ref),
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -9,7 +9,6 @@ import * as submitData from '../util/submit-data'
 import { getLocation } from '../url/location'
 import { getInfo, getConfigurationValue, getRuntime, getConfiguration } from '../config/config'
 import { cleanURL } from '../url/clean-url'
-import { now } from '../timing/now'
 import { eventListenerOpts } from '../event-listener/event-listener-opts'
 import { Obfuscator } from '../util/obfuscate'
 import { applyFnToProps } from '../util/traverse'
@@ -180,7 +179,7 @@ export class Harvest extends SharedContext {
       encodeParam('v', VERSION),
       transactionNameParam(info),
       encodeParam('ct', runtime.customTransaction),
-      '&rst=' + now(),
+      '&rst=' + this.timeKeeper.now(),
       '&ck=0', // ck param DEPRECATED - still expected by backend
       '&s=' + (runtime.session?.state.value || '0'), // the 0 id encaps all untrackable and default traffic
       encodeParam('ref', ref),
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -16,6 +16,7 @@ import { SharedContext } from '../context/shared-context'
 import { VERSION } from '../constants/env'
 import { isWorkerScope, isIE } from '../constants/runtime'
 import { warn } from '../util/console'
+import { TimeKeeper } from '../timing/time-keeper'
 
 const warnings = {}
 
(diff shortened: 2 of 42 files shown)


src/common/context/shared-context.js
```diff
@@ -17,5 +18,7 @@ export class SharedContext {
     } catch (err) {
       warn('An error occured while setting SharedContext', err)
     }
+
+    this.timeKeeper = TimeKeeper.getTimeKeeperByAgentIdentifier(this.sharedContext.agentIdentifier)
```
metal-messiah, line 22: This is fine but just wanted to mention that the way this shared context module currently works is it extracts pointers from the parent on L14-17. If `timeKeeper` was in the model on L4-6 it would automatically grab it from the feature and you wouldn't need to define it / fetch it here.  If that's a good or bad pattern is a different discussion point for a different time 😆

src/loaders/agent-base.js
```diff
@@ -4,14 +4,15 @@ import { warn } from '../common/util/console'
...
+import { globalScope } from '../common/constants/runtime'
 
 /**
  * @typedef {import('./api/interaction-types').InteractionInstance} InteractionInstance
  */
 
 export class AgentBase {
   agentIdentifier
-  timeKeeper = new TimeKeeper(this)
+  timeKeeper = new TimeKeeper(Math.floor(globalScope?.performance?.timeOrigin || globalScope?.performance?.timing?.navigationStart || Date.now()))
```
metal-messiah, line 15: This is fine but it represents the third place we've defined the origin time so we just need to follow up and remove `runtime.offset` and `runtime/constants.offset` at some time

src/common/timing/time-keeper.js
```diff
@@ -79,9 +88,10 @@ export class TimeKeeper {
    * Converts a page origin relative time to an absolute timestamp
    * corrected to NR server time.
    * @param relativeTime {number} The relative time of the event in milliseconds
-   * @returns {number} The correct timestamp as a unix/epoch timestamp value
+   * @returns {number} Corrected unix/epoch timestamp
    */
   convertRelativeTimestamp (relativeTime) {
+    if (!this.#correctedOriginTime) throw new Error('InvalidState: Timing correction attempted before NR time calculated.')
```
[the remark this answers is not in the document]
metal-messiah: These things should be mostly-irrelevant anyway right?  The agent shuts down if it doesnt have the correction factor, and we should have some mutual understanding that we cant fix timestamps until the aggregation phase anyway

src/common/harvest/harvest.js
```diff
@@ -177,7 +176,7 @@ export class Harvest extends SharedContext {
       encodeParam('v', VERSION),
       transactionNameParam(info),
       encodeParam('ct', runtime.customTransaction),
-      '&rst=' + now(),
+      '&rst=' + this.timeKeeper.now(),
```
metal-messiah, line 179: This would be the place that `sharedContext` would come into play with the comment above, ie. `this.sharedContext.timeKeeper`

src/loaders/agent-base.js
metal-messiah: `timing/now` as well

src/common/context/shared-context.js
patrickhousley: Honestly, I'm beginning to hate the word `context`. There is room for improvement here, the inheritance tree between our features and other code like the harvester is weird and confusing.

src/loaders/agent-base.js
patrickhousley: Thats the plan.

src/common/timing/time-keeper.js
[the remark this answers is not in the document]
patrickhousley: I can remove them once I have more tests in place, possibly. It is nice to have these though so we can get a more visible indication of when we have done something bad in our code. I wonder if there is a better way of excluding these in the final production build.

src/features/soft_navigations/aggregate/bel-node.js
```diff
@@ -1,12 +1,12 @@
...
+import { TimeKeeper } from '../../../common/timing/time-keeper'
 
 let nodesSeen = 0
 
 export class BelNode {
   belType
   /** List of other BelNode derivatives. Each children should be of a subclass that implements its own 'serialize' function. */
   children = []
-  start = now()
+  start = TimeKeeper.now()
```
cwli24, line 9: This is actually always overwritten by either the `AjaxNode` constructor or the `Interaction` constructor. I think it'd be better to just remove the default
```suggestion
  start
```

src/features/soft_navigations/instrument/index.js
```diff
@@ -23,7 +23,7 @@ export class Instrument extends InstrumentBase {
     const historyEE = wrapHistory(this.ee)
     const eventsEE = wrapEvents(this.ee)
 
-    const trackURLChange = () => handle('newURL', [now(), '' + window.location], undefined, this.featureName, this.ee)
+    const trackURLChange = () => handle('newURL', [TimeKeeper.now(), '' + window.location], undefined, this.featureName, this.ee)
```
cwli24, line 26: Can you actually change this **and** 'newDom' to use performance.now()? Accurate is important to interaction and tying events in. I think there's a larger cleanup audit needed to ensure JSErrors and Ajax are tracked more precisely as well when they pipe into SN for timeline evaluation, but for now this'll keep the precision of interactions to `DOMHighResTimestamp`.
(UI event which _starts_ interactions, and `trackURLChangeEvent` handler for popstate also uses the `timeStamp` property of events which are `DOMHighResTimestamp` -- this is what we want.)

src/common/url/beacon.js
cwli24, line 1: Empty file?

src/features/soft_navigations/aggregate/bel-node.js
patrickhousley: Can we look at that work as part of another ticket? Do you want to open a ticket for that?

src/features/soft_navigations/instrument/index.js
patrickhousley: Can you open a separate ticket for that work? I don't think it needs to be included here and I don't want to add more changes than necessary.

src/loaders/api/api.js
```diff
@@ -150,9 +150,9 @@ export function setAPI (agentIdentifier, forceDrain, runSoftNavOverSpa = false)
       var hasCb = typeof cb === 'function'
       handle(SUPPORTABILITY_METRIC_CHANNEL, ['API/createTracer/called'], undefined, FEATURE_NAMES.metrics, instanceEE)
       // Soft navigations won't support Tracer nodes, but this fn should still work the same otherwise (e.g., run the orig cb).
-      if (!runSoftNavOverSpa) handle(spaPrefix + 'tracer', [now(), name, contextStore], ixn, FEATURE_NAMES.spa, instanceEE)
+      if (!runSoftNavOverSpa) handle(spaPrefix + 'tracer', [TimeKeeper.now(), name, contextStore], ixn, FEATURE_NAMES.spa, instanceEE)
```
metal-messiah, line 153: This is a general comment, and not one that needs any action taken ATM

I've noticed that the inst files only need `.now` and the agg files are the only ones to actually use any correction methods.

We should probably follow up in another effort to see if we can split the timekeeper up to have an inst/agg type dichotomy as well to reduce build size impact for unused mechanisms.

inst == `.now`, agg == `correction methods`

metal-messiah: (or go back to .now() being its own file used in the inst/agg files, and move the timekeeper to the async chunk)

metal-messiah: (this would also eliminate the need to access the session entity pre-load, since we only need it at rum-call time (in the agg)

metal-messiah (approved):
code changes make sense, when tests can pass reliably i think it lgtm

src/loaders/api/api.js
cwli24: I'm on board with having un-floored highrestimestamps if there's nothing hurting it. I'll audit the places using timekeeper.now and open a ticket to clean up those.

PR #919, opened 2024-03-18
feat: Switch web vitals library to attribution build

cwli24:
Report extra attributes for each vital using web vital library's attribution build.
---

### Overview

Most "potentially-helpful" debugging information offered by CWV's attribution build is now included with the `PageViewTiming` event nodes. Many of them have the same property name as given by CWV.
Note that we may remap some or all of them to our own standardized names in the future. At the present, they can be NRQL for using those attribute names directly, like `SELECT firstByteToFCP FROM PageViewTiming WHERE timingName = 'firstContentfulPaint'`.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-191642
* Untracked chore: fix test matcher not considering Android versions

### Testing

Jest tests for each indiv vital & pvt component updated.


src/common/vitals/cumulative-layout-shift.js
```diff
@@ -1,13 +1,19 @@
...
 import { isBrowserScope } from '../constants/runtime'
 
 export const cumulativeLayoutShift = new VitalMetric(VITAL_NAMES.CUMULATIVE_LAYOUT_SHIFT, (x) => x)
 
 if (isBrowserScope) {
-  onCLS(({ value, entries }) => {
-    if (cumulativeLayoutShift.roundingMethod(value) === cumulativeLayoutShift.current.value) return
-    cumulativeLayoutShift.update({ value, entries })
+  onCLS(({ value, attribution, id }) => {
+    const attrs = {
```
patrickhousley, line 10: This is removing a logic check. Was that no longer needed or is there something else we need to put here to prevent duplicate reporting?

src/common/vitals/interaction-to-next-paint.js
```diff
@@ -7,7 +7,14 @@ export const interactionToNextPaint = new VitalMetric(VITAL_NAMES.INTERACTION_TO
 
 if (isBrowserScope) {
 /* Interaction-to-Next-Paint */
-  onINP(({ value, entries, id }) => {
-    interactionToNextPaint.update({ value, entries, attrs: { metricId: id } })
```
patrickhousley, line 11: This is changing the signature of the object we report out via the update method. Where does this change need to be accounted for and has it been?

src/common/vitals/largest-contentful-paint.js
```diff
@@ -7,22 +7,26 @@ import { cleanURL } from '../url/clean-url'
...
 if (isBrowserScope) {
-  onLCP(({ value, entries }) => {
+  onLCP(({ value, attribution }) => {
   /* Largest Contentful Paint - As of WV v3, it still imperfectly tries to detect document vis state asap and isn't supposed to report if page starts hidden. */
     if (initiallyHidden || largestContentfulPaint.isValid) return
 
-    const lcpEntry = entries[entries.length - 1] // this looks weird if we only expect one, but this is how cwv-attribution gets it so to be sure...
-    largestContentfulPaint.update({
-      value,
-      entries,
```
patrickhousley, line 17: This is changing the signature of the object we report out via the update method. Where does this change need to be accounted for and has it been?

src/common/vitals/vital-metric.js
```diff
@@ -20,7 +20,7 @@ export class VitalMetric {
     this.history.push(state)
     this.#subscribers.forEach(cb => {
       try {
-        cb(this.current)
+        cb(state)
```
patrickhousley, line 23: Where was `this.current` even defined?

src/features/page_view_timing/aggregate/index.js
```diff
@@ -43,6 +45,12 @@ export class Aggregate extends AggregateBase {
     timeToFirstByte.subscribe(({ entries }) => {
       this.addTiming('load', Math.round(entries[0].loadEventEnd))
     })
+    // *cli Mar'24 - CLS node won't be added until we fix the rounding problem in schema that's grounding the decimal value to 0
+    // subscribeToVisibilityChange(() => {
+    //   const { name, value, attrs } = cumulativeLayoutShift.current
+    //   if (value === undefined) return
+    //   this.addTiming(name, value * 1000, attrs) // downstream consumer interprets the value as ms-unit and converts it to seconds; cls score is neither and we need to negate that division
+    // }, true) // so CLS node only reports on vis change rather than on every change
```
patrickhousley, line 53: Are we OK with leaving commented out code in the repo?

tools/browsers-lists/utils.mjs
```diff
@@ -16,8 +16,8 @@ export function getBrowserName ({ browserName, platformName }) {
 }
 
 export function getBrowserVersion (capabilities) {
-  if (capabilities.platformName?.toLowerCase() === 'ios') {
-    return capabilities['appium:platformVersion'] || capabilities.platformVersion
+  if (['ios', 'android'].includes(capabilities.platformName?.toLowerCase())) {
+    return capabilities['appium:platformVersion'] || capabilities.version
```
patrickhousley, line 20: Why was this changed?

cwli24: Matcher wasn't respecting versioning for Android. One of the currently failing soft nav test needs to be excluded from Android 9, but `'android>9'` not working atm because of this.

src/common/vitals/vital-metric.js
cwli24: Right below in this file as a getter, so it basically just go and grab the state defined here again

src/features/page_view_timing/aggregate/index.js
cwli24: I'm expecting the fix needed from consumer to be done in the near future, I think. Just saves me later from re-writing/rethinking out what needs to be written lol.

src/common/vitals/first-input-delay.js
```diff
@@ -1,19 +1,24 @@
...
+      type: attribution.eventType,
+      fid: Math.round(value),
+      eventTarget: attribution.eventTarget,
+      loadState: attribution.loadState
+    }
 
     // CWV will only report one (THE) first-input entry to us; fid isn't reported if there are no user interactions occurs before the *first* page hiding.
     firstInputDelay.update({
-      value: entries[0].startTime,
-      entries,
```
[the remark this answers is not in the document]
cwli24: you're talking about `entries` right? Yeah it's never used except by the current `onTTFB` for the load node (and technically if we moved onTTFB to attribution, it wouldn't need `entries` either as `attribution.navigationEntry` would hold it).

Entries is a bit useless atm and just extra references to perf entries.

src/common/vitals/vital-metric.js
patrickhousley: Ah ok 👍

src/common/vitals/first-input-delay.js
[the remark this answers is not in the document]
patrickhousley: I put this comment on several changes. I just want to make sure that we are either not using the `entries` property or accounted for the change. Also, just want to make sure that the final harvest payload is accurate and not missing something because of this change. Just precautionary here.

src/common/vitals/cumulative-layout-shift.js
cwli24: The check's purpose was to not report on this callback trigger if `value` remains the same, like: [0.1, 0.2, 0.2] <- don't capture 2nd `0.2`. 
This is likely related to `reportAllChanges` being true.
However, web vitals already does not report if the value remains unchanged so we don't need this.

src/common/vitals/cumulative-layout-shift.test.js
```diff
@@ -59,13 +70,11 @@ describe('cls', () => {
...
-        expect(value).toEqual(1)
-        expect(triggered).toEqual(1)
-        setTimeout(() => {
-          expect(triggered).toEqual(1)
-          done()
-        }, 1000)
+        expect(value).toEqual(0.123)
       })
+      metric.update({ value: 0.123 })
+      expect(triggered).toBeGreaterThanOrEqual(2)
```
metal-messiah, line 76: I'm still curious why this behavior and test are changing to allow CLS to report identical values more than once

tests/specs/spa/soft-navigations.e2e.js
```diff
@@ -146,6 +146,7 @@ describe('Soft navigations', () => {
   })
 
   // This reproduction condition only happens for chromium. I.e. safari & firefox load still fires before they let ajax finish.
+  // Also, Android 9.0- is not happy with 1mb-dom.html, so that ought to be excluded from this test.
```
metal-messiah, line 149: Can you just not use the 1mb test file and instead switch it for the 64kb one?  That seems to pass locally for me

src/common/vitals/first-input-delay.js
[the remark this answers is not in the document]
cwli24: Good to double check. So only `onTTFB` is subscribed in more than the one PVT aggregate code. TTFB was not switched over here aka still uses the non-attribution build, and the `entries` for it was left alone as well as in the VitalMetric class, so yeah nothing's missing or breaking with this.
(onTTFB could be switched and this `entries` removed altogether really, but unnecessary work.)

src/common/vitals/cumulative-layout-shift.test.js
cwli24: The test description says reports more than once, not necessarily "except if the value is the same", so I'm just updating it to be more true to description.

Also, `onCLS` inherently will not report the same value more than once, so that behavior is baked in already

cwli24: We can keep that redundant check though. I'm just following the philosophy of letting CWV handle calculations and scoring.

tests/specs/spa/soft-navigations.e2e.js
cwli24: I'm still getting an error for that test on android 9, just a different one that violates one of the expects which passes on the other browsers.

src/common/vitals/cumulative-layout-shift.test.js
metal-messiah: Then IMO wouldn't test that it **can** report the same value twice, since i think that implies that it has some utility for us (where in fact the test used to ensure that it can not report the same value twice since we dont want to harvest identical nodes twice)

cwli24: You suggesting to remove it? It does serve to ensure there's no check blocking it from reporting twice.

src/common/vitals/first-input-delay.js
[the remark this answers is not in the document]
metal-messiah: As long as we have a predictable way to access data regardless of the metric type, i'm good with it.  I don;t love that most dont use entries, but one does.  if it feels disjointed as is we may need to revisit standardizing the attributes flowing through these emitters at a later time

src/common/vitals/cumulative-layout-shift.test.js
metal-messiah: I cant see any case where we would **want** it to report the same value twice, so we need to pick a behavior here.  Either we enforce it like it was before or not, which is fine, but testing that it **can** do something we don't want it to do is strange to me here.

src/common/vitals/cumulative-layout-shift.js
```diff
@@ -1,13 +1,19 @@
...
+      metricId: id,
```
metal-messiah, line 11: Also curious what the `metricId` is/will be used for here

src/features/page_view_timing/aggregate/index.js
metal-messiah: Along these lines, should the [CLS](https://github.com/newrelic/newrelic-browser-agent/pull/919/files#diff-c3fae4e9a454f8ce96b46ab1b4c0ec1cf02d4f5a52418979046ed371e4965d96R9-R17) change in the emitter also be removed/commented out since there isnt anything consuming it?

src/common/vitals/cumulative-layout-shift.test.js
patrickhousley: I don't normally advocate for writing tests for a third-party but, if we are assuming that the CWV library will not double-report the same value, it makes sense to me that we test that assumption just in case the library is updated later and breaks that assumption.

src/features/page_view_timing/aggregate/index.js
cwli24: Harmless to leave in really; the attrs just won't be consumed at all. I could comment out the attrs though if you feel like we should--only the value is needed/used atm with it as an attr under other nodes. So?

src/common/vitals/cumulative-layout-shift.js
cwli24: Similar to metricId for INP in that if you want the final CLS for a view, you have to query/filter the new cls nodes sent by `latest(metricId)` to get that value. This is the most accurate by-the-books approach

src/common/vitals/cumulative-layout-shift.test.js
metal-messiah: (or just go back to enforcing it ourselves and not need to worry)

cwli24: The test itself is really for cls vital metrics class and the callback as onCLS is mocked, not to be conflated with the behavior of the original onCLS per se. Even though it's not expected to double report the same value in the field, unit testing that it by _our_ callback definition is allowed isn't really counterintuitive. That said, since it raised confusion, I removed it.

src/common/vitals/first-input-delay.js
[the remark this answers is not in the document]
cwli24: Ported ttfb over and removed `entries` entirely at this time as it's not meaningfully used anywhere.

metal-messiah (approved):

PR #921, opened 2024-03-19
chore: add jserrors timeline test

metal-messiah:
Add a test to check that a captured error in SR "Error" mode occurs with the correct timing.
---

### DIFF
--- a/tests/specs/session-replay/adjacent-payloads.e2e.js
+++ b/tests/specs/session-replay/adjacent-payloads.e2e.js
@@ -1,0 +1,29 @@
+import { config, decodeAttributes } from './helpers'
+
+describe('errors', () => {
+  beforeEach(async () => {
+    await browser.enableSessionReplay()
+  })
+
+  afterEach(async () => {
+    await browser.destroyAgentSession()
+  })
+
+  it('error timestamp should be contained within replay timestamp', async () => {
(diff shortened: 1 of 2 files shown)


tests/specs/session-replay/adjacent-payloads.e2e.js
```diff
@@ -0,0 +1,29 @@
...
+    await browser.url(await browser.testHandle.assetURL('rrweb-instrumented.html', config({
+      session_replay: { error_sampling_rate: 100, sampling_rate: 0 }
+    })))
+      .then(() => browser.waitForAgentLoad())
+
+    await browser.testHandle.expectBlob(10000, true)
+
+    const [errorPayload, blobPayload] = await Promise.all([
+      await browser.testHandle.expectErrors(15000),
+      await browser.testHandle.expectBlob(15000)
```
patrickhousley, line 22: It looks like this test is going to take at least 15 seconds to run. Is there a better way to test this that does not rely on such a long timeout? Or is this the only way to verify the timestamp?

patrickhousley (approved):

tests/specs/session-replay/adjacent-payloads.e2e.js
metal-messiah: there's probably a better way. I'll work out speeding this up before merging

PR #922, opened 2024-03-19
feat: maintain a harvestCount for use later with harvestId

metal-messiah:
Add a counter which increments after every harvest. This will be used in a future effort to curate a unique harvest ID for API paging purposes.
---

### Overview
This PR adds a harvestCount int to the runtime object, which increments every time `_send` is completed.  `_send` is used under the hood by all harvest methods to send data.  The plan for the future will be to concat this harvestCount with the `ptid` and `session` values to create a unique harvest identifier on blob-consumer payloads.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-238623
https://new-relic.atlassian.net/browse/NR-238628

### Testing
A new, simple jest test has been added to confirm this behavior.  WDIO test should be added later when implemented in outgoing harvests

### DIFF
--- a/src/common/harvest/harvest.js
+++ b/src/common/harvest/harvest.js
@@ -160,6 +160,9 @@ export class Harvest extends SharedContext {
       }, eventListenerOpts(false))
     }
 
+    const runtime = getRuntime(this.sharedContext.agentIdentifier)
+    runtime.harvestCount++
+
     return result
   }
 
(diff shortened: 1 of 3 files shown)


src/common/harvest/harvest.js
```diff
@@ -160,6 +160,9 @@ export class Harvest extends SharedContext {
       }, eventListenerOpts(false))
     }
 
+    const runtime = getRuntime(this.sharedContext.agentIdentifier)
+    runtime.harvestCount++
```
patrickhousley, line 164: I was going to ask about making this a class private since everything has to come through this module but this makes sense because we dont want this module having to worry about the url/data it is transmitting.

patrickhousley (approved):

PR #938, opened 2024-04-01
chore: Convert browser/spa/api.browser.js

cwli24:
---

### Overview

Total PITA.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-138435

### Testing

Way too much.


src/features/spa/aggregate/index.js
```diff
@@ -745,7 +745,7 @@ export class Aggregate extends AggregateBase {
       else smCategory = 'Custom'
       handle(SUPPORTABILITY_METRIC_CHANNEL, [`Spa/Interaction/${smCategory}/Duration/Ms`, Math.max((interaction.root?.end || 0) - (interaction.root?.start || 0), 0)], undefined, FEATURE_NAMES.metrics, baseEE)
 
-      scheduler.scheduleHarvest(0)
+      scheduler?.scheduleHarvest(0)
```
metal-messiah, line 748: If there is a case here where we are trying to trigger a harvest of an IXN and the scheduler doesnt exist, that seems like an issue.  Is this a new finding?

cwli24: I don't think it's a problem in practice, but the way these tests are setup is granular and captures more than the regular flow. Strictly speaking, you just have to emit 'interaction' to the baseEE with a passable object to trigger this function, regardless of the class state.

metal-messiah (approved):

PR #949, opened 2024-04-05
feat: Allow unmasking elements with nr-unmask selectors

metal-messiah:
Add custom masking mechanism to allow unmasking elements with `nr-unmask` class selectors and `data-nr-unmask` data attribute selectors
---

### Overview
This adds a custom handler to override the recorder's decision to mask if the element in question has a relevant `nr-unmask` CSS selector.  This only works in rrweb >=12 so a version bump is required to do this work. This only applies to elements that would have already been masked (ie in the maskText or maskInput selection list).  It will NOT apply unmasking to `password` type inputs.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-256256

### Testing
New test cases have been added to validate this behavior


package.json
```diff
@@ -168,7 +168,7 @@
   "dependencies": {
     "core-js": "^3.26.0",
     "fflate": "^0.7.4",
-    "rrweb": "2.0.0-alpha.11",
+    "rrweb": "^2.0.0-alpha.12",
```
patrickhousley, line 171: Having the carat in the version worries me since I think we need to be in sync with the UI, right? Maybe I am wrong.

tests/specs/session-replay/rrweb-configuration.e2e.js
```diff
@@ -58,7 +58,7 @@ describe.withBrowsersMatching(notIE)('RRWeb Configuration', () => {
         .then(() => browser.waitForAgentLoad())
 
       const [{ request: { body } }] = await Promise.all([
-        browser.testHandle.expectBlob(),
+        browser.testHandle.expectBlob(10000),
```
patrickhousley, line 61: Just curious, why are we adding a timeout on these expects when we are not expecting them to fail? The timeout was original added so we could expect a timeout and not have to wait 65 seconds. I guess this is just so we don't have to wait 65 seconds to find out the harvest didn't happen?

patrickhousley (approved):

patrickhousley (approved):
Approved pending tests passing and the comment on the dependency version.

tests/specs/session-replay/rrweb-configuration.e2e.js
metal-messiah: yep exactly, i had some failures earlier when i was building this and this was just added to fail quicker

package.json
metal-messiah: fair point, should probably be more explicit

PR #951, opened 2024-04-07
feat: Adjust Session Replay Error Tracking

metal-messiah:
Adjust when errors are decorated with session replay identifiers to more closely match actual recording data
---

#### Why:
Errors currently get decorated with `hasReplay`, which informs the UI to search for replays.  This decoration happens retroactively in the `agg` module, after the buffer has drained.  This means that errors that happen pre-load can be decorated with `hasReplay` even though there is no replay at the time of the error occurring.  This ALSO means that if there is a large gap between when the pre-load error occurred -- and when the page LOAD event occurred -- the error and replay events could be so displaced that the default search constraints in the UI can fail (+/- 10s from error occurrence).

#### What:
This PR makes a few deliberate changes to the way JSErrors and SR interact:
* Errors will be decorated with `hasReplay` live (now), instead of retroactively after the buffer has drained (prev).
    * This means that pre-load errors will only get `hasReplay` IF the recorder was preloaded and is recording.
    * post-load errors will only get `hasReplay` IF the recorder is recording.
* Errors will also only kick the recorder out of `ERROR` mode into `FULL` mode when the recorder is already running
    * This used to be retroactive, and kick the recorder into `FULL` regardless of if the recorder existed at the time of the error

### Related Issue(s)
NR-256713

https://newrelic.slack.com/archives/C04QZSS6XPV/p1712358079773939?thread_ts=1712353671.785849&cid=C04QZSS6XPV

### Testing
New test cases have been added to account for the change


src/features/jserrors/aggregate/index.js
```diff
@@ -78,9 +81,20 @@ export class Aggregate extends AggregateBase {
...
     }
 
-    if (body && body.err && body.err.length && !this.errorOnPage) {
-      payload.qs.pve = '1'
-      this.errorOnPage = true
+    if (body && body.err && body.err.length) {
+      if (this.replayAborted) {
+        body.err.forEach((e, i, arr) => {
+          try {
+            delete arr[i].params.hasReplay
```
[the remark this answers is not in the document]
patrickhousley: If we do `delete e.params?.hasReplay` is the try catch necessary?

tools/wdio/plugins/custom-commands.mjs
```diff
@@ -193,6 +193,33 @@ export default class CustomCommands {
         })
     })
 
+    /**
+     * Waits for the session replay feature to initialize and start recording.
+     */
+    browser.addCommand('waitForPreloadRecorder', async function () {
```
patrickhousley, line 199: How is this different from `waitForSessionReplayRecording`?

metal-messiah: replayRecording can happen if the recorder is triggered in either the inst or the agg, while preloadRecorder is specific to just the the recorder running in the inst.  Probably need to go through and consolidate some of these commands at some point

patrickhousley: Sorry, I should probably know this but is the feat aggregate not loaded early when we need to start recording before page load?

metal-messiah: nope, only the recorder file.  The agg still imports at load time

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/951/files#diff-46275fa8336f66585ff8b5c90b61f397c044b2930b5b29c947ccc4e5614eadfaR52-R56

patrickhousley: Ok thanks. This makes sense then.

patrickhousley (approved):

PR #964, opened 2024-04-10
feat: Generate PTID in Agent

metal-messiah:
Generate PTID in agent instead of in RUM response.  This allows PageView events to be decorated with a page trace ID.
---

### Overview
This abstracts the action of generating our own PTID out of the ST merge PR into a standalone PR

### Related Issue(s)
NR-166514
NR-167098

### Testing
Affected tests have been updated

### DIFF
--- a/src/features/session_trace/aggregate/index.js
+++ b/src/features/session_trace/aggregate/index.js
@@ -47,6 +47,7 @@ export class Aggregate extends AggregateBase {
     this.trace = {}
     this.nodeCount = 0
     this.sentTrace = null
+    this.everSent = false
     this.harvestTimeSeconds = getConfigurationValue(agentIdentifier, 'session_trace.harvestTimeSeconds') || 10
     this.maxNodesPerHarvest = getConfigurationValue(agentIdentifier, 'session_trace.maxNodesPerHarvest') || 1000
     /**
@@ -99,7 +100,7 @@ export class Aggregate extends AggregateBase {
 
         if (prevMode === MODE.ERROR && this.#scheduler) {
           this.trimSTNs(ERROR_MODE_SECONDS_WINDOW) // up until now, Trace would've been just buffering nodes up to max, which needs to be trimmed to last X seconds
-          this.#scheduler.runHarvest()
+          this.#scheduler.runHarvest({})
         } else {
           controlTraceOp(MODE.FULL)
         }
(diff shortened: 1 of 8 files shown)


patrickhousley (approved):
Approved but I wonder if there is more we can do around not needing to wait for the response on the first STN call in the agent. Maybe some code we could remove as well. Not necessary in this PR.

patrickhousley (approved):

src/features/session_trace/aggregate/index.js
```diff
@@ -189,13 +190,12 @@ export class Aggregate extends AggregateBase {
...
     this.#scheduler.harvest.on('resources', this.#prepareHarvest.bind(this))
-    if (dontStartHarvestYet === false) this.#scheduler.runHarvest({ needResponse: true }) // sends first stn harvest immediately
+    if (dontStartHarvestYet === false) this.#scheduler.runHarvest({}) // sends first stn harvest immediately
     startupBuffer.decide(true) // signal to ALLOW & process data in EE's buffer into internal nodes queued for next harvest
   }
 
   #onHarvestFinished (result) {
-    if (result.sent && result.responseText && !this.ptid) { // continue interval harvest only if ptid was returned by server on the first
-      this.agentRuntime.ptid = this.ptid = result.responseText
+    if (result.sent && !result.failed && !this.#scheduler.started) { // continue interval harvest only after first call
```
cwli24, line 198: Use `everSent` to keep sync with L220
```suggestion
    if (result.sent && !result.failed && !this.everSent) { // continue interval harvest only after first call
```

src/features/session_trace/aggregate/index.js
```diff
@@ -99,7 +100,7 @@ export class Aggregate extends AggregateBase {
 
         if (prevMode === MODE.ERROR && this.#scheduler) {
           this.trimSTNs(ERROR_MODE_SECONDS_WINDOW) // up until now, Trace would've been just buffering nodes up to max, which needs to be trimmed to last X seconds
-          this.#scheduler.runHarvest({ needResponse: true })
+          this.#scheduler.runHarvest({})
```
cwli24, line 103: Why having to pass empty object in? `runHarvest` needs fixing its opts param if undefined

src/features/session_trace/aggregate/index.js
```diff
@@ -195,7 +195,7 @@ export class Aggregate extends AggregateBase {
   }
 
   #onHarvestFinished (result) {
-    if (result.sent && !result.failed && !this.#scheduler.started) { // continue interval harvest only after first call
+    if (result.sent && !result.failed && !this.everSent) { // continue interval harvest only after first call
```
metal-messiah, line 198:
```suggest
if (result.sent && !result.failed && !this.#scheduler.started) { // continue interval harvest only after first call
```

metal-messiah: I need to undo this suggestion because it breaks.  everSent is set to true before sending, making this statement always falsy if we use it here.

src/features/session_trace/aggregate/index.js
metal-messiah: functionally identical, doesn't matter to me.

cwli24 (approved):
Gogogo before Ashley slams the gate

PR #966, opened 2024-04-11
fix: stabilize timestamp corrections

metal-messiah:
Ensure Session Replay timestamps are only ever adjusted once and fix correction factor to alleviate timeOrigin inconsistencies
---

### Overview
This PR ensures that SR timestamps are only ever adjusted once, and that the origin time they use to correct the issue does not fall victim to a known timeOrigin bug in Chromium and Firefox browsers. 
* https://issues.chromium.org/issues/40866530
* https://bugzilla.mozilla.org/show_bug.cgi?id=1710762


metal-messiah:
Smoke test in Chrome passed locally, will run full suite on release PR
Screenshot 2024-04-11 at 2 59 34 PM

cwli24 (approved):

PR #974, opened 2024-04-12
chore: Fix lint error prefer const

patrickhousley:
---

### Overview

Removing the disable on the prefer-const rule and fixing all the errors.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-87760


cwli24 (approved):
Oh boy

metal-messiah (changes requested):
We should probably wait to do this until after the big ST pr merges (and just rerun the linting then) since this will likely cause lots of chaos there

PR #980, opened 2024-04-16
feat: Maintain calculated NR server time for session

patrickhousley:
This update changes the agent to store the New Relic server time calculated when session tracking is enabled. This will prevent the agent from re-calculating server time on subsequent page loads ensuring the timeline view of events in New Relic is more accurate.
---

### Overview

Store the NR server time in session if session tracking is enabled.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-238959


src/common/config/state/info.js
```diff
@@ -44,7 +44,15 @@ export function getInfo (id) {
...
+  _cache[id] = getModeledObject({
+    ...(_cache[id] || {}),
+    ...obj
+  }, model)
+
   const agentInst = getNREUMInitializedAgent(id)
-  if (agentInst) agentInst.info = _cache[id]
+  if (agentInst && !agentInst.info) {
+    Object.defineProperty(agentInst, 'info', {
+      get: () => getInfo(id)
```
patrickhousley, line 55: These config files were changed like this because I noticed `newrelic.initializedAgents[].runtime.session`  and `newrelic.initializedAgents[].runtime.timeKeeper` where always undefined.

src/common/timing/time-keeper.js
```diff
@@ -30,18 +35,25 @@ export class TimeKeeper {
...
 
-  constructor () {
-    this.#originTime = Date.now() - performance.now()
+  constructor (agentIdentifier) {
+    this.#session = getRuntime(agentIdentifier)?.session
+
+    if (this.#session) {
+      const ee = baseEE.get(agentIdentifier)
+      ee.on(SESSION_EVENTS.UPDATE, this.#processSessionUpdate.bind(this))
+      ee.on(SESSION_EVENTS.STARTED, () => {
```
metal-messiah, line 44: Does this only matter for "reset" sessions?  It will inherently miss the first started event, but if that is true would we still be safer to check the arg `isNew`?

metal-messiah: maybe it doesnt matter since it already checks itself for #ready as well

src/common/config/state/runtime.js
```diff
@@ -1,31 +1,35 @@
...
-  denyList: undefined,
-  harvestCount: 0,
-  timeKeeper: undefined
+const model = () => {
+  const readonly = {
+    buildEnv: BUILD_ENV,
+    distMethod: DIST_METHOD,
+    version: VERSION,
+    originTime
+  }
```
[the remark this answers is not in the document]
patrickhousley: You're right. The intention is to make them read only but this part of the code has changed a couple of times so I probably need to re-review it.

src/common/timing/time-keeper.js
patrickhousley: It is better that it checks `#ready` and is only reliant on itself. It doesn't need to capture the first call on an absolutely brand new session since the next statement just reads the state. This is meant just to capture session timeouts. I was going to use the `RESET` event but the session manager completely replaces the session state after all the reset handlers are ran.

src/common/config/state/runtime.js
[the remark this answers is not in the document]
patrickhousley: fixed

metal-messiah (approved):

cwli24 (approved):

PR #982, opened 2024-04-17
feat: Session Replay preload optimizations

metal-messiah:
Optimize early load error detection to mitigate certain race conditions when transitioning from error mode to full mode.
---

### Overview
This PR introduces a number of changes to the early lifecycle behaviors associated with detecting errors and triggering a replay to switch into full mode and/or decorating an error with hasReplay:

* Adds a buffered layer of "preloaded errors" that can be accessed by the agg module, eliminating the race between noticing the error in the inst class and importing the agg while preloading
* Stops the stop/start cycle reset when `switchToFull` is called, which causes arbitrary excess processing and is unreliable.  Instead it simply starts recording if it detects it is not already recording, and does nothing but switch the mode to `full` if it is already recording.
* Adds an else case to the `switchToFull` method which will run `init` if the agent has not already done so at the time of an error being observed.  This is suspected as a fail point in a race condition between an error being noticed and the agg running
* Moves any error handlers to `after` the start of the recorder, ensuring that the recording data can exist before deciding what to do
* Moves the event that informs other features (errors) that a replay is "running" until after it has actually successfully `stored` an event
* Introduces an early harvest for Snapshots.  This is an effort to get the most critical part of the replay data harvested immediately, instead of arbitrarily waiting.
* 
### Related Issue(s)
NR-260186

### Testing
Many tests have been updated to account for the new behaviors

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -54,7 +54,6 @@ export class Aggregate extends AggregateBase {
     this.timeKeeper = undefined
 
     this.recorder = args?.recorder
-    this.preloaded = !!this.recorder
     this.errorNoticed = args?.errorNoticed || false
 
     handle(SUPPORTABILITY_METRIC_CHANNEL, ['Config/SessionReplay/Enabled'], undefined, FEATURE_NAMES.metrics, this.ee)
@@ -112,7 +111,7 @@ export class Aggregate extends AggregateBase {
       this.forceStop(this.mode !== MODE.ERROR)
     }, this.featureName, this.ee)
 
-    registerHandler('preload-errs', e => {
+    registerHandler(SR_EVENT_EMITTER_TYPES.ERROR_DURING_REPLAY, e => {
       this.handleError(e)
     }, this.featureName, this.ee)
 
@@ -237,11 +236,6 @@ export class Aggregate extends AggregateBase {
     await this.prepUtils()
 
     if (!this.recorder.recording) this.recorder.startRecording()
-    if (!this.preloaded) {
-      this.ee.on('err', e => {
-        this.handleError(e)
-      })
-    }
 
     this.syncWithSessionManager({ sessionReplayMode: this.mode })
   }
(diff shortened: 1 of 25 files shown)


src/features/session_replay/instrument/index.js
```diff
@@ -25,11 +26,12 @@ export class Instrument extends InstrumentBase {
...
+      this.#startRecording(session?.sessionReplayMode).then(() => {
+        this.ee.on('err', (e) => {
+          this.errorNoticed = true
+          handle('preload-errs', [e], undefined, this.featureName, this.ee)
+        })
       })
-      this.#startRecording(session?.sessionReplayMode)
     } else {
       this.importAggregator()
     }
```
patrickhousley, line 37:
```
this.ee.on('err', e => {
  // If preload and featAgg doesn't exist OR featAgg exists
  handle('preload-errs', [e], undefined, this.featureName, this.ee)
})
    
    
if (this.#canPreloadRecorder(session)) {
} else {
  this.importAggregator()
}
```

Just calling out that making the instrument listen to the errors and check if it needs to be processed before passing the error to the aggregate via event emitter is more inline with how the other features work and would remove the two different channels for errors depending on if preloaded or not. We could remove the listener for `err` in the aggregate.

metal-messiah: this could be simplified by utilizing the REPLAY_RECORDING event in the instrument class

src/features/session_replay/aggregate/index.js
```diff
@@ -301,7 +303,7 @@ export class Aggregate extends AggregateBase {
   }
 
   getCorrectedTimestamp (node) {
-    if (!node.timestamp) return
+    if (!node?.timestamp) return
```
patrickhousley, line 306: So if there is a node without a timestamp?? I thought that wasn't possible.

patrickhousley (approved):
Pending tests and comments

src/features/session_replay/aggregate/index.js
metal-messiah: The only time i saw this come up was in a jest test, but i just added it while I was thinking about it

PR #983, opened 2024-04-17
feat: Decorate errors with hasReplay individually

metal-messiah:
Change the error aggregation method to consider errors with a hasReplay flag separately to ensure timestamps are always correct. Decorate each aggregation set with hasReplay directly instead of a request-level query param
---

### Overview
This PR abstracts the aggregation of errors with hasReplay to their own set, which ensures that the `firstOccurrenceTimestamp` is not applied to duplicate errors that may have happened outside of a replay's import logic.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-260137

### Testing
New tests have been added which check that duplicated errors are buckets and decorated correctly.

### DIFF
--- a/src/features/jserrors/aggregate/index.js
+++ b/src/features/jserrors/aggregate/index.js
@@ -172,7 +172,7 @@ export class Aggregate extends AggregateBase {
      * the canonical stack trace excludes items like the column number increasing the hit-rate of different errors potentially
      * bucketing and ultimately resulting in the loss of data in NR1.
      */
-    var bucketHash = stringHashCode(`${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}_${params.hasReplay}`)
+    var bucketHash = stringHashCode(`${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}_${params.hasReplay ? 1 : 0}`)
 
     if (!this.stackReported[bucketHash]) {
       this.stackReported[bucketHash] = true
--- a/src/features/jserrors/aggregate/index.js
+++ b/src/features/jserrors/aggregate/index.js
@@ -172,13 +172,14 @@ export class Aggregate extends AggregateBase {
     // Do not modify the name ('errorGroup') of params without DEM approval!
     if (filterOutput?.group) params.errorGroup = filterOutput.group
 
+    if (hasReplay && !this.replayAborted) params.hasReplay = hasReplay
     /**
      * The bucketHash is different from the params.stackHash because the params.stackHash is based on the canonicalized
      * stack trace and is used downstream in NR1 to attempt to group the same errors across different browsers. However,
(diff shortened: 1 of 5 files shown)


cwli24:
Notes: A specific JS error is aggregated onto itself no matter how many times it's been thrown. As such, any properties or params are determined by the very first instance of such error (see `Aggregator` class' `getBucket` fn). If "Error abc123" happens 1st _without_ replay and 2nd time _with_ replay, it will presently get lumped together into the same hash marked as "without replay" respective to the 1st throw. Given that we do not have a set plan to individualize each occurrence of the same error at the moment _but_ replays are requiring JS error have accurate `hasReplay` association with first occurrence timestamps, the bucket/hash is being split into two groups: occurrences of "Error abc123" when replay is not present and occurrences of the same error when replay _is_ present.

src/features/jserrors/aggregate/index.js
```diff
@@ -172,13 +172,14 @@ export class Aggregate extends AggregateBase {
...
 
+    if (hasReplay && !this.replayAborted) params.hasReplay = hasReplay
     /**
      * The bucketHash is different from the params.stackHash because the params.stackHash is based on the canonicalized
      * stack trace and is used downstream in NR1 to attempt to group the same errors across different browsers. However,
      * the canonical stack trace excludes items like the column number increasing the hit-rate of different errors potentially
      * bucketing and ultimately resulting in the loss of data in NR1.
      */
-    var bucketHash = stringHashCode(`${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}`)
+    var bucketHash = stringHashCode(`${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}_${params.hasReplay}`)
```
cwli24, line 182: Append if defined is less work:
```suggestion
    let hashStr = `${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}`
    if (params.hasReplay) hashStr += '_' + Number(hasReplay)
    const bucketHash = stringHashCode(hashStr)
```

metal-messiah: change template to use Number()

tests/assets/rrweb-duplicate-errors-split.html
```diff
@@ -0,0 +1,62 @@
...
+        right: 50px;
+        top: 200px;
+      }
+    </style>
+    <link rel="stylesheet" type="text/css" href="style.css" /> 
+    {init} {config} {loader}
+    <script>
+      pre = 0, post = 0
+      inv = setInterval(() => {
+        console.log(pre, post)
```
metal-messiah, line 26: remove this

src/features/jserrors/aggregate/index.js
metal-messiah: switched to ternary since Number(undefined) is NaN and does not alter the original logic

patrickhousley (approved):

PR #985, opened 2024-04-18
fix: Session replay preload without autoStart

patrickhousley:
In scenarios where SessionReplay was configured to not auto start and the user already has a replay recording from a previous page or after a page reload, calls to the `start('session_replay')` API made too soon after agent initialization would be ignored.
---

### Overview

After investigating, we found that the issue would only happen in the session replay feature on page refresh or navigation if a session replay was already running. This appears to be the cause of how the session replay preload functionality uses async/await to import the recorder. This causes the call to importAggregator to be scheduled on the JS stack later than normal allowing enough time for the start API to be called before the listener for ${this.featureName}-opt-in is setup.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-260131

### Testing

Unit tests for the instrument base and session replay shared utils have been updated. A new wdio test has been added to the session replay preload spec that verifies the replay recording continues to harvest after a page refresh that would use preload logic.


src/features/utils/instrument-base.js
```diff
@@ -50,6 +50,15 @@ export class InstrumentBase extends FeatureBase {
     if (getConfigurationValue(this.agentIdentifier, `${this.featureName}.autoStart`) === false) this.auto = false
     /** if the feature requires opt-in (!auto-start), it will get registered once the api has been called */
     if (this.auto) registerDrain(agentIdentifier, featureName)
+    else {
+      this.ee.on(`${this.featureName}-opt-in`, () => {
+        // register the feature to drain only once the API has been called, it will drain when importAggregator finishes for all the features
+        // called by the api in that cycle
+        registerDrain(this.agentIdentifier, this.featureName)
+        this.auto = true
+        this.importAggregator()
```
metal-messiah, line 59: We probably need to add a check to `importAggregator()` that early returns if its already staged (but has not actually imported the agg), so it doesnt set up two window listeners

src/features/session_replay/shared/utils.js
```diff
@@ -1,11 +1,13 @@
...
-export const enableSessionTracking = (agentId) => isBrowserScope && getConfigurationValue(agentId, 'privacy.cookies_enabled') === true
+export function enableSessionTracking (agentId) {
+  return isBrowserScope && getConfigurationValue(agentId, 'privacy.cookies_enabled') === true
+}
 
 function hasReplayPrerequisite (agentId) {
-  return originals.MO && // Session Replay cannot work without Mutation Observer
-  enableSessionTracking && // requires session tracking to be running (hence "session" replay...)
+  return !!originals.MO && // Session Replay cannot work without Mutation Observer
+  enableSessionTracking(agentId) && // requires session tracking to be running (hence "session" replay...)
```
cwli24, line 10: gah, noice eye 👍

src/features/utils/instrument-base.js
cwli24: importAggregator still needs a 
`if (!this.auto) return // instrument file does not import aggregate if feat waits for start API`

cwli24: the `startedAlready` flag can be inside this cb listener in Jordan's case

metal-messiah (approved):
pending passing test suite

PR #989, opened 2024-04-18
chore: Test Suite Optimization

metal-messiah:
Increase WDIO concurrency and decrease JIL concurrency to offset the runtime differences and decrease total runtime
---


patrickhousley:
Nothing really to gain here since iOS takes so long anyways.

PR #993, opened 2024-04-21
fix: Prevent null CLS

metal-messiah:
Add check to prevent null and undefined web-vitals values from being reported 
---
### Overview
This PR prevents the vital-metric class from updating/emitting if provided a nullish value.  This was only observed in practice during some wdio test runs (blocking some PRs intermittently) -- And we don't exactly understand why web-vitals would even emit a nullish value -- But this is a simple fix that should prevent issues later if this is actually happening in the wild since we should not be reporting null values.

### Related Issue(s)
NR-260221

### Testing
Unit test has been updated to ensure nullish values are not reported to the subscribers

### DIFF
--- a/src/common/vitals/vital-metric.js
+++ b/src/common/vitals/vital-metric.js
@@ -9,7 +9,7 @@ export class VitalMetric {
   }
 
   update ({ value, attrs = {} }) {
-    if (value < 0) return
+    if (value === undefined || value === null || value < 0) return
     const state = {
       value: this.roundingMethod(value),
       name: this.name,
--- a/tests/unit/common/vitals/cumulative-layout-shift.test.js
+++ b/tests/unit/common/vitals/cumulative-layout-shift.test.js
@@ -10,9 +10,10 @@ const clsAttribution = {
   largestShiftValue: 0.9712,
   loadState: 'dom-content-loaded'
 }
+let mockReturnVal = 0.123
 const getFreshCLSImport = async (codeToRun) => {
   jest.doMock('web-vitals/attribution', () => ({
-    onCLS: jest.fn(cb => cb({ value: 0.123, attribution: clsAttribution, id: 'beepboop' }))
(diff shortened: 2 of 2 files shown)


tests/unit/common/vitals/cumulative-layout-shift.test.js
cwli24, line 1: Could you add another test for 0 value, for extra assurance?

tests/unit/common/vitals/cumulative-layout-shift.test.js
```diff
@@ -60,4 +61,24 @@ describe('cls', () => {
...
+  test('undefined value is not reported', done => {
+    mockReturnVal = undefined
+    getFreshCLSImport(metric => {
+      metric.subscribe(({ value, attrs }) => {
+        console.log('should not have reported...')
+        expect(1).toEqual(2)
+      })
+    })
+    setTimeout(done, 1000) // should not get subscribe invokation
+  })
```
cwli24, line 83:
```suggestion
  })
  test('zero value IS reported', done => {
   mockReturnVal = 0
    getFreshCLSImport(metric => {
      metric.subscribe(({ value, attrs }) => {
        expect(value).toEqual(0)
        done()
      })
    })
```

tests/unit/common/vitals/cumulative-layout-shift.test.js
cwli24: -see suggestion-

metal-messiah: 👌

src/common/vitals/vital-metric.js
```diff
@@ -9,7 +9,7 @@ export class VitalMetric {
   }
 
   update ({ value, attrs = {} }) {
-    if (value < 0) return
+    if (value === undefined || value === null || value < 0) return
```
patrickhousley, line 12: Could just check if `value` is type of `number`
```suggestion
    if (typeof value !== 'number' || value < 0) return
```

cwli24 (approved):

PR #997, opened 2024-04-23
feat: JSErrors timestamped per harvest

patrickhousley:
JSErrors that happen multiple times throughout a page's life will now receive a timestamp unique to the agent harvest cycle instead of always reporting the timestamp of the first occurrence of that error.
---

### Overview

Modified the JSErrors feature to cache the error timestamp per harvest cycle instead of per page lifecycle.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-261440

### DIFF
--- a/src/features/jserrors/aggregate/index.js
+++ b/src/features/jserrors/aggregate/index.js
@@ -34,6 +34,7 @@ export class Aggregate extends AggregateBase {
 
     this.stackReported = {}
     this.observedAt = {}
+    this.harvestObservedAt = {}
     this.pageviewReported = {}
     this.bufferedErrorsUnderSpa = {}
     this.currentBody = undefined
@@ -85,6 +86,7 @@ export class Aggregate extends AggregateBase {
         this.errorOnPage = true
       }
     }
+    this.harvestObservedAt = {}
     return payload
   }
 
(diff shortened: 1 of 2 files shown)


src/features/jserrors/aggregate/index.js
```diff
@@ -85,6 +86,7 @@ export class Aggregate extends AggregateBase {
         this.errorOnPage = true
       }
     }
+    this.harvestObservedAt = {}
```
metal-messiah, line 89: For brevity, If the harvest fails and retries, what is the behavior we should expect for subsequent (duplicate) errors?

patrickhousley: I assume that the new instances of the error will get a new timestamp for the new harvest interval, even though the harvest will also contain data from a previous harvest that failed to send.

patrickhousley: OK so I did some testing and decided to just let the aggregator handle the timestamp field. If there are no objections, then the timestamp reported will be that of the first instance of the error for that specific harvest interval.

cwli24 (approved):
Confirmed to work as intended manual, plus verified retry logic

PR #1000, opened 2024-04-23
fix: Clean up xhrWrappable

cwli24:
Remove `xhrWrappable` which is an old variable checking for `addEventListener` on XHR. It is expected to have no compatibility impact on modern standard frameworks or browsers.
---

### Overview

Delete presumably dead and ineffectively redundant code blocking trace, ajax, and spa if the addEventListener on XHR prototype somehow isn't a function. We've had no report of such case. This is a remnant of old IE.

### Related Issue(s)

Follow-up of https://new-relic.atlassian.net/browse/NR-57395

### Testing

Standard

### DIFF
--- a/tests/functional/spa/fetch.test.js
+++ b/tests/functional/spa/fetch.test.js
@@ -78,7 +78,7 @@ testDriver.test('response size', supported, function (t, browser, router) {
   var testCases = [
     {
       name: 'with request that returns content-length header',
-      responseBodySize: 10,
+      responseBodySize: 'abc123'.length,
       asset: 'spa/fetch-simple.html'
     },
     {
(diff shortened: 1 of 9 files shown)


tests/functional/spa/fetch.test.js
```diff
@@ -78,7 +78,7 @@ testDriver.test('response size', supported, function (t, browser, router) {
   var testCases = [
     {
       name: 'with request that returns content-length header',
-      responseBodySize: 10,
+      responseBodySize: 'abc123'.length,
```
metal-messiah, line 81: why did this change?

metal-messiah (approved):

PR #1003, opened 2024-04-25
feat: Restructure AJAX Aggregate

cwli24:
Re-organization of the Ajax feature as part of agent maintenance. This has no expected impact on its functionality.
---

### Overview

Note the ajax aggregate class undergoes a full layout change with Chunk put into its own class/file and constructor nested functions made into member functions, plus additional refactoring and clean up.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-138416

### Testing

New jest + regression


src/features/ajax/aggregate/index.js
```diff
@@ -15,251 +14,183 @@ import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
 import { AggregateBase } from '../../utils/aggregate-base'
 import { parseGQL } from './gql'
 import { getNREUMInitializedAgent } from '../../../common/window/nreum'
+import Chunk from './chunk'
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
+  #agentRuntime
+  #beacon
+  #proxyBeacon
```
metal-messiah, line 23: Curious why the need to do this.  This is already stored in the runtime obj, so abstacting these to another var (esp a private one), is just code duplication

src/features/ajax/aggregate/index.js
```diff
@@ -15,251 +14,183 @@ import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
...
+      if (!this.spaAjaxEvents[interaction.id]) return
 
       if (!wasSaved) { // if the ixn was saved, then its ajax reqs are part of the payload whereas if it was discarded, it should still be harvested in the ajax feature itself
-        spaAjaxEvents[interaction.id].forEach(function (item) {
-          ajaxEvents.push(item)
-        })
+        this.spaAjaxEvents[interaction.id].forEach((item) => classThis.ajaxEvents.push(item))
       }
-      delete spaAjaxEvents[interaction.id]
+      delete this.spaAjaxEvents[interaction.id]
```
metal-messiah, line 52: The mix of `classThis` and `this.` together in this single callback is a little strange.  Can we stick to one?

src/features/ajax/aggregate/index.js
```diff
@@ -15,251 +14,183 @@ import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
...
     })
     // --- ^
     // --- v Used by new soft nav
-    registerHandler('returnAjax', event => ajaxEvents.push(event), this.featureName, this.ee)
+    registerHandler('returnAjax', event => this.ajaxEvents.push(event), this.featureName, this.ee)
     // --- ^
+    registerHandler('xhr', function () { // the EE-drain system not only switches "this" but also passes a new EventContext with info (horrible practice, I know)
```
metal-messiah, line 58: Not sure the point of putting things like `horrible practice` into our open source code TBH

metal-messiah, line 58:
```suggestion
    registerHandler('xhr', function () { // the EE-drain system not only switches "this" but also passes a new EventContext with info. Should consider platform refactor to another system which passes a mutable context around separately and predictably to avoid problems like this.
```

src/features/ajax/aggregate/index.js
metal-messiah: private methods in general just add extra overhead (esp. when paired with babel) and make things harder to test in general anyway (since now you cant just reference the property directly in a test), so we should be considerate about when and where they get used, is my only ask

src/features/ajax/aggregate/index.js
```diff
@@ -15,251 +14,183 @@ import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
...
-            /* if it is too big BUT it isnt length 1, we can split it down again,
-             else we just want to NOT push it into payload
-             because if it's length 1 and still too big for the maxPayloadSize
-             it cant get any smaller and we dont want to recurse forever */
-            tooBig = true
-            break
-          }
-        } else {
-          payload.push(currentChunk.payload)
+  #getPayload (events, numberOfChunks) {
```
metal-messiah, line 166: Again since its on topic, what is the designation for when a method should be private vs public?

tests/components/ajax/aggregate.test.js
```diff
@@ -0,0 +1,245 @@
...
+      decodedEvents.forEach(payload => {
+        payload.forEach(event => {
+          validateCustomAttributeValues(expectedCustomAttributes, event.children)
+          delete event.children
+
+          expect(event).toEqual(expected) // event attributes serialized correctly
+        })
+      })
+
+      mockCurrentInfo.jsAttributes = {} // reset jsattributes for other tests
```
metal-messiah, line 181: can we put stuff like this in the afterEach block so its more clear what sideEffects we are managing through the tests

tests/components/ajax/aggregate.test.js
```diff
@@ -0,0 +1,245 @@
...
+        requestedWith: 'XMLHttpRequest',
+        requestBodySize: 128,
+        responseBodySize: 256,
+        nodeId: '0',
+        guid: null,
+        traceId: null,
+        timestamp: null
+      }
+
+      ajaxAggregate.ee.emit('xhr', ajaxArguments, context)
```
metal-messiah, line 156: The [original test had two ajax's](https://github.com/newrelic/newrelic-browser-agent/blob/c487e04f46a606545c8172b6ed1c022b4dc1cac1/tests/browser/xhr/ajax-events.browser.js#L176-L177). Theres not a test in here that two ajaxs that arent "too big" are harvested correctly.  Should consider if that is a test case that needs evaluation

src/features/ajax/aggregate/index.js
cwli24: I could change these to be info, init, and runtime, but those references (instead of value) will still be stored as variables. Is that better?

src/features/ajax/aggregate/index.js
cwli24: Forgot to update this after changing to arrow

src/features/ajax/aggregate/index.js
```diff
@@ -15,251 +14,183 @@ import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
...
-    ee.on('interactionDone', (interaction, wasSaved) => {
-      if (!spaAjaxEvents[interaction.id]) return
+    this.ee.on('interactionDone', (interaction, wasSaved) => {
+      if (!this.spaAjaxEvents[interaction.id]) return
 
       if (!wasSaved) { // if the ixn was saved, then its ajax reqs are part of the payload whereas if it was discarded, it should still be harvested in the ajax feature itself
-        spaAjaxEvents[interaction.id].forEach(function (item) {
-          ajaxEvents.push(item)
-        })
+        this.spaAjaxEvents[interaction.id].forEach((item) => classThis.ajaxEvents.push(item))
```
cwli24, line 50:
```suggestion
        this.spaAjaxEvents[interaction.id].forEach((item) => this.ajaxEvents.push(item))
```

src/features/ajax/aggregate/index.js
cwli24: `getPayload` should not be an accessible function outside the Aggregate instance (as there's no use case for it alone at the moment; it's called by `storeXhr`). If it weren't for tests, maybe neither should `prepareHarvest`. The reason why it's not a separate function is because it's inherently tied to this class and isn't re-usable by other classes. It could exist as a pure fn though if we pass `this` to it strictly speaking, as this is a OOP principle more than a technical restriction.

tests/components/ajax/aggregate.test.js
cwli24: I took that as a typo. The test right underneath it `correctly serializes a very large AjaxRequest events payload` queues up 10 ajax events.

src/features/ajax/aggregate/index.js
cwli24: Done

PR #1005, opened 2024-04-26
feat: Centralize Supportability Metric Collection

metal-messiah:
Centralize collection of supportability metrics to aid adjacent teams in auditing and understanding their purpose over time
---


metal-messiah:
Will likely close this investigation if build-size issues cannot be mitigated, as those outweigh the benefits of centralizing all the tags.  Will consider alternative solution

PR #1008, opened 2024-05-01
fix: Session trace nodes de-duplication

cwli24:
Remove duplicated error and event nodes in traces. Each error noticed by the JS errors feature will now only be 1 node instead of 3. Each occurrence of any event fired on for a DOM node or XMLHttpRequest object is pared down to 1 instead of being directly proportional to the number of subscribed listeners.
---

### Overview

Trace was emitting more than one node for certain events. It has been audited and tested to not duplicate any node type, and the offending events has been capped to once per occurrence. This should decrease total trace payload size in proportion to the number of errors thrown and number of listeners being added to an event like `click` and reduce the "noise".

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-231236
https://new-relic.atlassian.net/browse/NR-231243

### Testing

Added


src/features/session_trace/aggregate/index.js
```diff
@@ -272,6 +274,8 @@ export class Aggregate extends AggregateBase {
   // Tracks the events and their listener's duration on objects wrapped by wrap-events.
   storeEvent (currentEvent, target, start, end) {
     if (this.shouldIgnoreEvent(currentEvent, target)) return
+    if (this.prevStoredEvents.has(currentEvent)) return // prevent multiple listeners of an event from creating duplicate trace nodes per occurrence
+    this.prevStoredEvents.add(currentEvent)
```
metal-messiah, line 278: reminder to one of us to port this over to the ST migration PR since this could be missed in the merge conflict

src/features/session_trace/aggregate/index.js
```diff
@@ -272,6 +274,8 @@ export class Aggregate extends AggregateBase {
   // Tracks the events and their listener's duration on objects wrapped by wrap-events.
   storeEvent (currentEvent, target, start, end) {
     if (this.shouldIgnoreEvent(currentEvent, target)) return
+    if (this.prevStoredEvents.has(currentEvent)) return // prevent multiple listeners of an event from creating duplicate trace nodes per occurrence
```
metal-messiah, line 277:
```suggestion
    if (this.prevStoredEvents.has(currentEvent)) return // prevent multiple listeners of an event from creating duplicate trace nodes per occurrence. Cleared every harvest. No change for re-duplication after clearing per harvest since the timestamps of the event are considered for uniqueness.
```

metal-messiah, line 277: Will the ST feature's emitters keep emitting if the feature becomes `blocked`? If so, this condition should also probably check if the feature is blocked before storing in the map to prevent data bloat

tests/specs/stn/trace-nodes.e2e.js
metal-messiah, line 1: This test will also need to be ported over to the new testing directory

src/features/session_trace/aggregate/index.js
cwli24: Resolved in huddle

metal-messiah (approved):

PR #1009, opened 2024-05-01
fix: Remove API start()'s features param

cwli24:
The `start()` method in agent API no longer accepts an argument. It will forcibly start all features that were marked `autoStart: false` in the configuration that has not yet started when it is called. If `page_view_event` had this property set to false, the agent will not send anything until `start` is called.
---

### Overview

No. more. feature-specific start(). This is due to use case feedback and unneeded complexity with all the possible auto/manual start permutations.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-260970

### Testing

Edited


metal-messiah (approved):
nice

PR #1014, opened 2024-05-03
fix: Improve Error Casting

metal-messiah:
Improve detection of line and column to allow errors that throw without a stack -- such as some chrome errors -- to still get basic error location information
---

### Overview
The primary goal of this PR was to address an issue where certain browsers (Chrome) can emit errors without a stack property under certain circumstances.  These errors are emitted with a line and column, but the previous implementation of the agent blindly ignored that data if the error was an instance of the Error class. This PR:
* Checks if the error has a stack before accepting it blindly
* Centralizes the error casting behavior
* Removes the need for intercepting fn-err messages for SPA decoration by diverting that behavior to the SPA instrument class
    - This was necessary to keep in the agent because function errors can propagate through the agent faster than the event handler can, including errors thrown in the createTracer API, and if we rely `solely` on the errorEvent, these errors can actually throw `after` the SPA interaction has already closed. The agent will decorate the errors directly in the SPA feature now, so that by the time they are seen in the error feature, they already have interaction information tied to them.
* Removes the need to de-dupe in the error instrument class since fn-err and errorEvent no longer clash.

### Related Issue(s)
NR-262956

### Testing
New unit & wdio tests have been added


src/features/jserrors/aggregate/index.js
```diff
@@ -206,6 +206,11 @@ export class Aggregate extends AggregateBase {
     // still send EE events for other features such as above, but stop this one from aggregating internal data
     if (this.blocked) return
 
+    if (err.__newrelic) {
```
patrickhousley, line 209: Is it possible that this will be a breaking point for multiple agents running on the same page?

patrickhousley (approved):
Looks good. I just have the one question about how our decoration of the error object could clash with multiple agents running on the same page.

PR #1015, opened 2024-05-06
fix: Run inside cross-origin iframes for firefox/safari

patrickhousley:
Resolve a bug that prevented the agent from running in cross-origin iFrames for Firefox and Safari due to how the browsers implement context security.
---

### Overview

If the agent is running inside an iframe, it will skip subscribing to the `onTTFB` event. If the iframe is cross-origin, some browsers will block access to the `responseStart` property of the navigation performance entry. This was caused by a change in the [WebVitals](https://github.com/GoogleChrome/web-vitals) library to treat `0` as an [invalid timestamp value](https://github.com/GoogleChrome/web-vitals/blob/main/src/lib/isInvalidTimestamp.ts#L21-L24).

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-262293
https://github.com/GoogleChrome/web-vitals/issues/275

### DIFF
--- a/tests/specs/rum/index.e2e.js
+++ b/tests/specs/rum/index.e2e.js
@@ -1,0 +1,37 @@
+import { checkRum } from '../../util/basic-checks'
+import { testAssetRequest } from '../../../tools/testing-server/utils/expect-tests'
+
+describe('basic pve capturing', () => {
+  ['same-origin', 'cross-origin'].forEach(page => {
+    it(`should send rum when ${page} page loads in an iframe`, async () => {
+      const [
+        rumResults
+      ] = await Promise.all([
+        browser.testHandle.expectRum(),
+        browser.url(await browser.testHandle.assetURL(`iframe/${page}.html`))
+      ])
+
+      checkRum(rumResults.request)
+    })
+  })
+
+  it('should capture page load timings', async () => {
+    await browser.testHandle.scheduleReply('assetServer', {
+      test: testAssetRequest,
+      permanent: true,
+      delay: 500
+    })
+
+    const [
+      rumResults
+    ] = await Promise.all([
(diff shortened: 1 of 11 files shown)


patrickhousley:
Note for future us....ios safari does not like class public properties using an imported function such as 

```
import { single } from '../../../common/util/invoke'

// Importing this file throws an error "single is not defined" only in ios safari

class {
  sendRum = single(() => {})
}

```

tests/specs/rum/index.e2e.js
```diff
@@ -0,0 +1,37 @@
...
+      rumResults
+    ] = await Promise.all([
+      browser.testHandle.expectRum(),
+      browser.url(await browser.testHandle.assetURL('64kb-dom.html'))
+    ])
+
+    checkRum(rumResults.request)
+    expect(parseInt(rumResults.request.query.be, 10)).toBeGreaterThanOrEqual(500)
+    expect(parseInt(rumResults.request.query.fe, 10)).toBeGreaterThanOrEqual(100)
+    expect(parseInt(rumResults.request.query.dc, 10)).toBeGreaterThanOrEqual(100)
```
metal-messiah, line 35: Is 100 specific here or just an arbitrary number?  Would it be more consistent if it just checked that it existed and was greater than `0`?

metal-messiah (approved):

tests/specs/rum/index.e2e.js
patrickhousley: The html asset is delayed by 500ms so it should technically be around 500ms but some browsers are a little troublesome so I just made it 100 to make sure the page was actually being delayed (ie not around 30 40 ms for an instantaneous server response).

PR #1022, opened 2024-05-07
chore: Audit timestamp flooring

cwli24:
---

### Overview

Clean up some usage of double precision performance.now() and make UI event soft navigations work with ms granularity.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-253743


patrickhousley (changes requested):
Test failures need to be looked at.

tests/components/soft_navigations/index.test.js
```diff
@@ -137,6 +138,17 @@ describe('soft navigations', () => {
       expect(softNavAggregate.interactionInProgress.trigger).toEqual('keydown')
     })
 
+    test('double precision Event DOMHighResTimestamp (microsec) is floored to ms and isActiveDuring with now() works', () => {
+      const preciseCurrentTime = performance.now()
+      const currentTime = now() // this value is expected to be a truncated version of preciseCurrentTime, e.g. 421 ms vs 421.291 ms
```
metal-messiah, line 143: > // this value is expected to be a truncated version of preciseCurrentTime, e.g. 421 ms vs 421.291 ms
Wouldnt it be more informative to just test that `now() === Math.floor(performance.now())` directly in the test then?

cwli24: That's rather out of scope, and is asserting `.start === floor(.start)`; a now unit test should assert that.
What _really_ matters here is the check on L148, and L149.

tests/components/soft_navigations/index.test.js
```diff
@@ -137,6 +138,17 @@ describe('soft navigations', () => {
...
+      expect(Number.isInteger(preciseCurrentTime)).toBeFalsy()
+      expect(Number.isInteger(currentTime)).toBeTruthy()
```
metal-messiah, line 145:
```suggestion
```

patrickhousley (approved):

PR #1035, opened 2024-05-14
chore: Convert spa/hashchange* browser

cwli24:
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-138439
https://new-relic.atlassian.net/browse/NR-138500


tests/specs/spa/capture-interactions.e2e.js
```diff
@@ -61,4 +63,74 @@ describe('spa captures interaction when', () => {
...
+        expect.objectContaining({
+          type: 'customTracer',
+          name: 'onPopstate',
+          children: []
+        })
+      ]
+    }))
+  })
+
+  it('hashchange in 2nd event listener', async () => {
```
patrickhousley, line 106: I think this name could be a little more clear. I am assuming that the purpose of this test is that the interaction is counting the window addEventListener callback in the interaction created by the click???

tests/specs/spa/capture-interactions.e2e.js
```diff
@@ -1,5 +1,7 @@
+import { notIE } from '../../../tools/browser-matcher/common-matchers.mjs'
+
 describe('spa captures interaction when', () => {
-  it('hashchange fires after finish', async () => {
+  it('hashchange fires after XHR loads', async () => {
```
patrickhousley, line 4: I know you only changed the name here but this test is sort of unnecessary and somehow expansive in what its testing at the same time. I don't think we need the XHR part of this test and could split the remaining into two parts: testing the interaction api and simply testing that the spa feature captures an interaction when the hash changes (separate from anything else).

patrickhousley (commented):
The `capture-interactions.e2e.js` and `harvesting.e2e.js` should also probably be combined in the `spa` directory, maybe into a file named `index.e2e.js`. Having spec files with just a few tests is more expensive unless there is a clear distinction in what is being tested...like splitting out a spec file dedicated to the interaction api.

tests/specs/spa/capture-interactions.e2e.js
```diff
@@ -61,4 +63,74 @@ describe('spa captures interaction when', () => {
       ]
     }))
   })
+
+  it.withBrowsersMatching(notIE)('hashchange is followed by a popstate', async () => {
```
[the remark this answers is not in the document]
metal-messiah: Group decided to keep these tests as 1:1 for conversion purposes and revisit usefulness another time

tests/specs/spa/capture-interactions.e2e.js
metal-messiah: Group decided to keep these tests as 1:1 for conversion purposes and revisit usefulness another time

tests/specs/spa/capture-interactions.e2e.js
metal-messiah:
```suggestion
  it('Spa does not prevent the bubbling of events', async () => {
```

patrickhousley (approved):

PR #1039, opened 2024-05-16
fix: Undefined stns in blob Trace

cwli24:
Fix an instance of undefined error in Session Trace feature introduced in v1.259.0. Harvest will now no longer be attempted if there are no trace nodes to send.
---

### Overview

Fix a scenario wherein `stns` is undefined and causing problem when `.length` property is accessed for it. Instead, prepareHarvest will early exit when such is the case.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-270488

### Testing

Component test added

### DIFF
--- a/src/features/session_trace/aggregate/trace/storage.js
+++ b/src/features/session_trace/aggregate/trace/storage.js
@@ -274,8 +274,7 @@ export class TraceStorage {
   // Ajax (FEATURE) events--XML & fetches--pipes into ST here.
   storeXhrAgg (type, name, params, metrics) {
     if (type !== 'xhr') return
-    this.storeSTN(new TraceNode('Ajax', metrics.time, metrics.time + metrics.duration, `${params.status} ${params.method}: ${params.host}${params.pathname
-    }`))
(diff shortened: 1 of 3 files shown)


src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -274,8 +274,7 @@ export class TraceStorage {
   // Ajax (FEATURE) events--XML & fetches--pipes into ST here.
   storeXhrAgg (type, name, params, metrics) {
     if (type !== 'xhr') return
-    this.storeSTN(new TraceNode('Ajax', metrics.time, metrics.time + metrics.duration, `${params.status} ${params.method}: ${params.host}${params.pathname
-    }`))
+    this.storeSTN(new TraceNode('Ajax', metrics.time, metrics.time + metrics.duration, `${params.status} ${params.method}: ${params.host}${params.pathname}`, 'ajax'))
```
metal-messiah, line 277: nice catch

metal-messiah (approved):
Approved pending test suite run

PR #1042, opened 2024-05-16
chore: Move ins test

metal-messiah:
---

### Related Issue(s)
[NR-136548](https://new-relic.atlassian.net/browse/NR-136548)
[NR-136547](https://new-relic.atlassian.net/browse/NR-136547)

### Testing
Old tests deleted, new tests created


tests/specs/ins/harvesting.e2e.js
cwli24, line 1: - Missing page action on final / unload harvest case. Should add this to final-harvesting.e2e.js actually -- there's no expectIns() in the test cases within that file, but there should be.
- Missing attribute precedence translation.

metal-messiah: thanks for catching that.  Ported those over

patrickhousley (approved):
Do we already have tests around the max nodes harvested?

PR #1044, opened 2024-05-17
chore: move pvt tests to wdio

metal-messiah:
---


tests/specs/pvt/retry-harvesting.e2e.js
```diff
@@ -0,0 +1,33 @@
+import { testTimingEventsRequest } from '../../../tools/testing-server/utils/expect-tests'
+
+describe('pvt harvesting', () => {
+  it('timings are retried when collector returns 429', async () => {
```
patrickhousley, line 4: The retry harvesting for the other features is a bit more robust around different response codes. Is that needed here and/or would that be useful?

tools/browser-matcher/common-matchers.mjs
```diff
@@ -110,3 +110,38 @@ export const supportsBFCache = new SpecMatcher()
...
+  .include('chrome>=60')
+  .include('edge>=79')
+  .include('android>9.0')
+
+export const supportsFirstContentfulPaint = new SpecMatcher()
+  .include('chrome>=60')
+  .include('edge>=79')
+  .include('android>9.0')
+  .include('firefox>=84')
+  .include('safari>15') // this should be >= 14.1 but safari 15 on Sauce hates FCP, and it destroys the other tests on the same thread too...
```
patrickhousley, line 124: Has this been re-reviewed with the move to wdio? If it still an issue, maybe we should get a ticket in to re-review this after the move to LT.

patrickhousley (approved):

tools/browser-matcher/common-matchers.mjs
metal-messiah: Probably doesn't matter anymore since the oldest Safari we test is 16 at this point, but i'll throw a ticket on the backlog to re-look into this

PR #1059, opened 2024-06-04
feat: Create logging feature

metal-messiah:
Add basis of logging feature for handling and buffering observed log data
---

### Overview
Add various parts necessary to enable a new feature for capturing logs in the agent. Harvesting mechanisms will be added in a future PR.

### Related Issue(s)
NR-251152
NR-251160

### Testing
Tests have been updated to account for tweaks to event emitter patterns

### DIFF
--- a/src/features/logging/aggregate/index.js
+++ b/src/features/logging/aggregate/index.js
@@ -1,7 +1,7 @@
 import { getInfo, getRuntime } from '../../../common/config/config'
 import { registerHandler } from '../../../common/event-emitter/register-handler'
 import { AggregateBase } from '../../utils/aggregate-base'
-import { FEATURE_NAME, LOGGING_EVENT_EMITTER_TYPES } from '../constants'
+import { FEATURE_NAME, LOGGING_EVENT_EMITTER_CHANNEL } from '../constants'
 import { Log } from '../shared/log'
 
 export class Aggregate extends AggregateBase {
@@ -21,7 +21,7 @@ export class Aggregate extends AggregateBase {
 
     this.waitForFlags([]).then(() => {
       /** emitted by instrument class (wrapped loggers) or the api methods directly */
-      registerHandler(LOGGING_EVENT_EMITTER_TYPES.LOG, this.handleLog.bind(this), this.featureName, this.ee)
+      registerHandler(LOGGING_EVENT_EMITTER_CHANNEL, this.handleLog.bind(this), this.featureName, this.ee)
       this.drain()
     })
   }
@@ -42,12 +42,12 @@ export class Aggregate extends AggregateBase {
     return {
       common: {
         attributes: {
-          entityGuid: (this.#agentRuntime.appMetadata?.agents?.[0] || {}).entityGuid,
+          entityGuid: this.#agentRuntime.appMetadata?.agents?.[0]?.entityGuid,
           session: {
             id: this.#agentRuntime?.session?.state.value || '0', // The session ID that we generate and keep across page loads
             hasReplay: this.#agentRuntime?.session?.state.sessionReplayMode === 1, // True if a session replay recording is running
             hasTrace: this.#agentRuntime?.session?.state.sessionTraceMode === 1, // True if a session trace recording is running
-            pageTraceId: this.#agentRuntime.ptid // The trace ID if a session trace is recording
+            pageTraceId: this.#agentRuntime.ptid // The page's trace ID (equiv to agent id)
           },
           agent: {
             appId: this.#agentInfo.applicationID, // Application ID from info object
--- a/src/features/logging/aggregate/index.js
+++ b/src/features/logging/aggregate/index.js
@@ -1,4 +1,4 @@
-import { getRuntime } from '../../../common/config/config'
+import { getInfo, getRuntime } from '../../../common/config/config'
 import { registerHandler } from '../../../common/event-emitter/register-handler'
 import { AggregateBase } from '../../utils/aggregate-base'
 import { FEATURE_NAME, LOGGING_EVENT_EMITTER_TYPES } from '../constants'
@@ -6,11 +6,18 @@ import { Log } from '../shared/log'
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
+  #agentRuntime
+  #agentInfo
   constructor (agentIdentifier, aggregator) {
     super(agentIdentifier, aggregator, FEATURE_NAME)
 
-    this.logs = []
-    this.agentRuntime = getRuntime(this.agentIdentifier)
+    /** held logs before sending */
+    this.bufferedLogs = []
+    /** held logs during sending, for retries */
+    this.outgoingLogs = []
+
+    this.#agentRuntime = getRuntime(this.agentIdentifier)
+    this.#agentInfo = getInfo(this.agentIdentifier)
 
     this.waitForFlags([]).then(() => {
       /** emitted by instrument class (wrapped loggers) or the api methods directly */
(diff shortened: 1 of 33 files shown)


src/features/logging/constants.js
```diff
@@ -0,0 +1,15 @@
...
+  ERROR: 'error',
+  WARN: 'warn',
+  INFO: 'info',
+  DEBUG: 'debug',
+  TRACE: 'trace'
+}
+
+export const LOGGING_EVENT_EMITTER_TYPES = {
+  LOG: 'log'
+}
```
cwli24, line 13: This can just be a constant

src/features/logging/instrument/index.js
```diff
@@ -0,0 +1,19 @@
...
+  static featureName = FEATURE_NAME
+  constructor (agentIdentifier, aggregator, auto = true) {
+    super(agentIdentifier, aggregator, FEATURE_NAME, auto)
+
+    const instanceEE = this.ee
+    /** emitted by wrap-logger function */
+    this.ee.on('wrap-logger-end', function handleLog ([message, ...args]) {
+      const { level } = this
+      bufferLog(instanceEE, message, { ...(!!args.length && { customAttributes: stringify(args) }) }, level)
+    })
```
cwli24, line 16: So the wrappedFn executes, ends, passes information that it ended through the EE just for the logging instrument to put it back in the EE for later? This seems very roundabout to just having wrap-logger be
```
originalFn = function() {
  originalFn(...)
  bufferLog(...)
}
```

src/features/logging/instrument/index.js
```diff
@@ -0,0 +1,19 @@
...
+export class Instrument extends InstrumentBase {
+  static featureName = FEATURE_NAME
+  constructor (agentIdentifier, aggregator, auto = true) {
+    super(agentIdentifier, aggregator, FEATURE_NAME, auto)
+
+    const instanceEE = this.ee
+    /** emitted by wrap-logger function */
+    this.ee.on('wrap-logger-end', function handleLog ([message, ...args]) {
+      const { level } = this
+      bufferLog(instanceEE, message, { ...(!!args.length && { customAttributes: stringify(args) }) }, level)
```
cwli24, line 15:
```suggestion
      bufferLog(instanceEE, message, args.length ? { customAttributes: stringify(args) } : {}, level)
```
is more readable + I'm a confused why the customAttributes/Log().attributes which is an object of args have a key of `customAttributes` with value of additional args stringified. Is that supposed to be the case?

src/features/logging/shared/log.js
```diff
@@ -0,0 +1,22 @@
...
+   * @param {{[key: string]: *}} attributes - other attributes
+   * @param {enum} level - Log level
+   */
+  constructor (timestamp, message, attributes, level) {
+    /** @type {long} */
+    this.timestamp = timestamp
+    this.message = message
+    this.attributes = attributes
+    this.logType = level
+    this.session = {
```
cwli24, line 18: Can we improve this naming?

src/features/logging/instrument/index.js
metal-messiah: I realized this inconsistency in a subsequent PR and thus this is fixed [here](https://github.com/newrelic/newrelic-browser-agent/pull/1061/files#diff-e37844b7cdcf3225ff308f042059b5b5c9d92ecdf60b9cc8c76fe3d2789c7a5cR12-R15) to match basically what you wrote, it aligns more nicely for sure.  unfortunate pattern of this split PR technique :/

src/features/logging/shared/log.js
metal-messiah: sure, but can you start by divulging what exactly is unclear here?

cwli24: "session" should be a restricted keyword at this point because there are many things tied to sessions and we have actual objects/variables that are named closely so, like `session` from the session-entity which is also on runtime.session.
It's unclear what `this.session` here signifies or why it's needed instead of simply `this.url` (what is this.url supposed to be too?) -- not that either of them appear to be referenced at all in this PR

src/features/logging/aggregate/index.js
```diff
@@ -0,0 +1,63 @@
...
+      )
+    )
+  }
+
+  preparePayload () {
+    this.outgoingLogs.push(...this.bufferedLogs.splice(0))
+    return {
+      common: {
+        attributes: {
+          entityGuid: (this.#agentRuntime.appMetadata?.agents?.[0] || {}).entityGuid,
```
cwli24, line 45:
```suggestion
          entityGuid: this.#agentRuntime.appMetadata?.agents?.[0]?.entityGuid,
```

src/features/logging/aggregate/index.js
```diff
@@ -0,0 +1,63 @@
...
+          session: {
+            id: this.#agentRuntime?.session?.state.value || '0', // The session ID that we generate and keep across page loads
+            hasReplay: this.#agentRuntime?.session?.state.sessionReplayMode === 1, // True if a session replay recording is running
+            hasTrace: this.#agentRuntime?.session?.state.sessionTraceMode === 1, // True if a session trace recording is running
+            pageTraceId: this.#agentRuntime.ptid // The trace ID if a session trace is recording
```
cwli24, line 50: ptid always exists now, regardless of trace, no?

src/features/logging/aggregate/index.js
```diff
@@ -0,0 +1,63 @@
...
+    )
+  }
+
+  preparePayload () {
+    this.outgoingLogs.push(...this.bufferedLogs.splice(0))
+    return {
+      common: {
+        attributes: {
+          entityGuid: (this.#agentRuntime.appMetadata?.agents?.[0] || {}).entityGuid,
+          session: {
```
cwli24, line 46: More of a question, but are there actual solid commitments to incorporate logging with replay and/or traces by the UI?

src/loaders/api/api.js
```diff
@@ -54,35 +54,17 @@ export function setAPI (agentIdentifier, forceDrain, runSoftNavOverSpa = false)
...
-      wrappedLoggers.add(parent[functionName])
-      instanceEE.on(`${functionName}-wrap-logger-end`, ([message, ...args]) => {
-        log(message, {
-          ...(!!args.length && { 'wrappedFn.args': stringify(args) }),
-          ...customAttributes
-        }, level)
-      })
-    }
+  apiInterface.wrapLogger = (parent, functionName, level = LOG_LEVELS.INFO) => {
+    const failureMessage = 'Failed to wrap: '
```
cwli24, line 64: (better to define constants in caps at the top of file by convention)

src/loaders/api/api.js
```diff
@@ -54,35 +54,17 @@ export function setAPI (agentIdentifier, forceDrain, runSoftNavOverSpa = false)
...
-  function log (message, customAttributes, level = 'info') {
-    handle(SUPPORTABILITY_METRIC_CHANNEL, [`API/logging/${level}/called`], undefined, FEATURE_NAMES.metrics, instanceEE)
-    handle('log', [now(), message, customAttributes, level], undefined, FEATURE_NAMES.logging, instanceEE)
-  }
-
   logApiMethods.forEach((method) => {
     apiInterface[method] = function (message, customAttributes = {}) {
-      log(message, customAttributes, method.toLowerCase().replace('log', ''))
+      bufferLog(instanceEE, message, customAttributes, method.toLowerCase().replace('log', ''))
     }
```
cwli24, line 60: Instead of exposing every log___ as api, why not just expose one `.log` and have the level be an arg? Idr if we talked about this and settled on one way or another.

src/features/logging/shared/log.js
metal-messiah: Ah i see your confusion here. -- this object is essentially the individual log Event, the properties here are what will show up for the DB item that is saved, 1:1.  The attributes here are the agreed-upon attributes that are supposed to constitute the shape of the log item.  In other words, the attribute `session.id` will be a queryable property in NR1.

src/features/logging/aggregate/index.js
```diff
@@ -0,0 +1,63 @@
...
+        this.#agentRuntime.timeKeeper.convertRelativeTimestamp(timestamp),
+        message,
+        attributes,
+        level
+      )
+    )
+  }
+
+  preparePayload () {
+    this.outgoingLogs.push(...this.bufferedLogs.splice(0))
```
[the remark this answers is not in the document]
metal-messiah: I cant merge your suggestion since it will break the behavior, (since outgoing logs only get pushed if retry is true, they would not get pushed on unload), but i can rework this to align with that pattern if that helps solidify things

src/features/logging/aggregate/index.js
metal-messiah: true, comment is incorrect

src/features/logging/aggregate/index.js
metal-messiah: yes

src/loaders/api/api.js
metal-messiah: I understand, and kind of agree TBH -- BUT... this is a pattern i'd like to avoid that we have fallen victim to in the past:
-- plan
-- agree on plan
-- write docs, CDDs, briefs and get approval from product
-- THEN decide to change major decisions while writing the code

Lets talk more about this internally

src/features/logging/shared/log.js
metal-messiah: i'll add some comments to that file to make it more explicit

src/features/logging/aggregate/index.js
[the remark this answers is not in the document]
metal-messiah: actually, it looks like this will work 👍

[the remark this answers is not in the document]
metal-messiah: This will be handled better in the log harvesting PR

src/loaders/api/api.js
metal-messiah: EM support gathered, will change to single API with arg

src/features/logging/instrument/index.js
cwli24: Not really worried but it's very extraneous. Would it make more sense to have wrap-function capable of buffering it and just having a register handler in the aggregate for wrap-logger-end directly? For thought.

cwli24 (approved):

PR #1089, opened 2024-06-26
feat: logging bundle optimization

metal-messiah:
Improve the bundle size of the logging feature by moving non-critical pieces to the aggregation portion of the agent code.
---

### Overview
This PR shifts deferrable portions of the logging feature logic to the aggregator portion of the agent to improve loader build size.

### Testing
Tests have been shifted to match new locations, but logic should remain


patrickhousley (approved):
I still wonder if we could do something like having two constants files, one with all the codes, and one that maps the codes to strings that gets async loaded in the metrics agg or something. I know we can't do just one constants file because webpack won't split individual js files.

PR #1102, opened 2024-07-03
fix: Release backlog memory when features are blocked by RUM

cwli24:
Plug a memory leak in Session Trace and other features. When entitlement or enable flag from RUM request returns a 0 for a feature, the agent will now release the reference to the events buffer under that feature so it can be garbage collected.
---

### Overview

The pre-load cycle of the agent sets up listeners and buffers events in a backlog and an internal handlers object to be flushed and processed after the page loads. It was discovered that if the RUM call getting made post-load returns a `0` for an applicable feature, those things are never cleared for such feature. This caused the backlog for instance to indefinitely hold onto events awaiting processing, consuming memory. This fix will flush those events in such cases when the features are not activated via the `deregisterDrain` call made by each.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-284960
This issue was also seen in other features and have been addressed together.

### Testing

New e2e on `-b *`: 
image

### DIFF
--- a/src/common/drain/drain.js
+++ b/src/common/drain/drain.js
@@ -113,7 +113,7 @@ function drainGroup (agentIdentifier, group, activateGroup = true) {
 
   if (!baseEE.isolatedBacklog) delete handlers[group]
   baseEE.backlog[group] = null
-  baseEE.emit('drain-' + group, []) // exists purely for a unit test
+  baseEE.emit('drain-' + group, []) // TODO: Code exists purely for a unit test and needs to be refined
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/common/drain/drain.js
```diff
@@ -86,29 +87,33 @@ function checkCanDrainAll (agentIdentifier) {
...
+        })
       })
-    })
+    }
   }
+
   if (!baseEE.isolatedBacklog) delete handlers[group]
   baseEE.backlog[group] = null
-  baseEE.emit('drain-' + group, [])
+  baseEE.emit('drain-' + group, []) // exists purely for a unit test
```
patrickhousley, line 116: Can we make this comment a TODO so we can look at removing it. Don't like having code around that is purely for testing.

```suggestion
  baseEE.emit('drain-' + group, []) // TODO: Code exists purely for a unit test and needs to be refined
```

patrickhousley:
I did some manual testing locally. Here are my findings.

Looks like the backlog is getting cleared correctly when SR and ST are enabled in the agent config but RUM responds with a 0.
image

patrickhousley (approved):

Issue #1095, opened 2024-06-28
Cannot read properties of undefined (reading 'stopTimer')


Pranav2111:
I am using browser agent for my Next JS application.
I am getting this error **_(Cannot read properties of undefined (reading 'stopTimer'))_** reported on my sentry very frequently.

patrickhousley:
@Pranav2111 we need more information to go on. What version of the agent are you using? Is this the NPM package or how is the agent installed? What version of node, npm, nextjs, etc? Drop a copy of your package.json and code where you are instrumenting the agent if it is the npm package. Whats the stack trace for the error? Please provide as much detail as possible when reporting issues to open source projects. Thanks.

cwli24:
We had another report of this same issue, and I suspected the area of source of it, but thank you for giving us the trace to confirm! Will be spinning up a bug ticket for this.

lluchez:
We are encountering the same issue, which started to happen on June 27 in our React App.

jaohaohsuan:
We have been encountering the same issue since June 28  in Angular 17 app.

```
T.abortin https://js-agent.newrelic.com/nr-spa-1.261.1.min.js at line 2 : 65690

n.<anonymous>in https://js-agent.newrelic.com/nr-spa-1.261.1.min.js at line 2 : 61759
Object.emitin inline code at line 37 : 9898
anonymousin https://js-agent.newrelic.com/nr-spa-1.261.1.min.js at line 2 : 94242
```

zboris:
The same error appears for the nr-spa.1.261.2 started 27 of June as well

<img width="1362" alt="image" src="https://github.com/newrelic/newrelic-browser-agent/assets/19600767/f1a9d0c6-7e18-46c6-98de-49f2842aa88c">

zboris:
this error is going alongside with `undefined is not an object (evaluating 'this.scheduler.stopTimer')` error as well which happens in Chrome and Safari on Mac only.

zboris:
i guess line 208 in the unminified code should be `this.scheduler?.stopTimer()` ?

YesThatAllen:
We've been getting this error sporadically starting June 27 at 3:34 AM US/Central. (Didn't seem high priority given the low % of page loads it was happening on.)


```
Unhandled error
TypeError: Cannot read properties of undefined (reading 'stopTimer')
Location
https://js-agent.newrelic.com/nr-spa-1.261.1.min.js:2 - T.abort
```

[...]

PR #1104, opened 2024-07-08
fix: address call to stopTimer on undefined scheduler

patrickhousley:
Fixed an issue where session trace stopping on one tab while the same site was opening in another tab could result in the second tab throwing an exception when attempting to call `stopTimer()` on an undefined harvest scheduler.
---

### Overview

Calls to `stopTimer`  from the `SESSION_EVENTS.UPDATE` ee event could result in attempting to call `stopTimer` on an undefined harvest scheduler. Added an optional chaining to the call.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-286373
#1095

### DIFF
--- a/src/features/session_trace/aggregate/index.js
+++ b/src/features/session_trace/aggregate/index.js
@@ -205,6 +205,6 @@ export class Aggregate extends AggregateBase {
     this.blocked = true
     this.mode = MODE.OFF
     this.agentRuntime.session.write({ sessionTraceMode: this.mode })
-    this.scheduler.stopTimer()
+    this.scheduler?.stopTimer()
   }
 }


src/features/session_trace/aggregate/index.js
```diff
@@ -201,10 +201,10 @@ export class Aggregate extends AggregateBase {
   }
 
   /** Stop running for the remainder of the page lifecycle */
-  abort (reason) {
+  abort () {
```
metal-messiah, line 204: We should probably have a separate ticket to add "abort reason" SMs here.  They have been very useful for SR and I suspect would have some utility with ST as well

metal-messiah (approved):

src/features/session_trace/aggregate/index.js
patrickhousley: I can revert the removal of the reason codes if you want. They were actually very helpful in determining where in the code the call to abort was coming from.

metal-messiah: Yeah that's reasonable for now

metal-messiah (approved):

PR #1108, opened 2024-07-10
chore: update ajax e2e tests - network capture

patrickhousley:
---

### Overview

- Updated all the `specs/ajax/*` tests to use network capturing. This resolves race conditioning issues within our tests.
- Updated cors settings on testing server to prevent caching and support logging of `OPTIONS` calls.
- Update server logging to log debug shim messages without having to specify the `-L` flag.
- Testing server network capture
  - Removing the `clear` method because of it's flakiness due to race conditions.
  - Fixed an issue where tests where not properly failing when the capture `await` was not resolving.
- WDIO
  - Removed the `quiet` flag due to no usage.
  - Improved testing server logging so the testing server can produce logs separate from WDIO verbose mode. The `-L` flag no longer implies `-v`.
  - Added a new test logger that will provide more verbose indication of which browser is being ran per thread and will log out the test name when each test in a spec file starts.
  - Added better error handling to the testing server and network capture connectors.
  - Re-enabling `network` and `terminal` log capturing when `-D` flag is provided.
  - Added code to clear the LT tunnel log before starting tests and enable verbose tunnel logging.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-287577


metal-messiah (dismissed):

tools/wdio/runner.mjs
```diff
@@ -28,11 +29,25 @@ const configFilePath = path.join(
...
+  '.lambdatest'
+)
+fs.ensureDirSync(lambdaTestOutputDir)
+const lambdaTestTunnelLog = path.join(
+  lambdaTestOutputDir,
+  'lt.log'
+)
+if (fs.pathExistsSync(lambdaTestTunnelLog)) {
+  fs.removeSync(lambdaTestTunnelLog)
+}
```
cwli24, line 50: Shouldn't do this. Leave it manual. It may also cause issues when executing 2 runners simul

patrickhousley: You mean executing wdio with the `-T` flag in two different terminals?

patrickhousley: Honestly, I added this to make capturing tunnel logs easier so I could give the right log to LT team. IDC either way but I would just point out that the LT plugin also downloads the tunnel binary to the same directory so I wouldn't advise running wdio in two different tunnels with the `-T` flag.

cwli24 (approved):
👍

Issue #787, opened 2023-10-24
Long error/log messages in critical path bundle


villelahdenvuo:
## Summary

I noticed that the bundle generated by webpack when importing the browser agent contains multiple long error / log strings e.g. 

> "Call to agent api setUserId failed. The js errors feature is not currently initialized."

> "API/noticeError/called"

> "Failed to execute setCustomAttribute.\nName must be a string type, but a type of <".concat(typeof v,"> was provided."

## Desired Behavior

Critical path bundle size should be minimized.

## Possible Solution

Use error codes instead and documentation to explain them in depth? More lazy loading?

## Additional context

JavaScript go brrr. 😄 
![image](https://github.com/newrelic/newrelic-browser-agent/assets/633524/4b2503bb-8462-46cf-be05-ef4ebe1f4d8a)

patrickhousley:
Hi @villelahdenvuo . We reviewed our code bundles and talked through this request internally and have decided not to pursue it at this time. The strings that could be changed to codes accounts for less than 5kb of the minified bundle. When compressed using gzip or brotli, that size goes even further down to less than 1kb. The amount of effort to implement and maintain this change plus the need for customers to go to the docs to research the codes compared to anticipated savings is too high for us to proceed at this time.

PR #1111, opened 2024-07-14
feat: Optimize bundle size with warning codes

metal-messiah:
Decrease bundle size by converting static strings to warning codes with an accompanying link to a web resource.
---

### Overview
This was conducted as part of an innovation project.


cwli24 (approved):

patrickhousley (approved):

villelahdenvuo:
Fixes #787 :)

src/common/config/state/init.js
```diff
@@ -89,15 +89,15 @@ const model = () => {
...
+        else if (val !== '') warn(6, val)
       },
       // password: must always be present and true no matter what customer sets
       get mask_input_options () {
         return hiddenState.mask_input_options
       },
       set mask_input_options (val) {
         if (val && typeof val === 'object') hiddenState.mask_input_options = { ...val, password: true }
-        else warn('An invalid session_replay.mask_input_option was provided and will not be used', val)
+        else warn(7, val)
```
villelahdenvuo, line 100: I would recommend using constants that get replaced during the build step e.g. `WARN_INVALID_SESSION_REPLAY_MASK_INPUT_OPTION`

patrickhousley: Thank you for the suggestion @villelahdenvuo . Using constants does not reduce the bundle size so we decided to go with codes. A link will appear in the console back to the docs in the repo that can be used to lookup the error code.

PR #1116, opened 2024-07-18
feat: Shut down agent if improperly configured

metal-messiah:
Shut down initialization of the agent if improperly configured. Agents must be provided a valid licenseKey and applicationID at runtime.
---

### Overview
Shut down the agent with a new warning code if invalid licenseKey or App ID are provided.  Also fixed an issue where the warning code did not link correctly.

### Related Issue(s)
NR-257211

### Testing
Tests have been updated to reflect the new behavior

### DIFF
--- a/src/features/utils/aggregate-base.js
+++ b/src/features/utils/aggregate-base.js
@@ -4,7 +4,6 @@ import { configure } from '../../loaders/configure/configure'
 import { gosCDN } from '../../common/window/nreum'
 import { drain } from '../../common/drain/drain'
 import { activatedFeatures } from '../../common/util/feature-flags'
-import { warn } from '../../common/util/console'
 
 export class AggregateBase extends FeatureBase {
   constructor (...args) {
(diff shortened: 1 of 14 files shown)


src/features/utils/aggregate-base.js
```diff
@@ -49,7 +50,9 @@ export class AggregateBase extends FeatureBase {
   checkConfiguration () {
     // NOTE: This check has to happen at aggregator load time
     if (!isConfigured(this.agentIdentifier)) {
-      let jsAttributes = { ...gosCDN().info?.jsAttributes }
+      const cdn = gosCDN()
+      if (!cdn.info?.licenseKey || !cdn.info?.applicationID) return warn(43)
```
patrickhousley, line 54: We may need to abort the agent/eventemitter and flush the backlog so we are no longer capturing/retaining data.

cwli24: the feature aggs are going to continue initializing so shouldn't flush backlog

metal-messiah: I think it should flush and abort the EE.  The feature aggs will never drain because waitForFlags will never fire

patrickhousley (changes requested):
I am still seeing harvests locally when I comment out the license key in the test server.

image

metal-messiah:
> I am still seeing harvests locally when I comment out the license key in the test server.
> image

Thanks, im going to move this logic to just before the rum call. itll be simpler and more effective for harvest cancelling

PR #1121, opened 2024-07-23
feat: Create generic events feature

metal-messiah:
Create a feature that will normalize, aggregate and harvest all generic events.
---

### Related Issue(s)
NR-291171

### Testing
Feature testing to be added when real data begins flowing through the new feature

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,11 @@
       "require": "./dist/cjs/features/ajax/index.js",
       "default": "./dist/esm/features/ajax/index.js"
     },
+    "./features/generic_events": {
+      "types": "./dist/types/features/generic_events/index.d.ts",
+      "require": "./dist/cjs/features/generic_events/index.js",
+      "default": "./dist/esm/features/generic_events/index.js"
+    },
     "./features/jserrors": {
       "types": "./dist/types/features/jserrors/index.d.ts",
       "require": "./dist/cjs/features/jserrors/index.js",
@@ -119,10 +124,20 @@
       "require": "./dist/cjs/features/page_view_timing/index.js",
       "default": "./dist/esm/features/page_view_timing/index.js"
     },
+    "./features/session_replay": {
+      "types": "./dist/types/features/session_replay/index.d.ts",
+      "require": "./dist/cjs/features/session_replay/index.js",
+      "default": "./dist/esm/features/session_replay/index.js"
+    },
     "./features/session_trace": {
       "types": "./dist/types/features/session_trace/index.d.ts",
       "require": "./dist/cjs/features/session_trace/index.js",
       "default": "./dist/esm/features/session_trace/index.js"
+    },
+    "./features/soft_navigations": {
+      "types": "./dist/types/features/soft_navigations/index.d.ts",
+      "require": "./dist/cjs/features/soft_navigations/index.js",
+      "default": "./dist/esm/features/soft_navigations/index.js"
     },
     "./features/spa": {
       "types": "./dist/types/features/spa/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -44,6 +44,9 @@
       "features/page_action": [
         "dist/types/features/page_action/index.d.ts"
       ],
+      "features/generic_events": [
+        "dist/types/features/generic_events/index.d.ts"
+      ],
       "features/page_view_event": [
         "dist/types/features/page_view_event/index.d.ts"
       ],
@@ -52,6 +55,9 @@
       ],
       "features/session_trace": [
         "dist/types/features/session_trace/index.d.ts"
+      ],
+      "features/session_replay": [
+        "dist/types/features/session_replay/index.d.ts"
       ],
       "features/spa": [
         "dist/types/features/spa/index.d.ts"
--- a/src/cdn/experimental.js
+++ b/src/cdn/experimental.js
@@ -20,6 +20,7 @@ import { Instrument as InstrumentSessionReplay } from '../features/session_repla
 import { Instrument as InstrumentGenericEvents } from '../features/generic_events/instrument'
 // import { Instrument as InstrumentSpa } from '../features/spa/instrument'
 import { Instrument as InstrumentSoftNav } from '../features/soft_navigations/instrument'
+import { Instrument as InstrumentPageAction } from '../features/page_action/instrument'
 import { Instrument as InstrumentLogs } from '../features/logging/instrument'
 
 new Agent({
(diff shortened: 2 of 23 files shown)


src/cdn/experimental.js
```diff
@@ -17,6 +17,7 @@ import { Instrument as InstrumentErrors } from '../features/jserrors/instrument'
 import { Instrument as InstrumentXhr } from '../features/ajax/instrument'
 import { Instrument as InstrumentSessionTrace } from '../features/session_trace/instrument'
 import { Instrument as InstrumentSessionReplay } from '../features/session_replay/instrument'
+import { Instrument as InstrumentGenericEvents } from '../features/generic_events/instrument'
 // import { Instrument as InstrumentSpa } from '../features/spa/instrument'
 import { Instrument as InstrumentSoftNav } from '../features/soft_navigations/instrument'
 import { Instrument as InstrumentPageAction } from '../features/page_action/instrument'
```
patrickhousley, line 23: Doesn't this need to be removed?

patrickhousley (changes requested):
Missing package.json updates for this feature and for soft nav. Should also update the npm tests.

src/cdn/experimental.js
metal-messiah: If you are referring to PageActions being removed from the CDN entries, IMO that should be done with the re-write of the feature for better consistency of the changes

patrickhousley: I just noticed you removed it in the pro polyfills loader. Just wondering if maybe one of these is a mistake?

cwli24: @metal-messiah better would be to leave this out then and replace PA with GE in that PR

src/features/generic_events/aggregate/index.js
```diff
@@ -0,0 +1,108 @@
...
+    if (options.retry) {
+      this.currentEvents = harvestEvents
+    }
+
+    return payload
+  }
+
+  onHarvestFinished (result) {
+    if (result && result.sent && result.retry && this.currentEvents) {
+      this.events = this.events.concat(this.currentEvents)
```
cwli24, line 67: currentEvents should maintain first in order

src/features/generic_events/aggregate/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+    /** should have been provided by reporting feature -- but falls back to now if not */
+    eventAttributes.timestamp ??= this.#agentRuntime.timeKeeper.convertRelativeTimestamp(now())
+
+    this.events.push(eventAttributes)
+
+    // check if we've reached the harvest limit...
+    if (this.events.length >= this.eventsPerHarvest) {
+      this.overflow = [...this.overflow, ...this.events.splice(0, Infinity)]
+      this.harvestScheduler.runHarvest()
```
cwli24, line 105: need to align on all the individual max buffer harvest logic. Low interval harvesting is becoming pointless if everything can just call harvest when full

metal-messiah: mainly there just protect unload behavior

src/features/generic_events/aggregate/index.js
metal-messiah: why does this matter?  Every event is timestamped

cwli24: just the pattern with all the other feature like trace; it's also implied when we troubleshoot and open payloads in dev tools to find the earliest at the top

src/cdn/experimental.js
metal-messiah: good catch

package.json
```diff
@@ -53,6 +56,9 @@
       "features/session_trace": [
         "dist/types/features/session_trace/index.d.ts"
       ],
+      "features/session_replay": [
+        "dist/types/features/session_replay/index.d.ts"
+      ],
```
patrickhousley, line 61: Still missing the typings entry for soft nav but thats not really part of this PR so I understand if you want to do that separately.

src/features/generic_events/aggregate/index.js
```diff
@@ -0,0 +1,114 @@
...
+      if (!ins) {
+        this.blocked = true
+        deregisterDrain(this.agentIdentifier, this.featureName)
+        return
+      }
+
+      // handle page actions and other generic events here
+      this.harvestScheduler = new HarvestScheduler('ins', { onFinished: (...args) => this.onHarvestFinished(...args) }, this)
+      this.harvestScheduler.harvest.on('ins', (...args) => this.onHarvestStarted(...args))
+      // this.harvestScheduler.startTimer(this.harvestTimeSeconds, 0)
```
patrickhousley, line 45: I assume this line will be uncommented once we add a sub-feature.

src/features/generic_events/aggregate/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+    for (let key in obj) {
+      let val = obj[key]
+      if (key === 'timestamp') val = this.#agentRuntime.timeKeeper.correctAbsoluteTimestamp(val)
+      obj[key] = (val && typeof val === 'object' ? stringify(val) : val)
+    }
+
+    const eventAttributes = {
+      /** Agent-level custom attributes */
+      ...(getInfo(this.agentIdentifier).jsAttributes || {}),
```
patrickhousley, line 94: Should jsAttributes be last in the list so that it gets truncated if the size is too large?

src/features/generic_events/aggregate/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+    /** should have been provided by reporting feature -- but falls back to now if not */
+    eventAttributes.timestamp ??= this.#agentRuntime.timeKeeper.convertRelativeTimestamp(now())
+
+    this.events.push(eventAttributes)
+
+    // check if we've reached the harvest limit...
+    if (this.events.length >= this.eventsPerHarvest) {
+      this.overflow = [...this.overflow, ...this.events.splice(0, Infinity)]
+      this.harvestScheduler.runHarvest()
```
patrickhousley, line 111: We should put in a ticket to centralize the harvest buffer mechanics so that we are not re-inventing the wheel in every feature around max harvest size, etc.

patrickhousley (approved):
All the comments I left were informational and don't necessarily require a change here. Approving.

src/features/generic_events/aggregate/index.js
metal-messiah: yes and also so that it matches the same attribute precedence

src/features/generic_events/aggregate/index.js
metal-messiah: yeah, needs a middle man to buffer instead of feature by feature so we can check, but agreed

PR #1122, opened 2024-07-24
fix: Prevent agent using invalid date header

patrickhousley:
Shut down the agent when unable to read a correctly formatted date header for the page view event harvest. This prevents harvesting data with wildly inaccurate timestamps. This should only happen with third-party or non-agent code has tampered with the ajax response headers.
---

### Overview

Adding supportability metrics to track when the calculated time diff is >= 12 hours, >= 6 hours, or >= 1 hour.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-295258
https://source.datanerd.us/agents/angler/pull/600

### DIFF
--- a/src/common/timing/time-keeper.js
+++ b/src/common/timing/time-keeper.js
@@ -1,6 +1,8 @@
 import { originTime } from '../constants/runtime'
 import { getRuntime } from '../config/config'
 
+const rfc2616Regex = /^(Mon|Tue|Wed|Thu|Fri|Sat|Sun), ([0-3][0-9]) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) ([0-9]{4}) ([01][0-9]|2[0-3])(:[0-5][0-9]){2} GMT$/
+
 /**
  * Class used to adjust the timestamp of harvested data to New Relic server time. This
  * is done by tracking the performance timings of the RUM call and applying a calculation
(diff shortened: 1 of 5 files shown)


src/features/page_view_event/aggregate/index.js
```diff
@@ -130,6 +130,15 @@ export class Aggregate extends AggregateBase {
           if (!this.timeKeeper.ready) throw new Error('TimeKeeper not ready')
 
           agentRuntime.timeKeeper = this.timeKeeper
+
+          // Check if the time diff is such that we need to capture a supportability metric
+          if (this.timeKeeper.localTimeDiff >= 12 * 60 * 60 * 1000) {
+            handle(SUPPORTABILITY_METRIC_CHANNEL, ['PVE/NRTime/Calculation/DiffExceed12'], undefined, FEATURE_NAMES.metrics, this.ee)
```
cwli24, line 136:
```suggestion
            handle(SUPPORTABILITY_METRIC_CHANNEL, ['PVE/NRTime/Calculation/DiffExceed12Hrs'], undefined, FEATURE_NAMES.metrics, this.ee)
```

src/common/timing/time-keeper.js
```diff
@@ -96,6 +101,8 @@ export class TimeKeeper {
 
   /** Process the session entity and use the info to set the main time calculations if present */
   processStoredDiff () {
+    if (this.#ready) return // Time diff has already been calculated
```
cwli24, line 104: This should make the PR a fix, and would require test for behavior

metal-messiah (approved):

PR #1140, opened 2024-08-07
fix: Disable session tracking when missing browser API

patrickhousley:
Disable session tracking when the browser is missing the PerformanceNavigationTiming api. This only happens with unsupported browsers and results in session trace data with the timing values in the wrong format.
---

### Overview

If `PerformanceNavigationTiming` is not defined, in the case of old unsupported browsers, session tracking will be disabled.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-297306


src/features/utils/feature-gates.js
cwli24, line 1: This affects session manager; it should only prevent traces & replay as discussed.

patrickhousley:
Closing in favor of #1142

PR #1142, opened 2024-08-08
fix: relative timestamps in session trace

patrickhousley:
In rare cases where PerformanceNavigationTiming API is undefined, session trace timing nodes would be reported using epoch timestamp values instead of relative timestamp values. This would result in a broken session trace experience in New Relic. If the PerformanceNavigationTiming API is unavailable, the epoch timestamps will now be converted to relative timestamps accounting for user clock difference with New Relic servers.
---

### Overview

Added code to convert epoch timestamps from `performance.timing` to relative timestamps correcting for NR server time.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-297306


src/common/timing/time-keeper.js
```diff
@@ -95,6 +95,10 @@ export class TimeKeeper {
     return Math.floor(this.#correctedOriginTime + relativeTime)
   }
 
+  convertAbsoluteTimestamp (timestamp) {
+    return Math.floor(this.correctAbsoluteTimestamp(timestamp) - this.#correctedOriginTime)
+  }
```
metal-messiah, line 100: I think we should probably consider keeping this method more pure.  I would keep the correction part as separate call and have whomever needs both correction `and` a conversion to call them both separately.  The reason is that if you accidentally pass a corrected unix stamp into this, it will break it before transforming it.

src/common/timing/time-keeper.js
```diff
@@ -95,6 +95,10 @@ export class TimeKeeper {
     return Math.floor(this.#correctedOriginTime + relativeTime)
   }
 
+  convertAbsoluteTimestamp (timestamp) {
```
metal-messiah, line 98:
```suggestion
/**
   * Converts a unix timestamp to a page origin relative time
   * @param timestamp {number} The unix/epoch timestamp
   * @returns {number} A relative timestamp in milliseconds from origin time
   */
  convertAbsoluteTimestamp (timestamp) {
```

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -164,6 +164,9 @@ export class TraceStorage {
       if (!(typeof val === 'number' && val >= 0)) continue
 
       val = Math.round(val)
+      if (this.parent.timeKeeper && this.parent.timeKeeper.ready && val >= this.parent.timeKeeper.correctedOriginTime) {
```
metal-messiah, line 167:
```suggestion
      // fallbacks from the performance.timing API supply values with unix timestamps that must be converted to relative ms from origin
      if (this.parent.timeKeeper && this.parent.timeKeeper.ready && val >= this.parent.timeKeeper.correctedOriginTime) {
```

tests/assets/stn/disable-performance-navigation-timing.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <script>
+    OriginalPerformanceNavigationTiming = PerformanceNavigationTiming
+    PerformanceNavigationTiming = undefined
```
metal-messiah, line 6: 👍

src/features/session_trace/aggregate/trace/storage.js
cwli24: I'm not certain `val >= this.parent.timeKeeper.correctedOriginTime` is always applicable or that every timing "val" is strictly always after the "correctedOriginTime".
I suggest just passing an arg to `storeTiming` that branches in this behavior.

src/common/timing/time-keeper.js
patrickhousley: I am a little torn on this comment. I get where you are coming from but `convertRelativeTimestamp` already converts a relative timestamp to a corrected epoch timestamp. Don't you think it would be a little weird to have the opposite conversion be different? Maybe both the `convert` methods should convert based on the original origin time instead of the corrected origin time?

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -164,8 +164,12 @@ export class TraceStorage {
       if (!(typeof val === 'number' && val >= 0)) continue
 
       val = Math.round(val)
-      if (this.parent.timeKeeper && this.parent.timeKeeper.ready && val >= this.parent.timeKeeper.correctedOriginTime) {
-        val = this.parent.timeKeeper.convertAbsoluteTimestamp(val)
+      if (this.parent.timeKeeper && this.parent.timeKeeper.ready && isAbsoluteTimestamp) {
+        val = this.parent.timeKeeper.convertAbsoluteTimestamp(
+          Math.floor(this.parent.timeKeeper.correctAbsoluteTimestamp(
+            this.parent.timeKeeper.correctAbsoluteTimestamp(val)
```
metal-messiah, line 170: is this double invokation of `correctAbsoluteTimestamp` intended here

patrickhousley: oops

PR #1149, opened 2024-08-13
security: apply obfuscation rules to session trace

patrickhousley:
A bug was introduced in version 1.259.0 where session trace data was not being processed through configured obfuscation rules. This fix ensures session trace data is properly obfuscated.

Potential areas where un-obfuscated data could be added to the session trace:

- When calling newrelic.addToTrace, the name of the trace event is a string provided by the customer.
- When capturing session trace nodes for ajax requests, the node will contain the current page URL excluding query parameters and hash fragments.
- When capturing resources being loaded on the page (javascript and css files), the session trace node will contain the URL of the resource excluding query parameters and hash fragments.
- When a user clicks a link on the page containing an href that includes a hash fragment and does not result in a hard page load, the href value of the link will be captured in a session trace node.
- When the browser agent captures an error on the page, the message of the error is included in a session trace node.
---

### Overview

Updating the agent so that each feature applies obfuscation rules to it's harvested data. This restores obfuscation on session traces where it was removed when the harvest was moved to the blob consumer. 

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-302660
https://new-relic.atlassian.net/browse/NR-302936


metal-messiah (dismissed):
we need to add a ticket to ensure all the payloads are run through this test.  PageActions and Logs are currently missing from this as well

patrickhousley:
> we need to add a ticket to ensure all the payloads are run through this test. PageActions and Logs are currently missing from this as well

Done https://new-relic.atlassian.net/browse/NR-302936

metal-messiah (approved):
changes make sense to me! when tests pass I think its g2g

PR #1152, opened 2024-08-16
feat: add payload size evaluation to generic events feature

metal-messiah:
Harvest early when the payload reaches 64kb and prevent harvests from sending if they exceed the 1MB ingest limit.
---

### Related Issue(s)
This was a missed condition of the feature https://new-relic.atlassian.net/browse/NR-290741

### Testing
New tests have been added to validate this behavior


src/features/generic_events/aggregate/event-buffer.js
```diff
@@ -0,0 +1,59 @@
...
+    this.buffer = prepend ? [...eventBuffer.buffer, ...this.buffer] : [...this.buffer, ...eventBuffer.buffer]
+    this.bytes = this.bytes + eventBuffer.bytes
+    return this
+  }
+
+  /**
+   * Returns a boolean indicating whether the size and buffer contain valid data
+   * @returns {boolean}
+   */
+  isValid () {
```
patrickhousley, line 47: This function name sort of implies that we are performing some sort of validation. Maybe `hasData` would be better?

src/features/generic_events/aggregate/event-buffer.js
```diff
@@ -0,0 +1,59 @@
...
+import { MAX_PAYLOAD_SIZE } from '../constants'
+
+/**
+ * A container that keeps an event buffer and size with helper methods
+ * @typedef {Object} EventBuffer
+ * @property {number} size
+ * @property {*[]} buffer
+ */
+
+export class EventBuffer {
```
patrickhousley, line 11: Could extend `Array` class.

src/features/generic_events/aggregate/event-buffer.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Adds an event object to the buffer while tallying size
+   * @param {Object} event the event object to add to the buffer
+   * @returns {EventBuffer} returns the event buffer for chaining
+   */
+  add (event) {
+    const size = stringify(event).length
+    if (!this.canMerge(size)) return
```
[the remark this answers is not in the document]
metal-messiah: ty, missed that one

src/features/generic_events/aggregate/event-buffer.js
metal-messiah: I think i have some other changes in mind with this as an innovation/POC so i'm gonna hold off on that for a moment

src/features/generic_events/aggregate/event-buffer.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Merges an EventBuffer into this EventBuffer
+   * @param {EventBuffer} events an EventBuffer intended to merge with this EventBuffer
+   * @param {boolean} prepend if true, the supplied events will be prepended before the events of this class
+   * @returns {EventBuffer} returns the event buffer for chaining
+   */
+  merge (eventBuffer, prepend = false) {
+    if (!this.canMerge(eventBuffer.bytes)) return
```
patrickhousley, line 45: Should this return `this` too?

```suggestion
    if (!this.canMerge(eventBuffer.bytes)) return this
```

src/features/generic_events/aggregate/event-buffer.js
patrickhousley: Sounds good. I also wonder if this is a precursor to the harvestor work we have been talking about. For example, this could be a generic event buffer that other buffers extend off of when they need additional logic. Just food for thought.

patrickhousley (dismissed):
Approved pending one more comment.

patrickhousley (approved):

PR #1162, opened 2024-08-27
chore: IFrame SMs

metal-messiah:
Add SMs detecting whether the agent is running in an iframe or if iframes are present on the tracked context
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-300493


patrickhousley (approved):
You could add some wdio tests around the new iframe metrics. Otherwise, looks good.

PR #1166, opened 2024-08-28
fix: fix duplicative log payloads on unload

metal-messiah:
Ensure agent clears buffer on page navigation unload harvests to prevent duplicative payloads from being sent.
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-307620


src/features/utils/event-buffer.js
patrickhousley, line 1: Should write a unit test for this file.

patrickhousley (approved):

PR #1171, opened 2024-09-01
feat: Centralize Event Buffers

metal-messiah:
Standardize and centralize the event buffer pattern across all features, to allow for consistent storage, harvesting, and cache management.
---

### Overview
All features would now have a `standard` way to store event data.  And external (future) modules would have a standard way to access that data, through the runtime event manager object.  All data stores are size-managed by the existing event-buffer class, which makes managing the data constraints and short-circuits easier. 

### Related Issue(s)
This was generated from innovation work

### Testing
Tests have had a major overhaul to support the change to the event buffer pattern

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,37 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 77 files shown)


metal-messiah:
putting back into draft mode until CDD is agreed upon

PR #1181, opened 2024-09-11
feat: Update TimeKeeper Source of Truth

metal-messiah:
Update the Agent's TimeKeeper to evaluate a timestamp provided directly by the consumer instead of a generic Date header, which should provide more control stability and granularity to the time synchronization processes.
---

### Related Issue(s)
NR-311277

### Testing
All existing tests should continue making sense for this and continue passing. This only changes the source of the time stamp, not the logic to manage timestamp corrections.

*Note: because the rumResponse now has a blocking value in the body, all tests that `mocked` the rumResponse had to be updated to include the new body fields.*

### DIFF
--- a/src/common/timing/time-keeper.js
+++ b/src/common/timing/time-keeper.js
@@ -61,6 +61,8 @@ export class TimeKeeper {
     this.processStoredDiff() // Check session entity for stored time diff
     if (this.#ready) return // Server time calculated from session entity
 
+    if (!nrServerTime) throw new Error('nrServerTime not found')
+
     const medianRumOffset = (endTime - startTime) / 2
     const serverOffset = startTime + medianRumOffset
 
@@ -69,7 +71,7 @@ export class TimeKeeper {
     this.#localTimeDiff = originTime - this.#correctedOriginTime
 
     if (isNaN(this.#correctedOriginTime)) {
-      throw new Error('invalid correctedOriginTime.')
+      throw new Error('invalid correctedOriginTime')
     }
 
     this.#session?.write({ serverTimeDiff: this.#localTimeDiff })
--- a/src/common/timing/time-keeper.js
+++ b/src/common/timing/time-keeper.js
@@ -1,8 +1,6 @@
 import { originTime } from '../constants/runtime'
 import { getRuntime } from '../config/runtime'
 
-const rfc2616Regex = /^(Mon|Tue|Wed|Thu|Fri|Sat|Sun), ([0-3][0-9]) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) ([0-9]{4}) ([01][0-9]|2[0-3])(:[0-5][0-9]){2} GMT$/
-
 /**
  * Class used to adjust the timestamp of harvested data to New Relic server time. This
  * is done by tracking the performance timings of the RUM call and applying a calculation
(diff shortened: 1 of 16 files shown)


src/common/timing/time-keeper.js
```diff
@@ -57,28 +55,23 @@ export class TimeKeeper {
...
     const serverOffset = startTime + medianRumOffset
 
     // Corrected page origin time
-    this.#correctedOriginTime = Math.floor(Date.parse(responseDateHeader) - serverOffset)
+    this.#correctedOriginTime = Math.floor(nrServerTime - serverOffset)
     this.#localTimeDiff = originTime - this.#correctedOriginTime
 
     if (isNaN(this.#correctedOriginTime)) {
-      throw new Error('Date header invalid format.')
+      throw new Error('invalid correctedOriginTime')
```
cwli24, line 74: Change this into something more public friendly

src/common/timing/time-keeper.js
```diff
@@ -57,28 +55,23 @@ export class TimeKeeper {
...
     if (this.#ready) return // Server time calculated from session entity
 
-    const responseDateHeader = rumRequest.getResponseHeader('Date')
-    if (!responseDateHeader) {
-      throw new Error('Missing date header on rum response.')
-    }
-    if (!rfc2616Regex.test(responseDateHeader)) {
-      throw new Error('Date header invalid format.')
-    }
+    if (!nrServerTime) throw new Error('nrServerTime not found')
```
cwli24, line 64: Should we have a default if BCS stops working aka giving nrservertime? So as not to "break" all new session pages on customers browser monitoring traffic.

tools/wdio/plugins/custom-commands.mjs
```diff
@@ -120,14 +120,12 @@ export default class CustomCommands {
      * to a specific value. Default is to set the header to one hour in the past.
      */
     browser.addCommand('mockDateResponse', async function (serverTime = Date.now() - (60 * 60 * 1000), opts = {}) {
-      const { flags } = opts
+      const { flags = {} } = opts
+      const body = JSON.stringify(rumFlags(flags, { nrServerTime: serverTime }))
```
cwli24, line 124: what's with the serverTime formula here? should just use date.now default, no?

src/common/timing/time-keeper.js
metal-messiah: All data therein would be invalid since the timestamps would be wrong so the accepted behavior up to this point is to kill the agent if a time correction cant be made.  That becomes a DEM problem if BCS does not work as intended

tools/wdio/plugins/custom-commands.mjs
metal-messiah: The testing server always sets the timestamp 1 hour back by default, but we have some tests that change the value dynamically so it needs to be parameterized.

cwli24: none of the mockdateresponse actually sets any time

metal-messiah: ah thanks for catching that, then thats a bug with the server-time tests.  We will need to fix that in a separate PR

metal-messiah: https://new-relic.atlassian.net/browse/NR-314521?atlOrigin=eyJpIjoiNzliODMwOGFjMzVjNDU2Zjg2MDE1ZDc5NmIwZTZkMDgiLCJwIjoiaiJ9

metal-messiah: actually.. lets just fix in this PR.  I'll do that

metal-messiah: This would be a lot of effort to rewrite the tests to support before and after.  It is checked in a unit test, so i think this is fine to leave alone.  The parameter isnt hurting anything and can be used later if we need to dynamically set the timestamp.

cwli24 (approved):

PR #1182, opened 2024-09-13
feat: Only decorate PageViews with "hasReplay" if the session is not new

metal-messiah:
Since the order of operations in the sampling mechanism forces a PageView event to harvest before a replay can be sampled, there is no way to tell if a valid replay will actually be harvested before the PageView is sent. This change prevents any PageView events on the first harvest from gaining the `hasReplay` flag for consistency in the UI.
---


metal-messiah:
closing for now, backend teams are trying to address there first

PR #1188, opened 2024-09-17
chore: automate uploading webview assets to lt

metal-messiah:
automate uploading webview assets to lt
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-314477


ptang-nr (approved):
I'm not familiar with how webview assets are used exactly by our tests yet but code-wise seems to look ok.  Voting to approve.

PR #1186, opened 2024-09-17
feat: Add UserAction to GenericEvents

cwli24:
Select user interactive events will be captured in this release. This is an extension of the GenericEvents feature meant to drive UI experience.
---

### Overview

Add `UserAction` event type going to `ins` endpoint. The list of select events are auto captured for DOM elements.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-304358

### Testing

Added

PR #1189, opened 2024-09-17
feat: Marks and measures

metal-messiah:
Add detection for marks and measures which captures data as a `BrowserPerformance` event.  This feature is being shipped in an experimental state and is `off` by default. To opt in to using this feature, it can be turned on by setting the config value `init.performance.capture_marks = true` or `init.performance.capture_measures = true` in Copy/Paste or NPM installations.
---

### Overview
- [x] This PR is branched from the [UserActions PR](https://github.com/newrelic/newrelic-browser-agent/pull/1186), and that PR should be reviewed and merged first.

### Related Issue(s)
NR-308305
NR-308314
NR-308319

### Testing
Component and E2E tests have been added


tests/specs/ins/harvesting.e2e.js
```diff
@@ -110,6 +110,53 @@ describe('ins harvesting', () => {
...
+      entryName: 'before-agent',
+      pageUrl: expect.any(String),
+      timestamp: expect.any(Number),
+      entryType: 'mark'
+    })
+    expect(insHarvest[1]).toMatchObject({
+      entryDuration: 0,
+      eventType: 'BrowserPerformance',
+      entryName: 'after-agent',
+      pageUrl: expect.any(String),
```
ptang-nr, line 134: Minor observation/Qs:

- Would we also want to test for `currentUrl`?  (i see it in the payload)
- This q is mostly out-of-scope but I started wondering about it while looking at the payloads. It seems like `currentUrl` and `pageUrl` are repetitive amongst the UserActions, as well as performance measures.  Would a single payload contain data from other urls / would we ever consider factoring out common url values within a payload to shrink down payload size?

ptang-nr (approved):
LGTM

tests/specs/ins/harvesting.e2e.js
ptang-nr: Discussed this on Slack yesterday - `currentUrl` may differ due to soft navs but `pageUrl` should theoretically stay the same.  So maybe it's something to look into in the future, should the need arise.

`currentUrl` is technically in the payload, but i don't consider it blocking or anything.  So marking as resolved.

PR #1196, opened 2024-09-23
Events storage

cwli24:
---


src/features/utils/processed-events-util.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Object} runtimePendingEvents - this agent's runtime.pendingEvents map
+ * @param {Object} newStorageOptions - options to pass to the ProcessedEvents constructor
+ * @returns {ProcessedEvents}
+ */
+export function getStorageInstance (featureName, runtimePendingEvents, newStorageOptions = {}) {
+  const harvestEndpoint = FEATURE_TO_ENDPOINT[featureName]
+  if (!runtimePendingEvents[harvestEndpoint]) runtimePendingEvents[harvestEndpoint] = []
+  let eventsStorage
+
+  if (harvestEndpoint === 'jserrors' && runtimePendingEvents[harvestEndpoint].length) {
```
metal-messiah, line 29: AJAX, Jserrors, and Metrics do not push events sequentially to a buffer like this, so not sure how you envision this to work without more info

metal-messiah: This also makes it really difficult to "take data back out" for a specific feature if that's a case we need to handle

src/common/storage/processed-events.js
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Cleanup task after a harvest.
+   * @param {Boolean} harvestFailed - harvester flag to restore events in main buffer for retry later if request failed
+   */
+  postHarvestCleanup (harvestFailed = false) {
+    const combinedSize = this.eventsBackupRawSize + this.eventsRawSize
+    if (harvestFailed && this.eventsBackup && (this.eventsBackupRawSize + this.eventsRawSize <= MAX_PAYLOAD_SIZE)) {
+      // If the combined payload would exceed the ingest limit, we would drop the old data.
+      this.events = [...this.eventsBackup, ...this.events]
```
metal-messiah, line 42: this loses all the built-in utility that the event-buffer has, to control whether a payload should even be allowed to re-merge due to size

src/features/page_view_timing/aggregate/index.js
```diff
@@ -113,7 +114,7 @@ export class Aggregate extends AggregateBase {
       attrs.cls = cumulativeLayoutShift.current.value
     }
 
-    this.timings.add({
+    this.timings.addEvent({
```
metal-messiah, line 117: This PR sort of feels like it is doing a lot of similar things to the central event buffer PR with different syntax, but it does not reinforce the concept that a feature agg class should have a standard way to store event data as opposed to the [central event buffer PR](https://github.com/newrelic/newrelic-browser-agent/pull/1171/files#diff-d4d922997ccbb1ffc77794d3ec5b070975a56d6e5cc7cc1c43887ed0be2127a2R19), since it is not built in as native part of the agg-base class.

PR #1203, opened 2024-09-27
chore: Add page url to Session Replay and Session Trace payloads

ptang-nr:
To aid with troubleshooting Session Replay and Session Trace, page url will be included in the respective payload's query string under `attributes.pageUrl`.
---

### Overview

In this PR, we are adding the page URL to the query string of the corresponding Session Replay and Session Trace payloads under `attributes.pageUrl`.  The page URL will be added last, so if the total payload size exceeds current limit (5000), then truncation may occur.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-268394

### Testing

- Added a test condition to the Session Replay standard payload unit test
- Manually tested and confirmed presence of URL in payload:
  
Session Replay
(screenshot - to be added)

Session Trace
(screenshot - to be added)

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -28,6 +28,7 @@ import { deregisterDrain } from '../../../common/drain/drain'
 import { now } from '../../../common/timing/now'
 import { buildNRMetaNode } from '../shared/utils'
 import { MAX_PAYLOAD_SIZE } from '../../../common/constants/agent-constants'
+import { cleanURL } from '../../../common/url/clean-url'
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
@@ -362,7 +363,8 @@ export class Aggregate extends AggregateBase {
           'rrweb.version': RRWEB_VERSION,
           'payload.type': recorderEvents.type,
           // customer-defined data should go last so that if it exceeds the query param padding limit it will be truncated instead of important attrs
-          ...(endUserId && { 'enduser.id': this.obfuscator.obfuscateString(endUserId) })
+          ...(endUserId && { 'enduser.id': this.obfuscator.obfuscateString(endUserId) }),
+          pageUrl: cleanURL(agentRuntime.origin)
           // The Query Param is being arbitrarily limited in length here.  It is also applied when estimating the size of the payload in getPayloadSize()
         }, QUERY_PARAM_PADDING).substring(1) // remove the leading '&'
       },
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -364,7 +364,7 @@ export class Aggregate extends AggregateBase {
           'payload.type': recorderEvents.type,
           // customer-defined data should go last so that if it exceeds the query param padding limit it will be truncated instead of important attrs
           ...(endUserId && { 'enduser.id': this.obfuscator.obfuscateString(endUserId) }),
-          pageUrl: cleanURL(agentRuntime.origin)
+          currentUrl: cleanURL('' + location)
           // The Query Param is being arbitrarily limited in length here.  It is also applied when estimating the size of the payload in getPayloadSize()
         }, QUERY_PARAM_PADDING).substring(1) // remove the leading '&'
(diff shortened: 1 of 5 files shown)


src/features/session_replay/aggregate/index.js
```diff
@@ -362,7 +363,8 @@ export class Aggregate extends AggregateBase {
           'rrweb.version': RRWEB_VERSION,
           'payload.type': recorderEvents.type,
           // customer-defined data should go last so that if it exceeds the query param padding limit it will be truncated instead of important attrs
-          ...(endUserId && { 'enduser.id': this.obfuscator.obfuscateString(endUserId) })
+          ...(endUserId && { 'enduser.id': this.obfuscator.obfuscateString(endUserId) }),
+          pageUrl: cleanURL(agentRuntime.origin)
```
metal-messiah, line 367: We might want to consider origin pageUrl vs "current" pageUrl, for SPA cases (not just here, but across the agent where we report it)

ptang-nr: @metal-messiah hey, I want to check my understanding... 

So I think the issue here is essentially that `agentRuntime` is  'fixed' after initialization (i totally missed that). Are we saying we might want to have a `currentUrl` available from `agentRuntime`? (so perhaps expose a method to update `model.currentUrl` and then we call it where needed?)

metal-messiah: Because if we dont, the ST/SR payload will always point at the origin, instead of the soft nav url, which could affect the ability to debug each payload (ie they look like they match urls but actually dont in reality due to soft navigations).

My question was more geared toward:
Should we report pageUrl (origin), as well as "currentUrl" (current spa route)

cwli24: - Seemingly no customer value here to pad extra data; solely for debugging?
- Does knowing the exact url on a replay chunk really help dbg? I'm not certain how so. But since you can get origin url by looking the associated PageView event, the current url if needed would be more helpful.
- The `PageAction` event's `pageUrl` should also be updated then to reflect current url (enhancement/feature change bump)

metal-messiah: pageAction already reports both the `pageUrl` (origin) and `currentUrl` (soft nav route)

metal-messiah: > Seemingly no customer value here to pad extra data; solely for debugging?

Definitely not, right now there are lots of joins that have to happen to get page view data to supplement replay data, that could be removed and rely on the payload as source of truth, which decreases load time in the UI and increases reliability of replays and traces.

cwli24: Then we should sync Logging's prop name, or value, no?

metal-messiah: > Does knowing the exact url on a replay chunk really help dbg? I'm not certain how so. But since you can get origin url by looking the associated PageView event, the current url if needed would be more helpful.

Yes because we often get cases where multiple replays are reporting simultaneously, and differentiating them becomes difficult. Think of the recent ST conversations where there are multiple payloads with the same PTID when we think it shouldnt be happening.  It would be nice to be able to rule out same page vs diff pages

metal-messiah: > Then we should sync Logging's prop name, or value, no?

Yeah we probably need to align all the payloads that report urls to report the same things

cwli24: Ok, I tickted the correction for logging in JIRA. Sounds like the actionable **here** is to change `pageUrl` & value to actually be `currentUrl` & corresponding value.

ptang-nr: (ok, sounds like we aren't asking for any change/addition to runtime obj right now)

 I prepped something locally on Friday and it was _pretty_ close.  I saw we were using the following in some places already, would this be acceptable? @cwli24 @metal-messiah 
`currentUrl: cleanURL('' + location)`

metal-messiah (approved):
Once this has a passing test run i think this is g2g

metal-messiah (approved):

PR #1207, opened 2024-10-02
fix: Always return a string for custom stringify method

ptang-nr:
An internal mechanism used to convert JS values to string was returning `undefined` for certain values, which may lead to errors for downstream callers when they are expecting a string.  This is part 2 of a patch to ensure a string is returned.

---

### Overview

Uses `''` as the fallback value whenever `JSON.stringify` returns `undefined`.

### Related Issue(s)

[NR-316691](https://new-relic.atlassian.net/browse/NR-316691)

### Testing

Adjusted unit tests

[NR-316691]: https://new-relic.atlassian.net/browse/NR-316691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/unit/common/util/stringify.test.js
```diff
@@ -19,10 +19,12 @@ test('should utilize JSON.stringify to serialize input data', () => {
...
   ['array', [0, 1, 'asdf', undefined, null, 'weee'], '[0,1,"asdf",null,null,"weee"]'],
   ['object', { a: 123, c: 'b', f: 'asdf', u: undefined, n: null }, '{"a":123,"c":"b","f":"asdf","n":null}'],
-  ['undefined', undefined, undefined],
+  ['undefined', undefined, ''],
   ['null', null, 'null'],
   ['number', 123, '123'],
-  ['function', function F () {}, undefined],
+  ['string', '0', '"0"'],
+  ['symbol', Symbol('foo'), ''],
+  ['function', function F () {}, ''],
```
ptang-nr, line 27: Assuming we are to follow the native `JSON.stringify` behavior for functions, which returns `undefined`, which in turn becomes `''` in our util method (as opposed to, the function `toString()` method which returns the entire function as a string).

tests/specs/err/scripts-and-callbacks.e2e.js
```diff
@@ -152,38 +152,41 @@ describe('JSE Error detection in various callbacks', () => {
...
+    { message: 'Unhandled Promise Rejection: {"a":1,"b":{"a":1}}', meta: 'nested obj' },
+    { message: 'Unhandled Promise Rejection: [1,2,3]', meta: 'array' },
+    { message: 'Unhandled Promise Rejection: test', meta: 'error with message' },
+    { message: 'test', meta: 'error with no setter with message' },
+    { message: 'Unhandled Promise Rejection', meta: 'undefined' },
+    { message: 'Unhandled Promise Rejection: null', meta: 'null' },
+    { message: 'Unhandled Promise Rejection: ', meta: 'error with no message' },
+    { message: 'Unhandled Promise Rejection: {}', meta: 'map object' },
+    { message: 'Unhandled Promise Rejection: {"abc":"Hello"}', meta: 'factory function' },
+    { message: 'Unhandled Promise Rejection: ', meta: 'uncalled function' },
```
ptang-nr, line 167: This was the culprit that was previously not passing.  Functions are now stringified to `''` instead of `undefined`.

tests/unit/common/util/stringify.test.js
metal-messiah: For brevity, can you also add a case to this unit test that implements the toJSON attribute on an object?

metal-messiah:
```suggestion
  ['function', function F () {}, ''],
  ['toJSON valid', { toJSON: () => ({foo:'bar'})}, '{"foo":"bar"}'],
  ['toJSON undefined', { toJSON: () => undefined}, ''],
  ['toJSON error', { toJSON: () => {throw new Error('toJSON')}}, ''],
```

metal-messiah (approved):

PR #1212, opened 2024-10-07
fix: Only ever allow session traces to capture page load timings once

metal-messiah:
An edge case issue was resolved where session traces could include duplicate copies of the page load timings obtained from the browser timing APIs.
---

### Related Issue(s)
NR-315831

### DIFF
--- a/src/features/session_trace/aggregate/index.js
+++ b/src/features/session_trace/aggregate/index.js
@@ -12,7 +12,6 @@ import { globalScope } from '../../../common/constants/runtime'
 import { MODE, SESSION_EVENTS } from '../../../common/session/constants'
 import { applyFnToProps } from '../../../common/util/traverse'
 import { cleanURL } from '../../../common/url/clean-url'
-import { single } from '../../../common/util/invoke'
 
 const ERROR_MODE_SECONDS_WINDOW = 30 * 1000 // sliding window of nodes to track when simply monitoring (but not harvesting) in error mode
 /** Reserved room for query param attrs */
@@ -65,6 +64,12 @@ export class Aggregate extends AggregateBase {
         // if another page's session entity has expired, or another page has transitioned to off and this one hasn't... we can just abort straight away here
         if (this.sessionId !== sessionState.value || (eventType === 'cross-tab' && this.scheduler?.started && sessionState.sessionTraceMode === MODE.OFF)) this.abort(2)
       })
+
+      if (typeof PerformanceNavigationTiming !== 'undefined') {
+        this.traceStorage.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
+      } else {
+        this.traceStorage.storeTiming(globalScope.performance?.timing, true)
+      }
     }
 
     /** ST/SR sampling flow in BCS - https://drive.google.com/file/d/19hwt2oft-8Hh4RrjpLqEXfpP_9wYBLcq/view?usp=sharing */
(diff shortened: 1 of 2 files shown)


src/features/session_trace/aggregate/index.js
```diff
@@ -93,11 +94,13 @@ export class Aggregate extends AggregateBase {
...
-    } else {
-      this.traceStorage.storeTiming(globalScope.performance?.timing, true)
-    }
+    single(() => {
+      if (typeof PerformanceNavigationTiming !== 'undefined') {
+        this.traceStorage.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
+      } else {
+        this.traceStorage.storeTiming(globalScope.performance?.timing, true)
+      }
+    })
```
cwli24, line 103: This doesn't work. Notwithstanding the fn returned by `single(...)` is never actually executed, every time `initialize()` is called, this'll simply get a single wrap of a "new" function. Make this an instance field defined above the constructor instead:
```suggestion
  collectNavTimingOnce = single(() => {
      if (typeof PerformanceNavigationTiming !== 'undefined') {
        this.traceStorage.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
      } else {
        this.traceStorage.storeTiming(globalScope.performance?.timing, true)
      }
    })
```
then call it here
```suggestion
    this.collectNavTimingOnce()
```

tests/components/session_trace/aggregate.test.js
cwli24, line 1: This need better more involved (prob e2e) test case(s) since the test suite didn't detect anything wrong even though the change is significantly non-functional.

src/features/session_trace/aggregate/index.js
cwli24: Example of what I mean:
image

src/features/session_trace/aggregate/index.js
```diff
@@ -93,11 +94,13 @@ export class Aggregate extends AggregateBase {
     registerHandler('trace-jserror', (...args) => this.traceStorage.storeErrorAgg(...args), this.featureName, this.ee)
     registerHandler('pvtAdded', (...args) => this.traceStorage.processPVT(...args), this.featureName, this.ee)
 
-    if (typeof PerformanceNavigationTiming !== 'undefined') {
-      this.traceStorage.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
-    } else {
-      this.traceStorage.storeTiming(globalScope.performance?.timing, true)
-    }
+    single(() => {
```
ptang-nr, line 97: Asking for my own understanding - are we using `single()` here because it's important to store the timing here instead of, say, in the `if !(this.initialized)` block?

metal-messiah: I think it could go either place, any strong preference one way vs another?

tests/components/session_trace/aggregate.test.js
metal-messiah: good call, will adjust

src/features/session_trace/aggregate/index.js
metal-messiah: ty

tests/components/session_trace/aggregate.test.js
metal-messiah: test added

src/features/session_trace/aggregate/index.js
metal-messiah: i actually feel like this is a little cleaner without needing an extra import or class attr, so going with this

tests/components/session_trace/aggregate.test.js
cwli24: I had in mind for e2e a more "natural" approach, like a workflow that caused initialize again, but if you're just calling `initialize()` again directly on the class instance, it can really just be a unit/component jest test with that approach.

cwli24: This is fine though as you won't get the actual timings on jest and would have to spyOn or something.

src/features/session_trace/aggregate/index.js
```diff
@@ -64,6 +64,12 @@ export class Aggregate extends AggregateBase {
...
       })
+
+      if (typeof PerformanceNavigationTiming !== 'undefined') {
+        this.traceStorage.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
+      } else {
+        this.traceStorage.storeTiming(globalScope.performance?.timing, true)
+      }
     }
 
     /** ST/SR sampling flow in BCS - https://drive.google.com/file/d/19hwt2oft-8Hh4RrjpLqEXfpP_9wYBLcq/view?usp=sharing */
```
cwli24, line 75: this is low impact but optimally L75-82 here can be moved up to before this block, so that the early return if applicable executes before we store stuff into `traceStorage` indefinitely in such case.

cwli24 (approved):

PR #1215, opened 2024-10-10
fix: Handle Session Replay Security Policy Errors

metal-messiah:
Handle Security Policy Errors encountered by session replay when browser APIs have been natively disabled.  This was also observed to happen in Chromium browsers that completely disabled cookie storage.  Supportability metrics about agent internal errors are now captured as a biproduct of this work to help facilitate future improvements.
---

### Overview
All errors thrown from either the `rrweb` library directly or the wrapping `nr-recorder` module will be handled through the new internal handler.
All instances of internal errors will now generate a supportability metric instead of an unused `ierr` harvest.  There are a few basic categories added as part of this pr.  More categories can be added later as needs arise

### Related Issue(s)
NR-316093

### Testing
Tests have been updated to evaluate the new behavior

### DIFF
--- a/docs/supportability-metrics.md
+++ b/docs/supportability-metrics.md
@@ -312,3 +312,9 @@ A timeslice metric is harvested to the JSE/XHR consumer. An aggregation service 
 * API/log/called
 <!--- newrelic.wrapLogger() was called --->
 * API/wrapLogger/called
+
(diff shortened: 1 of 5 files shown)


ptang-nr (dismissed):
Voting to approve 👍

PR #1216, opened 2024-10-10
feat: Relocate aggregator from loader and improve agent internals

cwli24:
Styling and architectural changes that will reduce agent loader size by moving the Aggregator module to async chunk and removing the need for some configuration getters in the features. No behavioral change or significant performance impact from the agent is expected. 
---

### Overview

`aggregator` no longer given to feature aggregate classes at initialization as it's widely unused. The couple features that uses the central aggregator can now reference the agent instance's `sharedAggregator` field. Further, configs can be grabbed off said instance directly rather than calling getX methods, so they have also been replaced as part of this.

Impact:

- aggregator code moved to async chunk from loader.
- since this modifies core code and every feature, this PR release should be treated as having mid to high risk.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-324625
https://new-relic.atlassian.net/browse/NR-322022

### DIFF
--- a/src/features/session_replay/instrument/index.js
+++ b/src/features/session_replay/instrument/index.js
@@ -17,16 +17,16 @@ export class Instrument extends InstrumentBase {
 
   #mode
   #agentRef
-  constructor (thisAgent, auto = true) {
-    super(thisAgent, FEATURE_NAME, auto)
+  constructor (agentRef, auto = true) {
+    super(agentRef, FEATURE_NAME, auto)
     let session
     this.replayRunning = false
-    this.#agentRef = thisAgent
+    this.#agentRef = agentRef
     try {
       session = JSON.parse(localStorage.getItem(`${PREFIX}_${DEFAULT_KEY}`))
     } catch (err) { }
 
-    if (hasReplayPrerequisite(thisAgent.agentIdentifier)) {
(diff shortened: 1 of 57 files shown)


src/features/session_replay/instrument/index.js
```diff
@@ -16,23 +16,25 @@ export class Instrument extends InstrumentBase {
...
 
   #mode
-  constructor (agentIdentifier, aggregator, auto = true) {
-    super(agentIdentifier, aggregator, FEATURE_NAME, auto)
+  #agentRef
+  constructor (agentRef, auto = true) {
+    super(agentRef, FEATURE_NAME, auto)
     let session
     this.replayRunning = false
+    this.#agentRef = agentRef
```
metal-messiah, line 24: Should the inst base just set `this.agentRef` like the agg base does?

cwli24: So far nothing else has needed it outside the constructors since the instrument files are not complicated enough to warrant complex stack calls. I think we leave this until the cost arises as the change would be very easy at that stage.

metal-messiah (approved):
Approved, looks good.  This needs to probably wait to merge and be released in its own release to mitigate risks

PR #1220, opened 2024-10-15
feat: Move deregisterDrain method to feature-base

ellisong:
Moves the `deregisterDrain` method in order to standardize the implementation pattern to match the drain method.
---

### Overview

Moves the `deregisterDrain` method in order to standardize the implementation pattern to match the drain method. Aggregate and instrument features should now use `this.deregisterDrain()` instead of importing the method itself. This is done to make it easier to call the `deregisterDrain` method regardless of feature in a consistent manner like the `drain` method (which currently only works for aggregate methods).

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-297543

### Testing

Make sure tests pass. The agent needs to function identically to before.

### DIFF
--- a/src/features/utils/aggregate-base.js
+++ b/src/features/utils/aggregate-base.js
@@ -3,7 +3,7 @@ import { getInfo, isValid } from '../../common/config/info'
 import { getRuntime } from '../../common/config/runtime'
 import { configure } from '../../loaders/configure/configure'
 import { gosCDN } from '../../common/window/nreum'
-import { deregisterDrain, drain } from '../../common/drain/drain'
+import { drain } from '../../common/drain/drain'
 import { activatedFeatures } from '../../common/util/feature-flags'
 import { Obfuscator } from '../../common/util/obfuscate'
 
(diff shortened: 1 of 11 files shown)


src/features/utils/aggregate-base.js
```diff
@@ -47,6 +47,10 @@ export class AggregateBase extends FeatureBase {
     this.drained = true
   }
 
+  deregisterDrain () {
+    deregisterDrain(this.agentIdentifier, this.featureName)
+  }
+
```
metal-messiah, line 53: I think we should consider putting this on the `FeatureBase` class instead of the `AggregateBase` class so that the instrument modules can use this as well (Both AggregateBase and InstrumentBase extend FeatureBase so they would then both have access).  

There are a couple instrument files that call `deregisterDrain`:
* session_trace/instrument
* generic_events/instrument

Thoughts about that?

ellisong: That's fine with me.

metal-messiah (approved):

metal-messiah (approved):

PR #1224, opened 2024-10-21
feat: Remove supportsSendBeacon runtime property

ellisong:
Removes the supportsSendBeacon runtime property from our agent code. We no longer officially support browsers that cannot use the sendBeacon method, so there's no point in checking for it.
---

### Overview

Removes the supportsSendBeacon runtime property from our agent code. This is done to tidy up the agent code and shorten the agent payload size a little bit.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-259634

### Testing

Make sure tests pass.


ptang-nr (approved):
LGTM

PR #1225, opened 2024-10-22
fix: Correct naming for logging pageUrl attribute

ellisong:
Corrects naming for the Logging pageUrl attribute by using the original url of the page instead of the url of when the event happens. Removes origin attribute from the runtime model object.
---

### Overview

Corrects the naming for the Logging `pageUrl` attribute by using the original url of the page instead of the url of when the event happens. The original url of the page is also referred to as `initialLocation`. A test has been added to check that the `pageUrl` attribute in the Log object is `initialLocation`.

Removes origin attribute from the runtime model object by cleaning up usage of `runtime.origin` in spa feature. The `initialLocation` constant is a direct replacement for it.

These changes are for cleaning up the agent code.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-320541

### Testing

Make sure tests pass. Agent should work identically to before except that the `pageUrl` for logging events is now corrected to the original URL.

### DIFF
--- a/tests/components/logging/aggregate.test.js
+++ b/tests/components/logging/aggregate.test.js
@@ -1,4 +1,5 @@
 import { getRuntime } from '../../../src/common/config/runtime'
+import { initialLocation } from '../../../src/common/constants/runtime'
 import { LOGGING_EVENT_EMITTER_CHANNEL } from '../../../src/features/logging/constants'
 import { Instrument as Logging } from '../../../src/features/logging/instrument'
 import { Log } from '../../../src/features/logging/shared/log'
@@ -181,6 +182,20 @@ describe('payloads', () => {
 
     loggingAggregate.ee.emit(LOGGING_EVENT_EMITTER_CHANNEL, [1234, Symbol('test'), {}, 'error'])
     expect(logs.pop().message).toEqual('Symbol(test)')
+  })
+
+  test('initialLocation should be in pageUrl of log object attributes', async () => {
+    const log = new Log(
+      Math.floor(runtime.timeKeeper.correctAbsoluteTimestamp(
+        runtime.timeKeeper.convertRelativeTimestamp(1234)
+      )),
+      'test message',
+      { },
+      'error'
+    )
+    const expected = initialLocation.toString()
+
+    expect(log.attributes.pageUrl).toEqual(expected)
   })
 })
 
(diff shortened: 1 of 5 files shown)


tests/components/logging/aggregate.test.js
```diff
@@ -182,6 +183,20 @@ describe('payloads', () => {
     loggingAggregate.ee.emit(LOGGING_EVENT_EMITTER_CHANNEL, [1234, Symbol('test'), {}, 'error'])
     expect(logs.pop().message).toEqual('Symbol(test)')
   })
+
+  test('initialLocation should be in pageUrl of log object attributes', async () => {
+    const log = new Log(
+      Math.floor(runtime.timeKeeper.correctAbsoluteTimestamp(
+        runtime.timeKeeper.convertRelativeTimestamp(1234)
+      )),
```
ptang-nr, line 191: For my understanding, why do we invoke timeKeeper methods here?  Could we have used just `1234` as the pretend timestamp?

ptang-nr (dismissed):
LGTM

tests/components/logging/aggregate.test.js
```diff
@@ -182,6 +183,20 @@ describe('payloads', () => {
     loggingAggregate.ee.emit(LOGGING_EVENT_EMITTER_CHANNEL, [1234, Symbol('test'), {}, 'error'])
     expect(logs.pop().message).toEqual('Symbol(test)')
   })
+
+  test('initialLocation should be in pageUrl of log object attributes', async () => {
```
cwli24, line 187: This test would be more meaningful if the window.location is changed to be different from initialLocation first

tests/components/logging/aggregate.test.js
cwli24: Yeah just use a random number will do instead of relying on these 2 timekeeper fns (side note: btw they have been replaced in src with correctRelativeTimestamp)

ptang-nr: ^ Good point.

hey @ellisong, I'll leave it up to you whether to make a cleanup here or not.  I'm going to re-approve the PR since this is not technically blocking and I _think_ y'all resolved the other comment.

ptang-nr (approved):

PR #1234, opened 2024-10-31
fix: Ensure all lazy loaded modules issue warning instead of errors

metal-messiah:
Lazy loaded modules that are blocked from loading by third party tools such as adblocking mechanisms will issue a warning to console instead of throwing an error.
---

### Related Issue(s)
NR-335501

### DIFF
--- a/src/features/utils/instrument-base.js
+++ b/src/features/utils/instrument-base.js
@@ -90,18 +90,29 @@ export class InstrumentBase extends FeatureBase {
         if (this.featureName === FEATURE_NAMES.sessionReplay) this.abortHandler?.() // SR should stop recording if session DNE
       }
 
-      // Create a single Aggregator for this agent if DNE yet; to be used by jserror endpoint features.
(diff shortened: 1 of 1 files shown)


cwli24 (approved):
We can slightly tune this more, but for simplicity sakes, this is good.

src/features/utils/instrument-base.js
cwli24, line 1: Actually throw on a jest test for this like what we have for session entity and/or the aggregate

JackWH:
@cwli24 Hey there, just found this after getting alerted to 5000+ JS errors in Bugsnag! It caused us to hit our rate limits and we're losing insight on stuff that matters. Do you know if this will be fixed soon, as it's now affecting customers in production... Thanks in advance 🙏

metal-messiah:
@JackWH Sorry about this, we are getting this rolled out now.  Should hopefully reach customer hands within a few hours

JackWH:
@metal-messiah Good to know, thanks for the speedy reply 😊

PR #1247, opened 2024-11-08
chore: Cleanup old browser test matchers

ellisong:
Cleanup old browser test matchers in our testing code. This helps with having less outdated matchers to worry about when writing tests.
---

### Overview

Cleanup old browser test matchers in our testing code. This helps with having less outdated matchers to worry about when writing tests.

Notable changes to the browser matchers (which take into account our current and future supported browsers) include:
- All supported browsers support the fetch API, both basic and extended
- All supported browsers support ES2022 (used in testing NPM agent)
- All supported browsers support back/forward cache. Edge browser has it enabled for the latest 10 browser versions.
- All supported browsers support First Contentful Paint timing
- A couple of unused browser matchers were removed (supportsMultipleTabs, notFirefox, notMobile)

All of the browsers in the matchers are reorganized in alphabetical order.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-283061

### Testing

Make sure tests pass, especially LambdaTest and integration tests. The agent code itself isn't changed at all.


metal-messiah:
This looks good, but can you change the `feat:` to a `chore:`.  I dont think this one needs to end up in the changelog since it only affects test runners (imo)

metal-messiah (approved):

PR #1250, opened 2024-11-13
fix: Harvest generic events when max size is reached

metal-messiah:
To avoid behavior that could lead to more network requests than necessary, generic events will now harvest early only if the max payload size is reached. This represents an increase from 64kb to 1MB.
---

### DIFF
--- a/.github/actions/build-ab/templates/released.js
+++ b/.github/actions/build-ab/templates/released.js
@@ -25,7 +25,7 @@ window.NREUM={
       enabled: true
     },
     performance: {
-      capture_marks: true, 
+      capture_marks: false, 
       capture_measures: true
     },
     proxy: {}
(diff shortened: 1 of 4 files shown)


.github/actions/build-ab/templates/released.js
cwli24, line 1: Only for dev or all env?

tests/specs/ins/harvesting.e2e.js
```diff
@@ -51,6 +51,7 @@ describe('ins harvesting', () => {
     ])
 
     const userActionsHarvest = insHarvests.flatMap(harvest => harvest.request.body.ins) // firefox sends a window focus event on load, so we may end up with 2 harvests
+    console.log(userActionsHarvest)
```
cwli24, line 54: Remove

.github/actions/build-ab/templates/released.js
metal-messiah: for all

tests/specs/ins/harvesting.e2e.js
metal-messiah: ty

cwli24 (approved):
Can you tune the PR summary to explicitly mention the generic events feature and prev 64kb harvest cap?

PR #1252, opened 2024-11-13
fix: Limit MicroAgent API based on available features

cwli24:
MicroAgent class now only shows the relevant API for use with NPM. The set is limited to methods that relate to features available with the MicroAgent.
---

### Overview

MicroAgent is trimmed down to the 10 methods, including `start`, that are relevant. The full set of `api.js` is still imported and set, but the public interface or users' IDE will not list the other methods used by non-MicroAgent features.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-289146

### DIFF
--- a/src/loaders/agent-base.js
+++ b/src/loaders/agent-base.js
@@ -4,11 +4,7 @@ import { warn } from '../common/util/console'
 import { SR_EVENT_EMITTER_TYPES } from '../features/session_replay/constants'
 import { generateRandomHexString } from '../common/ids/unique-id'
 
-/**
- * @typedef {import('./api/interaction-types').InteractionInstance} InteractionInstance
- */
-
-export class AgentBase {
+export class MicroAgentBase {
   agentIdentifier
 
   constructor (agentIdentifier = generateRandomHexString(16)) {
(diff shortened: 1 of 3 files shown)


src/loaders/agent-base.js
```diff
@@ -4,11 +4,7 @@ import { warn } from '../common/util/console'
 import { SR_EVENT_EMITTER_TYPES } from '../features/session_replay/constants'
 import { generateRandomHexString } from '../common/ids/unique-id'
 
-/**
- * @typedef {import('./api/interaction-types').InteractionInstance} InteractionInstance
- */
-
-export class AgentBase {
+export class MicroAgentBase {
```
metal-messiah, line 7: no personal preference here, but we haven't really mixed multiple classes in single files like this as a pattern in the agent up to now (i dont think) .  Could consider putting MicroAgentBase in its own named file.

cwli24: sure, does make it more PITA to read across 2 files than 1 for me but it's a toss up btwn convention & convenience

metal-messiah (approved):

PR #1253, opened 2024-11-13
feat: Aggregate log events based on rum flag

ptang-nr:
Aggregate logging events according to the `log` flag received from rum call to BCS.
---

### Overview

Part 2 of the auto-logging feature work. This work will enable the logging feature to be controlled via BCS, which is ultimately based on user settings from UI (on/off, sampling, log level).

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-323597

### Testing

Extended existing logging aggregate unit and e2e tests around handling the flag.  Includes tests for each logging mode to ensure log messages of equal or higher levels are collected.  For example, if the `log` flag = 3 (INFO), then collect `console.error`, `console.warn`, `console.info`, and `console.log` events.

Extended existing tests to ensure no harvests when rum flag = 0 and events are drained.

### DIFF
--- a/src/features/logging/aggregate/index.js
+++ b/src/features/logging/aggregate/index.js
@@ -5,7 +5,7 @@ import { warn } from '../../../common/util/console'
 import { stringify } from '../../../common/util/stringify'
 import { SUPPORTABILITY_METRIC_CHANNEL } from '../../metrics/constants'
 import { AggregateBase } from '../../utils/aggregate-base'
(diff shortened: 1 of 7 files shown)


metal-messiah (approved):

ptang-nr:
Mentioned in standup today (and got ack'd) that the default level was changed after the initial approval. We _will_ re-review again when officially merging feat branch to main, so gonna go ahead and merge now.

PR #1255, opened 2024-11-19
chore: improve testing server debug shim

metal-messiah:
The current logging shim output of the test server alters Ajax and resource payloads when enabled because it generates network requests to capture the log messages. This change introduces a global array that intercepts and stores log messages in it that can be fetched on the fly with a browser execute statement. 
---

### Overview
An example of how to use this would be something like:
1. *anywhere* in your test add the included helper method
    - `console.log(await getDebugLogs())`
This will return the current state of the log message buffer.  Anytime something like `console.log` is called in the test execution step (client side), the log buffer will be populated with the contents.  You must start the test server with the `-d` flag as before for the shim to be injected into the page.

### DIFF
--- a/tests/specs/util/helpers.js
+++ b/tests/specs/util/helpers.js
@@ -106,6 +106,18 @@ export function decodeAttributes (attributes) {
   return decodedObj
 }
 
+export async function getDebugLogs () {
+  return browser.execute(function () {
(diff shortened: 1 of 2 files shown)


ptang-nr (approved):
😎

PR #1260, opened 2024-11-20
chore: Allow logs api wrapper to add custom attributes

ptang-nr:
This will support the rollout of the auto-logging feature without disrupting customers that may currently be using `wrapLogger` to wrap the `console` object.
---

### Overview

Currently, once an object's method is wrapped, it is blocked from being re-wrapped.  With the auto-logging feature, the browser agent's internal wrapping for `console` will be triggered before any other wrappers that may be invoked via `wrapLogger`.  This creates a small potential change in behavior if customers are already using `wrapLogger` to wrap the `console` object. 

So the logging context is now modifiable, such that log api wrappers can overwrite the custom attributes sent along with the log events.  For a given wrapped method, attributes supplied by the last wrapper will "win".

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-323597

### Testing

Updated existing test regarding re-wrapping.  Attributes supplied by subsequent re-wrapping will be used instead of the initial ones.

### DIFF
--- a/src/common/wrap/wrap-function.js
+++ b/src/common/wrap/wrap-function.js
@@ -46,6 +46,11 @@ export function createWrapperWithEmitter (emitter, always) {
 
   return wrapFn
 
+  function getContextName (prefix, methodName) {
+    const contextPrefix = prefix ? prefix + methodName : methodName
+    return `${contextPrefix}-ctx`
+  }
+
   /**
    * Wraps a function with event emitting functionality.
    * @param {function} fn - The function to wrap.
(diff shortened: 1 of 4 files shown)


src/common/wrap/wrap-function.js
```diff
@@ -128,6 +140,9 @@ export function createWrapperWithEmitter (emitter, always) {
       const method = methods[i]
       const fn = obj[method]
 
+      if (overwriteContext && fn[flag]) {
+        fn[getContextName(prefix, method)] = getContext
```
metal-messiah, line 144: We may need to be careful here about mutating the original function like this. let's mob as a group about this PR

ptang-nr:
We talked about this during rodeo.  Closing in favor of the solution in [test-context](https://github.com/newrelic/newrelic-browser-agent/compare/main...test-context) branch.

PR #1261, opened 2024-11-20
fix: Ignore reserved attribute names on UserActions tied to the window

metal-messiah:
Force the agent to not capture target-level attributes for window-level UserActions. Customers can set window-level variables that mimic HTML element attributes like type or tagName. This change prevents the potential capture of those attributes as part of blur or focus user actions tied to the window.
---


ellisong:
Why is the `tests/assets/user-actions-modified-window.html` file unused?

metal-messiah:
> Why is the `tests/assets/user-actions-modified-window.html` file unused?

Thanks i think i missed wiring it up.  I'll fix the test to utilize it

ellisong (approved):
LGTM

PR #1264, opened 2024-11-25
feat: Persist logging flag in session

ptang-nr:
This is another part of the work for the auto-logging feature.  The auto-logging mode should stay consistent within a given session and will auto-disable upon session expiration after 4 hours or session inactivity after 30 minutes.
---

### Overview
The auto-logging mode should stay consistent within a given [session](https://docs.newrelic.com/docs/browser/browser-monitoring/page-load-timing-resources/cookie-collection-session-tracking/) and will auto-disable upon session expiration after 4 hours or session inactivity after 30 minutes.

If cookies are not enabled, then the logging mode for each page will be set independently based on the rum response.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-341932

### Testing

Added/updated relevant tests for logging with regards to session

### DIFF
--- a/src/features/logging/aggregate/index.js
+++ b/src/features/logging/aggregate/index.js
@@ -10,7 +10,9 @@ import { Log } from '../shared/log'
 import { isValidLogLevel } from '../shared/utils'
 import { applyFnToProps } from '../../../common/util/traverse'
 import { MAX_PAYLOAD_SIZE } from '../../../common/constants/agent-constants'
-import { FEATURE_TO_ENDPOINT } from '../../../loaders/features/features'
+import { FEATURE_NAMES, FEATURE_TO_ENDPOINT } from '../../../loaders/features/features'
+import { SESSION_EVENT_TYPES, SESSION_EVENTS } from '../../../common/session/constants'
+import { ABORT_REASONS } from '../../session_replay/constants'
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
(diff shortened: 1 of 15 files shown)


src/features/utils/instrument-base.js
```diff
@@ -139,6 +139,8 @@ export class InstrumentBase extends FeatureBase {
         return hasReplayPrerequisite(this.agentIdentifier) && !!session
       case FEATURE_NAMES.sessionTrace:
         return !!session
+      case FEATURE_NAMES.logging:
+        return !!session
```
metal-messiah, line 143: While this makes sense and I can see why you added this bit, product decided that the logging feature should still import and run if session is disabled.

src/features/logging/aggregate/index.js
```diff
@@ -10,21 +10,43 @@ import { Log } from '../shared/log'
...
         this.blocked = true
         this.deregisterDrain()
         return
       }
-      this.loggingMode = loggingMode
+      if (session.isNew || !canEnableSessionTracking(this.agentRef.agentIdentifier)) {
+        this.loggingMode = loggingMode
+        this.syncWithSessionManager({
+          loggingMode: this.loggingMode
+        })
```
metal-messiah, line 46: I think we need to double check that this wont cause issues. One of the two cases that could trigger this line is that session track is disabled, meaning there is no session object.  I fear calling this method could cause errors when that is the case

metal-messiah: Should probably only try to sync with the session manager if we know it exists for sure. Either handled here or in the base class method

tests/specs/logging/harvesting.e2e.js
```diff
@@ -11,12 +11,18 @@ describe('logging harvesting', () => {
     })
   })
 
+  afterEach(async () => {
+    // logging mode is sticky to the session, so we need to reset before the next test
+    await await browser.destroyAgentSession()
```
metal-messiah, line 16: is the double await here intended?

tests/specs/logging/harvesting.e2e.js
```diff
@@ -81,7 +87,7 @@ describe('logging harvesting', () => {
         it(`should harvest expected logs - ${type} post load - logging mode: ${mode}`, async () => {
           await mockRumResponse(logLevel)
           const [[{ request: { body } }]] = await Promise.all([
-            logsCapture.waitForResult({ totalCount: 1 }),
+            logsCapture.waitForResult({ totalCount: 1, timeout: 15000 }),
```
metal-messiah, line 90: Just curious what was the reason for needing to add a timeout on the various logsCapture calls across some of these tests

tests/specs/logging/harvesting.e2e.js
ptang-nr: Nope, that's most definitely a typo 👍

tests/specs/logging/harvesting.e2e.js
ptang-nr: I wanted to speed up the test failures.  When I was working through the failures, it was taking a long time for the tests to timeout + fail.  15 secs seemed enough to determine if it was going to pass or not.

src/features/logging/aggregate/index.js
```diff
@@ -123,4 +147,14 @@ export class Aggregate extends AggregateBase {
   queryStringsBuilder () {
     return { browser_monitoring_key: this.agentRef.info.licenseKey }
   }
+
+  /** Abort the feature, once aborted it will not resume */
+  abort (reason = {}) {
+    handle(SUPPORTABILITY_METRIC_CHANNEL, [`Logging/Abort/${reason.sm}`], undefined, FEATURE_NAMES.logging, this.ee)
+    this.blocked = true
+    this.loggingMode = LOGGING_MODE.OFF
+    this.syncWithSessionManager({ loggingMode: this.loggingMode })
```
metal-messiah, line 156: This call will also need to check if session manager exists, so it may make more sense for this method to do the check itself

metal-messiah: nvm, i see that you already did that 👍

src/features/logging/aggregate/index.js
```diff
@@ -10,21 +10,45 @@ import { Log } from '../shared/log'
...
+      if (this.loggingMode === LOGGING_MODE.OFF || (session.isNew && loggingMode === LOGGING_MODE.OFF)) {
         this.blocked = true
         this.deregisterDrain()
         return
       }
-      this.loggingMode = loggingMode
+      if (session.isNew || !this.isSessionTrackingEnabled) {
+        this.loggingMode = loggingMode
+
+        if (this.isSessionTrackingEnabled) {
```
metal-messiah, line 44: this if statement is redundant now (since syncWith.. does the same check), but nbd

metal-messiah (approved):

src/features/logging/aggregate/index.js
ptang-nr: yeah, i wasn't sure if it's worth blocking the extra call so added it.  Sounds like probably not worth it (?) I can avoid doing that in the future.

PR #1267, opened 2024-12-02
feat: Capture Nearest UserAction Fields

metal-messiah:
The UserAction feature has been extended to capture "nearest" field values when the UserAction target does not include the desired fields. To facilitate this, a configuration was added allowing the user to control what fields are collected, specified as `init.user_actions.elementAttributes: []`.  When empty, no extra fields will be collected.  When supplied, the target will be checked for each property. This configuration defaults to `['id', 'className', 'tagName', 'type']`, which matches the fields captured before this change. 

For each elementAttribute defined, if the property is not found, the agent will search up the parent tree of the target element, until the property is found.  If found up the parent tree, a property called "nearest" + fieldName will be added to the event.
---

### Overview
This was spawned from feedback and POC tied to internal NR1 use cases and product analytics.  Exposing to customer base to benefit all.

### Related Issue(s)
[NR-359755](https://new-relic.atlassian.net/browse/NR-359755)

### Testing
Test has been added

[NR-359755]: https://new-relic.atlassian.net/browse/NR-359755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ellisong (approved):
LGTM

PR #1282, opened 2024-12-12
feat: allow feature flags to control experimental features

metal-messiah:
Allow enablement of experimental features to be controlled by internal feature flags to support lower effort requirements from customers and support of APM-injected installations.
---

### Overview
This PR adds a check for feature flags so that experimental datasources can be triggered to on by internal feature flags that require less intervention from customers and supports APM injected environments.  The feature flag will enable the feature, even if the `enabled` value in the configuration is false.  Please see https://featureflag.datanerd.one/categories/968 for more info on the feature flags being applied.

### Testing
New tests have been added to the `/ins` e2e path that validate that feature flags can control the behavior


src/features/generic_events/instrument/index.js
```diff
@@ -7,22 +7,17 @@ import { handle } from '../../../common/event-emitter/handle'
...
-    /** config values that gate whether the generic events aggregator should be imported at all */
-    const genericEventSourceConfigs = [
-      agentRef.init.page_action.enabled,
-      agentRef.init.performance.capture_marks,
-      agentRef.init.performance.capture_measures,
-      agentRef.init.user_actions.enabled,
-      agentRef.init.performance.resources.enabled
-    ]
+
+    const { datasources, anyEnabled } = getEnabledDatasources(this.agentRef.init)
```
ptang-nr, line 17: I like the introduction of the utils function to help manage the logic.  I am wondering if this is also introducing a shift in the usage pattern (very superficially).  Would it be possible to keep the logic "hidden" within the model (it looks like we're mostly checking against `init.feature_flags`)?  I also find it easier to grok if `enabled` stays the source of truth

metal-messiah: this is a good idea.  I'll hit you on slack and maybe you can expound on that a bit more

src/features/generic_events/shared/utils.js
```diff
@@ -0,0 +1,32 @@
...
+  if (captureMarksEnabled) anyEnabled = true
+  const captureMeasuresEnabled = init.performance.capture_measures || init.feature_flags.includes(FEATURE_FLAGS.MEASURES) // experimental stage
+  if (captureMeasuresEnabled) anyEnabled = true
+  const pageActionsEnabled = init.page_action.enabled
+  if (pageActionsEnabled) anyEnabled = true
+  const resourcesEnabled = init.performance.resources.enabled || init.feature_flags.includes(FEATURE_FLAGS.RESOURCES) // experimental stage
+  if (resourcesEnabled) anyEnabled = true
+
+  return {
+    anyEnabled,
```
ptang-nr, line 23: Nit: I think simplify a bit to something like `anyEnabled: userActionsEnabled || the next feat, etc || ... `

tests/specs/ins/harvesting.e2e.js
```diff
@@ -333,102 +365,99 @@ describe('ins harvesting', () => {
...
-        timestamp: expect.any(Number),
-        transferSize: expect.any(Number),
-        workerStart: expect.any(Number)
-      }
-      const chromeFields = {
-        deliveryType: expect.any(String),
-        firstInterimResponseStart: expect.any(Number),
-        renderBlockingStatus: expect.any(String),
-        serverTiming: expect.any(String),
-        responseStatus: expect.any(Number)
```
ptang-nr, line 414: asking because I don't know - did `responseStatus` get dropped?

ptang-nr (approved):
LGTM!

PR #1283, opened 2024-12-16
chore: Fix gha tests

metal-messiah:
v22.12.0 was shown to cause issues with our testing server architecture.  Pinning all GHA runs to 22.11.0, which is the last known working version of Node with current testing architecture patterns. Please note the devEngines field in package.json.  This will force the dev env to use the same node version and install pattern that our testing server uses in GHA.  This is aimed at reducing discrepancies between local testing and GHA testing
---


ellisong (approved):
LGTM

PR #1298, opened 2024-12-24
feat: Centralized harvesting

cwli24:
Introduces a singleton harvester tasked with sending feature payloads rather than each feature utilizing their own scheduler. This reduces runtime space usage and the number of timers which proportionally scales as the number of features increases.
---

### Overview

This is a major restructuring affecting all features and the medium through which they harvest. It also add the capability for some features to buffer data under different apps and have those partitions be part of the harvest process as well.

Roughly, the data model has gone from
`[<feature>.each] -> [harvest-scheduler.js] -> [harvest.js]`
to
`[harvester.js]-->[<feature.each>]`

In addition, a layer has been added between each feature and its storage unit to manage multiple app data. In particular, those will now have the following model:
`[<feature>] -> [EventStoreManager] -> [EventBuffer]/[EventAggregator]`

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-340102
https://new-relic.atlassian.net/browse/NR-340106


src/common/harvest/harvester.js
```diff
@@ -0,0 +1,247 @@
...
+
+    /* Flush all buffered data if session resets and give up retries. This should be synchronous to ensure that the correct `session` value is sent.
+      Since session-reset generates a new session ID and the ID is grabbed at send-time, any delays or retries would cause the payload to be sent under the wrong session ID. */
+    agentRef.ee.on(SESSION_EVENTS.RESET, () => this.initializedAggregates?.forEach(aggregateInst => this.triggerHarvestFor(aggregateInst, { forceNoRetry: true })))
+  }
+
+  #startTimer (harvestInterval) {
+    const onHarvestInterval = () => {
+      this.initializedAggregates.forEach(aggregateInst => this.triggerHarvestFor(aggregateInst))
+      setTimeout(onHarvestInterval, harvestInterval * 1000) // repeat in X seconds
```
ptang-nr, line 48: Could this mean if (in theory, for whatever reason) `send` took a long time, that we could be accumulating data for longer than the set harvest interval? 🤔

cwli24: right now, per harvest, multiple send goes out, and none of it is tracked or chained on, so each harvest is treated independently. This brings up the concern that their event buffer is shared (primarily around how the current save & reloadsave mechanism works) if one-at-a-time requests isn't enforced, but no, there's no "accumulation" going on with this model.

src/common/harvest/harvester.js
```diff
@@ -0,0 +1,247 @@
...
+          case 429:
+          case 408:
+          case 500:
+          case 503:
+            cbResult.retry = true
+            break
+        }
+        if (localOpts.needResponse) {
+          cbResult.responseText = this.responseText
+        }
```
metal-messiah, line 166: consider wrapping this in a helper function since it is duplicated below

src/common/harvest/harvester.js
metal-messiah: I dont think the interval is considering response time, simply every x seconds (with inconsistencies with JS timers) it will try to send a new payload

src/features/logging/aggregate/index.js
```diff
@@ -10,26 +9,19 @@ import { Log } from '../shared/log'
...
-        retryDelay: this.harvestTimeSeconds,
-        getPayload: (options) => this.makeHarvestPayload(options.retry),
-        raw: true
-      }, this)
       /** emitted by instrument class (wrapped loggers) or the api methods directly */
       registerHandler(LOGGING_EVENT_EMITTER_CHANNEL, this.handleLog.bind(this), this.featureName, this.ee)
       this.drain()
       /** harvest immediately once started to purge pre-load logs collected */
-      this.scheduler.startTimer(this.harvestTimeSeconds, 0)
+      agentRef.runtime.harvester.triggerHarvestFor(this)
```
metal-messiah, line 24: i wonder if this should be the expected default behavior for all the features, since pre-load data is very important.

cwli24: Hmm, seems like a product choice really. Not all features certainly have things to send on load, but it also doesn't hurt to add that check if it's to be done. Let's ask if Cheryl has an opinion next Monday

src/common/harvest/harvester.js
ptang-nr: Ty both, I think it's a lot clearer now!  I was mostly checking my understanding - that using `setTimeout` enforces the 30 secs intervals between harvest triggers.  The part I got mixed up on -- in reality, we're not waiting on `send` so there's not really a scenario where we'd get held up and send more than 30secs worth of data at a time.

metal-messiah (approved):

PR #1300, opened 2024-12-27
chore: Finalize changes to build size action

metal-messiah:
Track build sizes using shared GHA
---
Track build sizes for:
* Pull request builds
* merges with main branch
* nightly checks
* new releases


.github/workflows/nightly.yml
```diff
@@ -14,6 +14,46 @@ jobs:
...
+          node-version: 22.11.0 # See package.json for the stable node version that works with our testing.  Do not change this unless you know what you are doing as some node versions do not play nicely with our testing server.
+      - name: Install project dependencies
+        run: npm ci
+      - name: Report lite size to NR
+        uses: newrelic/capture-build-size@main
+        with:
+          nr-api-key: ${{ secrets.INTERNAL_STAGING_INGEST_LICENSE_KEY }}
+          nr-account-id: ${{ secrets.INTERNAL_STAGING_INGEST_ACCOUNT_ID }}
+          nr-env: 'staging'
+          analysis-file-url: 'https://js-agent.newrelic.com/dev/nr-rum-standard.stats.json'
```
cwli24, line 35: Didn't make immediate sense to me why this grabs the stat file from CDN when there wasn't a build & publish step preceding this. While nightly runs on `main` which should be updated by publish-dev, I think this is either unclear or redundant entirely. Put another way, since publish-dev runs on all `main` commits and does this already, we don't need nightly to do this.

.github/workflows/publish-release.yml
```diff
@@ -61,6 +61,46 @@ jobs:
...
+      - name: Get version number
+        id: agent-loader-version
+        run: echo "results=$(cat package.json | jq -r '.version')" >> $GITHUB_OUTPUT
+      - name: Get local stats
+        id: get-stats
+        shell: bash
+        run: |
+          rum=$(cat ./build/nr-rum-standard.stats.json); echo "rum=$rum" >> $GITHUB_OUTPUT;
+          full=$(cat ./build/nr-full-standard.stats.json); echo "full=$full" >> $GITHUB_OUTPUT;
+          spa=$(cat ./build/nr-spa-standard.stats.json); echo "spa=$spa" >> $GITHUB_OUTPUT;
```
cwli24, line 73: The prod build has versions in the name!
image

.github/workflows/nightly.yml
cwli24, line 1: Are you using the `github.ref_name` aka "version" as the tag being reported in NR? ("main", "1293/merge")
Are you planning to visually display across versions via using the `trigger` ? Or can you easily distinguish where the version bumps are on the chart via file name that includes version?

.github/workflows/publish-release.yml
metal-messiah: ty for catching that

.github/workflows/nightly.yml
metal-messiah: prs will be named 1293/merge, the main branch will always be "main" and the releases will be `1.274.0` or whatever.  I think thats enough of a visual breakdown for our case, and the commit SHAs will be there if we need to dig into a particular change.  It will also create a deploy marker

.github/workflows/nightly.yml
metal-messiah: this was mainly to keep the dashboard hydrated at least once per day but is mainly unneeded if we dont feel like doing that

.github/workflows/publish-release.yml
```diff
@@ -61,6 +61,46 @@ jobs:
           fastly_key: ${{ secrets.FASTLY_PURGE_KEY }}
           fastly_service: staging-js-agent.newrelic.com
           asset_path: ${{ join(fromJson(steps.s3-upload.outputs.results).*.Key, ' ') }}
+      - name: Get version number
```
cwli24, line 64: Can this be spun off into a separate job in this flow? Publish-release is too important and rarely ran, and rest below (publish-current-to-s3) depends on this job. Just about to cut a release now and would rather not deal with this blocking the pipeline from running, to be selfish.

cwli24 (approved):

PR #1311, opened 2025-01-08
fix: Make Metrics harvest only on EoL for new Harvester

cwli24:
Prevent Metrics feature from harvesting on interval after changes in v1.278.0. This reverts its behavior back to sending only final harvests.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-356543

### Testing

New component test added for this case


metal-messiah (approved):
Approved once tests run

PR #1315, opened 2025-01-09
fix: Allow the page view feature to have access to an event buffer to prevent undefined checks at harvest time

metal-messiah:
Fix an issue where if the page unloaded before the page view event could "block" itself, an undefined error could be thrown. The page view feature now has access to the event buffer, to prevent that undefined check.  The event buffer will remain empty, since the page view feature does not carry state. This change rectifies an issue observed in 1.278.1.
---

### DIFF
--- a/tests/unit/features/utils/aggregate-base.test.js
+++ b/tests/unit/features/utils/aggregate-base.test.js
@@ -150,16 +150,16 @@ test('does not initialized Aggregator more than once with multiple features', as
   expect(mainAgent.mainAppKey).toBeUndefined()
 
   new AggregateBase(mainAgent, FEATURE_NAMES.pageViewEvent)
-  expect(EventStoreManager).toHaveBeenCalledTimes(1)
+  expect(EventStoreManager).toHaveBeenCalledTimes(2)
   expect(EventStoreManager).toHaveBeenCalledWith(mainAgent.mainAppKey, 2) // 2 = initialize EventAggregator
   expect(mainAgent.mainAppKey).toBeTruthy()
   expect(mainAgent.sharedAggregator).toBeTruthy()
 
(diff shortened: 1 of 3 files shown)


tests/unit/features/utils/aggregate-base.test.js
```diff
@@ -150,16 +150,16 @@ test('does not initialized Aggregator more than once with multiple features', as
   expect(mainAgent.mainAppKey).toBeUndefined()
 
   new AggregateBase(mainAgent, FEATURE_NAMES.pageViewEvent)
-  expect(EventStoreManager).toHaveBeenCalledTimes(1)
+  expect(EventStoreManager).toHaveBeenCalledTimes(2)
```
cwli24, line 153:
```suggestion
  expect(EventStoreManager).toHaveBeenCalledTimes(2) // once for runtime.sharedAgg + once for PVE.events
```

tests/unit/features/utils/aggregate-base.test.js
```diff
@@ -175,10 +175,9 @@ test('does initialize separate Aggregators with multiple agents', async () => {
 
   new AggregateBase(mainAgent, FEATURE_NAMES.pageViewEvent)
   new AggregateBase(mainAgent2, FEATURE_NAMES.pageViewEvent)
-  expect(EventStoreManager).toHaveBeenCalledTimes(2)
-  expect(EventStoreManager).not.toHaveBeenCalledWith(expect.any(Object), 1)
+  expect(EventStoreManager).toHaveBeenCalledTimes(4)
```
cwli24, line 178:
```suggestion
  expect(EventStoreManager).toHaveBeenCalledTimes(4) // runtime.sharedAgg + PVE.events but multiply by having 2 agents
```

cwli24 (approved):

PR #1319, opened 2025-01-11
feat: Remove FID

ptang-nr:
First Input Delay (FID) has been deprecated and replaced by Interaction To Next Paint (INP).   PageViewTiming events for first interactions are preserved by detecting the first `first-input` event  and will no longer output value for the `firstInputDelay` and `loadState` attributes.  For more info on PageViewTiming, see https://docs.newrelic.com/docs/browser/new-relic-browser/page-load-timing-resources/pageviewtiming-async-or-dynamic-page-details/.  For more info on INP and the FID deprecation, see https://web.dev/articles/inp.
---

### Overview

In this PR, we are proceeding with the removal of FID related logic from the browser agent.

Notes:
- The `firstInteraction` timing event should continue to be detected as-is, with the exception of no longer collecting or generating values for `firstInputDelay` and `loadState` attributes.  Supported in Chrome, Edge, Firefox and Android.  Not supported on Safari and iOS (see [PerformanceEventTming browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEventTiming#browser_compatibility))

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-355558

### Testing

Updated existing tests that dealt with FID.
Added unit tests for first interaction.
Manually tested 
- Detection of first input on Chrome
- No first input detected when page is opened in background on Chrome


src/common/vitals/first-interaction.js
metal-messiah, line 1: Im confused as to what this file is doing.  Is anything in the agent referencing this value?

ptang-nr: To round out the thread - we had a team discussion and the main reason for this is to preserve the `firstInteraction` timing event for customers.  (this is just me thinking - maybe we can revisit removing first interaction support as a separate effort.  My impression from the team discussion is we want to de-support this)

cwli24: I thought the outcome was to use INP, not revert FI into custom-rolled implementation

metal-messiah: I think there were findings with PM that INP support was too limited due to browser support

src/common/vitals/first-interaction.js
```diff
@@ -1,24 +1,34 @@
...
-    })
+        observer.disconnect()
+        if (!firstInteraction.isValid) {
+          firstInteraction.update({
+            value: firstInput.startTime,
+            attrs
+          })
+        }
+      })
+      observer.observe({ type: PERFORMANCE_ENTRY_TYPE.FIRST_INPUT })
```
metal-messiah, line 29: I think this probably needs to include the buffered argument to capture cases where an interaction has occurred before this module was executed

metal-messiah (approved):

PR #1326, opened 2025-01-13
feat: debounce window user actions

metal-messiah:
Debounce focus and blur events on the window for 500ms to prevent duplicate events from being harvested. All major browsers have their own quirky behaviors that generate duplicate blur and focus events under some conditions. This change ensures that only 1 blur or focus event can be captured every 500ms to reduce noise.
---

### Related Issue(s)
https://newrelic.slack.com/archives/G019T80B1EJ/p1736789211555589
https://stackoverflow.com/questions/62939627/blur-focus-event-getting-triggered-multiple-times

### Testing
A new test has been added that spams focus and blur events, but expects only 1 to be reported.


tests/specs/ins/harvesting.e2e.js
```diff
@@ -178,6 +178,25 @@ describe('ins harvesting', () => {
...
+        let i = 0; while (i++ < 10) {
+          window.dispatchEvent(new Event('focus'))
+          window.dispatchEvent(new Event('blur'))
+        }
+      })
+
+    ])
+
+    const userActionsHarvest = insHarvests.flatMap(harvest => harvest.request.body.ins) // firefox sends a window focus event on load, so we may end up with 2 harvests
+    expect(userActionsHarvest.filter(ua => ua.action === 'focus').length).toEqual(1)
```
cwli24, line 197: Should also check that there's 1 `blur` as well, and that both of them only have an array ms of length 1 too.

metal-messiah: ty

cwli24 (dismissed):

cwli24 (approved):

cwli24 (approved):

PR #1327, opened 2025-01-14
fix: Change obfuscator to read dynamically from configuration

metal-messiah:
Fix an issue where obfuscation rules are not honored if added after the agent has initialized.  Some user cases require obfuscation rules to be dynamically added, this change will allow that behavior -- as long as the obfuscation rules are added to `newrelic.initializedAgents[index].init.obfuscate`
---

### Related Issue(s)
This is related to an ongoing incident

### Testing
A new test has been added that validates that obfuscation rules can be added later and still affect outgoing payloads.

### DIFF
--- a/src/common/util/obfuscate.js
+++ b/src/common/util/obfuscate.js
@@ -22,6 +22,9 @@ import { warn } from './console'
 export class Obfuscator {
   constructor (agentRef) {
     this.agentRef = agentRef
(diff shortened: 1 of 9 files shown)


ptang-nr (approved):

ptang-nr (approved):
Thanks Jordan for tracking all of this down 🙏

PR #1331, opened 2025-01-15
feat: Add HTTP codes for harvest retry attempts

cwli24:
Agent will begin to retry harvests to the endpoints that failed for the following additional status codes: 502, 504, and 512-530. This is suggested by New Relic ingest teams as a possibility to recapture lost data in these marginal cases.
---

### Overview

Harvester now will attempt to retry the following codes in total: 408, 429, 500, 502-504, 512-530.
Metrics track if a former failed attempt subsequently succeeded or failed on next attempt, split by the former returned HTTP code.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-137499
Angler update: https://source.datanerd.us/agents/angler/pull/657

### Testing

Retry e2es were updated. New SMs are tested in the PVT retry spec file (only).

### DIFF
--- a/src/common/harvest/harvester.js
+++ b/src/common/harvest/harvester.js
@@ -1,6 +1,8 @@
-import { FEATURE_TO_ENDPOINT, JSERRORS, RUM, EVENTS } from '../../loaders/features/features'
+import { SUPPORTABILITY_METRIC_CHANNEL } from '../../features/metrics/constants'
+import { FEATURE_TO_ENDPOINT, JSERRORS, RUM, EVENTS, FEATURE_NAMES } from '../../loaders/features/features'
 import { VERSION } from '../constants/env'
 import { globalScope, isWorkerScope } from '../constants/runtime'
+import { handle } from '../event-emitter/handle'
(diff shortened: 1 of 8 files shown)


src/common/harvest/harvester.js
```diff
@@ -160,14 +172,12 @@ function send (agentRef, { endpoint, targetApp, payload, localOpts = {}, submitM
...
-      case 429:
       case 408:
+      case 429:
       case 500:
-      case 503:
         return true
-      default:
-        return false
     }
+    return (status >= 502 && status <= 504) || (status >= 512 && status <= 530)
```
metal-messiah, line 180: This is a nit pick, feel free to ignore, but something feels weird to read for the mixing of the switch statement and the other check at the bottom.  It works so do what you want with that comment

metal-messiah (approved):

src/common/harvest/harvester.js
cwli24: I know it sucks because listing out every code in this range would make the switch really large, but adding the single codes to an OR statement or if-else block would be very wordy too. Imma kick this can down

PR #1332, opened 2025-01-16
feat: Capture details in marks and measures

metal-messiah:
The `detail` property of `PerformanceMark` and `PerformanceMeasure` events will now be captured in New Relic `BrowserPerformance` under the `entryDetail` attribute.  If `detail` is an object, it and any nested objects within it will be flatten into string path keys prefixed with `entryDetail`. See [the PR](https://github.com/newrelic/newrelic-browser-agent/pull/1332) for elucidation.
---

### Overview
This PR is to support changes requested by the platform team to be utilized in conjunction with MFE POC.

Flattening examples:

- `detail: { foo: { bar: 'baz' }, other: 'hello' }` will create an attribute on the BrowserPerformance event called `entryDetail.foo.bar: 'baz'` and another attribute called `entryDetail.other: 'hello'`. 
- Other non-object data types provided to the `detail` option will be processed as-is.  `detail: 'hello'` will create an attribute called `entryDetail: 'hello'`. 

### Testing
 tests have been added/updated

### DIFF
--- a/src/features/generic_events/aggregate/index.js
+++ b/src/features/generic_events/aggregate/index.js
@@ -133,7 +133,6 @@ export class Aggregate extends AggregateBase {
 
                     function createDetailAttrs (obj, rootKey = '', output = {}) {
                       if (obj === null || obj === undefined) return output
-                      // if (typeof obj === 'string') output.entryDetail = obj
                       else if (typeof obj !== 'object') output.entryDetail = obj
                       else {
                         Object.entries(obj).forEach(([key, value]) => {
--- a/src/features/generic_events/aggregate/index.js
+++ b/src/features/generic_events/aggregate/index.js
@@ -15,6 +15,7 @@ import { applyFnToProps } from '../../../common/util/traverse'
 import { UserActionsAggregator } from './user-actions/user-actions-aggregator'
 import { isIFrameWindow } from '../../../common/dom/iframe'
 import { handle } from '../../../common/event-emitter/handle'
+import { isPureObject } from '../../../common/util/type-check'
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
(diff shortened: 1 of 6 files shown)


src/features/generic_events/aggregate/index.js
```diff
@@ -121,14 +121,27 @@ export class Aggregate extends AggregateBase {
...
                       timestamp: this.toEpoch(entry.startTime),
                       entryName: cleanURL(entry.name),
                       entryDuration: entry.duration,
-                      entryType: type,
-                      ...(entry.detail && { entryDetail: entry.detail })
+                      entryType: type
                     })
+
+                    function createDetailAttrs (obj, rootKey = '', output = {}) {
+                      if (obj === null || obj === undefined) return output
```
cwli24, line 135: This is a quite a convoluted method that uses an arg as output but also returns an output. Are you trying to flatten the objected into a single level of key paths and values?

cwli24:
Reminder: Fix typo in description and change/reorganize it for comprehension

src/features/generic_events/aggregate/index.js
metal-messiah: the DBC does not flatten a JSON string obj to attributes.  It will report it as a string

This helper function essentially turns
`{foo1: {foo2: {foo3: 'bar3'}, test: 'sup'}`
into
`entryDetail.foo1.foo2.foo3: 'bar3'`
`entryDetail.test: 'sup'`

cwli24: Ok, right. So the way `output` is both an IO arg and a returned val makes this wonky. Arrays are also not considered here so you can end up with `.foo1.0` or `foo1.1`.

Here's a suggested edit:
```js
function flattenJSON(nestedJSON, parentKey = 'entryDetail') {
  let items = {}
  Object.keys(nestedJSON).forEach(key => {
    let newKey = parentKey + '.' + key;
    if (typeof nestedJSON[key] === 'object' && nestedJSON[key] !== null && !Array.isArray(nestedJSON[key])) {
      Object.assign(items, flattenJSON(nestedJSON[key], newKey))
    } else {
      items[newKey] = nestedJSON[key];
    }
  })
  return items
}
```

cwli24 (approved):

PR #1344, opened 2025-01-24
chore: Support experiment uploads without deploy to internal site

ptang-nr:
Modify the existing `publish-experiment` workflow to support uploading experiments without deploying to internal site. This will come in handy for cases where we may wish to make experimental builds available on the CDN, to be used outside of internal experiments.
---


metal-messiah:
I think this PR will need to update `publish-dev` to also consider if the experiment was meant to make it to NR1. Wait to merge this until more validation is done

metal-messiah:
> I think this PR will need to update `publish-dev` to also consider if the experiment was meant to make it to NR1. Wait to merge this until more validation is done

To expand on this, when `publish-dev` runs automatically on merges to main, it will check the experiments folder and upload them to the a/b script still. I think the `standalone` experiments like this need to go to a new directory in S3

.github/workflows/publish-experiment.yml
```diff
@@ -15,6 +15,12 @@ on:
         options:
           - dev
           - staging
+          - standalone # Note: this option prevents the experiment from being added to dev or staging envs. See the `publish-dev` workflow and `build-ab` action.
```
ptang-nr, line 18: hey @metal-messiah, thanks for pointing out the gap with the `publish-dev` workflow!  I tried to make the s3 path dynamic but it didn't quite pan out.  I'm thinking adding a "standalone" folder will be the smallest change to get what we want.  What do you think?

metal-messiah: yeah i think this is the best outcome!

metal-messiah (approved):

PR #1359, opened 2025-01-31
fix: Roll back to previous FirstInteraction implementation

metal-messiah:
Setting agent back to previous FirstInteraction implementation after some reports of slow downs around the performance observer on the first input of a page.
---

### Overview
This is being rolled-back until we can evaluate and implement next steps.

### Related Issue(s)
This undos the changes in #1319 

### Testing
Tests have been reverted.


ellisong (approved):
LGTM

PR #1369, opened 2025-02-05
chore: Logs CI update

metal-messiah:
This updates the logs CI to facilitate logs testing in NR.  To faciliate that testing, we are removing the manual wrapping calls, and strictly relying on auto-wrapping to capture the log data.
---


ellisong (approved):
LGTM

PR #1380, opened 2025-02-11
chore: Add typing definition to Agent constructor options

cwli24:
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-362070
https://new-relic.atlassian.net/browse/NR-362072

### DIFF
--- a/src/common/config/info.js
+++ b/src/common/config/info.js
@@ -5,6 +5,27 @@
 import { defaults as nrDefaults, getNREUMInitializedAgent } from '../window/nreum'
 import { getModeledObject } from './configurable'
 
+/**
+ * @typedef {Object} Info
+ * @property {string} [beacon]
+ * @property {string} [errorBeacon] - Base URL endpoint for all data harvested by the agent. Proxies should be defined in the init instead.
+ * @property {string} licenseKey - New Relic license key provided by the website in user account.
+ * @property {string} applicationID - New Relic application ID provided when creating a browser entity in the UI.
+ * @property {number} [sa]
+ * @property {number} [queueTime]
+ * @property {number} [applicationTime]
+ * @property {string} [ttGuid]
+ * @property {string} [user]
+ * @property {string} [account]
+ * @property {string} [product]
+ * @property {string} [extra]
+ * @property {Object} [jsAttributes] - Custom attributes that are added to majority of agent's payloads. The `setCustomAttribute` API method affects this.
+ * @property {string} [userAttributes]
+ * @property {string} [atts]
+ * @property {string} [transactionName]
+ * @property {string} [tNamePlain]
+ */
+
 const model = {
   // preset defaults
   beacon: nrDefaults.beacon,
@@ -20,9 +41,7 @@ const model = {
   account: undefined,
   product: undefined,
   extra: undefined,
-  jsAttributes: {
-
-  },
+  jsAttributes: {},
   userAttributes: undefined,
   atts: undefined,
   transactionName: undefined,
--- a/src/common/config/init.js
+++ b/src/common/config/init.js
@@ -68,7 +68,7 @@ import { getModeledObject } from './configurable'
  * @property {boolean} [session_replay.enabled] - Turn on/off the session replay feature (off by default).
  * @property {boolean} [session_replay.preload] - If true, allow the agent to import and run rrweb recorder immediately instead of waiting until after the window.load event, for new sessions. Existing sessions ignore this setting.
  * @deprecated {number} [session_replay.sampling_rate] - This setting is deprecated and ineffective. Sampling is controlled in New Relic by server-side configuration.
- * @property {number} [session_replay.error_sampling_rate] - This setting is deprecated and ineffective.
+ * @deprecated {number} [session_replay.error_sampling_rate] - This setting is deprecated and ineffective.
  * @property {boolean} [session_replay.collect_fonts] - When true, serialize fonts for collection without public asset url. This is currently broken -- https://github.com/rrweb-io/rrweb/issues/1304.
  * @property {boolean} [session_replay.inline_images] - When true, serialize images for collection without public asset url. Not recommended for use. This is currently for TESTING as it easily generates payloads too large to be harvested.
  * @property {boolean} [session_replay.fix_stylesheets] - When true, tries to fetch any missing stylesheets again to inline in replayer.
--- a/src/common/config/init.js
+++ b/src/common/config/init.js
@@ -69,7 +69,7 @@ import { getModeledObject } from './configurable'
  * @property {boolean} [session_replay.preload] - If true, allow the agent to import and run rrweb recorder immediately instead of waiting until after the window.load event, for new sessions. Existing sessions ignore this setting.
  * @deprecated {number} [session_replay.sampling_rate] - This setting is deprecated and ineffective. Sampling is controlled in New Relic by server-side configuration.
  * @deprecated {number} [session_replay.error_sampling_rate] - This setting is deprecated and ineffective.
- * @property {boolean} [session_replay.collect_fonts] - When true, serialize fonts for collection without public asset url. This is currently broken -- https://github.com/rrweb-io/rrweb/issues/1304.
+ * @property {boolean} [session_replay.collect_fonts] - When true, serialize fonts for collection without public asset url. See issues with this feature at https://github.com/rrweb-io/rrweb/issues/1304.
  * @property {boolean} [session_replay.inline_images] - When true, serialize images for collection without public asset url. Not recommended for use. This is currently for TESTING as it easily generates payloads too large to be harvested.
  * @property {boolean} [session_replay.fix_stylesheets] - When true, tries to fetch any missing stylesheets again to inline in replayer.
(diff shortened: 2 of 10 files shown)


src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {boolean} [logging.enabled] - Turn on/off the logging feature (on by default).
+ * @property {boolean} [logging.autoStart] - If true, the agent will automatically start the logging feature. Otherwise, it will be in a deferred state until the `start` API method is called.
+ * @property {Object} [metrics]
+ * @property {boolean} [metrics.enabled] - Turn on/off the metrics feature (on by default).
+ * @property {boolean} [metrics.autoStart] - If true, the agent will automatically start the metrics feature. Otherwise, it will be in a deferred state until the `start` API method is called.
+ * @property {Array<Object>} [obfuscate] - Array of regexp and corresponding replacement patterns for obfuscating data.
+ * @property {Object} [page_action]
+ * @property {boolean} [page_action.enabled] - Must be true to allow PageAction events to be captured.
+ * @property {Object} [page_view_event]
+ * @property {boolean} [page_view_event.enabled] - This setting is ignored! PageViewEvent is always enabled by force.
```
metal-messiah, line 44: we should discuss if these types of configs (not intended for public mutation) are worth defining

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {boolean} [page_view_timing.enabled] - Turn on/off the page view timing feature (on by default).
+ * @property {boolean} [page_view_timing.autoStart] - If true, the agent will automatically start the page view timing feature. Otherwise, it will be in a deferred state until the `start` API method is called.
+ * @property {Object} [performance]
+ * @property {boolean} [performance.capture_marks] - If true, the agent will capture PerformanceMark events.
+ * @property {boolean} [performance.capture_measures] - If true, the agent will capture PerformanceMeasure events.
+ * @property {boolean} [performance.capture_detail] - If true, `BrowserPerformance` events from marks and measures will include, as attribute(s), the `detail` metadata provided to `markOptions` and `measureOptions`.
+ * @property {Object} [performance.resources]
+ * @property {boolean} [performance.resources.enabled] - If true, the agent will capture PerformanceResourceTiming entries.
+ * @property {Array<string>} [performance.resources.asset_types] - Array of `initiatorType` strings to filter the desired ResourceTiming entries. By default, all resource types are captured.
```
metal-messiah, line 55: This should note that an empty array also captures all types

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
 import { isValidSelector } from '../dom/query-selector'
 import { DEFAULT_EXPIRES_MS, DEFAULT_INACTIVE_MS } from '../session/constants'
 import { warn } from '../util/console'
 import { getNREUMInitializedAgent } from '../window/nreum'
 import { getModeledObject } from './configurable'
 
-const nrMask = '[data-nr-mask]'
+/**
+ * @typedef {Object} Init
+ * @property {Object} [ajax]
```
metal-messiah, line 14: Almost every property in this object is optional, so for almost all of these types, they should be noted as optional

`Object` --> `Object=`

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {string} [proxy.beacon] - Set value will be used to overwrite the endpoint URL to which we send analytics.
+ * @property {Object} [session]
+ * @property {number} [session.expiresMs] - When session tracking is on, this determines how long a session will last before expiring. Modifying this value is not recommended.
+ * @property {number} [session.inactiveMs] - When session tracking is on, this determines how long a session will last without user activity before expiring. Modifying this value is not recommended.
+ * @property {Object} [session_replay]
+ * @property {boolean} [session_replay.autoStart] - If true, the agent will automatically start the session replay feature. Otherwise, it will be in a deferred state until the `start` API method is called.
+ * @property {boolean} [session_replay.enabled] - Turn on/off the session replay feature (off by default).
+ * @property {boolean} [session_replay.preload] - If true, allow the agent to run rrweb recorder immediately instead of waiting until after the window.load event, for new sessions. Existing sessions ignore this setting.
```
metal-messiah, line 69:
```suggestion
 * @property {boolean} [session_replay.preload] - If true, allow the agent to import and run rrweb recorder immediately instead of waiting until after the window.load event, for new sessions. Existing sessions ignore this setting.
```

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {number} [session_replay.sampling_rate] - This setting is deprecated and ineffective. Sampling is controlled in New Relic by server-side configuration.
```
metal-messiah, line 70:
```suggestion
 * @deprecated {number} [session_replay.sampling_rate] - This setting is deprecated and ineffective. Sampling is controlled in New Relic by server-side configuration.
```

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {number} [session_replay.error_sampling_rate] - This setting is deprecated and ineffective.
```
metal-messiah, line 71:
```suggestion
 * @deprecated {number} [session_replay.error_sampling_rate] - This setting is deprecated and ineffective.
```

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {boolean} [session_replay.collect_fonts] - When true, serialize fonts for collection without public asset url. This is currently broken -- https://github.com/rrweb-io/rrweb/issues/1304.
```
metal-messiah, line 72:
```suggestion
 * @property {boolean} [session_replay.collect_fonts] - When true, serialize fonts for collection without public asset url. See issues with this feature at https://github.com/rrweb-io/rrweb/issues/1304.
```

src/common/config/init.js
```diff
@@ -3,15 +3,95 @@
...
+ * @property {boolean} [soft_navigations.autoStart] - If true, the agent will automatically start the soft navigations feature. Otherwise, it will be in a deferred state until the `start` API method is called.
+ * @property {Object} [spa]
+ * @property {boolean} [spa.enabled] - Turn on/off the single page application feature (on by default). NOTE: the SPA feature is deprecated and under removal procedure.
+ * @property {boolean} [spa.autoStart] - If true, the agent will automatically start the single page application feature. Otherwise, it will be in a deferred state until the `start` API method is called.
+ * @property {boolean} [ssl] - If explicitly false, the agent will use HTTP instead of HTTPS. This setting should NOT be used.
+ * @property {Object} [user_actions]
+ * @property {boolean} [user_actions.enabled] - Must be true to allow UserAction events to be captured.
+ * @property {Array<string>} [user_actions.elementAttributes] - List of HTML Element properties to be captured with UserAction events' target elements. This may help to identify the source element being interacted with in the UI.
```
metal-messiah, line 91:
```suggestion
 * @property {Array<string>} [user_actions.elementAttributes] - List of HTML Element properties to be captured with UserAction events' target or ancestor elements. This may help to identify the source element being interacted with in the UI.
```

metal-messiah:
Should also add tests for these types in the d.ts types test

src/common/config/info.js
```diff
@@ -5,6 +5,27 @@
...
+ * @property {string} [user]
+ * @property {string} [account]
+ * @property {string} [product]
+ * @property {string} [extra]
+ * @property {Object} [jsAttributes] - Custom attributes that are added to majority of agent's payloads. The `setCustomAttribute` API method affects this.
+ * @property {string} [userAttributes]
+ * @property {string} [atts]
+ * @property {string} [transactionName]
+ * @property {string} [tNamePlain]
+ */
```
metal-messiah, line 27: same goes for these, would be helpful to mark the optional ones vs required

src/common/config/init.js
cwli24: I think if it isn't used at all, it should be removed to begin with. The typings don't actually prevent people from plugging in the property anyways in JS. It's really just informational.

src/common/config/init.js
cwli24: They're already noted as optional by the `[]`. 
image

`=` is actually used to denote a default value. The by-effect makes it optional. It's more for `@param` function defs than type defs.

src/common/config/info.js
cwli24: Everything aside form licensekey & applicationId already marked as optional

src/common/config/init.js
cwli24:
```suggestion
 * @property {Array<string>} [performance.resources.asset_types] - Array of `initiatorType` strings to filter the desired ResourceTiming entries. By default, all resource types are captured, which is also the case if an empty array is provided.
```

src/common/config/init.js
metal-messiah: got it, didnt know about the [] in jsdoc, makes sense

src/common/config/init.js
cwli24: I had to revert this btw. Seems the @deprecated tag does not play nice with type definitions. It kept breaking the typing output (thereby breaking the testing of the custom options) to mark them as such, so I'm just going to leave them as @property and have the deprecated message.

metal-messiah (approved):

PR #1397, opened 2025-02-25
fix: Deduplicate Prefixes on Unhandled Promise Rejection Messages

metal-messiah:
Remove an issue that caused `Unhandled Promise Rejection:` prefixes on casted error messages to be duplicated if already supplied on rejection reason messages.
---

### Related Issue(s)
Screenshot 2025-02-25 at 11 39 41 AM

### Testing
A new unit test was added to validate that prefixes are not duplicated

### DIFF
--- a/src/features/jserrors/shared/cast-error.js
+++ b/src/features/jserrors/shared/cast-error.js
@@ -54,7 +54,7 @@ export function castPromiseRejectionEvent (promiseRejectionEvent) {
   }
 
   const error = castError(promiseRejectionEvent.reason)
-  if (!((error?.message || '').startsWith(prefix))) error.message = prefix + error?.message
+  if (!((error.message || '').startsWith(prefix))) error.message = prefix + error.message
   return error
 }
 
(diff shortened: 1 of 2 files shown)


src/features/jserrors/shared/cast-error.js
```diff
@@ -46,15 +46,15 @@ export function castPromiseRejectionEvent (promiseRejectionEvent) {
...
     } catch (e) {
-      return castError(promiseRejectionEvent.reason)
+      // failed to modify the message, do nothing else
     }
+    return castError(promiseRejectionEvent.reason)
   }
 
   const error = castError(promiseRejectionEvent.reason)
-  error.message = prefix + ': ' + error?.message
+  if (!((error?.message || '').startsWith(prefix))) error.message = prefix + error?.message
```
ptang-nr, line 57: I think `error` is defined at this point, so we can drop the `?`s

```suggestion
  if (!((error.message || '').startsWith(prefix))) error.message = prefix + error.message
```

ptang-nr (dismissed):

ptang-nr (approved):

PR #1410, opened 2025-03-10
fix: Obtain FirstInteraction directly from performance API

metal-messiah:
To alleviate the potential performance impact of creating new performance observers to support legacy FirstInteraction timings, a change has been made to query the performance API directly for the presence of first-input data without the use of an observer at logical checkpoints.
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-365650

### Testing
New tests and test fixes for e2e and unit


src/features/page_view_timing/aggregate/index.js
```diff
@@ -82,13 +84,36 @@ export class Aggregate extends AggregateBase {
...
       name,
       value,
       attrs
-    })
+    }
+    this.events.add(timing)
 
     handle('pvtAdded', [name, value, attrs], undefined, FEATURE_NAMES.sessionTrace, this.ee)
+
+    this.checkForFirstInteraction()
```
cwli24, line 96: Isn't this redundant if it's already being called during preharvest check?

ptang-nr: Also a bit curious about this call.  🤔 Wouldn't this check for fi on every timing event vs. at one point we talked about only checking for fi upon harvest (which happens more in batches).

tests/components/page_view_timing/index.test.js
```diff
@@ -47,6 +66,27 @@ describe('pvt aggregate tests', () => {
...
+            }]
+          })
+          setTimeout(callCb, 250)
+        }
+        setTimeout(callCb, 250)
+      },
+      disconnect: jest.fn()
+    }))
+    global.PerformanceObserver = mockPerformanceObserver
+    global.PerformanceObserver.supportedEntryTypes = ['first-input']
```
ptang-nr, line 88: This was my bad, you can drop this.  
```suggestion
```

src/features/page_view_timing/aggregate/index.js
metal-messiah: While redundant, it makes unit testing much easier to be able to skip the harvest cycle shenanigans.  Both approaches have an unknown number of times the hit the API (up to 6 or so times in the event driven model, potentially unlimited times in the pre-harvest check).  Once either of them finds a data source, they stop checking so its really whoever does it first, wins.  It's really not a performance burden either way imo.

ptang-nr (dismissed):

src/features/page_view_timing/aggregate/index.js
cwli24: Ok, well it's not major enough to matter, but I should say that prod code shouldn't be beholden to jest. I.e. the jest test should be re-written using mock harvester or fake timer instead of src code altered.

cwli24 (approved):

PR #1413, opened 2025-03-11
feat: Inspection events

ellisong:
Adds inspection events that detail when various agent-related events occur. These events can be listened to on the event listener 'newrelic' string type. Events include agent initialization, agent fully loaded, event buffered, event harvested, and api calls. 
---

### Overview

A test file is included to test whether the global dispatch events work at all. We can get more in-depth on the testing if desired but there are a _lot_ of events that happen after agent load.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-375506

### DIFF
--- a/src/common/event-emitter/contextual-ee.js
+++ b/src/common/event-emitter/contextual-ee.js
@@ -8,7 +8,6 @@ import { getOrSet } from '../util/get-or-set'
 import { getRuntime } from '../config/runtime'
 import { EventContext } from './event-context'
 import { bundleId } from '../ids/bundle-id'
-import { dispatchGlobalEvent } from '../dispatch/global-event'
 
 // create a unique id to store event context data for the current agent bundle
 const contextId = `nr@context:${bundleId}`
@@ -106,14 +105,6 @@ function ee (old, debugId) {
     var bufferGroup = getBuffer()[bufferGroupMap[type]]
     if (bufferGroup) {
       bufferGroup.push([emitter, type, args, ctx])
-      dispatchGlobalEvent({
-        agentIdentifier: emitter.debugId,
-        loaded: true,
-        type: 'data',
-        name: 'buffer',
-        feature: bufferGroupMap[type],
-        data: { type, args, force: !!force, bubble: !!bubble }
-      })
     }
 
     // Return the context so that the module that emitted can see what was done.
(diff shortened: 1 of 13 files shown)


src/common/event-emitter/contextual-ee.js
```diff
@@ -105,6 +106,14 @@ function ee (old, debugId) {
...
     if (bufferGroup) {
       bufferGroup.push([emitter, type, args, ctx])
+      dispatchGlobalEvent({
+        agentIdentifier: emitter.debugId,
+        loaded: true,
+        type: 'data',
+        name: 'buffer',
+        feature: bufferGroupMap[type],
+        data: { type, args, force: !!force, bubble: !!bubble }
+      })
```
metal-messiah, line 116: I understand the confusion here, but this dispatch should actually be in the event store manager.  Doing this in the event emitter directly will be way too noisy to be useful.

Something like 
```javascript dispatchGlobalEvent({
      agentIdentifier,
      loaded: true,
      type: 'data',
      name: 'buffer',
      feature: this.StorageClass === EventAggregator ? 'shared_aggregator' : this.featureName,
      data: event
    })
    ```
    in https://github.com/newrelic/newrelic-browser-agent/blob/29e82a7b06a1697750ed058bec81930a8c552137/src/features/utils/event-store-manager.js#L47-L50

Comment (ellisong) [reply src/common/event-emitter/contextual-ee.js]: moved to store manager

File: src/loaders/api/api.js [hunk 2]
```diff
@@ -194,6 +195,14 @@ export function setAPI (agentIdentifier, forceDrain, runSoftNavOverSpa = false)
   function apiCall (prefix, name, notSpa, bufferGroup) {
     return function () {
       handle(SUPPORTABILITY_METRIC_CHANNEL, ['API/' + name + '/called'], undefined, FEATURE_NAMES.metrics, instanceEE)
+      dispatchGlobalEvent({
+        agentIdentifier,
+        loaded: true,
```
metal-messiah, line 200: I think after you pointed out the flakiness of the tests, the `loaded` attribute here (and maybe in all the emitters needs to be "smarter".  It could probably just check to see if the exported `activatedFeatures` object has been set yet -- if no, loaded = `false`, if yes, loaded = `true`.   The API can technically be called before the window/agent has fully loaded.

metal-messiah: this could be improved in a future ticket, because i think there are some things we could do to unify all the dispatch calls across the agent to provide common attributes (agentIdentifier, loaded) without requiring the individual modules to pass that stuff around so much -- like make the dispatcher a higher order function stored in the runtime or something.  I'll add a separate ticket for that work that can extend from this as to not derail this work

metal-messiah (approved):
looks good!

PR #1414, opened 2025-03-13
feat: Mfe ajax detection

metal-messiah:
Enable agent to capture scoped AJAX data via `X-Newrelic-Entity-Guid` request header.
---


metal-messiah:
This PR will need to wait until decisions in MFE dacis are reached about how/when entity guids are passed around

metal-messiah:
closed in favor of #1612

PR #1416, opened 2025-03-15
fix: Session Replay text masking for whitespace

ptang-nr:
There is an edge case where whitespace was being masked, which interferes with the layout/styling in session replays.
---

### Overview

In this PR, we address the edge case by allowing the whitespace to return as-is when it's not a password field.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-379491

### Testing

Added unit test for `customMasker` function
Manually tested session replay on customer's site using rrweb REPL

### DIFF
--- a/src/features/session_replay/shared/utils.js
+++ b/src/features/session_replay/shared/utils.js
@@ -28,3 +28,15 @@ export function buildNRMetaNode (timestamp, timeKeeper) {
     originTimeDiff: Math.floor(originTime - timeKeeper.correctedOriginTime)
   }
 }
+
+export function customMasker (text, element) {
+  try {
+    if (typeof element?.type === 'string') {
+      if (element.type.toLowerCase() === 'password') return '*'.repeat(text?.length || 0)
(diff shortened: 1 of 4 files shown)


src/features/session_replay/shared/utils.js
```diff
@@ -28,3 +28,15 @@ export function buildNRMetaNode (timestamp, timeKeeper) {
...
+export function customMasker (text, element) {
+  try {
+    if (typeof element?.type === 'string') {
+      if (element.type.toLowerCase() === 'password') return '*'.repeat(text?.length || 0)
+      if (element?.dataset?.nrUnmask !== undefined || element?.classList?.contains('nr-unmask')) return text
+    }
+  } catch (err) {
+    // likely an element was passed to this handler that was invalid and was missing attributes or methods
+  }
+  return typeof text === 'string' ? text.replace(/[\S]/g, '*') : '*'.repeat(text?.length || 0)
```
cwli24, line 41: This doesn't seem to match the PR description. Isn't whitespaces still being replaced by * here, aka being masked?

cwli24 (dismissed):

cwli24 (approved):

PR #1418, opened 2025-03-18
feat: Handle duplicate agent APIs

ellisong:
Top-level interaction api calls are only responded to by the first exposed non-micro agent. This was changed to prevent issues for customers that run multiple agents together. The top-level interaction api should only be used to communicate with the main agent on the webpage. API calls on micro agents must instead be directly called on the respective micro agents.
---

### Overview

I added a new multi-agent.html in our browser-agent-wrapper test build. This lets a main browser agent and a micro agent run together on the same web page.

I'd recommend a notice on documentation notifying customers of this change; it can break experiences in some rare situations.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-361249

### Testing

Unit and wdio tests were added/changed, make sure they pass.


metal-messiah (approved):
lgtm!

PR #1423, opened 2025-03-21
fix: Get string className for SVG elements

cwli24:
Change the `nearestClass` field for SVG elements to be a meaningful string name. The `className` of such elements is of type `SVGAnimatedString` and previously resulted in an empty object string.
---

### Overview

Use `.baseVal` property of SVGAnimatedString object to represent className for SVG. This is the most common way to make it a string.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-376964

### Testing

Added


src/common/dom/selector-path.js
```diff
@@ -34,9 +34,7 @@ export const generateSelectorPath = (elem, targetFields = []) => {
   try {
     while (elem?.tagName) {
       const { id, localName } = elem
-      targetFields.forEach(field => {
-        nearestFields[nearestAttrName(field)] ||= elem[field]
-      })
+      targetFields.forEach(field => { nearestFields[nearestAttrName(field)] ||= (elem[field]?.baseVal || elem[field]) })
```
metal-messiah, line 37: Are there are any other non-SVG attributes that could have this property that we need to be aware of?

PR #1424, opened 2025-03-24
feat: Reduce noise from `mousemove` events

ptang-nr:
There may be an edge case where a site performance can be impacted if the browser agent picked up `mousemove` events via `addEventListener`.
---

### Overview

In this PR, we are excluding `mousemove` events from SessionTrace instrumentation to reduce potential processing burden from high occurrences of said event.  

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-381325

### Testing

- Manually tested via adding a event listener for the `mousemove` event and confirming the event has been ignored successfully.
- Added an e2e test to ensure no `mousing` event is detected for `mousemove` event

### DIFF
--- a/src/common/wrap/wrap-events.js
+++ b/src/common/wrap/wrap-events.js
@@ -45,7 +45,8 @@ export function wrapEvents (sharedEE) {
   ee.on(ADD_EVENT_LISTENER + '-start', function (args, target) {
     var originalListener = args[1]
     if (originalListener === null ||
-      (typeof originalListener !== 'function' && typeof originalListener !== 'object')
+      (typeof originalListener !== 'function' && typeof originalListener !== 'object') ||
+      args[0] === 'mousemove'
     ) {
       return
     }
--- a/src/common/wrap/wrap-events.js
+++ b/src/common/wrap/wrap-events.js
@@ -45,8 +45,7 @@ export function wrapEvents (sharedEE) {
   ee.on(ADD_EVENT_LISTENER + '-start', function (args, target) {
     var originalListener = args[1]
     if (originalListener === null ||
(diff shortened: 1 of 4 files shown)


src/common/wrap/wrap-events.js
```diff
@@ -45,7 +45,8 @@ export function wrapEvents (sharedEE) {
   ee.on(ADD_EVENT_LISTENER + '-start', function (args, target) {
     var originalListener = args[1]
     if (originalListener === null ||
-      (typeof originalListener !== 'function' && typeof originalListener !== 'object')
+      (typeof originalListener !== 'function' && typeof originalListener !== 'object') ||
+      args[0] === 'mousemove'
```
cwli24, line 49: I think wrap-events doesn't need changing. It just needs to be excluded from Traces. There's no event specific logic in here

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -15,7 +15,7 @@ const SUPPORTS_PERFORMANCE_OBSERVER = typeof globalScope.PerformanceObserver ===
 
 const ignoredEvents = {
   // we find that certain events make the data too noisy to be useful
-  global: { mouseup: true, mousedown: true },
+  global: { mouseup: true, mousedown: true, mousemove: true },
```
cwli24, line 18: in `shouldIgnoreEvent` fn, also move the 2nd line up to reduce processing from `eventOrigin()`
image

ptang-nr: Ah yeah, good call!

cwli24 (approved):

PR #1425, opened 2025-03-27
feat: Erase .api property on agent instance

cwli24:
The `.api` property of full and micro agents have been removed. The browser API methods are instead moved to the agent instance itself for ease of targeted access. This includes a minor fix of runtime `loaderType` for agents that are re-configured after load.
---

### Overview

Instead of:
```js
agent: {
 features: ...,
 info: ...,
 api: {
  setErrorHandler: () => {},
  setAPI: () => {},
  ...
 },
 ...
}
```
The API will now be:
```js
agent: {
 features: ...,
 info: ...,
 setErrorHandler: () => {},
 setAPI: () => {},
 ...
}
```

`mainOrMicroAgent.api.___` will still work as it will redirect up a level and call `mainOrMicroAgent.___`.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-172955

### DIFF
--- a/src/features/utils/aggregate-base.js
+++ b/src/features/utils/aggregate-base.js
@@ -136,14 +136,14 @@ export class AggregateBase extends FeatureBase {
       } catch (err) {
         // do nothing
       }
-      configure({ agentIdentifier: this.agentIdentifier }, {
+      configure(existingAgent, {
         ...cdn,
         info: {
           ...cdn.info,
           jsAttributes
         },
         runtime: existingAgent.runtime
-      })
+      }, existingAgent.runtime.loaderType)
     }
   }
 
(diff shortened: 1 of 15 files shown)


src/features/utils/aggregate-base.js
```diff
@@ -136,14 +136,14 @@ export class AggregateBase extends FeatureBase {
       } catch (err) {
         // do nothing
       }
-      configure({ agentIdentifier: this.agentIdentifier }, {
+      configure(existingAgent, {
```
metal-messiah, line 139: if `configure(...` has a reference to `existingAgent` already, do we need the third argument?

src/features/utils/nr1-debugger.js
metal-messiah, line 1: Is this dead code?

src/features/utils/aggregate-base.js
cwli24: Yes, since agent.js and micro-agent.js passes in a 3rd argument with different defaulted values. I left that unchanged.

src/features/utils/nr1-debugger.js
cwli24: I think so. I wasn't sure if there's a reason it was kept around or if it needed to be kept.

metal-messiah (approved):

PR #1426, opened 2025-03-27
feat: Prevent storing session data past session expiry

ptang-nr:
There may be edge cases where some Session Replay/Session Trace data could be stored even though the agent session has expired.  Add instrumentation to detect such occurrences and skip storing these events for the time being.
---

### Overview

In this PR, 

- add supportability metrics (once per harvest) regarding the presence of Session Replay/Session Trace data after agent session expiry
- prevent storing such event data

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-341205

### Testing

Manually tested on local test-server
Added unit tests to test data is not stored when session is expired


src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -139,6 +141,16 @@ export class TraceStorage {
...
   }
 
+  isBeforeSessionExpiry (entryTimestamp) {
+    let isValidTimingEntry = entryTimestamp < DEFAULT_EXPIRES_MS
+
+    if (!isValidTimingEntry && !this.eventsSeenAfterSessionExpire) {
+      this.eventsSeenAfterSessionExpire = true
+    }
+
+    return isValidTimingEntry
```
metal-messiah, line 151: Something like this might be less verbose
```suggestion
    return this.eventsSeenAfterSessionExpire ||= entryTimestamp < DEFAULT_EXPIRES_MS
```

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -49,6 +50,7 @@ export class TraceStorage {
       const openedSpace = this.trimSTNs(ERROR_MODE_SECONDS_WINDOW) // but maybe we could make some space by discarding irrelevant nodes if we're in sessioned Error mode
       if (openedSpace === 0) return
     }
+    if (!this.isBeforeSessionExpiry(stn.s)) return
```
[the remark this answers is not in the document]
metal-messiah: If thats the case, you will need to reference the session entity itself to see exactly when it will expire

src/features/session_trace/aggregate/index.js
```diff
@@ -124,6 +124,10 @@ export class Aggregate extends AggregateBase {
     const endUserId = this.agentRef.info.jsAttributes['enduser.id']
     const entityGuid = this.agentRef.runtime.appMetadata.agents?.[0]?.entityGuid
 
+    if (this.events.eventsSeenAfterSessionExpire) {
+      this.parent.reportSupportabilityMetric('Session/Expired/SessionTrace/Seen')
```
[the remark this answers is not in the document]
ptang-nr: Oof, thanks for catching that!

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -39,6 +39,13 @@ export class TraceStorage {
 
   constructor (parent) {
     this.parent = parent
+    this.eventsSeenAfterSessionExpire = false
+  }
+
+  isAfterSessionExpiry (entryTimestamp) {
+    // eslint-disable-next-line no-return-assign
+    return this.eventsSeenAfterSessionExpire ||= this.parent.timeKeeper?.ready && this.parent.agentRef.runtime?.session?.state?.expiresAt &&
+      this.parent.timeKeeper.convertRelativeTimestamp(entryTimestamp) >= this.parent.agentRef.runtime.session.state.expiresAt
```
metal-messiah, line 48: I'm still a bit hesitant to know if this will actually work in reality.  When the session entity expires, it "resets", meaning the `expiresAt` will be pushed into the future at that time, creating a race between the feature storing data and the session entity pushing its expiry stamp back.  I wonder if it would be more pragmatic to manage this at the session level, by checking add a flag to the session-entity when it resets, and just checking that directly

ptang-nr: As I was looking at session-entity and walking through the reset, I _think_ we can leverage the sessionId (?) So that's what I did in the latest commit.  Let me know if that's not on the right track 🤔 (i did see afterwards that we have a similar check inside the agg so taking that as a sign right now :) )

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -44,8 +44,10 @@ export class TraceStorage {
 
   isAfterSessionExpiry (entryTimestamp) {
     // eslint-disable-next-line no-return-assign
-    return this.eventsSeenAfterSessionExpire ||= this.parent.timeKeeper?.ready && this.parent.agentRef.runtime?.session?.state?.expiresAt &&
-      this.parent.timeKeeper.convertRelativeTimestamp(entryTimestamp) >= this.parent.agentRef.runtime.session.state.expiresAt
+    return this.eventsSeenAfterSessionExpire ||=
+      (!!this.parent.agentRef.runtime?.session?.state?.value && this.parent.sessionId !== this.parent.agentRef.runtime.session.state.value) || // diff session ids means session has rotated
+      (this.parent.timeKeeper?.ready && this.parent.agentRef.runtime?.session?.state?.expiresAt &&
+        this.parent.timeKeeper.convertRelativeTimestamp(entryTimestamp) >= this.parent.agentRef.runtime.session.state.expiresAt)
```
metal-messiah: I think if this works and passes tests it looks good, my only final ask is if we can please extract `this.parent.agentRef.runtime?.session?.state?` to a variable to make this big statement a little easier to read through

ptang-nr: This ended up shrinking a bit after the move.  Lemme know if you spot anything else!

src/common/session/session-entity.js
```diff
@@ -176,6 +179,7 @@ export class SessionEntity {
       if (this.isExpired(obj.expiresAt)) {
         this.collectSM('expired')
         this.collectSM('duration', obj, true)
+        this.state.numOfExpirations++
```
metal-messiah, line 182: This needs to get applied to the `inactive timer` as well since that can also cause a session to reset

metal-messiah: maybe just move the incrementer to the `reset()` call itself?

metal-messiah (approved):

PR #1430, opened 2025-04-01
feat: Remove agent config cache and getters/setters

cwli24:
Removes agent internal configuration utility functions. No impact to agent functionality is expected. Build size have been reduced by this coding style change.
---

### Overview

Agent arch changed to storing & using config object references directly rather than storing and fetching them using a module-level map via `agentIdentifier` string. This removes a bunch of bloat and reduces size.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-340432


src/common/event-emitter/contextual-ee.js
```diff
@@ -34,7 +33,7 @@ function ee (old, debugId) {
     // We only want to check the runtime configuration for `isolatedBacklog` if the event emitter belongs to a feature.
     // For feature event emitters, the debugId will be an agentIdentifier with a length of 16. In contrast, some of our
     // tests do not namespace the event emitter with an agentID and just use the parent (which has no ID).
-    isolatedBacklog = debugId.length !== 16 ? false : getRuntime(debugId).isolatedBacklog
+    isolatedBacklog = debugId.length !== 16 ? false : nr.initializedAgents?.[debugId]?.isolatedBacklog
```
metal-messiah, line 36: this should probably have a fallback of `false` if it fails to find `initializedAgents` -- though it likely doesnt matter much if its just used as a truthy check and not explicitly checking for `false` so that just needs to be checked

metal-messiah:
Theres a lot of merge conflicts in this one so i will wait to review until those are consolidated

src/common/event-emitter/contextual-ee.js
cwli24: Right, no usage of isolatedBacklog does an explicit check for `false`, so the fallthrough to undefined in that case works fine.

cwli24: I did figure out this is causing the test failure because it's missing the `.runtime` but that's another issue... heh

src/features/soft_navigations/aggregate/bel-node.js
```diff
@@ -14,9 +14,9 @@ export class BelNode {
   callbackDuration = 0
   nodeId = ++nodesSeen
 
-  constructor (agentIdentifier) {
-    if (!agentIdentifier) throw new Error('Interaction is missing core attributes')
-    this.agentIdentifier = agentIdentifier
+  constructor (agentRef) {
+    this.obfuscator = agentRef.runtime.obfuscator
+    this.info = agentRef.info
```
metal-messiah, line 19: These feel weird getting set in the base class, since it doesn't actually use anything from the agent ref. Could just set the relevant values in the outer classes -- Doesn't necessarily need to change, but just wanted to point that out

metal-messiah (approved):

metal-messiah:
Needs a suitable release notes blurb in the description before merging

cwli24:
> Needs a suitable release notes blurb in the description before merging

This is a chore PR; it'll get disregarded

src/features/soft_navigations/aggregate/bel-node.js
cwli24: Valid, but I'll save the duplication on the derived classes since they share using these assignments

PR #1433, opened 2025-04-07
feat: API Code Splitting

metal-messiah:
Split the API method definitions per feature instead of a single file shared across all loader types.  This allows each loader to only instantiate the necessary APIs for the features included, and reduce overall bundle size for the "Lite" build as well as custom-built agents with NPM.  A shell will still exist for non-initialized APIs to prevent errors from throwing on APIs used without the needed features. 
---

### Overview
Each API has been moved to an individual file.  Each feature can then import and setup the APIs relevant to that feature.

### Related Issue(s)
Relates to #1443 , #1447 
NR-398729

### Testing
Tests should continue to pass


ptang-nr (approved):
Reviewed to the best that I can

PR #1435, opened 2025-04-08
fix: Reset `notified` when replay stops recording

ptang-nr:
Resolves an issue where JavaScriptError events would be missing the `hasReplay` attribute after session recording is resumed (for example, from user clicking on another tab/tab going idle and then coming back to it).  
---

### Overview

Session replay records user interactions and could be a powerful tool to understand user experience.  
- If/when an error occurs and a session replay was present, the error would normally be decorated with an `hasReplay` attribute.
- If user has opened multiple tabs, session replay will only record from the latest opened tab,
- When a tab is in a 'hidden' state, the session replay will stop and errors will NOT be decorated with `hasReplay`
- When a user returns to a previously opened tab,
    - if it is the latest opened tab, session recording will resume and any errors should once again be decorated with `hasReplay`
    - if it is NOT the latest opened tab, session recording will NOT resume and errors will NOT be decorated with `hasReplay`

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-400822

### Testing

- Manual testing using `rrweb-record.html`
    - Opened in a single tab, then visit an unrelated site in another tab and come back (recording should resume, jserrors should have `hasReplay = true`)
    - Opened in a single tab, then opened the same page again in a second and third tab (recording only resumes in the third tab, only errors in third tab is decorated)
- Created a new E2E test for this scenario under `tests/specs/session-replay`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -54,6 +54,7 @@ module.exports = {
       rules: {
         'no-console': ['error'],
         'n/no-callback-literal': 'off', // This is not NodeJS code and should not be forced to adhere to NodeJS callback parameter pattern
+        'n/handle-callback-err': 'off', // eslint interprets our noticeError method as a callback, but it is not
         'headers/header-format': [
           'error',
           {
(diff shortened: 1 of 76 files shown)


ptang-nr:
`tests/components/soft_navigations/api.test.js` possibly [flaky](https://github.com/newrelic/newrelic-browser-agent/actions/runs/14342700213/job/40205671665)
Screenshot 2025-04-08 at 1 19 06 PM

Screenshot 2025-04-08 at 1 17 55 PM

metal-messiah (approved):

PR #1439, opened 2025-04-10
chore: Improve test result console report

ellisong:
Improves console reports for test results by only printing spec files containing test failures instead of all spec files.
---

### Overview

This PR changes a config setting in the spec reporter called `onlyFailures`. When set to `true`, it only prints out spec reports for test files that have a failing test. It used to print every single spec file result at the end regardless of passing or failure which made it difficult to find the tests that failed. This change results in a lot less noise at the bottom of the console log which makes it easy now to find test failures.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-250731

### Testing

Look at the console report at the end of testing. Should now only have test failures reported (if there are any). It will still show a count of how many tests failed/passed.


ptang-nr (approved):
LGTM :shipit:

PR #1443, opened 2025-04-14
feat: Improve lazy-loading optimization

metal-messiah:
Optimize our pattern to ensure that code-bundlers such as webpack can more effectively tree-shake the output to only include the relevant files needed to run the agent.  This specifically helps reduce the bundle size, reduce the amount of lazy files generated, and streamlines "custom" agent builds used with NPM.
---

### Overview
Agent behavior should be unchanged.

### Related Issue(s)
Related to #1447, #1433
NR-408546

### Testing
Tests should continue to pass


src/features/utils/instrument-base.js
```diff
@@ -52,36 +51,45 @@ export class InstrumentBase extends FeatureBase {
...
-        this.auto = true
-        this.importAggregator(agentRef)
-      }))
+          registerDrain(agentRef.agentIdentifier, this.featureName)
+          resolve()
+        }))
+      })
+    } else {
+      /** if the feature requires opt-in (!auto-start), it will get registered once the api has been called */
+      registerDrain(agentRef.agentIdentifier, featureName)
```
cwli24, line 67: I don't quite get the need for this change

metal-messiah: IMO it makes it a little simpler to not need to mock out auto and pass it around, and makes it so that importAggregator is always called in just one place.  Just wait for the deferment promise to resolve, it either does so immediately if auto, or after the start api if not

cwli24: Yeah nvm, I think cutting down importAggregator to 1 call makes it simpler right.

cwli24 (approved):

Issue #1450, opened 2025-04-17
Uncaught (in promise) TypeError: Unhandled Promise Rejection: Cannot read properties of undefined (reading 'add')


agsbs:
## Description

When enabling the Ajax feature of the newrelic-browser-agent package, I’m encountering the following error:

```
Uncaught (in promise) TypeError: Unhandled Promise Rejection: Cannot read properties of undefined (reading 'add')
```

## Steps to Reproduce

Instantiate the `BrowserAgent `

```
new BrowserAgent(options)
```

Or instantiate the `Agent` with `Ajax` feature.

```
new Agent({
      ...options,
      features: [
        Ajax,
      ],
    });
```

The issue didn't occur if `Ajax` is excluded.

## Expected Behavior
It shouldn't be throwing an error

## Relevant Logs / Console output

```
index.js:106 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'add')
    at Aggregate.storeXhr (index.js:106:19)
    at EventContext.<anonymous> (index.js:37:17)
    at drain.js:137:15
    at Array.forEach (<anonymous>)
    at emitEvent (drain.js:130:44)
    at drainGroup (drain.js:104:9)
    at drain.js:81:7
    at Array.forEach (<anonymous>)
    at checkCanDrainAll (drain.js:79:11)
    at drain (drain.js:69:3)
```

## Your Environment

The error is occurring on Chrome and Firefox, potentially other browsers too.

metal-messiah:
Thanks for reporting this, we are looking into this now and will keep you posted in this issue thread.

metal-messiah:
Addressed as part of #1454

Issue #1446, opened 2025-04-15
Uncaught TypeError: Cannot read properties of undefined (reading 'clear')


DJoyekurun:
## Description

We have a rails application hosted on heroku with the new relic gem. We recently started getting the following error (see logs) without deploying any new change. We are on the latest version of the gem.

The configuration file contains browser_monitoring: auto_instrument: true.

The error is occurring almost every time the app is opened and is flooding our error logs.

## Expected Behavior

What can be the cause of this error, and how can we fix this?

## Relevant Logs / Console output

```
TypeError: Cannot read property 'clear' of undefined
y.abort(https://js-agent.newrelic.com/nr-full-1.287.0.min.js:2)

n.<anonymous>(https://js-agent.newrelic.com/nr-full-1.287.0.min.js:2)

Object.emit(https://app.xxx.com/post_message:4)

[anonymous](https://js-agent.newrelic.com/nr-full-1.287.0.min.js:2)

nrWrapper(https://app.xxx.com/post_message:4)
```

Here is the minified code from `nr-full-1.287.0.min.js:2`:
```
 abort(e={}) {
                this.reportSupportabilityMetric("Logging/Abort/".concat(e.sm)),
                this.blocked = !0,
                this.events.clear(),
                this.events.clearSave(),
                this.updateLoggingMode(o.A$.OFF),
                this.deregisterDrain()
```

## Your Environment

The error is occurring on all browsers/devices that we could test with.

metal-messiah:
Thanks for reporting this, this has been addressed as part of #1454 .  Please reach back out if further issues occur!

PR #1452, opened 2025-04-17
fix: Ensure event buffer always exists

metal-messiah:
Ensure event buffers always exist to prevent race conditions between agent life-cycles and data storage.
---

### Related Issue(s)
This PR aims to stabilize #1446, #1450, NR-405039 & NR-404490

### Testing
Existing tests should continue to pass

### DIFF
--- a/src/features/utils/event-store-manager.js
+++ b/src/features/utils/event-store-manager.js
@@ -38,11 +38,13 @@ export class EventStoreManager {
   }
 
   setEventStore (targetEntityGuid) {
+    /** we should already have an event store for the default */
+    if (!targetEntityGuid) return
     /** if the target is the container agent, SHARE the default storage -- otherwise create a new event store */
     const eventStorage = (isContainerAgentTarget(this.entityManager.get(targetEntityGuid), this.agentRef))
       ? this.appStorageMap.get(DEFAULT_KEY)
       : new this.StorageClass()
-    this.appStorageMap.set(targetEntityGuid || DEFAULT_KEY, eventStorage)
+    this.appStorageMap.set(targetEntityGuid, eventStorage)
   }
 
   // This class must contain an union of all methods from all supported storage classes and conceptualize away the target app argument.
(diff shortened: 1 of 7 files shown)


tools/wdio/config/base.conf.mjs
```diff
@@ -46,7 +46,7 @@ export default function config () {
     mochaOpts: {
       ui: 'bdd',
       timeout: args.timeout,
-      retries: args.retry ? 3 : 0
+      retries: args.retry ? 5 : 0
```
ptang-nr, line 49: Curious to know why we're bumping the number of retries.  Any concerns this can make the overall test suite take even longer to fail?  (i think I've seen 38m for one of the browser types to fail before 🫠)

src/features/utils/event-store-manager.js
```diff
@@ -30,10 +31,15 @@ export class EventStoreManager {
    */
   #getEventStore (targetEntityGuid) {
     if (!targetEntityGuid) return this.defaultEntity
-    if (!this.appStorageMap.has(targetEntityGuid)) this.appStorageMap.set(targetEntityGuid, new this.StorageClass())
+    if (!this.appStorageMap.has(targetEntityGuid)) this.setEventStore(targetEntityGuid)
     return this.appStorageMap.get(targetEntityGuid)
   }
 
+  setEventStore (targetEntityGuid) {
+    const eventStorage = (!!this.defaultEntity && isContainerAgentTarget(this.entityManager.get(targetEntityGuid), this.agentRef)) ? this.defaultEntity : new this.StorageClass()
```
ptang-nr, line 39: 🤔 Would `this.defaultEntity` always exist at this point?

metal-messiah: It will only not exist in the case that the eventStoreManager was provided an entityGuid in its constructor, such as if a feature were deferred until after the page had loaded fully

tools/wdio/config/base.conf.mjs
metal-messiah: This was to account for the flaky retry harvest tests in mobile.  Its slower to retry 5 times vs 3 times, but faster than re-running the full suite

ptang-nr (approved):
Voting 👍

PR #1457, opened 2025-04-23
chore: Store recorder status under runtime

ptang-nr:
Centralize the recorder status under runtime.  This will continue to reflect the status on a per-agent instance basis (vs. the sessionReplayMode).
---

### Overview

This is a follow-up to make improvements around SR recording.

### Related Issue(s)

[NR-400822](https://new-relic.atlassian.net/browse/NR-400822)

### Testing

- Manually smoke tested on rrweb-record.html (invoked `newrelic.noticeError`)
- Updated affected tests

[NR-400822]: https://new-relic.atlassian.net/browse/NR-400822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/features/session_replay/shared/recorder.js
```diff
@@ -105,9 +103,7 @@ export class Recorder {
     })
 
     this.stopRecording = () => {
-      this.recording = false
-      this.notified = false
-      this.parent.ee.emit(SR_EVENT_EMITTER_TYPES.REPLAY_RUNNING, [false, this.parent.mode])
+      this.parent.agentRef.runtime.isRecording = false
```
metal-messiah, line 106: We probably need to add this line in `session_replay/aggregate` in the `abort` method as well, just to ensure that if an error is thrown in `const stop = recorder({ ...` aka before the stopRecording method is set, it can still ensure the recording state is accurate.  Maybe a try/catch in the recorder's `startRecording` method as well

ptang-nr: Just saw this 👀

ptang-nr: Added - I threw in a try/catch.  For now, I assumed it's treated as an internal error.  Lmk if we wanna do something different!

src/features/session_replay/aggregate/index.js
```diff
@@ -369,6 +369,7 @@ export class Aggregate extends AggregateBase {
     this.reportSupportabilityMetric(`SessionReplay/Abort/${reason.sm}`, data)
     this.blocked = true
     this.mode = MODE.OFF
+    this.agentRef.runtime.isRecording = false
```
cwli24, line 372: `stopRecording` in the following line will set this to false already, no?

src/loaders/api/constants.js
```diff
@@ -5,7 +5,7 @@
 export const prefix = 'api-'
 export const spaPrefix = prefix + 'ixn-'
 
-export const replayRunning = {}
+// export const replayRunning = {}
```
metal-messiah, line 8:
```suggestion
```

ptang-nr: Thanks for catching that, missed cleanup 😞

src/features/session_replay/aggregate/index.js
ptang-nr: @cwli24 yeah, I added it based on some of the feedback.  Maybe now that we have the try/catch, this becomes redundant 🤔

ptang-nr: Removed!

metal-messiah (approved):

PR #1463, opened 2025-04-30
feat: Set UserAction currentUrl before aggregation end

ellisong:
Set currentUrl on the first event of a user action instead of at the end of aggregation.
---

### Overview

User actions we collect are aggregated, so once the aggregation stops (e.g. a new event comes in that doesn't match), that's when the currentUrl is assigned to the event. A navigation can happen before the next user action is dispatched, so the currentUrl gets set incorrectly if it is still in an aggregation state during the navigation. This PR assigns currentUrl on registering the first event node instead of assigning it at the end of aggregation.

This issue can happen when replacing URL state in the history object using `history.replaceState()` instead of pushing or popping it.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-398667

### Testing

Added a new test. The currentUrl should be before the URL history is changed.


ptang-nr (approved):
LGTM

PR #1466, opened 2025-05-02
chore: enable prerelease npm action to actually publish to npm

metal-messiah:
Enable pre-release NPM action to actually publish instead of dry-run
---


.github/actions/prerelease-npm-version/action.yml
cwli24, line 1: You can just pass in this arg when calling the action, no?

metal-messiah: Thats true, but this PR also enables it to start working for publish-dev automatically to skip the manual call

PR #1467, opened 2025-05-05
feat: Control log API through separate RUM flag

cwli24:
Logs captured by the `newrelic.log` method will now be configurable through settings and controlled separately from wrapped logs. This allows one to be disabled without affecting the other, as in the case that only `newrelic.log` is desired and not auto instrumented `console`.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-399179

### Testing

Added

### DIFF
--- a/src/features/logging/aggregate/index.js
+++ b/src/features/logging/aggregate/index.js
@@ -20,31 +20,37 @@ export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
   constructor (agentRef) {
     super(agentRef, FEATURE_NAME)
-    this.isSessionTrackingEnabled = canEnableSessionTracking(agentRef.init) && agentRef.runtime.session
+    const updateLocalLoggingMode = (auto, api) => {
+      this.loggingMode = { auto, api }
+      // In agent v1.290.0 & under, the logApiMode prop did not yet exist, so need to account for old session state being in-use.
+      if (api === undefined) this.loggingMode.api = auto
+    }
 
     // The SessionEntity class can emit a message indicating the session was cleared and reset (expiry, inactivity). This feature must abort and never resume if that occurs.
     this.ee.on(SESSION_EVENTS.RESET, () => {
       this.abort(ABORT_REASONS.RESET)
     })
-
     this.ee.on(SESSION_EVENTS.UPDATE, (type, data) => {
       if (this.blocked || type !== SESSION_EVENT_TYPES.CROSS_TAB) return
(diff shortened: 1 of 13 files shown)


src/features/logging/aggregate/index.js
```diff
@@ -166,13 +168,18 @@ export class Aggregate extends AggregateBase {
       this.events.clear()
       this.events.clearSave()
     }
-    this.updateLoggingMode(LOGGING_MODE.OFF)
+    this.loggingMode = {
```
metal-messiah, line 171: Should this just call `this.updateLocalLoggingMode` to reduce the places that `this.loggingMode` is set?

src/loaders/api/log.js
```diff
@@ -14,5 +14,5 @@ export function setupLogAPI (agent) {
 }
 
 export function log (message, { customAttributes = {}, level = LOG_LEVELS.INFO } = {}, agentRef, targetEntityGuid, timestamp = now()) {
-  bufferLog(agentRef.ee, message, customAttributes, level, targetEntityGuid, timestamp)
+  bufferLog(agentRef.ee, message, customAttributes, level, false, targetEntityGuid, timestamp)
```
metal-messiah, line 17: what is `<api>.wrapLogger` setting for this?  Shouldnt that honor the API settings as well?

cwli24: I wasn't sure. It wasn't clear if the goal was to configure just .log separate or .wrapLogger as well--or whether those two themselves should also be separate.

At the moment, only individual .log are controlled by the new flag, while wrapLogger is lumped with console

metal-messiah: I believe the original "[auto logging](https://docs.newrelic.com/docs/browser/browser-monitoring/browser-pro-features/browser-logs/get-started/#automatic-log)" documentation has solely described auto-implementation of the console APIs. I think it would make more sense to include `wrapLogger` in with the API sampling but I will defer to @ptang-nr who is leading the feature.

ptang-nr: Talked with PM, `wrapLogger` should also follow API sampling 👍

Kai-Richardson:
> Currently, I find myself stuck having to allow my browser apps to ingest millions of unused logs monthly, just to get a few hundred relevant logs that I need to know about.

A workaround, though obviously not the best solution, is to use version `1.282` - since that's right before the auto-logging was added.

src/features/logging/aggregate/index.js
cwli24: Could, but would have to change that to be an instance method though. Very negligible whether to pick 1 or the other

ptang-nr (dismissed):
LGTM

metal-messiah:
Blocked until API team is out of code freeze and can implement settings changes.

PR #1471, opened 2025-05-08
chore: Do not warn unless config actually fails to exist

metal-messiah:
This PR updates the configure call to only warn if the attribute fails to exist.  In the off chance that `checkConfiguration` is called in the aggregate base and forces a re-configure, the `getter` used for `harvestCount` can be attempted to be overset and throw a console warning. This warning is misleading, as that getter does not cause issues and does in fact still exist. You can see this in action [here](https://docs.newrelic.com/). A new test has been added to ensure this case does not log a warning anymore.
---


ptang-nr (approved):
LGTM

PR #1476, opened 2025-05-16
feat: Create the measure API

ellisong:
---


src/loaders/api/measure.js
```diff
@@ -0,0 +1,47 @@
...
+    if (typeof returnObj.customAttributes !== 'object' || typeof name !== 'string' || name.length === 0) {
+      warn(57)
+      return
+    }
+
+    const getValueFromTiming = (timing, d) => {
+      if (timing == null) return d
+      if (typeof timing === 'number') return timing
+      if (typeof PerformanceMark === 'function' && timing instanceof PerformanceMark) return timing.startTime
+      return Number.NaN
```
ptang-nr, line 27: I was considering the use of default parameter because it's similar to our `null` check and noticed `undefined` would return NaN.  Why are they treated differently here? (too bad default param doesn't cover `null` 😞)

ellisong: I intentionally used double equals on `timing == null` to catch undefined being passed in.

ptang-nr: Ah - i had missed that `==`.  Nice!

PR #1480, opened 2025-05-21
Fix: Silent setting getter only harvestCount message

cwli24:
The warning that results from the agent attempting to overwrite the runtime `harvestCount` will be correctly silenced.
---

### Related Issue(s)

This is actually a proxy PR to get https://github.com/newrelic/newrelic-browser-agent/pull/1471 into production via patch version release.


ellisong (approved):
LGTM

PR #1483, opened 2025-05-23
fix: Harvest first session trace payload immediately

metal-messiah:
For better consistency, implemented old behavior where the first ST payload is harvested immediately when the page loads.  This helps ensure cases where replays are harvested but traces are not happen less frequently.
---

### Overview
- Harvest session trace immediately, and bump that to staging ✅
- Emit NR event for internal errors ✅
- Update postamble to not care about hipaa check in staging ✅
- Message when SR sends when ST hasn’t sent ✅
- Message if ST fails for any reason ✅


src/features/session_trace/aggregate/index.js
```diff
@@ -95,6 +96,8 @@ export class Aggregate extends AggregateBase {
         if (this.mode === MODE.ERROR) this.switchToFull()
       }, this.featureName, this.ee)
     }
+    /** try to harvest immediately. This will not send if the trace is not running in FULL mode due to the pre-harvest checks. */
+    this.agentRef.runtime.harvester.triggerHarvestFor(this)
```
cwli24, line 100: I think this needs to be after drain so any other nodes can be processed.

src/features/session_replay/aggregate/index.js
```diff
@@ -262,6 +262,10 @@ export class Aggregate extends AggregateBase {
     if (recorderEvents.type === 'preloaded') this.agentRef.runtime.harvester.triggerHarvestFor(this)
     payloadOutput.payload = payload
 
+    if (!this.agentRef.runtime.session.state.traceHarvestStarted) {
+      warn(57, JSON.stringify(this.agentRef.runtime.session.state))
```
ptang-nr, line 266: I have some doubts around whether this can potentially dump out `custom` attributes.

ptang-nr: Actually, nvm - i was initially wondering about logging out non-agent data but should not be an issue since the data either goes to console or loops thru logging, which gets obfuscated 🤔 (size shouldn't be a huge concern either)

PR #1485, opened 2025-05-23
chore: bump safari for test suite

metal-messiah:
---


tools/browsers-lists/lt-desktop-latest-vers.json
cwli24, line 1: Wait, doesn't this file update automatically from one of the scripts? Last I recall

metal-messiah: Apparently it must not work for Safari, because 18 has been available in LT for almost a year and it hasnt updated to use it

metal-messiah: I found what happened.  Starting with version 18, lambda's API started reporting `macOS` as `MacOS` and our matcher was not including it.  Will update the PR to fix that

PR #1486, opened 2025-05-23
feat: reduce harvest interval to 10 seconds

metal-messiah:
Reduce the standard harvest interval to 10 seconds to aid with late-page harvesting and unloading behaviors
---


metal-messiah:
closing for now as DEM is safe mode and an alternative plan is being discussed

PR #1490, opened 2025-05-30
fix: Add safeguards for addToTrace

ellisong:
Events created with addToTrace API, which have an invalid UNIX timestamp, emit a warning and return without creating the event.
---

### Overview

Added safeguards for `addToTrace()` API call. It was possible to have negative timestamps before with misuse. This PR does not create the event if the timestamps are negative. A warning is emitted of error code 61 if the timestamps are invalid.

At first I thought that the requirements were to only check that the arguments of the method are valid timestamps, but it makes more sense to me to check that the timestamps in the event creation are valid.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-366078

### Testing

Make sure tests pass

### DIFF
--- a/docs/warning-codes.md
+++ b/docs/warning-codes.md
@@ -121,3 +121,5 @@
 `Session replay harvested before a session trace payload could be sent. This could be problematic for replays that rely on a trace`
 ### 60
 `Session trace aborted`
+### 61
+`Timestamps must be valid UNIX timestamps.`
--- a/docs/warning-codes.md
+++ b/docs/warning-codes.md
@@ -122,4 +122,4 @@
 ### 60
 `Session trace aborted`
 ### 61
-`Timestamps must be valid UNIX timestamps.`
+`Timestamps must be non-negative and end time cannot be before start time.`
--- a/src/loaders/api/addToTrace.js
+++ b/src/loaders/api/addToTrace.js
@@ -21,7 +21,7 @@ export function setupAddToTraceAPI (agent) {
       t: 'api'
     }
 
-    if (report.s < 0 || report.e < 0) {
+    if (report.s < 0 || report.e < 0 || report.e < report.s) {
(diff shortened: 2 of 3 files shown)


docs/warning-codes.md
```diff
@@ -120,4 +120,6 @@
 ### 59
 `Session replay harvested before a session trace payload could be sent. This could be problematic for replays that rely on a trace`
 ### 60
-`Session trace aborted`
\ No newline at end of file
+`Session trace aborted`
+### 61
+`Timestamps must be valid UNIX timestamps.`
```
cwli24: The warning doesn't match the behavior. There's nothing actually checking they're valid _Unix timestamps_, only that they're non-negative.

src/loaders/api/addToTrace.js
```diff
@@ -20,6 +21,11 @@ export function setupAddToTraceAPI (agent) {
       t: 'api'
     }
 
+    if (report.s < 0 || report.e < 0) {
```
cwli24: I think you can also add a check that the end isn't strictly greater than the start. (equal is ok?)

docs/warning-codes.md
ellisong: What should the warning text be?

src/loaders/api/addToTrace.js
metal-messiah: Any opposition to just making these checks happen in the agg module to save loader build size?

cwli24: I think that's a little too extreme for the minified savings. Not worth the added complexity.

cwli24 (approved):
Do you want to make msg 61 specific to the Trace API? Or keep it generic so that maybe we can reuse this warning message for other checks in the agent.

metal-messiah (approved):

PR #1491, opened 2025-06-04
fix: Prevent ST from holding onto Event refs in memory when aborted

cwli24:
Releases the `Event` entries the agent hold onto in Trace after it has aborted. This addresses a known case of memory leak caused by continuing to add those references to a Set that is never cleared.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-420780

### Testing

Jest added

### DIFF
--- a/src/features/session_trace/aggregate/trace/storage.js
+++ b/src/features/session_trace/aggregate/trace/storage.js
@@ -53,7 +53,6 @@ export class TraceStorage {
 
   /** Central internal function called by all the other store__ & addToTrace API to append a trace node. They MUST all have checked #canStoreNewNode before calling this func!! */
   #storeSTN (stn) {
-    if (!this.#canStoreNewNode()) return
     if (this.trace[stn.n]) this.trace[stn.n].push(stn)
     else this.trace[stn.n] = [stn]
 


ellisong (dismissed):
LGTM

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -135,6 +139,11 @@ export class TraceStorage {
     }
   }
 
+  storeNode (node) {
```
metal-messiah, line 142: Is this new method necessary since storeSTN also makes this check?

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -135,6 +139,11 @@ export class TraceStorage {
...
+    if (!this.#canStoreNewNode()) return
```
metal-messiah, line 142: Is this necessary since `#storeSTN` makes this check?

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -160,13 +169,16 @@ export class TraceStorage {
           Math.floor(this.parent.timeKeeper.correctAbsoluteTimestamp(val))
         )
       }
-      this.storeSTN(new TraceNode(key, val, val, 'document', 'timing'))
+      if (!this.#canStoreNewNode()) return // at any point when no new nodes can be stored, there's no point in processing the rest of the timing entries
```
metal-messiah, line 171: same -- Is this necessary since `#storeSTN` makes this check?

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -216,36 +228,42 @@ export class TraceStorage {
 
   // Tracks when the window history API specified by wrap-history is used.
   storeHist (path, old, time) {
-    this.storeSTN(new TraceNode('history.pushState', time, time, path, old))
+    if (!this.#canStoreNewNode()) return
```
metal-messiah, line 230: Same - Is this necessary since `#storeSTN` makes this check?

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -216,36 +228,42 @@ export class TraceStorage {
...
+    }
 
     this.#laststart = resources[resources.length - 1].fetchStart | 0
   }
 
   // JavascriptError (FEATURE) events pipes into ST here.
   storeErrorAgg (type, name, params, metrics) {
     if (type !== 'err') return // internal errors are purposefully ignored
-    this.storeSTN(new TraceNode('error', metrics.time, metrics.time, params.message, params.stackHash))
+    if (!this.#canStoreNewNode()) return
```
metal-messiah, line 257: same - Is this necessary since `#storeSTN` makes this check?

src/features/session_trace/aggregate/trace/storage.js
cwli24: Ooops; I forgot to remove it from storeSTN, thanks for the catch.
The idea is we don't need to double-check (hence having a private method), and the check needs to go earlier depending on which caller it is.

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -41,15 +41,19 @@ export class TraceStorage {
...
       const openedSpace = this.trimSTNs(ERROR_MODE_SECONDS_WINDOW) // but maybe we could make some space by discarding irrelevant nodes if we're in sessioned Error mode
-      if (openedSpace === 0) return
+      if (openedSpace === 0) return false
     }
+    return true
+  }
 
+  /** Central internal function called by all the other store__ & addToTrace API to append a trace node. They MUST all have checked #canStoreNewNode before calling this func!! */
+  #storeSTN (stn) {
+    if (!this.#canStoreNewNode()) return
```
cwli24:
```suggestion
```

metal-messiah (approved):

PR #1495, opened 2025-06-07
feat: Mfe refactor

metal-messiah:
---


metal-messiah:
Screenshot 2025-09-25 at 12 12 05 PM
Something flaky in GH instance.  Passing locally

src/features/session_replay/aggregate/index.js
```diff
@@ -232,19 +235,18 @@ export class Aggregate extends AggregateBase {
...
     if (len > MAX_PAYLOAD_SIZE) {
       this.abort(ABORT_REASONS.TOO_BIG, len)
-      return [payloadOutput]
+      return
     }
+
     // TODO -- Gracefully handle the buffer for retries.
     if (!this.agentRef.runtime.session.state.sessionReplaySentFirstChunk) this.syncWithSessionManager({ sessionReplaySentFirstChunk: true })
     this.recorder.clearBuffer()
-    if (recorderEvents.type === 'preloaded') this.agentRef.runtime.harvester.triggerHarvestFor(this)
```
ptang-nr, line 240: Why is the preloaded harvest being removed?

tests/components/soft_navigations/aggregate.test.js
```diff
@@ -6,7 +6,6 @@ import { INTERACTION_STATUS, NO_LONG_TASK_WINDOW, POPSTATE_MERGE_WINDOW, POPSTAT
 let mainAgent
 
 beforeAll(() => {
-  jest.useFakeTimers({ doNotFake: ['nextTick', 'performance'] }) // to aid new long task window heuristics as default ixn will need to wait additional 5s to finish
```
ptang-nr, line 9: It looks like there are a series of changes that moved away from using jest fake timers in this test.  I didn't fully follow what was going on/why and was just curious about what prompted the change.

tests/specs/session-replay/rrweb-configuration.e2e.js
```diff
@@ -93,7 +93,8 @@ describe('RRWeb Configuration', () => {
 
       const sessionReplaysHarvests = await sessionReplaysCapture.waitForResult({ timeout: 10000 })
       expect(sessionReplaysHarvests.length).toBeGreaterThanOrEqual(1)
-      expect(decodeAttributes(sessionReplaysHarvests[0].request.query.attributes).hasSnapshot).toEqual(true)
```
ptang-nr, line 96: Also curious about the change here.  It looks(?) like the test used to indirectly assert the first harvest to be snapshot but that got relaxed, sort of.  We are still only capturing the first SR payload tho, so I'm not sure what has changed.

src/features/session_replay/aggregate/index.js
metal-messiah: preload harvests were already removed in a previous PR, this was just a remnant that got missed

tests/components/soft_navigations/aggregate.test.js
metal-messiah: If i remember correctly (its been a minute) -- fake timers were causing failures/wrong data to happen in these tests and these are "component" tests, meaning mock as little as possible so I decided to actually wait the required time for them to run.

tests/specs/session-replay/rrweb-configuration.e2e.js
metal-messiah: I think i just did a blanket find/replace since the "new" selector will work with both timeout: x as well as totalCount: x, it is more extensible later on if needed

metal-messiah: down on 403 is where it was actually necessary

tests/components/soft_navigations/aggregate.test.js
ptang-nr: (ooh, good to know! Adding this to my "internal checklist" when I write tests <3)

tests/specs/npm/registered-entity.e2e.js
```diff
@@ -1,96 +1,158 @@
...
+            expect(log.attributes['parent.id']).toEqual(containerAgentEntityGuid)
+          }
+          expect(ranOnce(id, 'log')).toEqual(true)
+          expect(Number(id)).toEqual(Number(log.message))
+        })
+      })
+
+      function ranOnce (appId, type) {
+        if (tests[appId][type] > 1) return false
+        countRuns(appId, type)
```
ptang-nr: If `tests[appId][type]` was 1, wouldn't this bump the run count to 2 and `ranOnce` still returns true?

ptang-nr: (it's a bit unclear if `ranOnce` is still testing for "ran once".  Is it possible to separate out the run count increment?)

metal-messiah: yeah totally could separate that out to make it more clear

metal-messiah: [a10d95d](https://github.com/newrelic/newrelic-browser-agent/commit/a10d95d1c1117ae4cd2a83cf281de6040b6c55bf)

src/common/config/init.js
```diff
@@ -48,7 +48,7 @@ const InitModelFn = () => {
   return {
     ajax: { deny_list: undefined, block_internal: true, enabled: true, autoStart: true },
     api: {
-      allow_registered_children: true,
+      allow_registered_children: false,
```
cwli24: 1. also change the default in `testing-server/constants` to match this to keep in sync
2. GE, jserror, and logging instrument all setup register method by default. It appears to me each of them will execute `invalidApiResponse` and spam warning # 55 with this change so this needs a tweaking.

metal-messiah: I'm not following here -- the 55 error code will only emit if `newrelic.register(...)` is called and the agent is not configured (and doesnt have the feature flag)

src/common/util/mfe.js
```diff
@@ -0,0 +1,34 @@
...
+  if (!isValidMFETarget(target)) {
+    return {
+      'entity.guid': containerAgentEntityGuid
+    }
+  }
+  return {
+    'mfe.id': target.id, // these field names may change as the schema is finalized
+    'mfe.name': target.name, // these field names may change as the schema is finalized
+    eventSource: 'MicroFrontendBrowserAgent', // these field names may change as the schema is finalized
+    'parent.id': containerAgentEntityGuid
```
cwli24: Is it too late to have synthesis also use "entity.guid" instead of "parent.id"?

docs/warning-codes.md
cwli24, line 1: What's the system to delete old warning codes? Or are they permanently baked here, and the number just continues to go up? (# 56 now falls out of use)

src/features/generic_events/aggregate/index.js
```diff
@@ -242,7 +242,6 @@ export class Aggregate extends AggregateBase {
         this.addEvent(event)
       }, this.featureName, this.ee)
 
-      agentRef.runtime.harvester.triggerHarvestFor(this)
```
cwli24, line 245: Should this go after drain instead of removal?

src/features/generic_events/aggregate/index.js
```diff
@@ -278,7 +277,9 @@ export class Aggregate extends AggregateBase {
       timestamp: Math.floor(this.agentRef.runtime.timeKeeper.correctRelativeTimestamp(now())),
       /** all generic events require pageUrl(s) */
       pageUrl: cleanURL('' + initialLocation),
-      currentUrl: cleanURL('' + location)
+      currentUrl: cleanURL('' + location),
+      /** Specific attributes only supplied if harvesting to endpoint version 2 */
+      ...(getVersion2Attributes(target, this))
```
cwli24, line 282: Question: right now without the attributes, it goes to the main app anyways. With `entity.guid` it still goes to the main app... what really is the difference? Do we even need `entity.guid` ?

src/features/logging/aggregate/index.js
```diff
@@ -100,24 +107,25 @@ export class Aggregate extends AggregateBase {
...
+  serializer (eventBuffer) {
     const sessionEntity = this.agentRef.runtime.session
     return [{
       common: {
         /** Attributes in the `common` section are added to `all` logs generated in the payload */
         attributes: {
-          'entity.guid': target.entityGuid, // browser entity guid as provided API target OR the default from RUM response if not supplied
+          ...(this.harvestEndpointVersion === 1 && {
+            'entity.guid': this.agentRef.runtime.appMetadata.agents[0].entityGuid,
+            appId: this.agentRef.info.applicationID
```
cwli24, line 122: - is potentially missing `appId` going to be an issue?
- this split logic is weird. These two props should be merged with `getVersion2Attributes` check above. (maybe that helper could be repurposed)

src/loaders/api/register.js
```diff
@@ -35,130 +35,99 @@ export function setupRegisterAPI (agent) {
...
+  target.licenseKey ||= agentRef.info.licenseKey // will inherit the license key from the container agent if not provided for brevity. A future state may dictate that we need different license keys to do different things.
+  target.blocked = false
+
+  /** @type {Function} a function that is set and reports when APIs are triggered -- warns the customer of the invalid state  */
+  let invalidApiResponse = () => {}
+  /** @type {Array} the array of registered target APIs */
+  const registeredEntities = agentRef.runtime.registeredEntities
+
+  /** if we have already registered this target, go ahead and re-use it */
+  const preregisteredEntity = registeredEntities.find(({ metadata: { target: { id, name } } }) => id === target.id && name === target.name)
```
cwli24: is the entityGuid sufficient? Does licenseKey also need to differ (can guid be the same)?

src/loaders/api/register.js
```diff
@@ -35,130 +35,99 @@ export function setupRegisterAPI (agent) {
...
+   * Block the API, and supply a warning function to display a message to end users
+   * @param {Function} warning
+   */
+  const block = (warning) => {
+    target.blocked = true
+    invalidApiResponse = warning
+  }
+
+  /** primary cases that can block the register API from working at init time */
+  if (!agentRef.init.api.allow_registered_children && !agentRef.init.feature_flags.includes('register')) block(() => warn(55))
```
cwli24: Previously, `allow_registered_children` needed to be true to continue this fn. Now even if it's false, if feature flag has "register", this will effectively ignore the init setting (truth) and proceed anyways. This is counterintuitive

src/common/util/mfe.js
metal-messiah: subject to change pre GA

src/features/generic_events/aggregate/index.js
metal-messiah: applied to all features in waitForDrain

src/features/logging/aggregate/index.js
metal-messiah: @metal-messiah should go back in and add appId as metadata for v2 events that are container events to keep parity with existing agents

src/loaders/api/register.js
metal-messiah: @metal-messiah use ID as unique matcher --- overwrite name if changed

src/loaders/api/register.js
metal-messiah: @metal-messiah to move the feature_flag lookup for this behavior to the `init` section

PR #1498, opened 2025-06-10
chore: fix broken gql test service

metal-messiah:
Point the graphql tests to our internal testing service with a new mocked endpoint instead of relying on a demo service, which has recently gone down and caused test failures.
---

### Testing
`graphql-metadata.e2e.js` should now pass


ellisong (approved):
LGTM

PR #1501, opened 2025-06-11
chore: Track click handlers of interactive elements

ptang-nr:
Extends `wrap-events` to track which buttons and links have click event handlers. 
---

### Overview

This is a first step towards helping track dead clicks on interactive elements, which are mainly buttons and links at the moment. Any click event handlers added via `addEventListener` for buttons and links will be stored in a lookup, key = element reference, value = handler function reference.  The lookup can be used later on to verify if a user click is going to "do something" vs. is a dead click.

### Related Issue(s)

[NR-422169](https://new-relic.atlassian.net/browse/NR-422169)

### Testing

- Added unit tests for InteractiveElements 

[NR-422169]: https://new-relic.atlassian.net/browse/NR-422169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/features/generic_events/aggregate/user-actions/interactive-elements.js
+++ b/src/features/generic_events/aggregate/user-actions/interactive-elements.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+class InteractiveElements {
+  constructor () {
+    this.elements = new WeakMap()
+  }
+
+  add (target, listener) {
+    if (!isInteractiveElement(target)) {
+      return
+    }
+    if (!this.elements.has(target)) {
+      this.elements.set(target, new Set())
+    }
+    this.elements.get(target).add(listener)
+  }
+
+  remove (target, listener) {
+    if (this.elements.has(target)) {
+      const handlers = this.elements.get(target)
+      handlers.delete(listener)
+      if (handlers.size === 0) {
+        this.elements.delete(target)
+      }
+    }
+  }
+
+  has (target) {
+    return this.elements.has(target)
+  }
+
+  get (target) {
+    return this.elements.get(target)
+  }
+}
+
+export const interactiveElems = new InteractiveElements()
+
+function isInteractiveElement (target) {
+  const tagName = (target && target.nodeType === 1 && target.tagName?.toLowerCase()) ?? ''
+  return tagName !== '' &&
+    (tagName === 'button' ||
+      tagName === 'a' ||
+      (tagName === 'input' && target.type === 'button'))
+}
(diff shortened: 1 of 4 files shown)


src/features/generic_events/aggregate/user-actions/interactive-elements.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+class InteractiveElements {
+  constructor () {
+    this.elements = new WeakMap()
```
metal-messiah: If this is acting as a private attribute it should probably be marked as one for consistency

metal-messiah: Either that or maybe something like extending WeakMap as the base class to keep things as primitive as possible

```javascript
class InteractiveElements extends WeakMap {
  constructor (iterable) {
    super(iterable)
  }

  add (target, listener) {
    if (!isInteractiveElement(target)) {
      return
    }
    if (!super.has(target)) {
      super.set(target, new Set())
    }
    super.get(target).add(listener)
  }

  delete (target, listener) {
    if (super.has(target)) {
      const handlers = super.get(target)
      handlers.delete(listener)
      if (handlers.size === 0) {
        super.delete(target)
      }
    }
  }
}
```

src/features/generic_events/aggregate/user-actions/interactive-elements.js
```diff
@@ -0,0 +1,47 @@
...
+    if (!isInteractiveElement(target)) {
+      return
+    }
+    if (!this.elements.has(target)) {
+      this.elements.set(target, new Set())
+    }
+    this.elements.get(target).add(listener)
+  }
+
+  remove (target, listener) {
```
metal-messiah: Should this be `delete` since the other keys match the WeakSet syntax?

src/features/generic_events/aggregate/user-actions/interactive-elements.js
ptang-nr: I can't really think of how InteractiveElements would evolve beyond a WeakMap, so went with the extend option for now.  It seems cleaner to only re-define where needed 👍

metal-messiah (approved):
Looks good to me, once the test suite is re-ran and passing, good to go!

ptang-nr:
@metal-messiah I tried running PR checks with `main` latest [here](https://github.com/newrelic/newrelic-browser-agent/actions/runs/15688021060).  

I seemed to have messed up the PR changes in doing so though 😓, so I backed out the merge from `main` to keep the PR changes legible. I'll merge `main` into `dead_clicks` separately to keep the feature branch updated.

Issue #1502, opened 2025-06-12
newrelic.finished() reports incorrect timeSinceload value


johnjardinemd:
It seems as though newrelic.finished()'s behaviour has changed the reported timeSinceload between versions 1.289.0 and 1.290.0

## Description

Prior to version, 1.290.0 if I were to execute newrelic.finished(Date.now()) and query timeSinceload on the event, I would receive a small number indicating the number of miiseconds it took the page to load. But now it returns a large number that resembles a unix timestamp.

## Steps to Reproduce

In SPA application, call `newrelic.finished(Date.now())` after the page finishes loading.

Query NRQL
`select timeSinceload from PageAction where appName = 'MyApp' AND actionName = 'finished' 
`

and receive the result of `1746835141.301`

## Expected Behavior

I would expect a number the matches the amount of time it took the page to load, e.g. `29.228`

## Your Environment

Browser: Chrome (latest--evergreen)
Operating System: Windows 10
Browser agent version involved: 1.290.0+
APM language and version: APM Agent 10.41.0.0 (.NET 9)

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-426133

johnjardinemd:
> https://new-relic.atlassian.net/browse/NR-426133

How does one, outside of New Relic, get access to your JIRA?
Can I assume this is a previously known or fixed issue?

johnjardinemd:
> Thanks for reporting this [@johnjardinemd](https://github.com/johnjardinemd) . Our Jira is an automated system that isnt exposed to the public. I will be keeping this issue open and working on a hotfix for this immediately.

Ah, ok. Excellent, thank you very much!

PR #1503, opened 2025-06-13
fix: Fix .finished API timeSinceLoad value

metal-messiah:
Fixes an issue where the `finished` api was creating a `timeSinceLoad` value for PageAction data as the unix timestamp instead of a relative time value in seconds from the page origin.
---

### Related Issue(s)
#1502 

### Testing
Tests have been updated to check that page actions create the correct timeSincePageLoad values


metal-messiah:
Addresses #1502

cwli24 (approved):

PR #1506, opened 2025-06-17
feat: Add "long tasks" internal message

metal-messiah:
Add an internal message indicating if a wrapped function qualifies as a "long task".  This is to be used later in agent features to enhance future behaviors.
---

### Overview
1. This PR adds a *new* message through the contextual EE when a long task is detected (>50ms): `long-task`
  - This can be subscribed to by any feature (though none are currently subscribed)
  - The metadata emitted contains task duration and method information, including errors thrown and fn name if supplied
2. The PR also adds the task metadata to the `-end` event, so that multiple pathways are supported for evaluating a task's execution metadata, including tasks that were not "long".

### Related Issue(s)
NR-414034

### Testing
New jest tests have been added, existing tests modified to check for existence of long task metadata.

### DIFF
--- a/src/common/wrap/wrap-function.js
+++ b/src/common/wrap/wrap-function.js
@@ -76,6 +76,7 @@ export function createWrapperWithEmitter (emitter, always) {
       var originalThis
       var ctx
       var result
+      let thrownError
 
       try {
         originalThis = this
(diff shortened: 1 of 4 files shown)


src/common/wrap/wrap-function.js
```diff
@@ -93,17 +94,28 @@ export function createWrapperWithEmitter (emitter, always) {
...
         safeEmit(prefix + 'err', [args, originalThis, err], ctx, bubble)
-
         // rethrow error so we don't effect execution by observing.
-        throw err
+        thrownError = err
+        throw thrownError
       } finally {
-        // happens no matter what.
-        safeEmit(prefix + 'end', [args, originalThis, result], ctx, bubble)
+        const duration = performance.now() - fnStartTime
```
cwli24: I think duration needs to move up further (& move result outside) like:
```js
 result = fn.apply(...)
 fnEndTime = performance.now()
} catch (err) {
 fnEndTime = performance.now()
 safeEmit(...) // this can add to the overall time if included
 thrownError = err
 throw thrownError
} finally {
 const duration = fnEndTime - fnStartTime
 ...
}
return result
```

metal-messiah: Yeah this makes sense, ty.  Will rework this

cwli24 (approved):

PR #1508, opened 2025-06-17
feat: Detect dead clicks for links and textboxes

ptang-nr:
User actions involving clicks on links and textboxes that will result in no changes will be marked as dead clicks.  

---

### Overview

This is part 1 of implementing dead clicks. 

Dead clicks will only be processed on visible elements and are defined as follows:
- For links, if there is no `href or click handlers defined
- For textboxes, it is readonly

The underlying processing will be in-place and running silently.  Dead clicks are only marked on UAs if the `user_frustrations` FF is in-effect.  

### Related Issue(s)

[NR-422170](https://new-relic.atlassian.net/browse/NR-422170)

### Testing

- Manual testing using `tests/assets/user-frustrations/instrumented-dead-clicks.html`
- Added `tests/specs/user-frustrations/dead-clicks.e2e.js`

[NR-422170]: https://new-relic.atlassian.net/browse/NR-422170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common/dom/selector-path.js
```diff
@@ -49,12 +58,34 @@ export const generateSelectorPath = (elem, targetFields = []) => {
...
     if (originalFieldName === 'tagName') originalFieldName = 'tag'
     if (originalFieldName === 'className') originalFieldName = 'class'
     return `nearest${originalFieldName.charAt(0).toUpperCase() + originalFieldName.slice(1)}`
   }
+
+  /**
+   * Checks if the element is an interactive link. Elements that are not visible will not be checked for interactivity.
+   * A link is considered interactive if it has an `href` attribute, an `onclick` handler, or any click event listener(s).
+   * @param elem
+   * @returns {boolean|undefined} undefined if not a visible link, false if the link is not interactive, true if it is interactive.
```
metal-messiah: > undefined if not a visible link, false if the link is not interactive, true if it is interactive.

Is there a reason to differentiate between undefined and false here?

src/common/dom/selector-path.js
```diff
@@ -49,12 +58,34 @@ export const generateSelectorPath = (elem, targetFields = []) => {
...
+   */
+  function isInteractiveLink (elem) {
+    try {
+      if (!elem ||
+        elem.nodeType !== Node.ELEMENT_NODE ||
+        elem.tagName?.toLowerCase() !== 'a' ||
+        !isVisible(elem)) return undefined
+      return typeof elem.onclick === 'function' || interactiveElems.has(elem) || elem.hasAttribute('href')
+    } catch (err) {
+      warn(63, {
```
metal-messiah: is there any worry of this warning happening a lot on certain apps?

src/features/generic_events/aggregate/user-actions/aggregated-user-action.js
```diff
@@ -37,4 +40,19 @@ export class AggregatedUserAction {
...
+
+  /**
+   * Determines if the click is a dead click, which is defined as a click on an element that is not interactive.
+   * @returns {boolean}
+   */
+  isDeadClick () {
+    try {
+      return this.event.type === 'click' && (this.hasActLink === false ||
+        (this.event.target && this.event.target.nodeType === Node.ELEMENT_NODE && isVisible(this.event.target) &&
+          this.event.target.tagName === 'INPUT' && this.event.target.type === 'text' && this.event.target.readOnly))
```
metal-messiah: This `if` statement is pretty hard to read what is actually being checked, FWIW.  Maybe easier to read if its break it up into individual `return false` statements.  Up to you on that one

metal-messiah: Could also maybe try to reduce the amount of times `this.event.target` is repeated, that might make it easier to read as well

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -24,12 +24,13 @@ export class UserActionsAggregator {
...
    * Process the event and determine if a new aggregation set should be made or if it should increment the current aggregation
    * @param {Event} evt The event supplied by the addEventListener callback
+   * @param {Array<string>} [targetFields=[]] specifies which fields to gather from the nearest element in the path
    * @returns {AggregatedUserAction|undefined} The previous aggregation set if it has been completed by processing the current event
    */
   process (evt, targetFields) {
     if (!evt) return
-    const { selectorPath, nearestTargetFields } = getSelectorPath(evt, targetFields)
-    const aggregationKey = getAggregationKey(evt, selectorPath)
+    const selectorInfo = getSelectorPath(evt, targetFields)
```
metal-messiah: I think this is supposed to be `getSelectorPathInfo`

metal-messiah: nvm

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -48,20 +49,18 @@ export class UserActionsAggregator {
  * Generates a selector path for the event, starting with simple cases like window or document and getting more complex for dom-tree traversals as needed.
  * Will return a random selector path value if no other path can be determined, to force the aggregator to skip aggregation for this event.
  * @param {Event} evt
- * @returns {string}
+ * @param {Array<string>} [targetFields=[]] specifies which fields to gather from the nearest element in the path
+ * @returns {{ selectorPath: (undefined|string), nearestTargetFields: {}, hasActLink: boolean}}
  */
 function getSelectorPath (evt, targetFields) {
```
metal-messiah: we should maybe consider renaming `getSelectorPath` and `gatherSelectorPathInfo` to something more informative of what it is doing, since it is returning a lot more than the selector path these days

src/features/generic_events/aggregate/user-actions/utils.js
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Checks if the element is visible in the DOM
+ * @param elem
+ * @returns {boolean|boolean}
+ */
+export function isVisible (elem) {
```
metal-messiah: is it possible to "click" on an element that is hidden with these criteria?

src/loaders/configure/configure.js
```diff
@@ -56,6 +56,8 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
 
     // Update if soft_navigations is allowed to run AND part of this agent build, used to override old spa functions.
     agent.runSoftNavOverSpa &&= (updatedInit.soft_navigations.enabled === true && updatedInit.feature_flags.includes('soft_nav'))
+
+    agent.runUserFrustrations ||= updatedInit.feature_flags.includes('user_frustrations')
```
metal-messiah: is there a need to keep this in the loader portion of the agent?

src/common/dom/selector-path.js
ptang-nr: I wanted to avoid evaluating dead click on elements that are not visible, since it would not be a normal/human usage.

ptang-nr: Separately, I also found I needed a way to tell apart when a click occurred on a child span inside a link vs. a click that occurred on a standalone span.  So I was hoping to use `undefined` to effectively ignore an element for dead click eval.

| hasActLink | Description | Evaluation |
| --- | --- | --- |
| true | link was interactive | not a dead click |
| false | link was not interactive| a dead click |
| undefined | no (visible) link detected | not a dead click |

ptang-nr: 🤔 Maybe I only need to know if an element is interactive or not.  Let me try something out.

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
ptang-nr: Yeah, I was struggling a bit with the naming.  We are building up the path and at the same time analyzing the path for metadata.  

(Gonna try and brainstorm some more.  Also open to suggestions!)

src/features/generic_events/aggregate/user-actions/utils.js
ptang-nr: No, at least not by a human user.  During the spike, I found it was still possible to trigger the 'click' events while an element is hidden, so was thinking I need to check for visibility to ensure it's a legitimate user interaction.  Not sure if I'm overthinking it 😬

src/loaders/configure/configure.js
ptang-nr: No, I was only mostly following `soft_nav`.  Moved the check to the UA agg.

src/common/dom/selector-path.js
ptang-nr: I was hoping I could get away with just keeping track of whether there were any interactive elements, but I still needed to differentiate between a true dead click vs. a standalone, non-interactive element.  So updated to keep track of `hasInteractiveElems`, `hasVisibleLink`, and `hasVisibleTextbox` (my reasoning: it's probably better to keep the metadata info more straightforward, and then interpret if it's a dead click)

src/common/dom/selector-path.js
ptang-nr: I was initially mostly worried an error could throw from the visibility check.  Ended up dropping the try/catch after beefing up the visibility check.

src/features/generic_events/aggregate/user-actions/utils.js
metal-messiah: Would it be simpler to check the event's `.isTrusted` property before running the DOM tree parsing?

ptang-nr: Ooh, I did not know about `.isTrusted`.  That would fit the bill perfectly! 

(ok, so I tried using `.isTrusted` but ran into a new problem where the dead clicks e2e test no longer works.  From searching online, it sounds like in theory I can override the property but I have yet to get a working override.  In light of this, I'm opting to drop the visibility check as it's not really the right thing to check and more of an edge case)

PR #1513, opened 2025-06-20
feat: Harvest early

metal-messiah:
---

### DIFF
--- a/docs/warning-codes.md
+++ b/docs/warning-codes.md
@@ -125,3 +125,5 @@
 `Timestamps must be non-negative and end time cannot be before start time.`
 ### 62 
 `Timestamp must be a unix timestamp greater than the page origin time`
+### 63
+`Single event was larger than the maximum allowed payload size`
--- a/src/common/config/info.js
+++ b/src/common/config/info.js
@@ -2,6 +2,7 @@
  * Copyright 2020-2025 New Relic, Inc. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
+import { stringify } from '../util/stringify'
 import { defaults as nrDefaults } from '../window/nreum'
 import { getModeledObject } from './configurable'
 
@@ -42,6 +43,7 @@ const InfoModel = {
   product: undefined,
   extra: undefined,
   jsAttributes: {},
+  jsAttributesBytes: 0,
   userAttributes: undefined,
   atts: undefined,
   transactionName: undefined,
@@ -57,5 +59,22 @@ export function isValid (info) {
 }
 
 export const mergeInfo = (info) => {
-  return getModeledObject(info, InfoModel)
+  const modeledObject = getModeledObject(info, InfoModel)
+
+  // proxy jsAttributes to calculate its size when changed
+  const originalJsAttributes = modeledObject.jsAttributes
+  modeledObject.jsAttributes = new Proxy(originalJsAttributes, {
+    set (target, prop, value) {
+      target[prop] = value
+      modeledObject.jsAttributesBytes = stringify(target).length
+      return true
+    },
+    deleteProperty (target, prop) {
+      const deleted = delete target[prop]
+      modeledObject.jsAttributesBytes = stringify(target).length
+      return deleted
+    }
+  })
+
+  return modeledObject
 }
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -89,7 +89,7 @@ export class Aggregate extends AggregateBase {
       value,
       attrs
     }
-    this.events.add(timing)
+    this.handleData(timing, true)
 
     handle('pvtAdded', [name, value, attrs], undefined, FEATURE_NAMES.sessionTrace, this.ee)
 
--- a/src/features/utils/aggregate-base.js
+++ b/src/features/utils/aggregate-base.js
@@ -18,6 +18,8 @@ import { EventBuffer } from './event-buffer'
 import { handle } from '../../common/event-emitter/handle'
 import { SUPPORTABILITY_METRIC_CHANNEL } from '../metrics/constants'
 import { EventAggregator } from '../../common/aggregate/event-aggregator'
+import { IDEAL_PAYLOAD_SIZE } from '../../common/constants/agent-constants'
+import { stringify } from '../../common/util/stringify'
 
 export class AggregateBase extends FeatureBase {
   /**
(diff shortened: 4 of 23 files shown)


src/features/utils/aggregate-base.js
```diff
@@ -60,17 +62,51 @@ export class AggregateBase extends FeatureBase {
...
+    const getSMTag = (type) => {
+      return `${this.featureName}/Harvest/${type}/Seen`
+    }
+    // this will be false when the event was dropped due to exceeding max size
+    if (!this.events.add(data)) {
+      if (!this.events.isEmpty()) {
+        // if the buffer is not empty, we can try to harvest what's in there before trying to add the new event again
+        this.reportSupportabilityMetric(getSMTag('Max'), getBytesEstimate())
+        this.agentRef.runtime.harvester.triggerHarvestFor(this)
+        this.handleData(data) // try to add the event again after harvest
```
cwli24: It's not obvious the `triggerHarvest` affects or clears `this.events`, and there's potential for failure in it doing so. Meaning this could cause infinite recursion at surface perspective. I think this secondary add after harvest attempt should just be an add rather than a recursive call.

src/features/utils/aggregate-base.js
```diff
@@ -60,17 +62,51 @@ export class AggregateBase extends FeatureBase {
...
     Its easier to just keep an empty event buffer in place. */
       default:
-        this.events = new EventStoreManager(this.agentRef, EventBuffer, entityGuid, this.featureName)
+        this.events = new EventStoreManager(this.agentRef, EventBuffer, entityGuid, this)
         break
     }
   }
 
+  handleData (data, customAttributesAreSeparate = false) {
+    /** aggregated timeslice metrics dont support byte counting at the moment */
```
metal-messiah: check `this.blocked` for safety

docs/warning-codes.md
```diff
@@ -125,3 +125,5 @@
 `Timestamps must be non-negative and end time cannot be before start time.`
 ### 62 
 `Timestamp must be a unix timestamp greater than the page origin time`
+### 63
+`Single event was larger than the maximum allowed payload size`
```
metal-messiah: Change verbiage here to account for all cases that an event couldnt be added

src/features/utils/aggregate-base.js
metal-messiah:
```suggestion
        if (!this.add(data)) {
         // SMS
```

src/common/config/info.js
```diff
@@ -57,5 +59,22 @@ export function isValid (info) {
...
 export const mergeInfo = (info) => {
-  return getModeledObject(info, InfoModel)
+  const modeledObject = getModeledObject(info, InfoModel)
+
+  // proxy jsAttributes to calculate its size when changed
+  const originalJsAttributes = modeledObject.jsAttributes
+  modeledObject.jsAttributes = new Proxy(originalJsAttributes, {
+    set (target, prop, value) {
+      target[prop] = value
+      modeledObject.jsAttributesBytes = stringify(target).length
```
cwli24: change this to increment & decrement based on stringifying the key-value pair rather than the whole object for performance purposes.

src/common/harvest/harvester.js
```diff
@@ -35,6 +35,14 @@ export class Harvester {
       this.initializedAggregates.forEach(aggregateInst => this.triggerHarvestFor(aggregateInst, { isFinalHarvest: true }))
       /* This callback should run in bubble phase, so that that CWV api, like "onLCP", is called before the final harvest so that emitted timings are part of last outgoing. */
     }, false)
+
+    this.agentRef.ee.on('event-buffered', (featureAgg) => {
```
cwli24: This system can be reworked to avoid ee, so do consider that approach.

- This func can be a helper method _somewhere_ (maybe this isn't the best place) that's called directly when an event is added
- The process of adding an event should pass the `evaluateHarvest` returned info to this helper
- Since `EventBuffer.add` is modified specifically, `(!this.events || this.events.StorageClass !== EventBuffer)` check wouldn't be necessary if the work done here is executed in there instead since it's implied that `this.events` would exist and it's that of an EventBuffer.

There's a path here to lessen passing featureAgg around and making this more code than it needs to be.

docs/warning-codes.md
```diff
@@ -125,3 +125,7 @@
 `Timestamps must be non-negative and end time cannot be before start time.`
 ### 62 
 `Timestamp must be a unix timestamp greater than the page origin time`
+### 63
+`A single event was larger than the maximum allowed payload size`
+### 64
```
cwli24: This isn't a helpful message for readers, and it really isn't necessary in the code based on usage (e.g. `agent.info` being undefined would cause major issues elsewhere in the program flow).

src/features/utils/event-store-manager.js
```diff
@@ -101,6 +101,8 @@ export class EventStoreManager {
       // should be the key that is honored to get the event store to ensure a valid connection was made.
       if (targetEntityGuid === DEFAULT_KEY) return
       const targetApp = this.entityManager.get(targetEntityGuid)
+      // const data = eventStore.get(opts)
+      // if (targetApp && (data.length || Object.keys(data).some(key => data[key].length))) allPayloads.push({ targetApp, data })
```
ellisong: remove commented code here

ellisong (commented):
see comment

src/features/page_view_timing/aggregate/index.js
```diff
@@ -135,6 +137,7 @@ export class Aggregate extends AggregateBase {
 
   // serialize array of timing data
   serializer (eventBuffer) {
+    if (!eventBuffer?.length === 0) return ''
```
ptang-nr: Can this be simplified to `!eventBuffer?.length` or `eventBuffer?.length === 0`?  (Assuming we're trying to return early if eventBuffer is empty)

src/features/utils/aggregate-base.js
```diff
@@ -76,15 +78,17 @@ export class AggregateBase extends FeatureBase {
...
    */
-  evaluateHarvest () {
-    const output = { shouldHarvestEarly: false, estimatedSize: 0 }
-    if (!this.events || this.events.StorageClass !== EventBuffer) return output
-    output.estimatedSize = this.events.byteSize() + (this.customAttributesAreSeparate ? this.agentRef.runtime.jsAttributesMetadata.bytes : 0)
-    output.shouldHarvestEarly = output.estimatedSize > IDEAL_PAYLOAD_SIZE
-    return output
+  decideEarlyHarvest () {
+    if (!this.canHarvestEarly) return
+    const estimatedSize = this.events.byteSize() + (this.customAttributesAreSeparate ? this.agentRef.runtime.jsAttributesMetadata.bytes : 0)
```
cwli24, line 87: I know this is only really called by EventBuffer, since this exist in the aggregate-base class, you should ensure byteSize for EventAggregator doesn't cause a problem later. Either:

1. change `EventStoreManager.byteSize` to `return this.#getEventStore(targetEntityGuid).byteSize?.() || 0`, OR
2. add a `byteSize` func in `EventAggregator` class that just returns 0

PR #1518, opened 2025-06-27
feat: Emit SM on rum response invalid timestamps

ellisong:
Emit a supportability metric on receiving an invalid timestamp in the rum call response.
---

### Overview

Whenever processRumRequest() in the time keeper inside the PageViewAggregate gives a timestamp back from the server that is behind the agent, we consider the timestamp invalid which causes data to be discarded. This new SM named `Generic/TimeKeeper/InvalidTimestamp/Seen` with a value of the time difference will give us an idea on how often this is happening before considering a solution to improve timestamp accuracy.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-341093

### Testing

Make sure component tests pass.


src/features/page_view_event/aggregate/index.js
```diff
@@ -152,9 +152,17 @@ export class Aggregate extends AggregateBase {
...
     try {
+      const wasReady = this.agentRef.runtime.timeKeeper.ready
+
       // will do nothing if already done
       this.agentRef.runtime.timeKeeper.processRumRequest(xhr, this.rumStartTime, rumEndTime, app.nrServerTime)
       if (!this.agentRef.runtime.timeKeeper.ready) throw new Error('TimeKeeper not ready')
+
+      // If timeKeeper's origin time is ahead of nrServerTime, then the timestamp is invalid. Report a supportability metric.
+      const timeDiff = this.agentRef.runtime.timeKeeper.correctedOriginTime - app.nrServerTime
+      if (wasReady && timeDiff > 0) {
```
ptang-nr, line 163: Minor - I think we can just check `timeDiff` since there's already a timekeeper readiness check on line 159

ellisong: I don't want to emit the SM the first time that it's ready.

PR #1522, opened 2025-07-01
chore: DEM Spammer

metal-messiah:
Create an action runner to load test downstream services with mock data.
---


.github/actions/consumer-load-test/action.yml
```diff
@@ -0,0 +1,43 @@
+# This composite action is used to load test DEM Consumers and NR1 UIs
+
+name: 'Consumer Load Test'
+
+inputs:
```
ellisong, line 5: Do the rest of the inputs need `type` definitions, such as for `license-key`, `app-id`, `entity-guid`, `minutes` ?

.github/actions/consumer-load-test/index.cjs
```diff
@@ -0,0 +1,105 @@
...
+        payloadMetadata
+      })
+
+      /** SNAPSHOT HARVEST */
+      if (args.sessionReplay) await harvestSessionReplay({
+        licenseKey,
+        appId,
+        timeNow,
+        offset,
+        harvestCount,
```
ellisong: harvestCount doesn't add up here, snapshot harvest will always be behind by one compared to the mutation harvest.

.github/actions/consumer-load-test/action.yml
metal-messiah: if omitted, GHA inputs default to `string` type inputs

.github/actions/consumer-load-test/index.cjs
metal-messiah: ty

ellisong (approved):
LGTM

PR #1526, opened 2025-07-03
fix: report empty previousUrl as undefined

metal-messiah:
The `previousUrl` attribute on BrowserInteraction events was reported as an empty string is cases where document.referrer is empty.  This caused URL grouping to apply grouping on an empty string, leading to meaningless groups being appended to previousUrl.  `previousUrl` is now reported as `undefined` to allow URL grouping to skip empty URLs.
---

### Overview
This is behavior required to support Browser User Journeys.

### Related Issue(s)
https://newrelic.slack.com/archives/C08235DFQU9/p1750874467526829

### Testing
Tests have been updated to account for this behavior.


ptang-nr (dismissed):
LGTM

PR #1527, opened 2025-07-07
fix: Add monkey-patching warnings

metal-messiah:
Added detection and warnings for when expected globals have been monkey-patched before the agent has been executed on the page.  This can cause unexpected behaviors and the intent is to inform customers of this behavior.
---

### Overview
Should expect a new warning code when expected globals have been modified before the agent has executed.

### Related Issue(s)
NR-435349

### Testing
Tests have been updated

### DIFF
--- a/tests/specs/nr-server-time.e2e.js
+++ b/tests/specs/nr-server-time.e2e.js
@@ -418,6 +418,21 @@ describe('NR Server Time', () => {
       expect(subsequentServerTimeDiff).toEqual(initialServerTimeDiff)
       expect(subsequentSession.localStorage.serverTimeDiff).toEqual(initialSession.localStorage.serverTimeDiff)
     })
+
+    it('should warn if the time globals are monkey-patched', async () => {
+      const url = await browser.testHandle.assetURL('monkey-patched.html')
+      await browser.url(url).then(() => browser.waitForAgentLoad())
+
+      const logs = await browser.execute(function () {
+        return window.logs
+      })
+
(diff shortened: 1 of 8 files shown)


tests/specs/nr-server-time.e2e.js
```diff
@@ -418,6 +418,21 @@ describe('NR Server Time', () => {
...
+
+    it('should warn if the time globals are monkey-patched', async () => {
+      const url = await browser.testHandle.assetURL('monkey-patched.html')
+      await browser.url(url).then(() => browser.waitForAgentLoad())
+
+      const logs = await browser.execute(function () {
+        return window.logs
+      })
+
+      expect(logs[0][0].includes('New Relic Warning') && logs[0][0].includes('63')).toEqual(true) // 63 is the warning code for monkey-patched globals
```
ellisong: This test needs to use error code 64 instead of 63.

tests/specs/nr-server-time.e2e.js
```diff
@@ -418,6 +418,21 @@ describe('NR Server Time', () => {
       expect(subsequentServerTimeDiff).toEqual(initialServerTimeDiff)
       expect(subsequentSession.localStorage.serverTimeDiff).toEqual(initialSession.localStorage.serverTimeDiff)
     })
+
+    it('should warn if the time globals are monkey-patched', async () => {
```
ellisong: This test seems to be testing monkey patching of performance API and the console functions rather than the time globals.

ellisong (changes requested):
see comments

PR #1531, opened 2025-07-24
feat: Add error cause detection

metal-messiah:
Detect and report the cause attribute on captured JavaScriptError events.  If the cause is an instance of an Error, the stack trace will be captured, otherwise the stringified contents of the cause will be reported. This will be queryable in JavaScriptError events under the attribute `cause`.
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-414116?atlOrigin=eyJpIjoiODY2NDUxZTIwMTUxNGQ0YWFkMDU0M2JmMDkxNzBkOWYiLCJwIjoiaiJ9

### Testing
New tests have been added to validate the various cause cases.


ptang-nr (approved):
LGTM

PR #1532, opened 2025-07-28
save history

metal-messiah:
---

### DIFF
--- a/src/features/session_trace/aggregate/index.js
+++ b/src/features/session_trace/aggregate/index.js
@@ -13,7 +13,7 @@ import { applyFnToProps } from '../../../common/util/traverse'
 import { cleanURL } from '../../../common/url/clean-url'
 import { warn } from '../../../common/util/console'
 
-const ERROR_MODE_SECONDS_WINDOW = 30 * 1000 // sliding window of nodes to track when simply monitoring (but not harvesting) in error mode
+const SUPPORTS_PERFORMANCE_OBSERVER = typeof globalScope.PerformanceObserver === 'function'
 /** Reserved room for query param attrs */
 const QUERY_PARAM_PADDING = 5000
 export class Aggregate extends AggregateBase {
@@ -29,8 +29,8 @@ export class Aggregate extends AggregateBase {
     this.everHarvested = false
     /** If the harvest module is harvesting */
     this.harvesting = false
-    /** TraceStorage is the mechanism that holds, normalizes and aggregates ST nodes.  It will be accessed and purged when harvests occur */
-    this.events = new TraceStorage(this)
+    /** TraceStorage is a middleware that decides how to format data before passing events to `this.events` */
+    this.traceStorage = new TraceStorage(this)
 
     /* This agg needs information about sampling (sts) and entitlements (st) to make the appropriate decisions on running */
     this.waitForFlags(['sts', 'st'])
@@ -64,9 +64,9 @@ export class Aggregate extends AggregateBase {
       })
 
       if (typeof PerformanceNavigationTiming !== 'undefined') {
-        this.events.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
+        this.traceStorage.storeTiming(globalScope.performance?.getEntriesByType?.('navigation')[0])
       } else {
-        this.events.storeTiming(globalScope.performance?.timing, true)
+        this.traceStorage.storeTiming(globalScope.performance?.timing, true)
       }
     }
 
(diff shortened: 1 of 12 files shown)


ptang-nr (approved):
Voting +1, may be good to get another set of eyes on it.  I mostly studied the changes in storage.js to learn what ST was doing :)  Not sure if we want to make any callouts in the release notes about changes to "smearing" (iiuc, it now happens when data is added, whereas it used to happen when data was taken out).

src/features/session_trace/aggregate/index.js
```diff
@@ -29,8 +29,8 @@ export class Aggregate extends AggregateBase {
     this.everHarvested = false
     /** If the harvest module is harvesting */
     this.harvesting = false
-    /** TraceStorage is the mechanism that holds, normalizes and aggregates ST nodes.  It will be accessed and purged when harvests occur */
-    this.events = new TraceStorage(this)
+    /** TraceStorage is a middleware that decides how to format data before passing events to `this.events` */
+    this.traceStorage = new TraceStorage(this)
```
cwli24, line 32: I'm not sure why you didn't run into issues but `events` is the common name for all feature data buffers, so this shouldn't be changed.

The base AggregateBase helpers references .events.

src/features/session_trace/aggregate/index.js
```diff
@@ -111,22 +111,31 @@ export class Aggregate extends AggregateBase {
       this.abort(3)
       return
     }
+
+    // stage timings if PO isn't supported, this checks resourcetiming buffer every harvest.
+    if (!SUPPORTS_PERFORMANCE_OBSERVER) {
```
cwli24: We shouldn't add backwards version oriented code (browser versions that didn't support POs).

On another note, I don't believe there are existing tests that checks for duplicated resource across harvests which is relevant to this kind of change.

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -265,38 +200,44 @@ export class TraceStorage {
...
-  save () {
-    this.#backupTrace = this.trace
+  trimSTNsByTime (lookbackDuration = ERROR_MODE_SECONDS_WINDOW) {
+    this.parent.events.clear({
+      clearBeforeTime: Math.max(now - lookbackDuration, 0),
+      timestampKey: 'e'
+    })
   }
 
-  get () {
```
cwli24, line 278: What happened to get? I'm not sure making ST so specialized like SR when it had a common interface to all the other features is the right move here. This pattern deviation is more devl bug prone. It also seems much beyond the scope of the PR ticket.

src/features/session_trace/aggregate/index.js
metal-messiah: not sure what you mean here.  The point of this work was to allow ST to start using the base `this.events` from the aggregate base file.  Before, it was being overridden with the TraceStorage class.

src/features/session_trace/aggregate/index.js
metal-messiah: `We shouldn't add backwards version oriented code (browser versions that didn't support POs).`
this is a direct port-over from the previous code.  We can remove if you think its no longer necessary

src/features/session_trace/aggregate/trace/storage.js
metal-messiah: Because Storage.js is no longer a state retrieval device.  it simply is a formatter for storing in the native `this.events` now, which means there's no point for things like `.get(), .save(), .isEmpty(), etc.`. I think we need to have a meeting on this to ensure that the changes are understood as to what they are doing.

src/features/session_trace/aggregate/trace/storage.js
```diff
@@ -21,125 +17,99 @@ const ignoredEvents = {
...
+const SMEARABLES = {
+  typing: 'typing',
+  scroll: 'scroll',
+  mousing: 'mousing',
+  touching: 'touching'
+}
+
+const GAPS = {
+  [SMEARABLES.typing]: 1000, // 1 second gap between typing events
+  [SMEARABLES.scrolling]: 100, // 100ms gap between scrolling events
```
metal-messiah, line 32: Fix to `scroll`

src/features/session_trace/aggregate/index.js
cwli24: Yeah let's just get rid of this. If we don't support IE at this point, no point to supporting code for features IE lacks.

ptang-nr (dismissed):
Voting 👍

src/features/utils/event-buffer.js
cwli24, line 1: Are there tests to cover the `merge` in tracestorage? I may be missing it on look over

cwli24 (dismissed):

src/features/utils/event-buffer.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1532/files#diff-14b5b1eca7ace0a715524c8f4a16a1899cfabb87e1a8f82f7e2e1354d20624b2R125-R144 is the existing spec test 

https://github.com/newrelic/newrelic-browser-agent/pull/1532/commits/806f6688edd492ec62787e3691ab80dbf62db31f for the new unit tests

PR #1533, opened 2025-07-28
feat: Add useConsentModel functionality

ellisong:
Adds use_consent_mode init property and functionality. Adds the consent() API call. The consent model, if enabled, disallows agent harvesting unless consent is given through the consent() API call.
---

### Overview

Adds useConsentModel init property and functionality (defaults to false). When turning `useConsentModel` on, the agent will not harvest (send over the network) until given "consent". Consent is given by the `consent` value in the session entity's local storage being set to `true`.

Adds an API method called `consent(accept?: boolean)` that changes the `consent` value in the session entity's local storage. Calling `true` or `undefined` for the argument grants consent, calling `false` in consent() rejects consent.

Agent features are still manually or automatically started regardless of consent being enabled or not, but they won't send data over the network unless consent is granted.

Also adds a .nvmrc file for node version 22.11

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-438120
JIRA: https://new-relic.atlassian.net/browse/NR-438139

### Testing

Tests are provided that tests the agent harvest blocking if the consent model is enabled in various situations. The default case (consent model not being enabled) is not tested since that's already covered.

### DIFF
--- a/src/features/page_view_event/aggregate/index.js
+++ b/src/features/page_view_event/aggregate/index.js
@@ -27,7 +27,7 @@ export class Aggregate extends AggregateBase {
   constructor (agentRef) {
     super(agentRef, CONSTANTS.FEATURE_NAME)
 
-    this.sentRum = false // flag to only call sendRum() once if successful, properties are by applicationID
+    this.sentRum = false // flag to facilitate calling sendRum() once externally (by the consent API in agent-session.js)
 
     this.timeToFirstByte = 0
     this.firstByteToWindowLoad = 0 // our "frontend" duration


src/common/harvest/harvester.js
```diff
@@ -56,6 +56,7 @@ export class Harvester {
    */
   triggerHarvestFor (aggregateInst, localOpts = {}) {
     if (aggregateInst.blocked) return false
+    if (this.agentRef.init.useConsentModel === true && localStorage.getItem('NrBrowserAgentConsent') !== 'true') return false
```
metal-messiah: Is there a reason to store this data under a different key in LS than the session entity? (`NRBA_SESSION`)

metal-messiah: We may want to evaluate whether or not reading from local storage for every feature on every harvest is going to be a performance burden (vs the way the session entity handles it by sync'ing state locally)

ellisong: No particular reason for the different key. It can be moved under `NRBA_SESSION`

src/common/session/session-entity.js
```diff
@@ -32,7 +32,8 @@ const model = {
   loggingMode: LOGGING_MODE.OFF,
   serverTimeDiff: null, // set by TimeKeeper; "undefined" value will not be stringified and stored but "null" will
   custom: {},
-  numOfResets: 0
+  numOfResets: 0,
+  consent: false
```
metal-messiah: this looks good, but with the session entity we will just need to make sure that `consent`'s value get passed to the next session object when the session "resets" https://github.com/newrelic/newrelic-browser-agent/pull/1533/commits/c486e4dc3e075422d44fcadf76049d2d2b740794#diff-b3e27fb4a8459e7a550bd391323f3a0b7fc10d4f028b019cb2154dc921c1519aR225

src/common/harvest/harvester.js
```diff
@@ -56,6 +56,7 @@ export class Harvester {
    */
   triggerHarvestFor (aggregateInst, localOpts = {}) {
     if (aggregateInst.blocked) return false
+    if (this.agentRef.init.useConsentModel && !this.agentRef.runtime.session.state.consent) return false
```
metal-messiah: I think we need to think through how this will affect the page_view_event aggregate feature.  That feature only ever calls `triggerHarvestFor` once, and blocks the other features.  When the consent API is called, we probably need to have the PVE feature try to harvest again.

To test this, we should add an e2e test that:
1. checks the expected behavior of the consent API when called `before` the page has loaded
2. checks the expected behavior of the consent API when called `after` the page has loaded
3. checks the expected behavior if the consent API is never called
4. checks the expected behavior in tandem with the `autoStart/.start()` behaviors of the agent
  - start called before consent
  - start called after consent
6. checks the expected behavior across page loads once consent has been given
7. checks the expected behavior when consent is `revoked`

cwli24: I'm also not comfortable with having harvest on/off as-is. If it's off, and we continue to buffer (global abort isn't done), this could cause memory/performance issues for customers. However, there may also be cases like near new session or page loads wherein we need to buffer that data for some time rather than do a global abort. The scenarios here need to be evaluated.

Also note that `.session` can be undefined, e.g. if session tracking is off or manager failed to initialize for w/e reason. So that requires an optional chaining `?.`

src/loaders/api/consent.js
```diff
@@ -0,0 +1,18 @@
...
+import { warn } from '../../common/util/console'
+
+export function setupConsentAPI (agent) {
+  setupAPI(CONSENT, function (accept) {
+    if (accept !== undefined && typeof accept !== 'boolean') {
+      warn(65, typeof accept)
+      return
+    }
+
+    agent.runtime.session.state.consent = accept === undefined ? true : accept
```
cwli24: Same here: `session` may not be defined

src/common/harvest/harvester.js
ellisong: I'll plan out the expected behavior and add tests for them. We have time due to some pending reviews from legal and compliance.

src/loaders/api/consent.js
ellisong: will fix

ellisong: need to do this asynchronously, registering an event handler

metal-messiah:
can you please add type-check conditions for your new API method to `api.test-d.ts` to ensure that the typings worked correctly.

metal-messiah:
we also haven't been very good about updating `init-types.js` but we probably should start to update that whenever we update the init.js file

src/common/config/init-types.js
```diff
@@ -84,6 +84,7 @@
  * @property {boolean} [spa.enabled] - Turn on/off the single page application feature (on by default). NOTE: the SPA feature is deprecated and under removal procedure.
  * @property {boolean} [spa.autoStart] - If true, the agent will automatically start the single page application feature. Otherwise, it will be in a deferred state until the `start` API method is called.
  * @property {boolean} [ssl] - If explicitly false, the agent will use HTTP instead of HTTPS. This setting should NOT be used.
+ * @property {boolean} [useConsentModel] - If true, the agent will use the consent model for whether to allow or disallow data harvest.
```
metal-messiah: can you update `agentoptions.test-d.ts` to check for this property

src/common/harvest/harvester.js
```diff
@@ -56,6 +56,7 @@ export class Harvester {
    */
   triggerHarvestFor (aggregateInst, localOpts = {}) {
     if (aggregateInst.blocked) return false
+    if (this.agentRef.init.useConsentModel && !this.agentRef.runtime?.session?.state?.consent) return false
```
metal-messiah: Just want to confirm here that consent model will not work (e.g. agent will never send data) if:
- enduser has local storage disabled 
- customer has session tracking disabled completely

ellisong: I'll add additional testing to be sure, but if they have local storage disabled then consent should not be possible to enable. The second one I'll need some context on how to disable session tracking.

src/common/config/init-types.js
ellisong: Sure thing. I didn't know we had these test-d files before.

src/features/utils/agent-session.js
```diff
@@ -56,6 +56,18 @@ export function setupAgentSession (agentRef) {
...
   }, 'session', sharedEE)
 
+  registerHandler('api-consent', (accept) => {
+    agentRef.runtime.session.write({ consent: accept === undefined ? true : accept })
+
+    // call sendRum if it wasn't called yet
+    const target = { licenseKey: agentRef.info.licenseKey, applicationID: agentRef.info.applicationID }
+    agentRef.features.page_view_event.onAggregateImported.then((loaded) => {
+      if (loaded) {
+        agentRef.features.page_view_event.featAggregate.sendRum(agentRef.info.jsAttributes, target)
```
metal-messiah: you could probably skip a bit of work/size here since the `sendRum` mechanism defaults to the data you are providing as arguments, it essentially is the same as 

```javascript
agentRef.features.page_view_event.featAggregate.sendRum()
```

src/loaders/api-base.js
```diff
@@ -226,4 +226,14 @@ export class ApiBase {
   measure (name, options) {
     return this.#callMethod(MEASURE, name, options)
   }
+
+  /**
+   * Accepts or rejects consent when the agent is configured to enable its consent model.
```
metal-messiah:
```suggestion
   * Accepts or rejects consent when the agent is configured to require consent before harvesting
```

src/common/harvest/harvester.js
metal-messiah: I think we should just approach this as notes in the documentation for customers

src/loaders/api-base.js
ellisong: will add in separate commit

src/common/harvest/harvester.js
ellisong: I'll check manually to make sure rather than writing the test out.

ellisong: It seems like the consent model (in Firefox at least) works even with local storage disabled, but it won't send data without consent being accepted. It's like the browser pretends to have local storage for the website but never actually stores it afterward between sessions.

src/features/page_view_event/aggregate/index.js
```diff
@@ -61,6 +64,8 @@ export class Aggregate extends AggregateBase {
    * @param {*} target The target to harvest to - Since we will not know the entityGuid before harvesting, this must be an object directly supplied from the info object or API, not an entityGuid string for lookup with the entityManager - Defaults to { licenseKey: this.agentRef.info.licenseKey, applicationID: this.agentRef.info.applicationID }
    */
   sendRum (customAttributes = this.agentRef.info.jsAttributes, target = { licenseKey: this.agentRef.info.licenseKey, applicationID: this.agentRef.info.applicationID }) {
+    if (this.sentRum[target.applicationID]) return
```
metal-messiah: I do wonder if two agent instances were running on the same page from the same bundle (NPM) if they would get blocked here.  It may be safer to use `agentIdentifier` as the key for `this.sendRum`

ellisong: added a note about this in epic ticket

metal-messiah:
Blocked until API team is out of code freeze and can implement settings changes.

metal-messiah (changes requested):
manually blocking to prevent accidental merge

metal-messiah (approved):

PR #1534, opened 2025-07-29
feat: Add dead click detection via DOM mutations

ptang-nr:
Add dead click detection via DOM mutations on user clicks.
---

### Overview
For a given user click action, start a dead click eval window for 2s.  
- If there's any DOM mutation during that time, then it is NOT a dead click.  
- If there're no DOM mutations within the eval window, then it will be marked as a dead click.  
- If another user action started before the eval window ends, then it is also considered NOT a dead click. 

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-443292

### Testing

- Added dead-clicks unit and e2e tests
- Manually tested on local test-server

### DIFF
--- a/src/features/generic_events/aggregate/index.js
+++ b/src/features/generic_events/aggregate/index.js
@@ -84,7 +84,8 @@ export class Aggregate extends AggregateBase {
                   if (canTrustTargetAttribute(field)) acc[targetAttrName(field)] = String(target[field]).trim().slice(0, 128)
                   return acc
                 }, {})),
-                ...aggregatedUserAction.nearestTargetFields
+                ...aggregatedUserAction.nearestTargetFields,
+                ...(agentRef.init.feature_flags.includes('user_frustrations') && aggregatedUserAction.deadClick && { deadClick: true })
               })
 
               /**
--- a/src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
+++ b/src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
@@ -3,13 +3,27 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { generateSelectorPath } from '../../../../common/dom/selector-path'
-import { OBSERVED_WINDOW_EVENTS } from '../../constants'
+import { FRUSTRATION_TIMEOUT_MS, OBSERVED_WINDOW_EVENTS } from '../../constants'
 import { AggregatedUserAction } from './aggregated-user-action'
+import { Timer } from '../../../../common/timer/timer'
 
 export class UserActionsAggregator {
   /** @type {AggregatedUserAction=} */
   #aggregationEvent = undefined
   #aggregationKey = ''
+  #deadClickTimer = undefined
+  #domObserver = {
+    running: false,
+    instance: undefined
+  }
+
+  constructor () {
+    if (MutationObserver) {
+      this.#domObserver.instance = new MutationObserver(() => {
+        this.#deadClickTimer?.clear()
+      })
+    }
+  }
 
   get aggregationEvent () {
     // if this is accessed externally, we need to be done aggregating on it
@@ -36,10 +50,45 @@ export class UserActionsAggregator {
     } else {
       // return the prev existing one (if there is one)
       const finishedEvent = this.#aggregationEvent
-      // then set as this new event aggregation
+      this.#deadClickCleanup()
+
+      // then start new event aggregation
       this.#aggregationKey = aggregationKey
       this.#aggregationEvent = new AggregatedUserAction(evt, selectorPath, nearestTargetFields)
+      if (evt.type === 'click') {
+        this.#deadClickSetup(this.#aggregationEvent)
+      }
       return finishedEvent
+    }
+  }
+
+  #deadClickSetup (userAction) {
+    if (this.#startObserver()) {
+      this.#deadClickTimer = new Timer({
+        onEnd: () => {
+          userAction.deadClick = true
+          this.#deadClickCleanup()
+        }
+      }, FRUSTRATION_TIMEOUT_MS)
+    }
+  }
+
+  #deadClickCleanup () {
+    this.#domObserver.instance?.disconnect()
+    this.#domObserver.running = false
+    this.#deadClickTimer?.clear()
+  }
+
+  #startObserver () {
+    if (!this.#domObserver.running && this.#domObserver.instance) {
+      this.#domObserver.running = true
+      this.#domObserver.instance.observe(document, {
+        attributes: true,
+        characterData: true,
+        childList: true,
+        subtree: true
+      })
+      return true
     }
   }
 }
--- a/src/features/generic_events/aggregate/user-actions/aggregated-user-action.js
+++ b/src/features/generic_events/aggregate/user-actions/aggregated-user-action.js
@@ -15,6 +15,7 @@ export class AggregatedUserAction {
     this.rageClick = undefined
     this.nearestTargetFields = nearestTargetFields
     this.currentUrl = cleanURL('' + location)
+    this.deadClick = false
   }
 
   /**
--- a/src/features/generic_events/constants.js
+++ b/src/features/generic_events/constants.js
@@ -12,6 +12,8 @@ export const OBSERVED_WINDOW_EVENTS = ['focus', 'blur']
 export const RAGE_CLICK_THRESHOLD_EVENTS = 4
 export const RAGE_CLICK_THRESHOLD_MS = 1000
 
+export const FRUSTRATION_TIMEOUT_MS = 2000
+
 export const RESERVED_EVENT_TYPES = ['PageAction', 'UserAction', 'BrowserPerformance']
 
 export const FEATURE_FLAGS = {
--- a/tests/specs/user-frustrations/dead-clicks.e2e.js
+++ b/tests/specs/user-frustrations/dead-clicks.e2e.js
@@ -1,0 +1,51 @@
+import { testInsRequest } from '../../../tools/testing-server/utils/expect-tests'
+
+const loaderTypes = ['full', 'spa'] // 'rum' is excluded as UserActions is not supported
+
+describe('User Frustrations - Dead Clicks', () => {
+  loaderTypes.forEach(loaderType => {
+    it('should decorate dead clicks - ' + loaderType, async () => {
+      const [insightsCapture] = await browser.testHandle.createNetworkCaptures('bamServer', [
+        { test: testInsRequest }
+      ])
+      await browser.url(await browser.testHandle.assetURL('test-builds/vite-react-wrapper/index.html'))
+        .then(() => browser.waitForAgentLoad())
+
+      await browser.execute(function () {
+        const USER_TIMEOUT = 2000
+        const commands = [
+          // [0] - span with no click handler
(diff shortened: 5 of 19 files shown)


ptang-nr:
Tested `dead-clicks.e2e.js` against android from local and passed.
`$ npm run wdio -- -r false -T tests/specs/user-frustrations/dead-clicks.e2e.js -b android@latest

> @newrelic/browser-agent@1.294.0 wdio
> node --max-old-space-size=8192 tools/wdio/bin/cli.js -r false -T tests/specs/user-frustrations/dead-clicks.e2e.js -b android@latest


Execution of 1 workers started at 2025-08-01T18:53:39.027Z

Current version of Node Tunnel:  4.0.9
Verifying credentials
Auth succeeded
Checking for updates
Binary already at latest version
Starting tunnel
Tunnel successfully initiated. You can start testing now
[0-0] RUNNING in Pixel 7 - file:///tools/wdio/plugins/mocha-globals/globals.mjs, file:///tests/specs/user-frustrations/dead-clicks.e2e.js
[0-0] 2025-08-01T18:54:40.201Z INFO test-logger: Executing in: Pixel 7 16
[0-0] 2025-08-01T18:54:41.489Z INFO test-logger: Running Test: User Frustrations - Dead Clicks.should decorate dead clicks - full
[0-0] 2025-08-01T18:55:07.762Z INFO test-logger: Running Test: User Frustrations - Dead Clicks.should decorate dead clicks - spa
[0-0] PASSED in Pixel 7 - file:///tools/wdio/plugins/mocha-globals/globals.mjs, file:///tests/specs/user-frustrations/dead-clicks.e2e.js

Spec Files:      1 passed, 1 total (100% completed) in 00:02:01  
`

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -2,14 +2,28 @@
...
+  #deadClickTimer = undefined
+  #domObserver = {
+    running: false,
+    instance: undefined
+  }
+
+  constructor () {
+    if (MutationObserver) {
+      this.#domObserver.instance = new MutationObserver(() => {
+        this.#deadClickTimer?.clear()
```
metal-messiah: Should we call `#deadClickCleanup` here instead so that we can also shut off the mutation observer since we know it is no longer needed?

tests/unit/common/dom/selector-path.test.js
metal-messiah, line 1: Can we extend this unit test to check the cases for detection of `hasButton: <boolean>, hasLink: <boolean>`

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
cwli24: and does deadClick need to be set here explicitly to `= false`? Or would that affect what we talked about in that if any click as part of a sequence results in a dead click, it stays a dead click via L69?

ptang-nr: > does deadClick need to be set here explicitly to = false?

@cwli24 I want to say we shouldn't need to. After the group huddle, the logic was amended to keep the UA as a "live" click instead.  So currently in a sequence of clicks, if any of them caused a DOM mutation to occur before the timer finished, the aggregated UA stays as a "live" click (`deadClick` is initialized as `false` in `AggregatedUserAction` constructor and stays as `false`. It will be set to `true` only if the timer finishes counting down).

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -2,14 +2,27 @@
...
   #aggregationEvent = undefined
   #aggregationKey = ''
+  #ufEnabled = false
+  #deadClickTimer = undefined
+  #domObserver = {
+    instance: undefined
+  }
+
+  constructor (userFrustrationsEnabled) {
+    if (userFrustrationsEnabled && MutationObserver) {
```
metal-messiah: we may need to check that we are in browser scope here.  Accessing the MutationObserver global on a worker thread would throw errors.

I'd suggest either using the original nreum.MO and/or check `isBrowserScope` as part of this statement/class

src/common/dom/selector-path.js
```diff
@@ -5,13 +5,14 @@
...
- * @param {boolean} includeId
- * @param {boolean} includeClass
- * @returns {string|undefined}
+ * @param {Array<string>} [targetFields=[]] specifies which fields to gather from the nearest element in the path
+ * @returns {{path: (undefined|string), nearestFields: {}, hasButton: boolean, hasLink: boolean}}
  */
-export const generateSelectorPath = (elem, targetFields = []) => {
-  if (!elem) return { path: undefined, nearestFields: {} }
+export const analyzeElemPath = (elem, targetFields = []) => {
+  const result = { path: undefined, nearestFields: {}, hasButton: false, hasLink: false }
```
metal-messiah, line 18: Nitpick: this object is never mutated for `path` or `nearestFields` so it feels weird to keep it organized like this and later just overridden with spread operators.

perhaps line 34 and 56 should be modified to update the result object instead of standalone vars.

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -28,40 +41,76 @@ export class UserActionsAggregator {
...
-  // if still no selectorPath, generate one from target tree that includes elem ids
-  else {
-    const { path, nearestFields } = generateSelectorPath(evt.target, targetFields)
-    selectorPath = path
-    nearestTargetFields = nearestFields
-  }
-  // if STILL no selectorPath, it will return undefined which will skip aggregation for this event
-  return { selectorPath, nearestTargetFields }
+function gatherSelectorPathInfo (evt, targetFields) {
+  const result = { path: undefined, nearestFields: {}, hasButton: false, hasLink: false }
```
metal-messiah: Nitpick: Sort of the same comment about the same `result` object above.  Maybe this should be passed into `analyzeElemPath` so we aren't keeping duplicates around.

src/features/generic_events/aggregate/index.js
```diff
@@ -84,7 +84,8 @@ export class Aggregate extends AggregateBase {
                   if (canTrustTargetAttribute(field)) acc[targetAttrName(field)] = String(target[field]).trim().slice(0, 128)
                   return acc
                 }, {})),
-                ...aggregatedUserAction.nearestTargetFields
+                ...aggregatedUserAction.nearestTargetFields,
+                ...(agentRef.init.feature_flags.includes('user_frustrations') && aggregatedUserAction.deadClick && { deadClick: true })
```
metal-messiah: is the featureflag check here still necessary since it is passed into the ua agg constructor and gates the logic there?

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
ptang-nr: Aah, I had wrongly assumed it'd get converted to a falsy value (tried this out and see it's a ReferenceError, TIL!)  Thank youuuu!!!

src/features/generic_events/aggregate/index.js
ptang-nr: Doh, yeah that should be removed! 👍

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
ptang-nr: Yeah, there was a weird kind of coupling going on.  I _think_ maybe it’s really a single method got split into two (?). 

I moved some stuff around and it feels a little better.  Let me know if you continue to see something weird here and maybe we can mob a bit on it :+1:

src/common/dom/selector-path.js
```diff
@@ -4,62 +4,69 @@
  */
 
 /**
- * Generates a CSS selector path for the given element, if possible
+ * Generates a CSS selector path for the given element, if possible.
  * Also gather metadata about the element's nearest fields, and whether there are any links or buttons in the path.
+ *
+ * Starts with simple cases like window or document and progresses to more complex dom-tree traversals as needed.
+ * Will return selectorPath: undefined if no other path can be determined.
```
metal-messiah:
```suggestion
 * Will return path: undefined if no other path can be determined.
```

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -2,14 +2,28 @@
...
+import { Timer } from '../../../../common/timer/timer'
+import { gosNREUMOriginals } from '../../../../common/window/nreum'
 
 export class UserActionsAggregator {
   /** @type {AggregatedUserAction=} */
   #aggregationEvent = undefined
   #aggregationKey = ''
+  #ufEnabled = false
+  #deadClickTimer = undefined
+  #domObserver = {
```
cwli24: is there a reason domobserver needs to be an object with an instance property instead of just being the observer itself?

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
```diff
@@ -28,40 +42,61 @@ export class UserActionsAggregator {
...
-  let selectorPath; let nearestTargetFields = {}
-  if (OBSERVED_WINDOW_EVENTS.includes(evt.type) || evt.target === window) selectorPath = 'window'
-  else if (evt.target === document) selectorPath = 'document'
-  // if still no selectorPath, generate one from target tree that includes elem ids
-  else {
-    const { path, nearestFields } = generateSelectorPath(evt.target, targetFields)
-    selectorPath = path
-    nearestTargetFields = nearestFields
+  #deadClickSetup (userAction) {
+    if (this.#startObserver()) {
```
cwli24: This is convoluted. The short circuit logic is double nested & not obvious. I suggest you check for the falsey case right here and let startObserver simply do as its name implies rather than actually be "start observer but only if there isn't a dead click timer")

src/features/generic_events/aggregate/user-actions/user-actions-aggregator.js
ptang-nr: It was an artifact from earlier iteration.  It's been removed since then.

(edit: double-checked, yeah, it's gone.  Just missed a fetch)

src/common/url/extract-url.js
```diff
@@ -0,0 +1,17 @@
...
+export function extractUrl (target) {
+  if (typeof target === 'string') {
+    return target
+  } else if (typeof target === 'object' && target instanceof gosNREUMOriginals().o.REQ) {
+    return target.url
+  } else if (globalScope?.URL && typeof target === 'object' && target instanceof URL) {
+    return target.href
+  }
+  return undefined
+}
```
metal-messiah, line 17: Might need to fact check me on this but I think this can be reduced some since the instanceof checks are inherently checking if the target is an object.

```suggestion
/**
 * Extracts a URL from various target types.
 * @param {string|Request|URL} target - The target to extract the URL from. It can be a string, a Fetch Request object, or a URL object.
 * @returns {string|undefined} The extracted URL as a string, or undefined if the target type is not supported.
 */
export function extractUrl(target) {
  if (typeof target === 'string')  return target
  else if (target instanceof gosNREUMOriginals().o.REQ) return target.url
  else if (globalScope?.URL && target instanceof URL) return target.href
}
```

PR #1539, opened 2025-08-04
chore: fix mobile test suites

metal-messiah:
IOS tests for latest browser should now run successfully
---

### DIFF
--- a/tools/browsers-lists/utils.mjs
+++ b/tools/browsers-lists/utils.mjs
@@ -1,3 +1,14 @@
+// syncronously generate the list of supported mobile browsers for the test run
+const user = process.env.LT_USERNAME || process.env.LAMBDA_USERNAME
+const key = process.env.LT_ACCESS_KEY || process.env.LAMBDA_ACCESS_KEY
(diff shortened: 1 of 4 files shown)


tools/browsers-lists/utils.mjs
```diff
@@ -1,3 +1,14 @@
+// syncronously generate the list of supported mobile browsers for the test run
+const user = process.env.LT_USERNAME || process.env.LAMBDA_USERNAME
+const key = process.env.LT_ACCESS_KEY || process.env.LAMBDA_ACCESS_KEY
+const mobileDevices = await fetch('https://mobile-api.lambdatest.com/mobile-automation/api/v1/list?region=us', {
```
cwli24, line 4: Rather than doing this, is it possible to update `lt-mobile-supported.json` list instead to have the right name? That seems like a more permanent fix.

metal-messiah: Problem (i thought) here was that the active list of available device names could potentially change run by run, meaning we'd have to do this locally, change the file on each run, and potentially cause merge conflicts in PRs.  Really all we need to do is get the device name to run against since the version itself isn't changing, but I am open to other solutions for sure.

PR #1540, opened 2025-08-05
fix: Remove event buffer inspection event

ellisong:
Removes the event buffer inspection event to avoid a memory leak on using console.log with the `newrelic` window event listener.
---

### Overview

Inspection events can cause an endless loop if a console log is detected and reported in the callback, causing a memory leak. To remedy this, we remove the buffer event as it is not worth the risk to customer applications.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-444882

### Testing

Make sure tests pass, especially inspection events tests.


cwli24:
Should probably also throw a code comment in `add` to prevent future repeat in case someone forget or is new

PR #1544, opened 2025-08-13
chore: Fix LT mobile web tests again

cwli24:
I am once again asking for LT to work
---


ptang-nr (approved):
PR checks looked good (i think there might be a race in one of the tests, so not counting that one).

PR #1548, opened 2025-08-19
feat: Prevent early harvests when in retry period

metal-messiah:
Prevent the agent from harvesting early when being rate limited.  If the ingest service returns a 429, the agent will replace its data and try the harvest again later.  While that is happening, it will no longer be able to harvest early as new data comes in, until it successfully harvests.
---

### Overview
`isRetrying` controls the gating behavior, and is set to true when a feature is "retrying" and set to false when a payload is successfully sent.  This creates a very easy to follow flow for controlling whether a feature can harvest early or not.

### Testing
New e2e test was added to ensure new early harvests are not made while retrying

### DIFF
--- a/tests/specs/harvesting/harvest-early.e2e.js
+++ b/tests/specs/harvesting/harvest-early.e2e.js
@@ -50,7 +50,9 @@ describe('should harvest early', () => {
     /** not harvesting early in retry mode */
     const [smHarvest] = await Promise.all([
       metricsCapture.waitForResult({ totalCount: 1 }),
-      $('button').click()
+      browser.execute(function () {
+        newrelic.addPageAction('test')
+      })
         .then(() => browser.pause(10000))
         .then(() => browser.refresh())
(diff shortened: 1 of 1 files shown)


cwli24 (dismissed):

cwli24 (approved):

tests/assets/harvest-early-block-internal.html
```diff
@@ -14,12 +14,12 @@
     {loader}
   </head>
   <body>
+    <button>Click me for ajax</button>
```
cwli24: This button doesn't do anything anymore though?

metal-messiah: true, i can remove

cwli24:
```suggestion
```

cwli24 (approved):

cwli24 (approved):

PR #1549, opened 2025-08-19
feat: Factor long tasks into XHR callback time if part of interaction

cwli24:
Soft navigation feature now back fills and reports `jsDuration`, `timeToLastCallbackEnd`, and `timeToSettle` attributes from former SPA feature for XHR `AjaxRequest` events. As before, these values only exist when the request is identified as part of an interaction, and not when it is standalone. Their calculations are newly determined by any long tasks resulting from their event listeners.
---

### Overview

(2 of 3 modifying soft navigation heuristic.) In addition to extending `BrowserInteraction` events from part 1, long tasks resulting from XHR event listeners will also be reflected in the `AjaxRequest` callback timings when reported under SPA. This means the `callbackEnd` and `callbackDuration` property reported under each ajax node as children of spa interaction events.
image

This PR is a checkpoint that substitutes in long task for the former Spa v1 spa-ajax callback timings. Note that this affects **only XHR** at this time, not fetch requests.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-438509
https://new-relic.atlassian.net/browse/NR-451766


tests/components/soft_navigations/aggregate.test.js
```diff
@@ -249,48 +249,59 @@ describe('getInteractionFor', () => {
...
     softNavAggregate.ee.emit('newUIEvent', [{ type: POPSTATE_TRIGGER, timeStamp: 105 }])
     softNavAggregate.ee.emit('newURL', [110, 'myurl.com'])
 
-    expect(softNavAggregate.interactionInProgress.trigger).toEqual(POPSTATE_TRIGGER)
+    expect(softNavAggregate.interactionInProgress.trigger).toEqual('click')
     expect(softNavAggregate.interactionInProgress.oldURL).toEqual(window.location.href)
     expect(softNavAggregate.interactionInProgress.newURL).toEqual('myurl.com')
   })
 
+
```
metal-messiah:
```suggestion
```

metal-messiah (approved):

PR #1550, opened 2025-08-22
feat: remove newrelic meta attribute

metal-messiah:
Remove the reporting of the __newrelic meta attribute from outgoing replay payloads.  This is being conducted as a measure to help save billable bytes for customers and reduce cost of use for Session Replay functionalities.
---

### Overview
This PR removes the __newrelic attribute from outgoing payloads, but as a byproduct of that, a number of internal mechanisms were able to be changed, simplified, or removed in a sort of "domino effect" to reduce build burden and performance impacts on enduser machines:
1. The agent now uses the contextual ee to buffer preloaded recording data.
2. Because the agent can buffer the data until the features are fully initialized (by way of the ee), it can now: 
  - assume that the timekeeper should always exist when processing rrweb data.
  - get rid of the need for separate buffers for preload vs standard datasets
  - remove the requirement to loop through preload harvests at startup
3. Because the timekeeper can be assumed to exist, we no longer need to "correct" timestamps at harvest time
4. Because we don't need to correct timestamps at harvest time, we no longer need to pass the `__newrelic` object around internally, so it was just completely removed.

See the following visuals around the changes to the buffer mechanics here that kick off the "domino" effect.
### before the change
before

### after the change
after

### Related Issue(s)
NR-446285

### Testing
many unit and e2e tests have been added/modified to adjust for the lack of the __newrelic node.  Existing tests should continue to pass.

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -18,7 +18,6 @@ import { MODE, SESSION_EVENTS, SESSION_EVENT_TYPES } from '../../../common/sessi
 import { stringify } from '../../../common/util/stringify'
 import { stylesheetEvaluator } from '../shared/stylesheet-evaluator'
 import { now } from '../../../common/timing/now'
-import { buildNRMetaNode } from '../shared/utils'
 import { MAX_PAYLOAD_SIZE } from '../../../common/constants/agent-constants'
 import { cleanURL } from '../../../common/url/clean-url'
 import { canEnableSessionTracking } from '../../utils/feature-gates'
@@ -94,8 +93,7 @@ export class Aggregate extends AggregateBase {
         }
         return
       }
-      this.drain()
-      this.initializeRecording(srMode)
+      this.initializeRecording(srMode).then(() => { this.drain() })
     }).then(() => {
       if (this.mode === MODE.OFF) {
         this.recorder?.stopRecording() // stop any conservative preload recording launched by instrument
@@ -121,7 +119,7 @@ export class Aggregate extends AggregateBase {
   }
 
   handleError (e) {
-    if (this.recorder) this.recorder.currentBufferTarget.hasError = true
+    if (this.recorder) this.recorder.events.hasError = true
     // run once
     if (this.mode === MODE.ERROR && globalScope?.document.visibilityState === 'visible') {
       this.switchToFull()
@@ -171,10 +169,10 @@ export class Aggregate extends AggregateBase {
 
     if (!this.recorder) {
       try {
-      // Do not change the webpackChunkName or it will break the webpack nrba-chunking plugin
+        // Do not change the webpackChunkName or it will break the webpack nrba-chunking plugin
         const { Recorder } = (await import(/* webpackChunkName: "recorder" */'../shared/recorder'))
         this.recorder = new Recorder(this)
-        this.recorder.currentBufferTarget.hasError = this.errorNoticed
+        this.recorder.events.hasError = this.errorNoticed
       } catch (err) {
         return this.abort(ABORT_REASONS.IMPORT)
       }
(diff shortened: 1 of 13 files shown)


src/features/session_replay/aggregate/index.js
```diff
@@ -231,24 +224,10 @@ export class Aggregate extends AggregateBase {
...
       len = payload.body.length
     } else {
-      payload.body = payload.body.map(({ __serialized, ...node }) => {
-        if (node.__newrelic) return node
-        const output = { ...node }
-        output.__newrelic = buildNRMetaNode(node.timestamp, this.timeKeeper)
-        output.timestamp = this.timeKeeper.correctAbsoluteTimestamp(node.timestamp)
-        return output
-      })
+      payload.body = payload.body.map(({ __serialized, ...node }) => ({ ...node }))
```
cwli24: bit of optimization here: looping through and `delete __serialized` would make this intention clearer and also more performant

src/features/session_replay/instrument/index.js
```diff
@@ -78,7 +78,7 @@ export class Instrument extends InstrumentBase {
 
       // If startReplay() has been used by this point, we must record in full mode regardless of session preload:
       // Note: recorder starts here with w/e the mode is at this time, but this may be changed later (see #apiStartOrRestartReplay else-case)
-      this.recorder ??= new Recorder({ mode: this.#mode, agentIdentifier: this.agentIdentifier, trigger, ee: this.ee, agentRef: this.#agentRef })
+      this.recorder ??= new Recorder({ ...this, mode: this.#mode, trigger, agentRef: this.#agentRef, timeKeeper: this.#agentRef.runtime.timeKeeper }) // if TK exists due to deferred state, pass it
```
cwli24: It doesn't look like `trigger` is used

src/features/session_replay/shared/recorder.js
```diff
@@ -152,25 +139,18 @@ export class Recorder {
...
-    }
+    /** because we've waited until draining to process the buffered rrweb events, we can guarantee the timekeeper exists */
+    event.timestamp = this.parent.timeKeeper.correctAbsoluteTimestamp(event.timestamp)
     event.__serialized = stringify(event)
     const eventBytes = event.__serialized.length
     /** The estimated size of the payload after compression */
     const payloadSize = this.getPayloadSize(eventBytes)
     // Checkout events are flags by the recording lib that indicate a fullsnapshot was taken every n ms. These are important
-    // to help reconstruct the replay later and must be included.  While waiting and buffering for errors to come through,
+    // to help reconstruct the100 replay later and must be included.  While waiting and buffering for errors to come through,
```
cwli24: introduced typo

src/features/session_replay/shared/recorder.js
```diff
@@ -34,47 +30,38 @@ export class Recorder {
...
-    this.#backloggedEvents = new RecorderEvents(this.shouldFix)
-    this.#preloaded = [new RecorderEvents(this.shouldFix)]
-    /** The pointer to the current bucket holding rrweb events */
-    this.currentBufferTarget = this.#events
-    /** Only set to true once a snapshot node has been processed.  Used to block preload harvests from sending before we know we have a snapshot */
+
+    /** Each page mutation or event will be stored (raw) in this array. This array will be cleared on each harvest */
+    this.events = new RecorderEvents(this.shouldFix)
+    /** Backlog used for a 2-part sliding window to guarantee a 15-30s buffer window */
+    this.backloggedEvents = new RecorderEvents(this.shouldFix)
```
cwli24, line 37: I'm not really certain why this exists

metal-messiah: every 15 seconds backlogged events just gets assigned to the pointer of the current this.events, and then this.events resets to a new empty event store.  This ensures that we can store 15-30s of nodes during error mode without needing to reindex things constantly

src/features/session_replay/aggregate/index.js
metal-messiah: also noticed we double serialize when storing in the event buffer so that has also been adjusted in d311e4c9cf472c3ae9fba678cde80faf73c61775

src/features/session_replay/instrument/index.js
metal-messiah: 4681f33943ec646644b68706382630efc2282bb5

src/features/session_replay/shared/recorder.js
metal-messiah: 4e74b026d46501897b898e9ac025b5afaabc8754

src/features/session_replay/shared/recorder.js
cwli24: then this doesn't need to be initialize to a new recorder events, no?

metal-messiah: I think this was originally added just to ensure that even if empty, the processing logic can always apply and it doesnt need to do any truthy checks

cwli24 (approved):

PR #1551, opened 2025-08-25
chore: fix ios test inconsistency

metal-messiah:
This PR fixes an issue that causes iOS to fail sporadically due to races between the mobile container and the window being ready to receive commands compared to other browsers.
---


metal-messiah:
https://github.com/newrelic/newrelic-browser-agent/actions/runs/17216405559

PR #1556, opened 2025-08-27
feat: Use xhr/fetch to gauge click liveness

ptang-nr:
Assess click "liveness" for buttons and links. If an XHR or fetch is detected within 2 seconds of a user click on a button or link, the click will not be marked as a dead click.
---
### Overview

In this PR, we implement dead click detection for buttons and links via network requests. If an XHR or fetch is not detected within 2 seconds of a user click, it will be marked as a dead click.  Calls sent by agent will also be excluded from consideration.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-443293

### Testing

Added unit/component/e2e tests


src/features/ajax/instrument/index.js
```diff
@@ -403,6 +405,12 @@ function subscribeToEvents (agentRef, ee, handler, dt) {
 
     ctx.loadCaptureCalled = true
   }
+
+  function evalIfNonAgentTraffic (host) {
```
metal-messiah: For the same reasons as error clicks relying on the jserror feature, this introduces a dependency on the AJAX feature for user frustrations to work accurately.

We should consider moving to the following:
1. Add a `.on` call in the generic_events/instrument file, that is listening for `send-xhr-start`
2. Add a wrapXhr() call in the generic_events/instrument file to ensure xhr is wrapped even without AJAX
3. add the existing `netReq` handle call inside the `.on` handler for `send-xhr-start` in the generic_events instrument file
4. register handler as you have already done in the generic events aggregate class.

src/features/ajax/instrument/index.js
```diff
@@ -403,6 +405,12 @@ function subscribeToEvents (agentRef, ee, handler, dt) {
...
+    try {
+      if (host && !agentRef.beacons.includes(host)) handle('netReq', [], undefined, FEATURE_NAMES.genericEvents, ee)
```
metal-messiah: IDK what `netReq` means here, maybe this event name should be more descriptive

src/loaders/agent.js
```diff
@@ -87,6 +88,15 @@ export class Agent extends AgentBase {
...
   }
 
+  get beacons () {
+    if (!this.#beacons) {
+      const beacons = new Set([this.info.beacon, this.info.errorBeacon])
+      if (this.init.proxy?.beacon) beacons.add(this.init.proxy.beacon)
+      this.#beacons = [...beacons]
+    }
+    return this.#beacons
+  }
```
metal-messiah: Instead of duplicating this, can you share/expose the `internalTrafficList` that is built in `configure.js` and use that?

src/loaders/agent.js
```diff
@@ -40,6 +40,7 @@ import { setupStartAPI } from './api/start'
  * sensitive to network load, this may result in smaller builds with slightly lower performance impact.
  */
 export class Agent extends AgentBase {
+  #beacons
```
ptang-nr:
```suggestion
```

src/features/ajax/instrument/index.js
ptang-nr: Just a note -  I was unable to grab the url in `send-xhr-start`, so ended up using `open-xhr-start` to grab the `host` value and then setting up the handler with `send-xhr-start`.  Lmk if I should do something different there!

(PS. Keeping the features as independent as possible made a lot of sense! We even did something similar for the nav changes.  Don't know why that didn't cross my mind 😑)

src/features/generic_events/instrument/index.js
```diff
@@ -71,6 +75,28 @@ export class Instrument extends InstrumentBase {
...
     }, eventListenerOpts(false, this.removeOnAbort?.signal))
 
+    wrapFetch(this.ee)
+    wrapXhr(this.ee)
+    this.ee.on('open-xhr-start', (args) => {
+      this.parsedUrl = parseUrl(args[1])
+    })
+    this.ee.on('send-xhr-start', () => {
+      emitIfNonAgentTraffic.call(this, this.parsedUrl)
+    })
```
metal-messiah: I think this method leaves you vulnerable to misinterpretation if a 2nd xhr is opened before the first one sends

metal-messiah: send-xhr-start is supplied with the `xhr` context, which has the responseURL property.  Although that includes redirects, i wonder if that would be accurate enough to warrant skipping the two-step method here

ptang-nr: Yes, I was able to switch over.  (I remember seeing once before when I first looked at how Ajax worked, but subsequently forgot about it when I found the URL in the args 😞)

ptang-nr:
Consolidating to https://github.com/newrelic/newrelic-browser-agent/pull/1534

PR #1557, opened 2025-08-27
chore: Fix hidden deps between component tests

ptang-nr:
---

### Overview

Discovered while investigating a Jest test [failure](https://github.com/newrelic/newrelic-browser-agent/actions/runs/17276249751) in another PR.

- Switched to spies to allow `triggerHarvestFor` to run for other tests
- Restored PO to not trigger extra BrowserPerformance entries in later tests

**Issue 1**
`should record marks when enabled` + `should record measures when enabled` tests would fail when run as standalone tests but pass if run as part of the component test

The mock restore wasn't taking effect in the `'should harvest early if will exceed 1mb'` and `'should not harvest if single event will exceed 1mb'` tests.

(in this screenshot, the `'should harvest early if will exceed 1mb'` and `'should not harvest if single event will exceed 1mb'` tests have been commented out, showing the hidden dependency)
Screenshot 2025-08-27 at 3 44 30 PM

**Issue 2**
Initially I ran the added `user_frustrations` tests as a standalone set and they passed.  When the CI pipeline ran the component test suite, they failed.  There were extra BrowserPerformance entries being created.

### Testing

- Tested the failing tests as standalone: `should record marks when enabled`, `should record measures when enabled`
- Tested the overall generic_events agg component test suite
- Tested the component test passes in the `NR-443293-network-requests` branch after applying the same fix.
Screenshot 2025-08-27 at 4 10 20 PM

### DIFF
--- a/tests/components/generic_events/aggregate/index.test.js
+++ b/tests/components/generic_events/aggregate/index.test.js
@@ -56,32 +56,29 @@ test('should warn if invalid event is provide', async () => {
 })
 
 test('should harvest early if will exceed 1mb', async () => {
-  mainAgent.runtime.harvester.triggerHarvestFor = jest.fn()
-  expect(mainAgent.runtime.harvester.triggerHarvestFor).toHaveBeenCalledTimes(0)
+  const triggerHarvestSpy = jest.spyOn(mainAgent.runtime.harvester, 'triggerHarvestFor').mockImplementation(() => {})
+  expect(triggerHarvestSpy).toHaveBeenCalledTimes(0)
   genericEventsAggregate.ee.emit('rumresp', [{ ins: 1 }])
   await new Promise(process.nextTick)
-  expect(mainAgent.runtime.harvester.triggerHarvestFor).toHaveBeenCalledTimes(1)
+  expect(triggerHarvestSpy).toHaveBeenCalledTimes(1)
 
   genericEventsAggregate.addEvent({ name: 'test', eventType: 'x'.repeat(15000) })
 
-  expect(mainAgent.runtime.harvester.triggerHarvestFor).toHaveBeenCalledTimes(1)
+  expect(triggerHarvestSpy).toHaveBeenCalledTimes(1)
   genericEventsAggregate.addEvent({ name: 1000, eventType: 'x'.repeat(100000) })
-  expect(mainAgent.runtime.harvester.triggerHarvestFor).toHaveBeenCalledTimes(2)
-
-  mainAgent.runtime.harvester.triggerHarvestFor.mockRestore()
+  expect(triggerHarvestSpy).toHaveBeenCalledTimes(2)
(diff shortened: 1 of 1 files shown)


tests/components/generic_events/aggregate/index.test.js
```diff
@@ -251,9 +244,11 @@ describe('sub-features', () => {
       disconnect: jest.fn()
     }))
 
-    global.PerformanceObserver = mockPerformanceObserver
     global.PerformanceObserver.supportedEntryTypes = ['mark']
 
+    // temporarily swap out triggerHarvestFor so the buffer doesn't get emptied before we can check
+    jest.spyOn(mainAgent.runtime.harvester, 'triggerHarvestFor').mockImplementation(() => {})
```
ptang-nr: Open to suggestions.  Couldn't figure out a better way to preserve/peek at the buffer before it gets emptied out while still in the generic event agg constructor.

cwli24 (commented):
I don't understand this change. Is issue that the 30s interval harvest somehow triggering during the test?

tests/components/generic_events/aggregate/index.test.js
metal-messiah: isnt `mainAgent.runtime.harvester.triggerHarvestFor` already mocked by the agentMock?  Could you just implement the mockImplementation on that or does that muddy everything up too much

cwli24:
> @cwli24 - no... the problem was M&M tests were failing when run as standalone tests but passing when run as part of the entire test suite. That points to some sort of hidden dependency between tests.
> 
> Longer version: when stepping through the calls, I can see the data in the event buffer but the data never makes it back to the test. Turns out, M&M tests would fail when run by themselves because `triggerHarvestFor()` would empty out the buffer before the GE agg constructor finished (GE agg constructor --> calls PO.observe and generate the event data --> triggerHarvestFor --> makeHarvestPayload empties the buffer --> test finishes constructing GE agg and tries to check the contents of the buffer and fail because it received an undefined value).

So the trigger harvest from the waitForFlags? In that case, you can simply replace the `mainAgent.runtime.harvester.triggerHarvestFor` with a jest fn after `beforeAll` hook in the original code.

tests/components/generic_events/aggregate/index.test.js
ptang-nr: I'm not following, did you mean I can use the spy object and check on the call param?  I tried the following but I'm not sure how to get at the `event` buffer content cleanly.  I think I'd have to peel back `appStorageMap` next and that's starting to get into the guts of the storage component.

Screenshot 2025-08-28 at 3 55 50 PM

(Or did you mean to change from `.mockImpl(() => {})` to `.mockImpl(something else here)`?  I couldn't really think of how to change it to help fish out event data 😖)

metal-messiah (approved):

PR #1558, opened 2025-09-03
chore: fix broken SR test

metal-messiah:
---

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -28,6 +28,7 @@ export class Aggregate extends AggregateBase {
 
   // pass the recorder into the aggregator
   constructor (agentRef, args) {
+    console.log('SR AGG constructor!')
     super(agentRef, FEATURE_NAME)
     /** Set once the recorder has fully initialized after flag checks and sampling */
     this.initialized = false
@@ -83,7 +84,9 @@ export class Aggregate extends AggregateBase {
 
     const { error_sampling_rate, sampling_rate, autoStart, block_selector, mask_text_selector, mask_all_inputs, inline_images, collect_fonts } = agentRef.init.session_replay
 
+    console.log('wait for flags...')
     this.waitForFlags(['srs', 'sr']).then(([srMode, entitled]) => {
+      console.log('got flags!', srMode, entitled)
       this.entitled = !!entitled
       if (!this.entitled) {
         this.deregisterDrain()
@@ -93,7 +96,11 @@ export class Aggregate extends AggregateBase {
         }
         return
       }
-      this.initializeRecording(srMode).then(() => { this.drain() })
+      this.ee.on('drain-' + this.featureName, () => {
+        console.log('GOT DRAIN EVENT')
+      })
+      console.log('CALL initializeRecording... has drained?', this.drained)
+      this.initializeRecording(srMode).then(() => { console.log('start draining'); this.drain(); console.log('drained') })
     }).then(() => {
       if (this.mode === MODE.OFF) {
         this.recorder?.stopRecording() // stop any conservative preload recording launched by instrument
@@ -134,6 +141,7 @@ export class Aggregate extends AggregateBase {
       if (!this.agentRef.runtime.isRecording) this.recorder.startRecording()
       this.syncWithSessionManager({ sessionReplayMode: this.mode })
     } else {
+      console.log('switching to full by calling initializeRecording')
       this.initializeRecording(MODE.FULL, true)
     }
   }
@@ -145,6 +153,7 @@ export class Aggregate extends AggregateBase {
    * @returns {void}
    */
   async initializeRecording (srMode, ignoreSession) {
+    console.log('initialize recording!', this.initialized, this.entitled)
     this.initialized = true
     if (!this.entitled) return
 
@@ -176,9 +185,8 @@ export class Aggregate extends AggregateBase {
       } catch (err) {
         return this.abort(ABORT_REASONS.IMPORT)
       }
-    } else {
-      this.recorder.parent = this
     }
+    this.recorder.parent = this
 
     // If an error was noticed before the mode could be set (like in the early lifecycle of the page), immediately set to FULL mode
     if (this.mode === MODE.ERROR && this.errorNoticed) this.mode = MODE.FULL
--- a/src/features/session_replay/instrument/index.js
+++ b/src/features/session_replay/instrument/index.js
@@ -19,6 +19,8 @@ export class Instrument extends InstrumentBase {
 
   #mode
   #agentRef
+  #importAbortController
+
   constructor (agentRef) {
     super(agentRef, FEATURE_NAME)
 
@@ -40,7 +42,7 @@ export class Instrument extends InstrumentBase {
       this.#mode = session?.sessionReplayMode
       this.#preloadStartRecording()
     } else {
-      this.importAggregator(this.#agentRef, () => import(/* webpackChunkName: "session_replay-aggregate" */ '../aggregate'))
+      this.#importAbortController = this.importAggregator(this.#agentRef, () => import(/* webpackChunkName: "session_replay-aggregate" */ '../aggregate'))
     }
 
     /** If the recorder is running, we can pass error events on to the agg to help it switch to full mode later */
@@ -84,6 +86,8 @@ export class Instrument extends InstrumentBase {
     } catch (err) {
       this.parent.ee.emit('internal-error', [err])
     }
+    /** cancel the original import before preloading the agg if it's already been staged */
+    this.#importAbortController?.abort()
     this.importAggregator(this.#agentRef, () => import(/* webpackChunkName: "session_replay-aggregate" */ '../aggregate'), { recorder: this.recorder, errorNoticed: this.errorNoticed })
   }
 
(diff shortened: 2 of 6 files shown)


metal-messiah:
https://github.com/newrelic/newrelic-browser-agent/actions/runs/17445179259/job/49537744181

cwli24 (commented):
Btw, if you're changing src code, this needs to be a full feat/fix PR. Usually would want a ticket too but if you make the description detail, can forego creating one.

src/features/session_replay/aggregate/index.js
```diff
@@ -176,9 +176,8 @@ export class Aggregate extends AggregateBase {
       } catch (err) {
         return this.abort(ABORT_REASONS.IMPORT)
       }
-    } else {
-      this.recorder.parent = this
     }
+    this.recorder.parent = this
```
cwli24: Did doing this even if recorder was newly initialized solve an edge case? Which?

metal-messiah: When first troubleshooting, I was seeing that it had a race happening when the trigger attr was removed, where the parent was still the instrument and it could throw errors, so i just did it for ensuring that it is always the agg class when making it this far

metal-messiah: long story short on this PR -- i was troubleshooting the wrong problem (the double import), fixed it, and then realized the actual problem was the `trigger` and fixed that and figured i should just keep the fixes for the double import as well instead of scrapping them

src/features/session_replay/instrument/index.js
```diff
@@ -78,12 +80,14 @@ export class Instrument extends InstrumentBase {
...
       // Note: recorder starts here with w/e the mode is at this time, but this may be changed later (see #apiStartOrRestartReplay else-case)
-      this.recorder ??= new Recorder({ ...this, mode: this.#mode, agentRef: this.#agentRef, timeKeeper: this.#agentRef.runtime.timeKeeper }) // if TK exists due to deferred state, pass it
+      this.recorder ??= new Recorder({ ...this, mode: this.#mode, agentRef: this.#agentRef, trigger, timeKeeper: this.#agentRef.runtime.timeKeeper }) // if TK exists due to deferred state, pass it
       this.recorder.startRecording()
       this.abortHandler = this.recorder.stopRecording
     } catch (err) {
       this.parent.ee.emit('internal-error', [err])
     }
+    /** cancel the original import before preloading the agg if it's already been staged */
+    this.#importAbortController?.abort()
```
cwli24: I think this is all wrong. If the import has already happen, i.e. replay was already allowed to run or running, then it should just keep running. Rather, the API call should not reset and start over or do something. startReplay should only restart replays that are paused or stopped.

src/features/session_replay/aggregate/index.js
metal-messiah: now that I am re-looking at this code, it actually would be better served by moving up into the first check for the recorder, so as to also ensure it is set for cases where initializeRecording can return early

src/features/session_replay/instrument/index.js
metal-messiah: i'm not following your comment here.  This is the staging of the aggregate module import, not the recorder. As such it doesn't affect what is recorded, it affects whether the SR agg file is imported and executed twice vs once.

src/features/session_replay/instrument/index.js
```diff
@@ -78,12 +80,14 @@ export class Instrument extends InstrumentBase {
 
       // If startReplay() has been used by this point, we must record in full mode regardless of session preload:
       // Note: recorder starts here with w/e the mode is at this time, but this may be changed later (see #apiStartOrRestartReplay else-case)
-      this.recorder ??= new Recorder({ ...this, mode: this.#mode, agentRef: this.#agentRef, timeKeeper: this.#agentRef.runtime.timeKeeper }) // if TK exists due to deferred state, pass it
+      this.recorder ??= new Recorder({ ...this, mode: this.#mode, agentRef: this.#agentRef, trigger, timeKeeper: this.#agentRef.runtime.timeKeeper }) // if TK exists due to deferred state, pass it
```
cwli24: I'm a little confused how this is causing test issues... because I don't see `trigger` being used at all in `Recorder`. If this is like a test spy, can it be removed or altered, so the src code isn't conforming to the test (seems backwards).

src/features/session_replay/instrument/index.js
metal-messiah: though i guess thinking about this module, we don't really need to care `when` the aggregate is imported, so we should import it like the rest of the features do, instead of in multiple places. I can make that change

cwli24: Ah so then in that case, the API can start the recorder, but still not reset the aggregator import. The wiring for Aggregate args should be modified instead if need be.

src/features/session_replay/instrument/index.js
metal-messiah: its used in the agg class. https://github.com/newrelic/newrelic-browser-agent/pull/1558/files#diff-71dce754271298de345c546458d504ef1905e201e05215a8a0e06ea91065be25R163

src/features/session_replay/instrument/index.js
metal-messiah: yep exactly.  just stage the agg import once like the other features do, but handle "early" importing the recorder for preload/api

metal-messiah: this will also make it so we dont need to pass around/return an abort controller for the inst. file and that can be removed as well

src/features/session_replay/instrument/index.js
cwli24: DOH, missed that

src/features/session_replay/aggregate/index.js
```diff
@@ -156,6 +156,10 @@ export class Aggregate extends AggregateBase {
     // session replays can continue if already in progress
     const { session, timeKeeper } = this.agentRef.runtime
     this.timeKeeper = timeKeeper
+
+    /** if recorder already exists, it could have been initialized by the inst. class.  Ensure its parent is always the agg class at this stage */
+    if (this.recorder) this.recorder.parent = this
```
cwli24: Are you sure this isn't going to cause problems? Like if the recorder is on? The next line checks for trigger, and I suspect you'd lose API trigger this way.

metal-messiah: oh i missed line 163.  I actually moved trigger into the recorder itself, so i need to re-route that

src/features/utils/instrument-base.js
```diff
@@ -99,7 +101,7 @@ export class InstrumentBase extends FeatureBase {
       } catch (e) {
         warn(20, e)
         this.ee.emit('internal-error', [e])
-        if (this.featureName === FEATURE_NAMES.sessionReplay) this.abortHandler?.() // SR should stop recording if session DNE
+        this.ee.emit('session-error')
```
cwli24: This is actually quite awkward because the SR instrument emits an event that it catches itself then calls handle again which does another emit (+buffer).

Better to change L104 here to `handle(SR_EVENT_EMITTER_TYPES.SESSION_ERROR, [], undefined, this.featureName, this.ee)` targeted at the SR feature.

cwli24: There's another problem here in that instrument base is used by all the features, meaning this can happen multiple times with or without delay in between. This is why the old code targeted one feature specifically.

src/features/session_replay/instrument/index.js
```diff
@@ -57,54 +60,50 @@ export class Instrument extends InstrumentBase {
...
-  async #preloadStartRecording (trigger) {
-    if (this.#alreadyStarted) return
-    this.#alreadyStarted = true
+  async importRecorder () {
+    /** if the import has never been staged, it will just resolve immediately.
+     * Once it has, it will be assigned as the actual import promise.
+     * This is to prevent multiple imports from being triggered before the previous has had a chance to resolve */
+    await this.#stagedImport
+    /** if the recorder has already been imported, return it */
+    if (this.recorder) return this.recorder
```
cwli24: You can also just set #stagedImport to be undefined at first, then return here if it's defined (to become the import) so that only 1 importRecorder is ran. This is a simpler approach that avoids unnecessary promises.

src/features/utils/instrument-base.js
metal-messiah: i think to your point though, if you handle the event with the feature name bucket, the recorder would only see one event, and the others pointed at other feature buckets would be discarded at drain time. That means you don't need to be selective, just emit it to its own feature bucket.  So that's likely the best path forward here.  I'll just skip the instrument setup and handle straight from the base.

cwli24: How does that work? You can emit duplicated events to the same feature bucket. How does the recorder only see one event if multiple instrument class unconditionally handles a session error?

src/features/session_replay/instrument/index.js
cwli24: nvm

src/features/utils/instrument-base.js
metal-messiah: each instrument base is extended by the feature instrument, meaning the handle call will have a distinct `featureName` value for each "importLater" that is invoked -- the __recorder__ is the one who is only looking for session_replay bucket events

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1558/files#diff-2570c3511d39007f17cfc32db2e68245b4b415a90668bf9246ef3e058d2cb5ecR106 <--- 'session-error' message with `this.featureName` bucket will be emitted once for every feature name (including `session_replay`)

https://github.com/newrelic/newrelic-browser-agent/pull/1558/files#diff-8daecbb02cdf0d67a386150b56086ed34331012e202bc0bb5d636e16dd79f37fR55-R58 <--- `recorder` is listening for `session-error` messages under the `session_replay` bucket only

PR #1562, opened 2025-09-10
feat: add interim session behavior

metal-messiah:
Introduces a change to session behavior wherein if a session expires **live** on the page, the new session is flagged as "interim". This flag controls behavior in the agent where the next **hard** page load will trigger a new fresh session to start, which will allow new sampling decisions for session replay, session trace and logging features.
---

### Overview
This is innovation work to stabilize session behavior to alleviate customer complaints around not being able to capture logs, traces or replays after a session has reset on a page.  Currently, these features are only re-initialized for a specific user after a dead expiration is observed (while the page is not being observed by the agent). A nice side-effect of this work is this provides us and customers a sure-fire (but internal) way to debug and reset their sessions in a way that the next page load will trigger sampling

```javascript
Object.values(newrelic.initializedAgents)[0].runtime.session.reset(true)
```

### Related Issue(s)
See [this thread](https://newrelic.slack.com/archives/G019T80B1EJ/p1757082558448299) for context around the innovation work

interim

### Testing
New tests have been added to validate that interim sessions can spawn new sessions on the next page load.

### DIFF
--- a/tests/components/session-entity.test.js
+++ b/tests/components/session-entity.test.js
@@ -190,7 +190,7 @@ describe('isNew', () => {
     expect(sessionInstance.value).not.toEqual(value)
   })
 
-  test('is true if existing session is not interim', () => {
+  test('is false if existing session is not interim', () => {
     storage.set(`${PREFIX}_${key}`, { ...model, interim: false, value, expiresAt: Date.now() + 100000, inactiveAt: Date.now() + 100000, updatedAt: Date.now() })
     const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
     expect(sessionInstance.isNew).toEqual(false)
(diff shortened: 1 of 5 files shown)


tests/components/session-entity.test.js
```diff
@@ -174,6 +182,19 @@ describe('isNew', () => {
...
   })
+
+  test('is true if existing session is interim', () => {
+    storage.set(`${PREFIX}_${key}`, { ...model, interim: true, value, expiresAt: Date.now() + 100000, inactiveAt: Date.now() + 100000, updatedAt: Date.now() })
+    const sessionInstance = new SessionEntity({ agentIdentifier, key, storage })
+    expect(sessionInstance.isNew).toEqual(true)
+    expect(sessionInstance.value).not.toEqual(value)
+  })
+
+  test('is true if existing session is not interim', () => {
```
metal-messiah:
```suggestion
  test('is false if existing session is not interim', () => {
```

metal-messiah:
closing until decision in DACI is reached

PR #1567, opened 2025-09-12
chore: Allow test server to report to BAM

metal-messiah:
Test server can report its data to BAM directly from pages.
---

### Overview
The testing server will now support reporting its activity to BAM instead of local network when supplied with the `-B` flag and a file path to a config file (.JSON format) or a JSON string as input.

An example config would look like:
```json
{
  "licenseKey": "...",
  "applicationID": "...",
  "accountID": "...",
  "agentID": "...",
  "trustKey": "...",
  "beacon": "...",
  "errorBeacon": "..."
}
```
These values are most easily obtained by inspecting an existing browser entity in NR1 or using NerdGraph to retrieve the configs for an entity.

This was added to support more easily smoke testing certain features with concurrent changes to a BAM consumer that is deployed to a staging cell.

### Related Issue(s)
[NR-482345](https://new-relic.atlassian.net/browse/NR-482345)

### Testing
try it: 
#### with file path
```bash
npm run test-server -- -B <path_to_your_config_json_file>
```
#### with json string argument
```bash
npm run test-server -- -B '{"licenseKey":"...","applicationID":"...","accountID":"...","agentID":"...","trustKey":"...","beacon":"...","errorBeacon":"..."}'
```

[NR-482345]: https://new-relic.atlassian.net/browse/NR-482345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/testing-server/constants.js
```diff
@@ -1,16 +1,39 @@
...
+      agentConfig = JSON.parse(fs.readFileSync(args.B, 'utf-8'))
+    } catch (err) {
+      console.error('Failed to parse BAM config:', err)
+    }
+  }
+}
+
+/** fallback if no override was provided or override outright failed to parse correctly */
+if (!agentConfig) {
+  agentConfig = {
```
cwli24, line 22: Really to send harvest, only the application ID & licenseKey are needed. To make this .B easier to use, you shouldn't need the rest of the properties but simply provide those 2.

metal-messiah: definitely, its up to you to provide what you want the `info` section to be here, its flexible (and the tests might require some of that extra info to "work" correctly).  Just know that it expects an info object string.

cwli24 (approved):

PR #1577, opened 2025-09-19
chore: SR SMs revision

metal-messiah:
Only capture `/jserror` payloads if 'err' payload is present 
---

### DIFF
--- a/src/common/harvest/harvester.js
+++ b/src/common/harvest/harvester.js
@@ -162,19 +162,7 @@ export function send (agentRef, { endpoint, targetApp, payload, localOpts = {}, 
         cbFinished(cbResult)
 
         /** temporary audit of consistency of page metadata flags */
-        if (!shouldRetry(this.status)) {
-          const hasReplay = baseParams.includes('hr=1')
-          const hasTrace = baseParams.includes('ht=1')
-          const hasError = qs?.attributes?.includes('hasError=true')
-
-          handle('page-metadata', [{
-            [featureName]: {
-              ...(hasReplay && { hasReplay }),
-              ...(hasTrace && { hasTrace }),
-              ...(hasError && { hasError })
-            }
-          }], undefined, FEATURE_NAMES.metrics, agentRef.ee)
-        }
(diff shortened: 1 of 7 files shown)


cwli24 (dismissed):

src/common/harvest/harvester.js
```diff
@@ -135,22 +135,24 @@ export function send (agentRef, { endpoint, targetApp, payload, localOpts = {},
...
-    // Warn--once per endpoint--if the agent tries to send large payloads
-    if (body.length > 750000 && (warnings[endpoint] = (warnings[endpoint] || 0) + 1) === 1) warn(28, endpoint)
-  }
+
+  // all gzipped data is already in the correct format and needs no transformation
+  // all features going to 'events' endpoint should already be serialized & stringified
+  let stringBody = gzip || endpoint === EVENTS ? body : stringify(body)
+
+  // Warn--once per endpoint--if the agent tries to send large payloads
+  if (endpoint !== BLOBS && stringBody.length > 750000 && (warnings[endpoint] = (warnings[endpoint] || 0) + 1) === 1) warn(28, endpoint)
```
cwli24: The next line ensures that stringBody is defined. Is it possible to be undefined? Because it'd throw on `stringBody.length`?

cwli24: Also to be cautious when changing conditions, is "not gzip" strictly equivalent to "not BLOBS endpoint"?

metal-messiah: for now, yes, but to be specific, we only want to warn for large harvests for non-blob requests since those are more or less expected to be large

metal-messiah: ```The next line ensures that stringBody is defined. Is it possible to be undefined? Because it'd throw on stringBody.length?```

Technically, i think PVE can send an "empty" body but it does so with an empty object like `{}` due to the `cleanPayload` call, which is the reason the test suite still passes, since length is a valid property of `"{}"`.  That said I'll just re-order these lines for safety.

metal-messiah:
Screenshot 2025-09-22 at 1 15 23 PM

Issue #1590, opened 2025-10-14
interaction waitForEnd


geminixandroid:
How to set waitForEnd for interaction() using typescript?
[interaction (SPA API)](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/interaction/)

PR #1591, opened 2025-10-14
fix: Update typings for interaction API to reflect waitForEnd

metal-messiah:
Fixes an issue where the typing file defining the `.interaction` API was not decorated correctly to account for the optional `waitForEnd` argument as [documented](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/interaction/#parameters).
---

### Related Issue(s)
#1590 

### Testing
The typings tests were adjusted to account for this, and a new change has been made to enforce that the typings tests are run at commit time to help prevent misses like this in the future.

### DIFF
--- a/src/loaders/api-base.js
+++ b/src/loaders/api-base.js
@@ -197,7 +197,7 @@ export class ApiBase {
   /**
    * Returns a new API object that is bound to the current SPA interaction.
    * {@link https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/interaction/}
-   * @param {Object} [opts] The options to apply when creating the interaction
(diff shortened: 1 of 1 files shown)


src/loaders/api-base.js
```diff
@@ -197,12 +197,14 @@ export class ApiBase {
   /**
    * Returns a new API object that is bound to the current SPA interaction.
    * {@link https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/interaction/}
+   * @param {Object} [opts] The options to apply when creating the interaction
```
cwli24:
```suggestion
   * @param {Object} [opts] Options to configure the new or existing interaction with
```

PR #1592, opened 2025-10-15
chore: Add rrweb node types event size SM

ptang-nr:
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-458681
Related Angler PR: https://source.datanerd.us/agents/angler/pull/754

### DIFF
--- a/src/features/session_replay/shared/recorder.js
+++ b/src/features/session_replay/shared/recorder.js
@@ -18,7 +18,6 @@ import { single } from '../../../common/util/invoke'
 import { registerHandler } from '../../../common/event-emitter/register-handler'
 
 const RRWEB_DATA_CHANNEL = 'rrweb-data'
-const payloadSizeSMTag = ['', 'rrweb/node/1/bytes', 'rrweb/node/2/bytes', 'rrweb/node/3/bytes', 'rrweb/node/4/bytes']
 
 export class Recorder {
   /** flag that if true, blocks events from being "stored".  Only set to true when a full snapshot has incomplete nodes (only stylesheets ATM) */


src/features/session_replay/shared/recorder.js
```diff
@@ -180,6 +181,9 @@ export class Recorder {
     const eventBytes = event.__serialized.length
     /** The estimated size of the payload after compression */
     const payloadSize = this.getPayloadSize(eventBytes)
+    if (event.type < 5 && event.type > 0) {
```
metal-messiah: Are there cases where rrweb emits these node types that we need to check this for?

ptang-nr: Initially I'd assumed based on these rrweb event [types](https://github.com/rrweb-io/rrweb/blob/76df9799ecc14930fa914e5623a73ea7726e3747/packages/types/src/index.ts#L1-L9) without fully knowing how/when they'd get emitted 🤔

I took a closer look at DOMContentLoaded, and looks like that may be related to the `recordAfter` [option](https://github.com/rrweb-io/rrweb/blob/76df9799ecc14930fa914e5623a73ea7726e3747/guide.md#options) so maybe that won't get emitted(?)  And in theory we probably also don't have any `Custom` or `Plugin`.

src/features/session_replay/shared/recorder.js
```diff
@@ -180,6 +181,7 @@ export class Recorder {
     const eventBytes = event.__serialized.length
     /** The estimated size of the payload after compression */
     const payloadSize = this.getPayloadSize(eventBytes)
+    handle(SUPPORTABILITY_METRIC_CHANNEL, [payloadSizeSMTag[event.type], eventBytes], undefined, FEATURE_NAMES.metrics, this.ee)
```
metal-messiah: Sorry this is a bit nitpicky, but it should be "smaller" in build burden


```suggestion
    handle(SUPPORTABILITY_METRIC_CHANNEL, ['rrweb/node/' + event.type + '/bytes', eventBytes], undefined, FEATURE_NAMES.metrics, this.ee)
```

src/features/session_replay/shared/recorder.js
```diff
@@ -18,6 +18,7 @@ import { single } from '../../../common/util/invoke'
 import { registerHandler } from '../../../common/event-emitter/register-handler'
 
 const RRWEB_DATA_CHANNEL = 'rrweb-data'
+const payloadSizeSMTag = ['', 'rrweb/node/1/bytes', 'rrweb/node/2/bytes', 'rrweb/node/3/bytes', 'rrweb/node/4/bytes']
```
metal-messiah:
```suggestion
```

PR #1593, opened 2025-10-16
feat: Switch to using NR fork of rrweb

cwli24:
Browser agent will now switch to using New Relic's forked version of rrweb for Session Replay instead of the original repository. This will allow browser team to custom tailor hotfixes for issues.
---

### Overview

Switch BA rrweb dependency over from original repository version 2.0.0-alpha.18 to forked version 1.0.1 (published on NPM). Note that the `rrweb.version` string in SR payload has also been updated from that alpha versioning to `@newrelic/rrweb@1.0.1` pattern, wherein the version will change as new versions on the fork are released and published.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-452639


ellisong (approved):
LGTM

src/common/constants/rrweb.js
```diff
@@ -0,0 +1,21 @@
...
+// (or if the env var was not injected) we fall back to '0.0.0'. We intentionally avoid
+// importing this repo's package.json or the rrweb package's package.json here to keep
+// the distributed ESM free of deep relative path resolution that might fail in consumer
+// bundlers. The actual version is sourced centrally in babel.config.js so both CDN and
+// NPM builds receive the installed package's concrete version string.
+// The Babel inline env plugin replaces process.env.RRWEB_VERSION with the concrete version string at build time.
+// NOTE: In distributed artifacts the literal version should appear; if not, check babel.config.js include list.
+// After Babel inline substitution this becomes: export const RRWEB_VERSION = "<resolved-version>" || '0.0.0'
+// which evaluates to the concrete version string.
+export const RRWEB_VERSION = process.env.RRWEB_VERSION || '0.0.0'
```
metal-messiah: Since this was removed from the env.js, what happens here in customer NPM builds that don't use our babel config? (this is why it read from package.json instead of the babel replacement in env.js)

PR #1597, opened 2025-10-21
fix: memoize promise context propagation to avoid safari hangs

shakespear-dev:
Caching the resolved SPA context for wrapped Promises prevents from re-traversing long promise chains on every propagation, eliminating the freeze on Safari and keeping third-party promise-heavy workflows responsive (e.g. with `html2pdf.js`).

### Overview

- Cache the resolved context once inside [wrap-promise](https://github.com/newrelic/newrelic-browser-agent/blob/796d970ce69e4130b6c71ef798c0a68820ffb033/src/common/wrap/wrap-promise.js) propagation handlers so downstream `.then` chains no longer walk the entire parent chain.
- Add a regression test that ensures the parent context is only resolved once per chain and that subsequent accesses reuse the cached value.

### Related Issue(s)

- Repro sandbox: [CodeSandbox](https://codesandbox.io/p/devbox/broken-river-46523k?workspaceId=ws_38qrSMMjqQpsmcCa3FfJkv)
- Preview: [https://46523k-5173.csb.app/](https://46523k-5173.csb.app/) (Press `Export as PDF` in Safari on iOS or macOS)

### Testing

Added a regression test in [wrap-promise.test.js](https://github.com/newrelic/newrelic-browser-agent/blob/796d970ce69e4130b6c71ef798c0a68820ffb033/tests/components/wrappings/wrap-promise.test.js) that validates context caching prevents repeated parent traversal in long promise chains.

### DIFF
--- a/src/common/wrap/wrap-promise.js
+++ b/src/common/wrap/wrap-promise.js
@@ -140,13 +140,23 @@ export function wrapPromise (sharedEE) {
 
     promiseEE.on('propagate', function (val, overwrite, trigger) {
       if (!this.getCtx || overwrite) {
-        this.getCtx = function () {
-          // eslint-disable-next-line
-          if (val instanceof Promise) {
-            var store = promiseEE.context(val)
-          }
+        const selfStore = this
+        const parentStore =
+          val instanceof Promise ? promiseEE.context(val) : null
+        let cachedCtx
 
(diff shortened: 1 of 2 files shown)


shakespear-dev:
@ptang-nr awesome, happy to help!

PR #1598, opened 2025-10-21
fix: Improve accuracy in standalone AjaxRequest event start times

metal-messiah:
Adjusts a problem in the agent where start times of certain standalone AjaxRequest event payloads were reported relative to the page origin instead of relative to the first event in the payload, causing discrepancies when visualizing in NR1.
---

### Overview
Standalone ajax nodes indexed 1+ in BEL7.2 event arrays must reference the first node in the series as the relative timestamp.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-479069?atlOrigin=eyJpIjoiMzc5NzA3YjE0ZGE0NGFhMjhiNjU3NThmN2E0ZTI4MDgiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

### Testing
New tests have been added to validate the timestamps are correct before sending.

### DIFF
--- a/src/features/ajax/aggregate/index.js
+++ b/src/features/ajax/aggregate/index.js
@@ -143,7 +143,7 @@ export class Aggregate extends AggregateBase {
       const event = eventBuffer[i]
       // ajax nodes are relative to the first node (or page origin if no previous node), so we need to calculate the relative start time
       const relativeStartTime = event.startTime - firstTimestamp
-      firstTimestamp ||= event.startTime
+      if (i === 0) firstTimestamp = event.startTime
 
       const fields = [
         numeric(relativeStartTime),
--- a/tests/assets/spa/dt/xhr-dt-crossorigin-load.html
+++ b/tests/assets/spa/dt/xhr-dt-crossorigin-load.html
@@ -22,13 +22,17 @@
       window.rawTimes.push(Math.floor(performance.now()))
       xhr.send()
 
-      for (var i = 0; i < 10; i++) {
-        var xhr = new XMLHttpRequest()
+      let offset = 100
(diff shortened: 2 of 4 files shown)


src/features/ajax/aggregate/index.js
```diff
@@ -137,10 +137,16 @@ export class Aggregate extends AggregateBase {
     const addString = getAddStringContext(this.agentRef.runtime.obfuscator)
     let payload = 'bel.7;'
 
+    let firstTimestamp = 0
+
     for (let i = 0; i < eventBuffer.length; i++) {
       const event = eventBuffer[i]
+      // ajax nodes are relative to the first node (or page origin if no previous node), so we need to calculate the relative start time
+      const relativeStartTime = event.startTime - firstTimestamp
+      firstTimestamp ||= event.startTime
```
cwli24: `eventBuffer[0].startTime` in theory could be 0. It's more secure to check "if i = 0, firstTimestamp = event.startTime" on this line.

tests/assets/spa/dt/xhr-dt-crossorigin-load.html
```diff
@@ -13,12 +13,23 @@
...
+      window.rawTimes = []
       var xhr = new XMLHttpRequest()
 
       var url = 'http://' + NREUM.info.beacon + '/dt/' + NREUM.info.licenseKey
 
       xhr.open('GET', url)
+      window.rawTimes.push(Math.floor(performance.now()))
       xhr.send()
+
+      for (var i = 0; i < 10; i++) {
```
cwli24: 10 might be excessive. It's better to test an ajax that isn't immediately fired after the last so that the floored start diff isn't 0.

PR #1600, opened 2025-10-22
fix: User frustrations logic should be gated

ptang-nr:
Move user frustrations logic to browser scope only with FF gate in Generic Events instrumentation.  This should avoid running extraneous wrapping logic and events related to user frustrations.
---

### Overview

Avoid running extraneous logic if `user_frustrations` FF is not enabled.

### Related Issue(s)

NR-443293

### Testing

Adjusted the component test to include the FF (it caught the logic stopped running w/o the FF).

### DIFF
--- a/src/features/ajax/aggregate/index.js
+++ b/src/features/ajax/aggregate/index.js
@@ -137,10 +137,16 @@ export class Aggregate extends AggregateBase {
     const addString = getAddStringContext(this.agentRef.runtime.obfuscator)
     let payload = 'bel.7;'
 
+    let firstTimestamp = 0
+
     for (let i = 0; i < eventBuffer.length; i++) {
       const event = eventBuffer[i]
+      // ajax nodes are relative to the first node (or page origin if no previous node), so we need to calculate the relative start time
+      const relativeStartTime = event.startTime - firstTimestamp
+      if (i === 0) firstTimestamp = event.startTime
+
       const fields = [
-        numeric(event.startTime),
+        numeric(relativeStartTime),
         numeric(event.endTime - event.startTime),
         numeric(0), // callbackEnd
         numeric(0), // no callbackDuration for non-SPA events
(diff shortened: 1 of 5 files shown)


src/features/generic_events/instrument/index.js
```diff
@@ -52,7 +52,46 @@ export class Instrument extends InstrumentBase {
...
         // Capture is not used here so that we don't get element focus/blur events, only the window's as they do not bubble. They are also not cancellable, so no worries about being front of line.
         )
+
+        if (agentRef.init.feature_flags.includes('user_frustrations')) {
+          globalScope.addEventListener('error', () => {
+            handle('uaErr', [], undefined, FEATURE_NAMES.genericEvents, this.ee)
+          }, eventListenerOpts(false, this.removeOnAbort?.signal))
+
+          wrapFetch(this.ee)
+          wrapXhr(this.ee)
```
cwli24: I have other wrappings that I'm going to introduce and it'll be cleaner to have them all in one location rather than split apart which makes it harder to find visually. Are you able to move this out into a separate section above this block (similar to the `setup____` section), even if it'll be some redundant checks? (My wrapping will be behind a check anyways)

cwli24: Same with wrapHistory which I didn't notice at first. Just make a `let historyEE`.

cwli24: so
```
const ufEnabled = agentRef.init.feature_flags.includes('user_frustrations')
let historyEE
if (isBrowserScope && ufEnabled) {
 wrapFetch
 wrapXhr
 historyEE = wrapHistory
}
// --> I add: if (blah_enabled) wrapBlah
...
if (ufEnabled)) {
          globalScope.addEventListener('error', () => {
...etc
```

metal-messiah (approved):

cwli24 (approved):

PR #1602, opened 2025-10-22
chore: fix flaky test

metal-messiah:
This change helps stabilize a flaky test.  The test would FAIL occasionally ONLY in github runners, which was confusing.  I realized it was due to the arbitrary 0, 300, 700 timestamps, which if the test would execute quicker than those arbitrary marks, would invoke newrelic.interaction() BEFORE the actual arbitrary timestamp, and would break things. See example screenshot

Screenshot 2025-10-22 at 4 22 55 PM

---

### DIFF
--- a/tests/components/soft_navigations/api.test.js
+++ b/tests/components/soft_navigations/api.test.js
@@ -7,7 +7,7 @@ import querypack from '@newrelic/nr-querypack'
  * Test `.interaction gets ixn retroactively too when processed late after ee buffer drain` is a bit
  * flaky so add a retry for this file.
  */
-jest.retryTimes(0)
+jest.retryTimes(5)
 
 const INTERACTION_API = 'api-ixn'
 let mainAgent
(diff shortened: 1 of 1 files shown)


tests/components/soft_navigations/api.test.js
```diff
@@ -1,11 +1,13 @@
...
 import { resetAgent, setupAgent } from '../setup-agent'
 
+import querypack from '@newrelic/nr-querypack'
+
 /**
  * Test `.interaction gets ixn retroactively too when processed late after ee buffer drain` is a bit
  * flaky so add a retry for this file.
  */
-jest.retryTimes(5)
+jest.retryTimes(0)
```
cwli24: You should remove this

tests/components/soft_navigations/api.test.js
```diff
@@ -281,33 +283,48 @@ test('.actionText and .setAttribute add attributes to ixn specifically', () => {
...
+  const deserializedPayload = querypack.decode(harvestPayloadBody)
+  const starts = [0, 300, 700]
+  deserializedPayload.forEach(payload => {
+    expect(payload.start).toEqual(starts.shift())
+  })
   // WARN: Double check decoded output & behavior or any introduced bugs before changing the follow line's static string.
-  expect(softNavAggregate.makeHarvestPayload().body).toEqual("bel.7;1,,,5k,,,'api,'http://localhost/,1,1,,2,!!!!'some_id,'1,!!;;1,,8c,5k,,,'api,'http://localhost/,1,1,,2,!!!!'some_other_id,'2,!!;;1,,jg,8c,,,'api,'http://localhost/,1,1,,2,!!!!'some_another_id,'3,!!;")
-})
+  expect(harvestPayloadBody).toEqual("bel.7;1,,,5k,,,'api,'http://localhost/,1,1,,2,!!!!'some_id,'1,!!;;1,,8c,5k,,,'api,'http://localhost/,1,1,,2,!!!!'some_other_id,'2,!!;;1,,jg,8c,,,'api,'http://localhost/,1,1,,2,!!!!'some_another_id,'3,!!;")
+}, 30000)
```
cwli24: what is this 30000 for btw?

tests/components/soft_navigations/api.test.js
```diff
@@ -281,33 +283,48 @@ test('.actionText and .setAttribute add attributes to ixn specifically', () => {
...
   await new Promise(process.nextTick)
 
   expect(softNavAggregate.interactionsToHarvest.get().length).toEqual(3)
+
+  const harvestPayloadBody = softNavAggregate.makeHarvestPayload().body
+  const deserializedPayload = querypack.decode(harvestPayloadBody)
+  const starts = [0, 300, 700]
+  deserializedPayload.forEach(payload => {
+    expect(payload.start).toEqual(starts.shift())
+  })
```
cwli24, line 321: This part isn't really necessary because the strict payload check already incorporate these expects by effect. If you change anything of the start or end time, it changes the querypack encode offsets which fails to match L326. Like I changed the first start time from `0` to `100` and everything is off:
image

tests/components/soft_navigations/api.test.js
```diff
@@ -281,33 +283,48 @@ test('.actionText and .setAttribute add attributes to ixn specifically', () => {
...
+  softNavAggregate.ee.emit(`${INTERACTION_API}-get`, [jest.now()])
   ixnContext = getIxnContext(mainAgent.interaction())
   ixnContext.associatedInteraction.nodeId = 3
   ixnContext.associatedInteraction.id = 'some_another_id'
   ixnContext.associatedInteraction.forceSave = true
-  softNavAggregate.ee.emit(`${INTERACTION_API}-end`, [1000], ixnContext)
+  jest.advanceTimersByTime(300)
+  softNavAggregate.ee.emit(`${INTERACTION_API}-end`, [jest.now()], ixnContext)
+  jest.useRealTimers()
+  jest.clearAllTimers()
```
cwli24: So I don't believe this actually fixes the underlying problem with real time being used to seek the fake-time ixn.
I also wouldn't mess with the faketimer set up for the rest of the file by switching back to real timer after this test.
--let's talk about this when you have a moment.

cwli24: reminder for shifting to this style: image

tests/components/soft_navigations/api.test.js
metal-messiah: removed

PR #1605, opened 2025-10-27
feat: Retry initial connect call

metal-messiah:
To help prevent data loss, the agent will now retry the original "RUM" call an extra time for retryable status codes.
---

#### Angler PR
https://source.datanerd.us/agents/angler/pull/757
#### Issue
https://new-relic.atlassian.net/browse/NR-137656?atlOrigin=eyJpIjoiNjI5ODU3OTNkMDA3NDk3ZTg5ODk2YzEwYjgzMjQxNjciLCJwIjoiaiJ9

### Testing
New tests have been added to ensure the RUM request is retried and is identical, and subsequent feature will harvest.

### DIFF
--- a/src/common/harvest/types.js
+++ b/src/common/harvest/types.js
@@ -24,7 +24,6 @@
  * @property {HarvestEndpointIdentifier} endpoint The endpoint to use (jserrors, events, resources etc.)
  * @property {HarvestPayload} payload Object representing payload.
  * @property {object} localOpts Additional options for sending data
- * @property {boolean} localOpts.needResponse Specify whether the caller expects a response data.
  * @property {boolean} localOpts.isFinalHarvest Specify whether the call is a final harvest during page unload.
  * @property {boolean} localOpts.sendEmptyBody Specify whether the call should be made even if the body is empty. Useful for rum calls.
  * @property {boolean} localOpts.forceNoRetry Don't save the buffered data in the case of a need to retry the transmission.
--- a/tests/components/page_view_event/aggregate.test.js
+++ b/tests/components/page_view_event/aggregate.test.js
@@ -23,7 +23,6 @@ test('PageViewEvent does not throw on Harvester driven processes', () => {
 
   pveAggregate.events.add(undefined) // ensure request even if sendRum() puts empty body into buffer
   expect(mainAgent.runtime.harvester.triggerHarvestFor(pveAggregate, {
-    needResponse: true,
     sendEmptyBody: true
   }).ranSend).toEqual(true) // mimics the manual trigger in PVE `sendRum`; this should return true as it actually tries to "send"
 })


src/common/harvest/harvester.js
```diff
@@ -271,7 +275,7 @@ function baseQueryString (agentRef, qs, endpoint) {
     param('v', VERSION),
     transactionNameParam(),
     param('ct', agentRef.runtime.customTransaction),
-    '&rst=' + now(),
+    param('rst', now(), qs),
```
cwli24: what's with the `qs` here?

src/features/page_view_event/aggregate/index.js
```diff
@@ -107,24 +115,30 @@ export class Aggregate extends AggregateBase {
     queryParameters.fp = firstPaint.current.value
     queryParameters.fcp = firstContentfulPaint.current.value
 
+    this.rumStartTime = now()
     const timeKeeper = this.agentRef.runtime.timeKeeper
     if (timeKeeper?.ready) {
-      queryParameters.timestamp = Math.floor(timeKeeper.correctRelativeTimestamp(now()))
+      queryParameters.timestamp = Math.floor(timeKeeper.correctRelativeTimestamp(this.rumStartTime))
+    } else {
+      queryParameters.rst = this.rumStartTime // we set this here in the feature instead of in the harvester to allow for a future retry to use the original RST
```
cwli24: I think you should report the rst as is at the time of retry. It's an accurate portrayal of what happened for that client because all the other events will follow this. You don't want a big time gap between PageView & rest.

src/features/page_view_event/aggregate/index.js
```diff
@@ -107,24 +115,30 @@ export class Aggregate extends AggregateBase {
...
+
+    this.triggerHarvestFor(this.payload)
+  }
 
+  triggerHarvestFor (payload) {
     this.agentRef.runtime.harvester.triggerHarvestFor(this, {
       directSend: {
-        target,
-        payload: { qs: queryParameters, body }
+        payload
```
cwli24: What if... we got rid of `directSend` which is only used by this feature and just buffer the body instead? ... let me whip something up.

src/common/harvest/harvester.js
metal-messiah: it allows rst to be overridden by the feature's query string

cwli24 (dismissed):

src/features/page_view_event/aggregate/index.js
```diff
@@ -110,24 +110,26 @@ export class Aggregate extends AggregateBase {
...
-
-    this.rumStartTime = now()
+    this.events.add(body)
 
     this.agentRef.runtime.harvester.triggerHarvestFor(this, {
-      directSend: {
-        target,
-        payload: { qs: queryParameters, body }
-      },
-      needResponse: true,
```
cwli24, line 125: I think there's a couple more places referencing 'needResponse' that can be removed as well

metal-messiah: [68c3206](https://github.com/newrelic/newrelic-browser-agent/pull/1605/commits/68c3206f7ae980040e61622fde0fd7a6170b035f)

PR #1609, opened 2025-10-28
chore: Test SR processes large CSS in "reasonable" time

ptang-nr:
Re-enabling the added e2e test to help catch future regressions.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-403003


ptang-nr:
Yeah, I tested it locally (it fails on v 1.300.0 but passes now).  Also ran the PR check this morning and this test passes.  (side note: running into some instability w/ Android test suite, but that should be unrelated)

PR #1611, opened 2025-10-30
feat: SMs for browser connect response

ellisong:
Adds supportability metrics for failed responses on initializing the page_view_event aggregate.
---

### Overview

Adds supportability metrics for failed responses from the browser connect service. We use this response to initialize our page view event feature as well as to know what aggregate features to start up for our remaining features. Any response that isn't a successful status code (>= http 400) will emit the following support metrics:

- `Browser/Supportability/BCS/Error/StatusCode`
  - The status code of the failed browser connect service response. Will always be a non-successful status code integer.
- `Browser/Supportability/BCS/Error/Dropped/Bytes`
  - Estimated amount of bytes dropped from our event backlog.
- `Browser/Supportability/BCS/Error/Duration/Ms`
  - The response time from sending the request to the browser connect service to receiving a response.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-428192

### Testing

Make sure WDIO tests pass.

Locally testing this outside WDIO is difficult since it requires getting an external service to intentionally fail its requests.

### DIFF
--- a/src/features/session_replay/aggregate/index.js
+++ b/src/features/session_replay/aggregate/index.js
@@ -7,7 +7,7 @@
  */
 
 import { registerHandler } from '../../../common/event-emitter/register-handler'
-import { ABORT_REASONS, FEATURE_NAME, QUERY_PARAM_PADDING, RRWEB_EVENT_TYPES, SR_EVENT_EMITTER_TYPES, TRIGGERS } from '../constants'
+import { ABORT_REASONS, ERROR_DURING_REPLAY, FEATURE_NAME, QUERY_PARAM_PADDING, RRWEB_EVENT_TYPES, TRIGGERS } from '../constants'
 import { AggregateBase } from '../../utils/aggregate-base'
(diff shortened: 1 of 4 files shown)


src/features/page_view_event/aggregate/index.js
```diff
@@ -136,6 +139,54 @@ export class Aggregate extends AggregateBase {
...
+
+        const encoded = textEncoder.encode(value)
+        return acc + encoded.byteLength
+      }, 0)
+
+      // Send SMs about failed RUM request
+      const body = {
+        sm: [{
+          params: {
+            name: 'Browser/Supportability/BCS/Error/StatusCode'
```
metal-messiah: This will need to change to reflect the code in the SM tag (params.name) instead of the stats.  The stats are aggregated on the backend and the original value will be lost if reported this way. Something like
```js
Browser/Supportability/BCS/Error/<status code>
```

ellisong:
angler PR: https://source.datanerd.us/agents/angler/pull/760

PR #1612, opened 2025-10-30
feat: Set up framework for detecting AJAX calls from MFEs

metal-messiah:
Adds capability to detect a custom header for detecting MFE-linked AJAX traffic.  This is in support of a future initiative and the specifics may change over time as the feature approaches GA.
---


src/features/utils/aggregate-base.js
```diff
@@ -70,7 +70,8 @@ export class AggregateBase extends FeatureBase {
 
   /** @type {Boolean} indicates if the feature supports registered entities and the harvest requirements therein. Also read by getter "harvestEndpointVersion". Controlled by feature flag in pre-release phase. */
   get supportsRegisteredEntities () {
-    return this.featureName in SUPPORTS_REGISTERED_ENTITIES && (SUPPORTS_REGISTERED_ENTITIES[this.featureName] || this.agentRef.init.feature_flags.includes('register.' + this.featureName))
+    if (!this.agentRef.init.api.allow_registered_children || !(this.featureName in SUPPORTS_REGISTERED_ENTITIES)) return false
+    return SUPPORTS_REGISTERED_ENTITIES[this.featureName] || this.agentRef.init.feature_flags.includes('register.' + this.featureName)
```
cwli24, line 74: This can be combined with harvestEndpointVersion below as it's not used elsewhere. I'll also note that this system seems overly complicated/overkill in all the checks that need to be done that it isn't obvious what exactly is required, devolving into another SR.

src/features/ajax/aggregate/index.js
```diff
@@ -170,7 +186,11 @@ export class Aggregate extends AggregateBase {
 
       // add custom attributes
       // gql decorators are added as custom attributes to alleviate need for new BEL schema
-      const attrParts = addCustomAttributes({ ...(jsAttributes || {}), ...(event.gql || {}) }, addString)
+      const attrParts = addCustomAttributes({
+        ...(jsAttributes || {}),
+        ...(event.gql || {}),
+        ...(getVersion2Attributes(event.target, this)) // event.target only exists if we saw the right headers on the request.  The helper will know what to do with that
```
cwli24, line 173: Is the events consumer already setup to synthesized based on a set of children nodes of each ajax nodes on payloads?

src/features/ajax/aggregate/index.js
```diff
@@ -120,15 +122,29 @@ export class Aggregate extends AggregateBase {
...
-      this.events.add(event)
+      this.events.add(containerAgentEvent)
+    }
+
+    /** make a copy of the event for the MFE target if it exists */
+    if (mfeTarget) {
+      this.events.add({
+        ...event,
+        target: mfeTarget // custom attrs are calculated later at serialization time, so just tack the target metadata on the event as is and let the serializer handle the v2 attributes
+      })
```
cwli24: Styling change: I'd just add this as a property directly. Just a general note that the spread operator is a more expensive operation than a single property set (O(n) vs O(1)), particularly if large objects. While negligible in modern JS engine, I notice a pattern of it being always used, even in this scenario, which I wanted to call out.
Same with `containerAgentEvent` above.

src/features/ajax/aggregate/index.js
```diff
@@ -90,6 +91,7 @@ export class Aggregate extends AggregateBase {
       return // do not send this ajax as an event
     }
 
+    // TODO, does product want to note that an AJAX is tied to a MFE in the session trace waterfall?
```
cwli24: is this comment intended to stay?

src/features/utils/aggregate-base.js
metal-messiah: its not immediately obvious here, but [other PRs](https://github.com/newrelic/newrelic-browser-agent/pull/1632/files#diff-248b37ea9426b8cda9e8e21a95ab7bd4fb57b99a8fe694b5e5e5a07092f444aaR28) will end up using this attr

src/features/ajax/aggregate/index.js
metal-messiah: no, but it is planned.  This feature is hidden under the `register.ajax` feature flag until that is finished on the backend.

src/common/wrap/wrap-fetch.js
```diff
@@ -72,14 +74,35 @@ export function wrapFetch (sharedEE) {
...
+        let mfeId
+        try {
+          const headers = args?.[0]?.headers || args?.[1]?.headers
+          if (headers) {
+            const isHeaderInstance = headers instanceof Headers
+            const entries = isHeaderInstance ? Array.from(headers.entries()) : Object.entries(headers)
+            for (const [key, val] of entries) {
+              if (String(key).toLowerCase() === NEW_RELIC_MFE_ID_HEADER) {
+                const stringVal = stringify(val)
+                if (stringVal) mfeId ??= stringVal // only supports the first value found
```
cwli24: just note that this "set first time" is not necessary if you have L92 break check

src/common/wrap/wrap-fetch.js
```diff
@@ -72,14 +74,35 @@ export function wrapFetch (sharedEE) {
...
+
         // we are wrapping args in an array so we can preserve the reference
         ee.emit(prefix + 'before-start', [args], ctx)
         var dtPayload
         if (ctx[contextId] && ctx[contextId].dt) dtPayload = ctx[contextId].dt
 
         var origPromiseFromFetch = fn.apply(this, args)
 
-        ee.emit(prefix + 'start', [args, dtPayload], origPromiseFromFetch)
+        ee.emit(prefix + 'start', [args, dtPayload, mfeId], origPromiseFromFetch)
```
cwli24, line 105: why isn't the ID here just assigned to context like XHR?

tests/specs/api.e2e.js
cwli24, line 1: Can you please run all these e2e tests you're adding with a soft_nav flag in your local testing server as constant? I'd prefer any SN issues to be resolved here rather than go through testing hell again on my merge later

src/common/wrap/wrap-fetch.js
metal-messiah: this particular method is emitting the original context (origPromiseFromFetch), so we shouldnt mutate that, and theres already a pattern of passing other args in the emit of this lifecycle checkpoint (dtPayload) so I just went with that

tests/specs/api.e2e.js
metal-messiah: agent only runs soft nav now

PR #1614, opened 2025-11-03
feat: Instrument native WebSocket API

cwli24:
Re-introduces `WebSocket` instrumentation to capture metrics about its usage. When enabled by the 'websockets' feature flag, the agent will wrap the native API to record usage data around each instance and send it to New Relic as a "WebSocket" event on close.
---

### Overview

- Added socketId to error objects for better linkage with WebSocket events.
- Removed deprecated WebSocket event handling code and associated metrics.
- Added handling for page navigation away and adjusted tests.
- Introduced new HTML test files for WebSocket error handling and multi-type message sending.
- Updated existing tests to validate WebSocket metrics and error linkage.
- Implemented a WebSocket echo server for testing various data types.
- Edit third-party WebSocket wrappers to support new feature flags.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-455564

SM: https://source.datanerd.us/agents/angler/pull/771


src/common/wrap/wrap-websocket.js
```diff
@@ -3,59 +3,167 @@
...
+      return '[object WebSocket]'
+    }
+
+    get [Symbol.toStringTag] () { // fake [object WebSocket] when Object.prototype.toString.call is used on instance
+      return WrappedWebSocket.name
+    }
+
+    // Private method to tag send, close, and event listener errors with WebSocket ID for JSErrors feature
+    #tagError (error) {
+      ;(error.__newrelic ??= {}).socketId = this.nrData.socketId
```
metal-messiah, line 54: Other implementations of this __newrelic wrapper use the scoped [agentIdentifier] namespace to avoid conflicts with other agents.  That should be considered here if needed

src/common/wrap/wrap-websocket.js
```diff
@@ -3,59 +3,167 @@
...
-      const socketId = generateRandomHexString(6)
-      this.report = reporter(socketId)
-      this.report('new')
-
-      const events = ['message', 'error', 'open', 'close']
-      /** add event listeners */
-      events.forEach(evt => {
-        this.addEventListener(evt, function (e) {
-          this.report(ADD_EVENT_LISTENER_TAG, { eventType: evt, event: e })
+      this.nrData = {
```
metal-messiah: Bit subjective, but perhaps we should consider codifying nrData into its own model/class etc. so its easier to track what properties are expected to exist here, as opposed to setting new properties dynamically in each event listener. (it also plays nicer with the IDE tools that way)

src/common/wrap/wrap-websocket.js
```diff
@@ -3,59 +3,167 @@
...
-      this.report('send', ...args)
+    send (data) {
+      // Only track metrics if the connection is OPEN; data sent in CONNECTING state throws, and data sent in CLOSING/CLOSED states is silently discarded
+      if (this.readyState === WebSocket.OPEN) {
+        const { type, size } = getDataInfo(data)
+        this.nrData.sendCount = (this.nrData.sendCount ?? 0) + 1
+        this.nrData.sendBytes = (this.nrData.sendBytes ?? 0) + size
+        this.nrData.sendBytesMin = Math.min(this.nrData.sendBytesMin ?? Infinity, size)
+        this.nrData.sendBytesMax = Math.max(this.nrData.sendBytesMax ?? 0, size)
+        if (!(this.nrData.sendTypes ?? '').includes(type)) {
```
metal-messiah, line 129: can we send different types after the connection has been established?

src/common/wrap/wrap-websocket.js
```diff
@@ -3,59 +3,167 @@
...
+  if (data instanceof Blob) {
+    return { type: 'Blob', size: data.size }
+  }
+  if (data instanceof DataView) {
+    return { type: 'DataView', size: data.byteLength }
+  }
+  if (ArrayBuffer.isView(data)) {
+    return { type: 'TypedArray', size: data.byteLength }
+  }
+  return { type: 'unknown', size: 0 }
```
metal-messiah, line 179: We should discuss fail path here, ie what to do with the socket instance if we cant calculate the message size since that is a major utility driver on this feature.

src/features/jserrors/aggregate/index.js
```diff
@@ -192,10 +192,13 @@ export class Aggregate extends AggregateBase {
     // still send EE events for other features such as above, but stop this one from aggregating internal data
     if (this.blocked) return
 
-    if (err?.__newrelic?.[this.agentIdentifier]) {
+    if (err.__newrelic?.[this.agentIdentifier]) {
       params._interactionId = err.__newrelic[this.agentIdentifier].interactionId
       params._interactionNodeId = err.__newrelic[this.agentIdentifier].interactionNodeId
     }
+    if (err.__newrelic?.socketId) {
```
metal-messiah, line 199: same question as prior: should this be scoped to the agentIdentifier like above

src/features/metrics/aggregate/websocket-detection.js
metal-messiah, line 1: Can be in a new ticket, but if one is not already accounted for, please add a new story to track websocket adoption success through SMs

src/common/wrap/wrap-websocket.js
```diff
@@ -3,59 +3,167 @@
...
+      })
+
+      this.addEventListener('close', (event) => {
+        this.nrData.closedAt = now()
+        this.nrData.closeCode = event.code
+        this.nrData.closeReason = event.reason
+        this.nrData.closeWasClean = event.wasClean
+        this.nrData.connectedDuration = this.nrData.closedAt - this.nrData.openedAt
+
+        subscribedFeatures.forEach(featureName => handle('ws-complete', [this.nrData], this, featureName, wsEE))
```
metal-messiah: Is there a reason to break the emit pattern here that the other wrappers use in favor of directly using handle? (All the other wrappers "emit" an unscoped message and the instrument modules are responsible for the "handle" behavior to buffer for the agg modules).

ex.
other wrappers:
instrument --calls--> wrap* --emits--> instrument --handle--> aggregate

this change:
instrument --calls--> wrapWebsocket(needs feature name now) --handle--> aggregate

src/common/wrap/wrap-websocket.js
metal-messiah: nvm your test case seems to prove that it can be changed

tests/specs/ins/websockets.e2e.js
```diff
@@ -0,0 +1,208 @@
...
+    const messageError = errors.find(err =>
+      err.params?.message?.includes('message handler')
+    )
+    expect(messageError).toBeTruthy()
+    expect(messageError.custom?.socketId).toBe(wsEvent.socketId)
+  })
+
+  // Test third-party WebSocket wrappers
+  ;['robust-websocket', 'reconnecting-websocket'].forEach((thirdPartyWSWrapper) => {
+    it('should work with known third-party WS wrapper - ' + thirdPartyWSWrapper, async () => {
```
metal-messiah, line 190: Do these suffice as a "page unload" type test or do we need a separate case here? (to ensure that a page unloading reports the event)

src/common/wrap/wrap-websocket.js
cwli24: So I wanted to keep wrap sockets to single agent (which it does via wrap once), so I don't think this is necessary. I don't think even applies since these errors are coming from each _ws instance_, meaning there can't be multiple socketId's per error.

src/common/wrap/wrap-websocket.js
cwli24: I mean it's just there for show. It's impossible for the native `send` to be sending any data type that isn't defined that we already covered. The API itself won't allow it. And given the specified accepted types, these are all reliable ways to get the size, maybe unless there's some weird monkey-patching of the native types involved.
We can throw error or something in such case if you want, although it's possible the native `send` is going to throw anyways.

src/features/metrics/aggregate/websocket-detection.js
cwli24: What do you mean? We specifically removed all the previous SMs. Isn't there a way to get GE events data from ingest queries?

src/common/wrap/wrap-websocket.js
cwli24: Exactly because as you shown, it's unnecessarily elaborate and extra layers of data passing we have to deal with. Instrument already calls "wrap___" so that's clear. Having another `.ee.on()` just to call `handle` for aggregate is very noisy.

I think it makes more sense for like ajax where you're tracking `-start`, `-end`, and doing other extra stuff in instrument.

tests/specs/ins/websockets.e2e.js
cwli24: Separate.

src/common/wrap/wrap-websocket.js
metal-messiah: Yeah ok that makes sense.

src/common/wrap/wrap-websocket.js
metal-messiah: Lets just plan to revisit this if it becomes an issue later

src/features/jserrors/aggregate/index.js
metal-messiah: see other comment

src/features/metrics/aggregate/websocket-detection.js
metal-messiah: The DBC transaction logs do not break down the event types inherently -- So either work needs to be defined to update DBC to add a new transaction attribute specifically for WebSockets, or we need to add a SM to capture it on our side.  I dont mind either way, but we should add a ticket to account for that as we will need one or the other to measure adoption here.

metal-messiah: https://new-relic.atlassian.net/browse/NR-484516

src/common/wrap/wrap-websocket.js
metal-messiah: i understand, but this is another case of introducing "this feature works this way while this feature works that way" instead of a consistent implementation pattern, which we should avoid doing if we can help it.

cwli24: Ok but just know that creating extraneous execution loops because of an old indirection system is simply inefficient.
Changing

metal-messiah: extracted this convo out to a separate standalone ticket https://new-relic.atlassian.net/browse/NR-484610

PR #1616, opened 2025-11-04
feat: Allow nested registrations

metal-messiah:
To facilitate inherent parent-child relationships with the planned register API, allow registered entities to expose their own `.register()` API for children of that entity to register with.  Entities who register with the container agent will be related to the container.  Entities that register under another registered entity will be related to both the container as well as the parent entity.
---

### Related Issue(s)
NR-484626

### Testing
new tests have been added

### DIFF
--- a/src/loaders/api/register.js
+++ b/src/loaders/api/register.js
@@ -28,7 +28,7 @@ import { recordCustomEvent } from './recordCustomEvent'
  */
 export function setupRegisterAPI (agent) {
   setupAPI(REGISTER, function (target) {
-    return buildRegisterApi(agent, target)
+    return register(agent, target)
   }, agent)
 }
 
@@ -36,15 +36,11 @@ export function setupRegisterAPI (agent) {
  * Builds the api object that will be returned from the register api method.
  * Also conducts certain side-effects, such as harvesting a PageView event when triggered and gathering metadata for the registered entity.
  * @param {Object} agentRef the reference to the base agent instance
- * @param {Object} handlers the shared handlers to be used by both the base agent's API and the external target's API
- * @param {Object} target the target information to be used by the external target's API to send data to the correct location
- * @param {string} [target.licenseKey] the license key of the target to report data to
- * @param {string} target.id the entity ID of the target to report data to
- * @param {string} target.name the entity name of the target to report data to
- * @param {string} [target.parentId] the ID of the parent target to report data to
+ * @param {import('./register-api-types').RegisterAPIConstructor} target
+ * @param {import('./register-api-types').RegisterAPIConstructor} [parent]
  * @returns {RegisterAPI} the api object to be returned from the register api method
  */
-export function buildRegisterApi (agentRef, target) {
+function register (agentRef, target, parent) {
   const attrs = {}
   warn(54, 'newrelic.register')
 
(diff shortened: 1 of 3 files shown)


tests/components/logging/aggregate.test.js
```diff
@@ -269,7 +269,8 @@ describe('payloads', () => {
           'mfe.id': registeredTarget.id,
           'mfe.name': registeredTarget.name,
           eventSource: registeredTarget.eventSource,
-          'parent.id': registeredTarget.containerId
+          'parent.id': registeredTarget.containerId,
+          'container.id': registeredTarget.containerId
```
metal-messiah: Remove the container.id attribute for now

src/loaders/api/register.js
```diff
@@ -40,15 +40,22 @@ export function setupRegisterAPI (agent) {
...
   const attrs = {}
   warn(54, 'newrelic.register')
 
   target ||= {}
+  target.eventSource = 'MicroFrontendBrowserAgent'
   target.licenseKey ||= agentRef.info.licenseKey // will inherit the license key from the container agent if not provided for brevity. A future state may dictate that we need different license keys to do different things.
   target.blocked = false
+  if (!target.parentId) {
+    Object.defineProperty(target, 'parentId', {
+      get: () => agentRef.runtime.appMetadata.agents?.[0]?.entityGuid // this is async so must be a getter
```
cwli24: I'm a bit confused as to how this supposed to work. If this method is called prior to RUM response, it's going to be undefined anyways, which means getV2 backfills with the container entity. That means you don't need to set a getter here because `target.parentId` is naturally undefined if it doesn't exist.
And if it's called after... well setting it has no behavioral difference. And if it's called by another MFE, then of course parentId is already provided.
So in summary, this is unnecessary?

metal-messiah: yeah it might be. I think i was going for a "single" source of truth, e.g. just always trust the target.parentId field, but i think the fallback is still in the helper method, so it might be superfluous. I'll double check that assumption and push up a new change without it

src/loaders/api/register.js
```diff
@@ -82,6 +84,7 @@ export function buildRegisterApi (agentRef, target) {
     log: (message, options = {}) => report(log, [message, { ...options, customAttributes: { ...attrs, ...(options.customAttributes || {}) } }, agentRef], target),
     measure: (name, options = {}) => report(measure, [name, { ...options, customAttributes: { ...attrs, ...(options.customAttributes || {}) } }, agentRef], target),
     noticeError: (error, attributes = {}) => report(noticeError, [error, { ...attrs, ...attributes }, agentRef], target),
+    register: (target = {}) => !isBlocked() && agentRef.register({ ...target, parentId: api.metadata.target.id }),
```
cwli24: Don't introduce one-off behaviors for only one method here. Either they all should check blocked or trust the register call itself will block if parent is blocked (might have to add that, but this is not a good place for it.)

metal-messiah: all the other methods check isBlocked as part of calling `report`. maybe theres a way to wrap this into `report` (i just didnt want it to "duplicate" as the other methods do.  I'll look at this

PR #1632, opened 2025-11-19
feat: Add "deregister" API

metal-messiah:
Adds support for deregistering a registered entity, for use with the upcoming Micro Front End support initiative.
---

### Overview
- A registered entity can "deregister" by way of the API.
- Added a "shared" flag to the register API, this will determine whether a new instance or a shared instance will be used for identical IDs.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-398453

### Testing
New tests have been added to validate deregister behaviors.


src/features/utils/event-buffer.js
```diff
@@ -44,18 +49,20 @@ export class EventBuffer {
...
+  add (event, target, evaluatedSize) {
     const addSize = evaluatedSize || stringify(event)?.length || 0 // (estimate) # of bytes a directly stringified event it would take to send
     if (this.#rawBytes + addSize > this.maxPayloadSize) {
       const smTag = inject => `EventBuffer/${inject}/Dropped/Bytes`
       this.featureAgg?.reportSupportabilityMetric(smTag(this.featureAgg.featureName), addSize) // bytes dropped for this feature will aggregate with this metric tag
       this.featureAgg?.reportSupportabilityMetric(smTag('Combined'), addSize) // all bytes dropped across all features will aggregate with this metric tag
       return false
     }
-    this.#buffer.push(event)
+    const hasV2Data = isValidMFETarget(target)
```
ptang-nr: Maybe consider adding an early return here if we already know target is undefined or null.  

Perhaps it'd be more performant if we can pre-determine the event's data version prior to calling `add` and cache that info somewhere so we don't have to check repeatedly 🤔  

```suggestion
    const hasV2Data = target ? isValidMFETarget(target) : false
```

metal-messiah: ```pre-determine the event's data version prior to calling add and cache that info somewhere```
I'm not really following what you have in mind here.  Can you give me an example of what you mean?

src/common/aggregate/aggregator.js
```diff
@@ -67,7 +67,17 @@ export class Aggregator {
...
   }
 
-  #getBucket (type, name, params, customParams) {
+  /**
+   * Returns the required harvesting version for the given dataset. If ANY piece of the payload has V2 data, returns 2, otherwise 1.
+   * @param {*} aggregatorTypes
+   * @returns {1|2}
+   */
+  getRequiredVersion (aggregatorTypes) {
+    if (Object.values((this.take(aggregatorTypes, false) || {})).some(arrayOfPayloads => arrayOfPayloads.some(payload => payload.hasV2Data === true))) return 2
```
cwli24: Are you unintentionally added an additional (unnecessary) property to outgoing payloads?

src/common/aggregate/aggregator.js
```diff
@@ -67,7 +67,17 @@ export class Aggregator {
     return hasData ? results : null
   }
 
-  #getBucket (type, name, params, customParams) {
+  /**
+   * Returns the required harvesting version for the given dataset. If ANY piece of the payload has V2 data, returns 2, otherwise 1.
+   * @param {*} aggregatorTypes
+   * @returns {1|2}
+   */
+  getRequiredVersion (aggregatorTypes) {
```
cwli24: This can just be in the EventAggregator as a sole fn; I don't see why it needs to be onion here.

src/common/util/mfe.js
```diff
@@ -24,14 +24,17 @@ export function hasValidValue (val) {
...
  */
 export function getVersion2Attributes (target, aggregateInstance) {
-  if (aggregateInstance?.harvestEndpointVersion !== 2) return {}
-  const containerAgentEntityGuid = aggregateInstance.agentRef.runtime.appMetadata.agents[0].entityGuid
-  if (!isValidMFETarget(target)) {
+  /** If the feature doesnt support registered entities (V2), no need to add any attributes at all */
+  if (!aggregateInstance.supportsRegisteredEntities) return {}
+  const containerAgentEntityGuid = aggregateInstance?.agentRef.runtime.appMetadata.agents[0].entityGuid
+  /** If the target in question is not an MFE (container), or its "blocked" (feature flag or register issue), just report it to the container */
+  if (!isValidMFETarget(target) || target.blocked) {
```
cwli24: Note that this block doesn't match the function description. Rather, this check belongs in whatever calls "get me version 2 attributes", and brings confusion by sometimes returning version 1 attributes instead.

src/features/generic_events/aggregate/index.js
```diff
@@ -291,7 +291,7 @@ export class Aggregate extends AggregateBase {
       ...obj
     }
 
-    this.events.add(eventAttributes)
+    this.events.add(event, target)
```
cwli24: I recall you just got rid of having the event buffer deal with `target`... only to reintroduce it yet again?

src/features/utils/event-buffer.js
ptang-nr: I am considering caching the value in `target` after the `isValidMFETarget` check in [src/loaders/api/register.js](https://github.com/newrelic/newrelic-browser-agent/blob/main/src/loaders/api/register.js#L76).

src/features/utils/event-buffer.js
```diff
@@ -25,12 +26,16 @@ export class EventBuffer {
...
+    return this.#buffer.some(({ hasV2Data }) => hasV2Data) ? 2 : 1
+  }
+
   isEmpty () {
     return this.#buffer.length === 0
   }
 
   get () {
-    return this.#buffer
+    return this.#buffer.map(({ event }) => event)
```
cwli24: get should be returning just whatever the buffer has. If there's bad data, then the problem to address is the bad .add

src/common/aggregate/aggregator.js
metal-messiah: thanks, nice catch -- i missed that

src/common/util/mfe.js
metal-messiah: These are the version 2 attributes for an invalid target tho... (ie, the container agent wont have a target, but if it still needs to report to v2 it needs those attributes as body attributes)

src/features/generic_events/aggregate/index.js
metal-messiah: yeah, i know.  We need some group feedback here, but essentially the `reason` is different in this PR. (before it was to scope to individual harvests, now it is determining which version to send to)

src/features/utils/event-buffer.js
metal-messiah: since this PR has the buffer now a nested object (event + hasV2data), that would be a much bigger change

src/loaders/api/register.js
```diff
@@ -49,17 +49,17 @@ function register (agentRef, target, parent) {
...
   const registeredEntities = agentRef.runtime.registeredEntities
 
-  /** if we have already registered this target, go ahead and re-use it */
-  const preregisteredEntity = registeredEntities.find(({ metadata: { target: { id, name } } }) => id === target.id)
-  if (preregisteredEntity) {
-    if (preregisteredEntity.metadata.target.name !== target.name) preregisteredEntity.metadata.target.name = target.name
-    return preregisteredEntity
+  if (target.shared) {
+  /** if we have already registered this shared target, go ahead and re-use it */
+    const sharedEntity = registeredEntities.find(({ metadata: { target: { id, name } } }) => id === target.id && !!target.shared)
```
ptang-nr:
```suggestion
    const sharedEntity = registeredEntities.find(({ metadata: { target: { id, name } } }) => id === target.id)
```

src/loaders/api/register-api-types.js
```diff
@@ -21,6 +22,7 @@
  * @typedef {Object} RegisterAPIConstructor
  * @property {string|number} id - The unique id for the registered entity. This will be assigned to any synthesized entities.
  * @property {string} name - The readable name for the registered entity. This will be assigned to any synthesized entities.
+ * @property {boolean} [shared] - Whether the registered entity is intended to be shared. Subsequent registrations with the same id will return the same registration instance.
```
cwli24: I don't understand what this is/means

metal-messiah: I suggest reading this documentation https://whimsical.com/shared-flag-E12Tg3wowMRjn5rEckrY3S which should help illustrate the end goal here

cwli24: I see, but my point was also that this isn't well explained or extracted from the code itself, leading to it being read as ambiguous. I think the description and/or naming can be improved here.

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1632/changes/122407c41a3c8923dbdf97210d55b8dd3d5c4c58

ptang-nr (approved):
LGTM

ptang-nr (approved):

PR #1634, opened 2025-11-20
feat: Ensure responseStart is valid before relying on onTTFB

metal-messiah:
Adds an extra validation step to ensure that responseStart of the nav timing entry is valid before relying on onTTFB to return valid data.  The web-vitals library depends on that value to operate correctly, and without a valid TTFB response, the agent will not send the RUM call.  This change ensures that fallback measures are taken when responseStart is invalid, to ensure a RUM call is sent more reliably, which helps compatibility with unsupported frameworks such as ElectronJS.
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-489088

### Testing
New tests have been added.  Also manually validated the changes work with Electron in original issue.


package.json
```diff
@@ -195,6 +195,7 @@
     "build:npm": "npm run npm:build:esm && npm run npm:build:cjs && npm run npm:build:types && npm run npm:pack",
     "cdn:build": "npm run cdn:build:prod",
     "cdn:build:local": "npm run cdn:webpack",
+    "cdn:build:local-external": "npm run cdn:webpack -- --env mode=local-external",
```
cwli24, line 198: what's this about?

metal-messiah: This allows the agent to be built in such a way that it can be referenced on a page that is not on the bam-test-1 domain but still hitting the local testing server (the normal local build has relative paths to the agent build dir that break when loaded externally, such as in an Electron container)

ptang-nr (dismissed):
🚢

PR #1636, opened 2025-11-22
fix: Observe document.body after document is interactive

ptang-nr:
Addresses the scenario where `MutationObserver.observe()` was issued before `document.body` is available, causing `Failed to execute 'observe' on 'MutationObserver': parameter 1 is not of type 'Node'.` error. 

---

### Overview

In the SPA browser agent, it may be possible to trigger this error if an interaction is triggered before the document body is ready, most likely by scripts in the header.  The fix effectively ignores any would-be SPA interactions that are not triggered by real users.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-491035

### Testing

Add mutation-observer-test.html for this specific scenario.  Manually verified this error can be triggered and that the fix addresses the scenario.  Added as a new e2e test.

Also added extra null checks based on error seen during troubleshooting.
Screenshot 2025-11-21 at 5 32 36 PM

### DIFF
--- a/tests/assets/no-body.html
+++ b/tests/assets/no-body.html
@@ -27,6 +27,7 @@
 
       }
     </script>
+    {init}
     {config}
     {loader}
   </head>
--- a/tests/assets/pagehide-beforeload.html
+++ b/tests/assets/pagehide-beforeload.html
@@ -7,6 +7,7 @@
   <head>
     <title>RUM Unit Test</title>
     <script>Object.defineProperty(document, 'visibilityState', {value: 'hidden', writable: true})</script>
+    {init}
     {config}
     {loader}
   </head>
--- a/tests/assets/rum-data.html
+++ b/tests/assets/rum-data.html
@@ -6,6 +6,7 @@
 <html>
   <head>
     <title>RUM Test</title>
+    {init}
     {config}
     {loader}
     <script>


src/features/soft_navigations/instrument/index.js
```diff
@@ -61,7 +61,7 @@ export class Instrument extends InstrumentBase {
 
     const processUserInteraction = debounce((event) => {
       handle('newUIEvent', [event], undefined, this.featureName, this.ee)
-      domObserver.observe(document.body, { attributes: true, childList: true, subtree: true, characterData: true })
+      domObserver.observe(document.documentElement, { attributes: true, childList: true, subtree: true, characterData: true })
```
metal-messiah: Cool idea. Worth asking:  This is supposed to encapsulate visible DOM changes i think -- so adding a meta tag or something to the head could false-positive trigger this.  do we need to be more selective?

cwli24: Rather, I think targeting documentElement here likely would exclude monitoring on the body in sync time, which just introduces another problem. If body isn't ready, the solution here would be to gate this behind that time, e.g. `DOMContentLoaded`. Allowing interactions to start via clicks before any element is rendered visible was nonsensical to begin with.

ptang-nr: @metal-messiah  Interesting, yeah, I've missed that nuance for SN.  In that case, listening on the body would make more sense 🤔

ptang-nr: @cwli24 - I pushed up the try/catch before I saw your comment.  I noticed we've done that in other places to prevent errors bubbling up.  And the observer continues to listen on the body in the meantime and will succeed once the body is ready.  (I'm wondering if that achieves the same outcome as delaying until `DomContentLoaded` or `document.body` is truthy 🤔)

cwli24: the problem with try-catch is that it's a blanket silent on the issue, which may cause soft navs to go unreported or breaking because the fundamental logic becomes unfulfilled. I think this approach only makes the feature state harder to troubleshoot and gray than being straightforward and enforcing steps.

src/features/soft_navigations/aggregate/index.js
```diff
@@ -229,7 +229,11 @@ export class Aggregate extends AggregateBase {
       if (!this.associatedInteraction) {
         // This new api-driven interaction will be the target of any subsequent .interaction() call, until it is closed by EITHER .end() OR the regular url>dom change process.
         this.associatedInteraction = thisClass.interactionInProgress = new Interaction(API_TRIGGER_NAME, time, thisClass.latestRouteSetByApi)
-        thisClass.domObserver.observe(document.body, { attributes: true, childList: true, subtree: true, characterData: true }) // start observing for DOM changes like a regular UI-driven interaction
+        try {
+          thisClass.domObserver.observe(document.body, { attributes: true, childList: true, subtree: true, characterData: true }) // start observing for DOM changes like a regular UI-driven interaction
```
cwli24: this isn't at risk of document.body not being fully loaded because Aggregate itself isn't executed until after page load event has well fired, so this definitely isn't subjected to the issue

metal-messiah: is there a chance that domObserver is not defined here

src/features/soft_navigations/instrument/index.js
```diff
@@ -61,11 +62,11 @@ export class Instrument extends InstrumentBase {
...
     const processUserInteraction = debounce((event) => {
       handle('newUIEvent', [event], undefined, this.featureName, this.ee)
-      try {
+
+      const startObserving = () => {
         domObserver.observe(document.body, { attributes: true, childList: true, subtree: true, characterData: true })
-      } catch (e) {
-        // do nothing
       }
+      document.body ? startObserving() : onDOMContentLoaded(startObserving)
```
metal-messiah: I know youre getting a bit of the "runaround" here, but IMO we shouldnt need to wait for an event process here.  just check `document readyState`. I worry that deferring the observer until the callback happens will have some side-effects we dont want for the normal processes

src/features/soft_navigations/aggregate/index.js
ptang-nr: Good point 👍   (Double-checked -- it looks like SN would try to "block" if MutationObserver is not defined [code](https://github.com/newrelic/newrelic-browser-agent/blob/main/src/features/soft_navigations/instrument/index.js#L31) so I want to say it shouldn't happen in theory)

src/features/soft_navigations/instrument/index.js
ptang-nr: I appreciate the reviews! 🙏

src/features/soft_navigations/instrument/index.js
```diff
@@ -61,7 +61,10 @@ export class Instrument extends InstrumentBase {
 
     const processUserInteraction = debounce((event) => {
       handle('newUIEvent', [event], undefined, this.featureName, this.ee)
-      domObserver.observe(document.body, { attributes: true, childList: true, subtree: true, characterData: true })
+
+      if (document.readyState !== 'loading') { // i.e. is "interactive" or "complete", "legit" user interaction
```
cwli24: This should be an early-return check to skip processing user interaction altogether (before L63)

tests/assets/spa/dt/fetch-dt-crossorigin-load-2.html
```diff
@@ -7,8 +7,8 @@
   <head>
     <meta charset="utf-8">
     <title>Example Page</title>
-    {config}
     {init}
+    {config}
```
cwli24, line 11: Does this end up affecting the configurations in any way? I removed the sslShim in my branch because it's inserted twice ultimately.

cwli24: in config-transform.js because it's already in loader-transform.js

ptang-nr: > Does this end up affecting the configurations in any way?

I think it should not break anything technically (there is no order dependency in the agent itself between the config and init sections).

I did notice sslShim is present in all 3 transformers (init, config, loader).  Not 100% clear on the historical reason, but seems like guarding of some sort to ensure `ssl` is always set (?)  

I contemplated restoring `init-transform.js` to the way it was before the refactor but mostly went this route to avoid undoing the refactor and potentially breaking other tests (I just had no way or knowing) and leaving the option open for test file setup to not require all 3 snippets.  This is at the cost of essentially requiring `{init}` to always be in front of `{config}`.  When I thumbed through our test files, that was the most used convention (init before config).

metal-messiah: `ssl` is a property in the config, so why don't we just migrate it out of a `shim` and just rely on `testing-server/constants` (where it is already set to `false`).  Any tests that specifically need to set `ssl` to be `true` should just do so in the `init` param like any other `init` param

PR #1640, opened 2025-12-01
feat: Improve agent compatibility with ChatGPT connector apps

metal-messiah:
Small tweaks were made to the agent to allow for better compatibility with the unique iframe constraints of ChatGPT connector applications.
---

### Overview
1. For whatever reason (still unknown) -- The GPT-specific iframe does not emit the window load event, which the current browser agent is subscribed to in many places.  Most notably in the lazy-loading phase of the agent's lifecycle which prevents it from downloading the rest of its code and causes it to never harvest anything.
  - Beyond this, the lack of this event also causes NR events like BrowserInteraction and some PageViewTiming events to never collect, since they also expected that event in their core logic
In Safari specifically, certain browser-level navigation timing APIs return undefined in the iframe, which caused the agent to under-report certain load timings
2. The agent has been updated to check multiple global events now to power its logic (the window load event, as well as the document readystate, for redundancy) as well as fallback APIs for timing logic.

### Testing
New tests have been added to ensure base changes are still congruent with general agent behaviors.

### DIFF
--- a/src/common/timing/nav-timing.js
+++ b/src/common/timing/nav-timing.js
@@ -71,7 +71,7 @@ export function addPN (pn, v) {
     handleValue(getPntType(pn.type), v, 'ty')
     handleValue(pn.redirectCount, v, 'rc')
   } catch (e) {
-    v.py = 0
+    v.ty = 0
     v.rc = 0
   }
   return v


src/common/timing/nav-timing.js
```diff
@@ -67,8 +67,13 @@ export function addPT (offset, pt, v = {}, isL1Api = false) {
 
 // Add Performance Navigation values to the given object
 export function addPN (pn, v) {
-  handleValue(getPntType(pn.type), v, 'ty')
-  handleValue(pn.redirectCount, v, 'rc')
+  try {
+    handleValue(getPntType(pn.type), v, 'ty')
+    handleValue(pn.redirectCount, v, 'rc')
+  } catch (e) {
+    v.py = 0
```
ptang-nr: Possible typo
```suggestion
    v.ty = 0
```

ptang-nr (approved):
LGTM

PR #1646, opened 2025-12-05
feat: Add resetSession option to setUserId() API

ptang-nr:
Introduces option to reset the browser agent session when updating a userid via the `setUserId()` API.
---

### Overview

Currently, the browser agent session may last up to 30 min of inactivity or 4 hours max total.  To better support customers who would like to have a single user per session, we are introducing a new optional param in the `setUserId()` API.

**Example 1**
The existing user is `user1` and should be updated to `user2`.  To also start a new session, call `newrelic.setUserId('user2', true)`.  Otherwise, call `newrelic.setUserId('user2') to update the user only.

**Example 2**
There is no existing user and customer wants to update to `user1`.  It will not be possible to start a new session as this will be the first user of the session.  Calling `newrelic.setUserId('user1', true)` or `newrelic.setUserId('user1')` will both result in updating the user only. 

**Example 3**
The existing user is `user1` and customer issues a `newrelic.setUserId('user1',true)` command.  This will result in no action - both the session and user will stay the same.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-482566
https://source.datanerd.us/agents/angler/pull/775 `merged`
https://github.com/newrelic/docs-website/pull/22455 `reviewed`

### Testing

Manual + added relevant setUserId() e2e test


src/loaders/api/setUserId.js
```diff
@@ -10,13 +10,22 @@ export function setupSetUserIdAPI (agent) {
...
     if (!(typeof value === 'string' || value === null)) {
       warn(41, typeof value)
       return
     }
-    return appendJsAttribute(agent, 'enduser.id', value, SET_USER_ID, true)
+
+    const currUser = agent.info.jsAttributes['enduser.id']
+
+    // only reset session if we are not setting the userid for the first time
+    const shouldReset = resetSession && agent.runtime.session && currUser != null && currUser !== value
```
metal-messiah: I think this would trigger a new session if currUser is `undefined` based on the shouldReset logic, right?

metal-messiah (dismissed):
Can you please update the interface files that control/check typings generation as part of this change? (jsdoc + args)

- api-base.js
- registered-entity.js
  - This will need to be "blocked" from resetting the session from a MFE.  That would not make sense to allow
- register-api-types.js
- api.test-d.ts
- regsitered-entity.test-d.ts
- api.test.js
- agent-session.test.js

We should also consider adding a new SM as part of this work to determine when a session has been reset due to the usage of this API

src/loaders/api/setUserId.js
ptang-nr: I learned this "trick" from another code review - apparently `!= null` covers both not `null` and not `undefined`.  I can make it explicit though if that's better for clarity.

metal-messiah: Oh i missed that bit, thanks for the clarity.  I'm surprised the linter let that get through because i think we follow [this rule.](https://eslint.org/docs/latest/rules/eqeqeq).  I'd just make it explicit if you don't mind, just for future readability.

src/loaders/api-base.js
```diff
@@ -84,9 +84,10 @@ export class ApiBase {
    * Adds a user-defined identifier string to subsequent events on the page.
    * {@link https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/setuserid/}
    * @param {string|null} value A string identifier for the end-user, useful for tying all browser events to specific users. The value parameter does not have to be unique. If IDs should be unique, the caller is responsible for that validation. Passing a null value unsets any existing user ID.
+   * @param {boolean} [resetSession=false] Optional param. When true, resets the current session (if any) when changing user id. Note: if setting user id for the first time, session will not be reset.
```
metal-messiah: <suggestion>
* @param {boolean} [resetSession=false] Optional param. When true, resets the current session when changing user id. Note: If the current user id is null when calling the API, the session can not be reset.
</suggestion>

metal-messiah: Maybe this should more specifically call out the null -> defined behavior

src/loaders/api/setUserId.js
```diff
@@ -5,18 +5,31 @@
...
       return
     }
-    return appendJsAttribute(agent, 'enduser.id', value, SET_USER_ID, true)
+
+    const currUser = agent.info.jsAttributes['enduser.id']
+
+    // reset session ONLY if we are updating the userid (from one value to another, or from a value to null/undefined)
+    const shouldReset = resetSession && agent.runtime.session && currUser !== undefined && currUser !== null && currUser !== value
+    if (shouldReset) {
+      agent.runtime.session.reset()
```
metal-messiah: We need to sync on this. There is an asynchronous nature to the session that needs to be accounted for here that isnt.

PR #1649, opened 2025-12-09
fix: Remove linebreak syntax from webpack output

ellisong:
Removes the newline character in webpack output to avoid issues with copy-pasting our script as a string inside code.
---

### Overview

This PR contains a new webpack plugin that can regex replace characters in our compiled assets. We use this to replace a newline character from templated webpack code to avoid an issue where the newline character breaks certain frameworks such as NextJS that reply on using our copy pasted agent as a string in their code.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-490425

### Testing

The compiled webpack files in the build directory (.js, .min.js, .js.map, .min.js.map) should have the newline removed. When searching for "Loading chunk" in build directory, results should say something like `'Loading chunk ' + chunkId + ' failed: ('` instead of `'Loading chunk ' + chunkId + ' failed\n('`.
The minified files will use double quotes instead of single quotes.

Make sure our source maps can be applied properly to our compiled js files. They were also altered.


tools/webpack/configs/common.mjs
```diff
@@ -72,6 +73,16 @@ export default (env, asyncChunkName) => {
...
+      new NRBAReplaceRegexPlugin({
+        test: /(['"] failed)\.\\n(\(["'])/g,
+        replace: '$1: $2',
+        include: /\.js$/
+      }),
+      new NRBAReplaceRegexPlugin({
+        test: /(['"] failed)\.\\\\n(\(["'])/g,
+        replace: '$1: $2',
+        include: /\.js.map$/
+      }),
```
metal-messiah, line 85: This seems like it solves the current problem but doesnt "future proof" us if webpack internals happen to change in the future.  Is there a more generic regex statement we should investigate using?

ellisong: If I make it more generic than this, we run the risk of replacing newlines that we shouldn't be. I searched for a very specific slice of string characters here that can't be minified to try to be as generic as possible.

metal-messiah (approved):

PR #1655, opened 2025-12-12
feat: Ajax deny list wildcard support

ellisong:
Adds wildcard character support, with the `*` character, to the AJAX deny list feature which allows to substitute one or more characters in your deny list filters.
---

### Overview

Adds wildcard support for AJAX deny list filters. We do this by modifying the deny list algorithms, using regex searching instead of length and direct string comparison.

When specifying `*` in your filter, it excludes all characters up to the character after the wildcard. We convert your filters into regular expression objects, escaping all regex characters except `*` which turns into `.*?`, and then search the AJAX request URL against each of the filters, denying any URLs that contain a positive match for both hostname and pathname.

Wildcards in our code are using lazy matching instead of greedy matching. This allows more fine-grained control of using wildcards in our AJAX deny list filters.

Existing deny list filters in our users current agent deployments will continue to work as normal with this change. No mandatory changes are needed on their end.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-491985

### Testing

Unit tests were added and expanded on for deny list. Make sure all tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,19 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [1.306.0](https://github.com/newrelic/newrelic-browser-agent/compare/v1.305.0...v1.306.0) (2025-12-16)
+
+
+### Features
+
+* Control log API through separate RUM flag ([#1467](https://github.com/newrelic/newrelic-browser-agent/issues/1467)) ([f9f1639](https://github.com/newrelic/newrelic-browser-agent/commit/f9f1639504a32b0f24baa117d287c47e209cc818))
+* Ensure responseStart is valid before relying on onTTFB ([#1634](https://github.com/newrelic/newrelic-browser-agent/issues/1634)) ([cd83f60](https://github.com/newrelic/newrelic-browser-agent/commit/cd83f605087a79cbc4817ddfb82d48c18eae8474))
+
+
+### Bug Fixes
+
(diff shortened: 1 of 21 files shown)


src/common/deny-list/deny-list.js
metal-messiah, line 21: We could have a short-circuit here, if someone supplied `["*"]` as a value, we could just short circuit and return `false` instead of looping through other values or applying regex searches

Could pair that with a change to `setDenyList` -- if someone supplies `["*"]` the deny list could just be a single "block-all" item, no need for others to be stored

src/common/deny-list/deny-list.js
```diff
@@ -68,46 +64,25 @@ export function setDenyList (denyListConfig) {
...
-  if (pattern === '') {
-    return true
-  }
-
-  return pattern === path
+function convertToRegularExpression (filter) {
+  const newFilter = filter
+    .replace(/[.+?^${}()|[\]\\]/g, (m) => '\\' + m) // use a replacer function to not break apm injection
+    .replace(/\*/g, '.*?') // use lazy matching instead of greedy
+  return new RegExp(newFilter)
```
metal-messiah: because this regex isnt anchored, it can select cases that are "wrong". I added an example test on my branch to illustrate this

Image

metal-messiah:
```suggestion
  return new RegExp('^' + newFilter + '$')
```

metal-messiah: applying this does seem to block old behavior though, like having `foo.com` apply to block `www.foo.com`.  Evaluate that

PR #1658, opened 2025-12-18
fix: Micro agent duplicate handlers

ptang-nr:
Fixes when micro agents are created in a staggered fashion, handlers for each agent is multiplied by the number of agents that follow.  For example if there were two agents, the handlers for the first agent was registered twice.  This had the effect of duplicating events + contents of harvest payloads.

---

### Overview

This PR avoids re-registering handlers as each micro agent goes through the drain process. Also introduces an update to the test setup to have some variation in the logging modes returned in the local test server.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-481854

### Testing

- Manual testing against the Rollup and browser-agent-wrapper micro agent examples
- Updated micro-agent.e2e.js test conditions

### DIFF
--- a/tests/components/logging/aggregate.test.js
+++ b/tests/components/logging/aggregate.test.js
@@ -32,15 +32,12 @@ afterEach(() => {
   jest.clearAllMocks()
 })
 
-const INIT_TIMEOUT = 20
-const wait = (ms) => new Promise(resolve => setTimeout(resolve, ms))
-
 const mockLoggingRumResponse = async (mode, apiMode) => {
   loggingAggregate.ee.emit('rumresp', [{
     log: mode,
     logapi: apiMode
   }])
-  return await wait(INIT_TIMEOUT)
+  return await new Promise(process.nextTick)
 }
 
 describe('class setup', () => {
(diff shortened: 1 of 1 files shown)


src/features/logging/aggregate/index.js
```diff
@@ -44,27 +44,47 @@ export class Aggregate extends AggregateBase {
...
+        }
+        updateLocalLoggingMode(currentState.loggingMode, currentState.logApiMode)
       }
-
-      /** emitted by instrument class (wrapped loggers) or the api methods directly */
-      registerHandler(LOGGING_EVENT_EMITTER_CHANNEL, this.handleLog.bind(this), this.featureName, this.ee)
-      this.drain()
-      /** harvest immediately once started to purge pre-load logs collected */
-      agentRef.runtime.harvester.triggerHarvestFor(this)
+      this.#completeInitialization(agentRef)
```
metal-messiah: Im confused why not handle this case more directly in the session event listener https://github.com/newrelic/newrelic-browser-agent/pull/1658/changes#diff-86f2cdde0a7f5866d64a9029ee41fc85c7798401971dd86e66018797b7753dacR38 

It naively seems like there could be a case for same-tab updates there that isnt arbitrary reads/delays

ptang-nr: Went back to take a second look - I think I couldn't find a way to make same-tab updates work for me.  Same-tab updates would only fire for each individual agent and not _across_ agents.  

Edit: I was using the wrong example for my test.  It looks like cross tab is firing in the near simultaneous case, so gonna explore this a bit more.
Edit 2: Never mind, the cross tab event may have a bad setup (?) 😵 I can't repro it after updating my test setup (custom-commands.mjs + testing-server/constants.js)
Edit 3: Ok, I think I've tracked the cross tab event down to other old tabs that I forgot to close 🫠

ptang-nr: Came back fresh today + use the right test file (browser-agent-wrapper/micro-agent.html) and confirmed I'm not seeing any session updates firing for the second micro agent.

https://github.com/user-attachments/assets/31cc10d6-27c2-432f-98d9-192ab6bc0f62

ptang-nr: Discussed offline - instead of timing-based (10ms is arbitrary), moving towards event-based checking would've been better.  But at this time, the changes that would be needed to enable cross-agent events (maybe via `globalEE`) does not align with the direction of the agent and we agreed this is falling under the "Not fixing" bucket, in favor of MFE where there should be only 1 agent on a page going forward.

PR #1659, opened 2025-12-19
chore: republish dev when deleting an experiment

metal-messiah:
We need to clear the CDN cache and republish dev assets when we delete experiment builds
---

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -19,7 +19,7 @@ concurrency:
 
 jobs:
   jest-tests:
-    if: github.ref == 'refs/heads/main'
+    if: github.ref == 'refs/heads/main' && github.event_name != 'workflow_call'
     name: Jest tests
     uses: ./.github/workflows/jest.yml
     secrets: inherit
@@ -28,6 +28,7 @@ jobs:
 
   # Build and publish the latest code from the main branch
   publish-dev-to-s3:
+    if: always() && !cancelled() && !contains(needs.*.result, 'failure')
     needs: [jest-tests]
     runs-on: ubuntu-latest
     timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-dev.yml
```diff
@@ -27,6 +28,7 @@ jobs:
 
   # Build and publish the latest code from the main branch
   publish-dev-to-s3:
+    if: always() && !cancelled() && !contains(needs.*.result, 'failure')
```
ptang-nr, line 31: What is happening here? Where does the search phrases `needs.*.result` and `failure` come from?

.github/workflows/publish-dev.yml
```diff
@@ -18,7 +19,7 @@ concurrency:
 
 jobs:
   jest-tests:
-    if: github.ref == 'refs/heads/main'
+    if: github.ref == 'refs/heads/main' && github.event_name != 'workflow_call'
```
ptang-nr, line 22: Just a thought - Based on the changes in this PR, I'm assuming so far `delete-experiment` is the only workflow calling `publish-dev`.  So skipping jest tests and publish agent2query makes sense.  Maybe just a mental note that we may have to revisit in the future if/when new workflows call `publish-dev`.

.github/workflows/publish-dev.yml
metal-messiah: These make it so that the job will at least evaluate conditions, since it has a `needs` and an `if` statement

The `needs: [jest-tests]` creates the dependency (aka, don't run the action until jest-tests runs), and then the `if` condition checks the results of those needed jobs via `needs.*.result`.

So `needs.*.result` evaluates to an array containing the result of the jest-tests job (either "success", "failure", "skipped", or "cancelled"). The condition then checks if any of those results is "failure" specifically (which would tell the job not to run). If the job is "skipped", this action will still be allowed to run.

Without `needs: [jest-tests]`, the `needs.*.result` would be empty and the condition wouldn't work as intended. The `needs` declaration is what populates the data that the if condition evaluates.

PR #1661, opened 2025-12-30
feat: add parent.type attribute to facilitate relationship synthesis

metal-messiah:
Adds the required parent.type attribute for future relationship synthesis during MFE ingest.
---

### Related Issue(s)
Please see https://github.com/newrelic/entity-definitions/pull/2544 for the sister PR facilitating relationship synthesis from attribution

### Testing
Updated relevant tests to ensure attribute is added correctly

### DIFF
--- a/tests/unit/common/util/v2.test.js
+++ b/tests/unit/common/util/v2.test.js
@@ -3,64 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { isValidMFETarget, hasValidValue, getVersion2Attributes } from '../../../../src/common/util/mfe'
+import { getVersion2Attributes } from '../../../../src/common/util/v2'
 
-describe('mfe utilities', () => {
-  describe('isValidMFETarget', () => {
-    test('returns true when target has id and name', () => {
-      expect(isValidMFETarget({ id: 'test-id', name: 'test-name' })).toBe(true)
-    })
-
-    test('returns false when target has id but no name', () => {
-      expect(isValidMFETarget({ id: 'test-id' })).toBe(false)
-    })
-
-    test('returns false when target has name but no id', () => {
-      expect(isValidMFETarget({ name: 'test-name' })).toBe(false)
-    })
-
-    test('returns false when target is empty', () => {
-      expect(isValidMFETarget({})).toBe(false)
-    })
-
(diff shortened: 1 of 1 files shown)


src/common/util/mfe.js
```diff
@@ -12,7 +12,7 @@ export function isValidMFETarget (target = {}) {
 }
 
 export function hasValidValue (val) {
-  return (typeof val === 'string' && val.trim().length < 501) || (typeof val === 'number')
+  return (typeof val === 'string' && !!val.trim() && val.trim().length < 501) || (typeof val === 'number')
```
ptang-nr: Just realized with this change, we can omit the `isValidMFETarget` call on https://github.com/newrelic/newrelic-browser-agent/blob/main/src/loaders/api/register.js#L76 🤔 it'll be redundant

metal-messiah: updated in [fbf5e79](https://github.com/newrelic/newrelic-browser-agent/pull/1661/commits/fbf5e7984dd38d95a938ef714db221ed634e40ff)

ptang-nr: Sorry lost track of this PR - re: the update, I see that `isValidMFETarget` is also called within `getVersion2Attributes`, which in turn is called by several feature aggs when collecting events.  

I'm hesitant about invoking this validation on a potentially hot path.  The validation/complexity isn't heavy, but it also seems like `target` is not expected to change throughout the lifetime of agents.  Is there a way we can avoid the extra checks?  Would having the validation during registration be good enough?

metal-messiah: yeah i think youre probably right about this.  We can probably just check if there is/isnt a target and trust that the register API did the validation.  I'll make that change

metal-messiah: Addressed in https://github.com/newrelic/newrelic-browser-agent/pull/1661/changes/fbf5e7984dd38d95a938ef714db221ed634e40ff..ba0b55609edd289391d6b894e87e61004ecd62bd

ptang-nr: Changes look good 👍  

One last thing for future consideration - it feels like we ought to formalize somewhere that `!target` means container agent.  Right now, it's more of an implicit understanding 🤔

PR #1663, opened 2025-12-31
fix: Clean WebSocket URLs

metal-messiah:
Remove hash fragments and query params from URLs reported on WebSocket events
---

### Related Issue(s)
NR-498982

### Testing
New tests have been added

### DIFF
--- a/tests/assets/websockets.html
+++ b/tests/assets/websockets.html
@@ -14,7 +14,7 @@
     <script>
       // preload socket
       window.bamServer = NREUM.info.beacon
-      window.preLoadSocket = new WebSocket(`ws://${bamServer}/websocket?loaded=pre`)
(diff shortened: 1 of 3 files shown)


tests/components/wrappings/wrap-websocket.test.js
```diff
@@ -65,6 +65,91 @@ describe('wrap-websocket', () => {
     expect(global.WebSocket).toBeUndefined()
   })
 
+  it('cleans URLs by removing query params and hash routes', () => {
+    const testCases = [
```
ptang-nr, line 69: May be missing the hash test case (?)

ptang-nr (approved):
LGTM

tests/components/wrappings/wrap-websocket.test.js
metal-messiah: hash is not supported here in the constructor

PR #1666, opened 2026-01-02
feat: Allow Session Replay to retry payloads

metal-messiah:
Payloads failed with retryable codes will be re-harvested by the agent. Further session replay data will not be collected until the harvest retries successfully.
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-489361

### Testing
New tests have been added.


src/features/session_replay/aggregate/index.js
```diff
@@ -209,6 +209,7 @@ export class Aggregate extends AggregateBase {
   }
 
   makeHarvestPayload () {
+    if (this.retrying && this.recorder.retryPayload) return this.recorder.retryPayload
```
cwli24:
```suggestion
    if (this.isRetrying) return this.recorder.retryPayload
```

src/features/session_replay/aggregate/index.js
```diff
@@ -338,9 +340,17 @@ export class Aggregate extends AggregateBase {
   }
 
   postHarvestCleanup (result) {
-    // The mutual decision for now is to stop recording and clear buffers if ingest is experiencing 429 rate limiting
-    if (result.status === 429) {
-      this.abort(ABORT_REASONS.TOO_MANY)
+    if (result.sent) {
+      if (result.retry) {
+        this.retrying = true
```
cwli24: use `isRetrying` from base class

src/features/session_replay/constants.js
```diff
@@ -54,5 +54,6 @@ export const TRIGGERS = {
   RESUME: 'resume',
   SWITCH_TO_FULL: 'switchToFull',
   INITIALIZE: 'initialize',
-  PRELOAD: 'preload'
+  PRELOAD: 'preload',
+  RETRY_SUCCESS: 'retrySuccess'
```
cwli24: Remove this change; unused

src/features/session_replay/aggregate/index.js
```diff
@@ -338,9 +340,17 @@ export class Aggregate extends AggregateBase {
...
+        this.forceStop()
+      } else {
+        this.recorder.retryPayload = undefined
+        if (this.retrying) {
+          this.retrying = false
+          this.switchToFull()
```
cwli24, line 352: We need to have a way to signal to UI or customers that their SR has a gap in it because of this failed harvest-retry cycle and put that in place prior to deploy.

src/features/session_replay/aggregate/index.js
metal-messiah: Session replay uses `this.retrying` to track the phase(s) of a retry operation - specifically to determine when to resume recording after a successful retry.  That means it cant be self-referential with the `.isRetrying` attribute. We need to be able to determine if we should stop recording (first phase) or resume recording (second phase).

src/features/session_replay/aggregate/index.js
metal-messiah: Added warning code 70

src/features/session_replay/aggregate/index.js
cwli24: From what I can see in the code, `this.isRetrying` is the same behavior. The check in `makeHarvestPayload` was also renamed so I'm confused why this is kept as `retrying`.

metal-messiah: We can overset the base attribute if you prefer, but it has to be set from the replay class and not the base in this case

cwli24 (approved):

PR #1668, opened 2026-01-05
chore: Silence test noise

metal-messiah:
Jest tests are very noisy with debug noise from the agent warn module. This PR silences those by setting the global console methods to mocked fns.
---

### Testing
See the [jest output](https://github.com/newrelic/newrelic-browser-agent/actions/runs/20723140266/job/59492011498?pr=1668) to see warnings are no longer spammed in the output -- compare to [non-silenced tests ](https://github.com/newrelic/newrelic-browser-agent/actions/runs/20720172764/job/59481500089?pr=1667)


tools/jest/globals.mjs
```diff
@@ -6,4 +6,10 @@ if (typeof window !== 'undefined') {
   window.Response = jest.fn()
 }
 
+/** silence unneeded console debug (warn(...)) noise during tests */
+if (typeof console !== 'undefined') {
+  console.error = jest.fn()
+  console.debug = jest.fn()
```
ptang-nr, line 12: Would `jest --silent` be an acceptable alternative?  

Agreed console prints from tests are noisy!  Not sure what they all are (should we clean them out or add lint rules?)  / if we'd ever want to reserve the ability to have debug prints(?)

PR #1673, opened 2026-01-08
chore: split api tests into separate files to increase run speed

metal-messiah:
By splitting the very long (1500+ lines) api.e2e.js test file into separate files, the work can be parallelized easily with the testing suite.  This causes ~80% faster runtime, reducing the test run from around 6 mins down to about 1 minute.

Screenshot 2026-01-08 at 3 28 01 PM

---


ptang-nr (approved):
LGTM

PR #1676, opened 2026-01-12
feat: Capture MFE timings as MicroFrontEndTiming Events

metal-messiah:
Captures the lifecycle timings of registered entities in preparation for future micro front end product
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-398453?atlOrigin=eyJpIjoiYWQ3ZDJjZGE5NjIwNDAzMzhjMDk3NTE1ZTQzODExYmYiLCJwIjoiaiJ9

### Testing
new tests have been added

### DIFF
--- a/src/loaders/api/register-api-types.js
+++ b/src/loaders/api/register-api-types.js
@@ -30,6 +30,12 @@
 /**
  * @typedef {Object} RegisterAPIMetadata
  * @property {Object} customAttributes - The custom attributes for the registered entity.
+ * @property {Object} timings - The timing metrics for the registered entity.
+ * @property {number} timings.registeredAt - The timestamp when the registered entity was created.
+ * @property {number} [timings.reportedAt] - The timestamp when the registered entity was deregistered.
+ * @property {number} timings.fetchStart - The timestamp when the registered entity began fetching.
+ * @property {number} timings.fetchEnd - The timestamp when the registered entity finished fetching.
+ * @property {Object} [timings.asset] - The asset path (if found) for the registered entity.
  * @property {Object} target - The options for the registered entity.
  * @property {string} [target.licenseKey] - The license key for the registered entity. If none was supplied, it will assume the license key from the main agent.
  * @property {string} target.id - The ID for the registered entity.


src/loaders/api/register.js
```diff
@@ -123,7 +133,29 @@ function register (agentRef, target, parent) {
...
+   * Reports the gathered timings for the registered entity through a custom event to the container agent. Only reports once
+   * by checking for the presence of the reportedAt timing.
+   * @returns {void}
+   */
+  function reportTimings () {
+    // only ever report the timings the first time this is called
+    if (timings.reportedAt) return
+    timings.reportedAt = now()
+    api.recordCustomEvent('MicroFrontEndTiming', {
+      duration: timings.reportedAt, // origin to reportedAt
```
cwli24: I'm not sure how this "duration" of MFE makes sense

src/common/util/script-tracker.js
```diff
@@ -0,0 +1,74 @@
...
+
+const chrome = /^\s*at (?:((?:\[object object\])?(?:[^(]*\([^)]*\))*[^()]*(?: \[as \S+\])?) )?\(?((?:file|http|https|chrome-extension):.*?)?:(\d+)(?::(\d+))?\)?\s*$/i
+const gecko = /^\s*(?:(\S*|global code)(?:\(.*?\))?@)?((?:file|http|https|chrome|safari-extension).*?):(\d+)(?::(\d+))?\s*$/i
+
+/** The scripts loaded on to the page, as observed by the PerformanceObserver */
+const scripts = new Set()
+
+if (globalScope.PerformanceObserver?.supportedEntryTypes.includes('resource')) {
+  const scriptTracker = new PerformanceObserver((list) => {
+    list.getEntries().forEach(entry => scripts.add(entry))
```
cwli24: Extra memory hog overtime as a page loads more and more script is a concern here

src/common/util/script-tracker.js
```diff
@@ -0,0 +1,74 @@
...
+ * Given an error stack, return script timing information if a script can be found with the resource timing API
+ * @param {Error.stack} stack
+ * @returns {{fetchStart: number, fetchEnd: number}}
+ */
+export function findScriptTimingsFromStack (stack) {
+  const timings = { fetchStart: 0, fetchEnd: 0 }
+  if (scripts.size < 1 || !stack) return timings
+
+  try {
+    const stackUrls = extractUrlsFromStack(stack)
```
cwli24: I don't quite understand the intention here. There can be various script source from an error stack. Is there a proof that the MFE source is on the stack & parse-able such that this doesn't end up grabbing the wrong script?

Also, the error is generated in the agent code... wouldn't the first line be pointing to the agent script itself?

src/loaders/api/register.js
```diff
@@ -123,7 +133,29 @@ function register (agentRef, target, parent) {
...
+      timeToLoad: timings.registeredAt - timings.fetchStart, // fetchStart to registeredAt
```
cwli24, line 151: Have you taken into account prefetched / prerendered scripts and what that can mean for time to load?

src/common/util/script-tracker.js
metal-messiah: switch to use static lookup of resources at register time instead of curating a set of data over time from PO

src/loaders/api/register.js
metal-messiah: removing, as discussed

src/common/util/script-tracker.js
metal-messiah: switching to static lookup

src/loaders/api/register.js
```diff
@@ -52,6 +54,16 @@ function register (agentRef, target, parent) {
   target.parent = parent || {}
   if (typeof target.tags !== 'object' || target.tags === null || Array.isArray(target.tags)) target.tags = {}
 
+  // set the stack limit higher to capture more stack frames for script tracking, then set it back to original after
+  const originalStackLimit = Error.stackTraceLimit
+  Error.stackTraceLimit = 50
```
cwli24: Image

Image
I feel like using this is going to cause issues

src/common/util/script-tracker.js
```diff
@@ -0,0 +1,60 @@
...
+import { globalScope } from '../constants/runtime'
+import { cleanURL } from '../url/clean-url'
+import { chrome, gecko } from './browser-stack-matchers'
+
+/**
+ * Extracts URLs from stack traces using the same logic as compute-stack-trace.js
+ * @param {string} stack The error stack trace
+ * @returns {string[]} Array of cleaned URLs found in the stack trace
+ */
+function extractUrlsFromStack (stack) {
```
cwli24, line 15: Instead of this func, you should just re-use computeStackTrace to get your frames/urls

src/loaders/api/register.js
metal-messiah: Yeah maybe.   I tested it in all the browsers, for firefox it doesn't throw errors, it just "does nothing"

Let me know what you think we should do there @cwli24 -- Do you want to check if its defined first? In FF it returns undefined while other browsers return a value

Image

src/common/util/script-tracker.js
```diff
@@ -0,0 +1,60 @@
...
+ * @returns {{fetchStart: number, fetchEnd: number}}
+ */
+export function findScriptTimingsFromStack (stack) {
+  const timings = { fetchStart: 0, fetchEnd: 0, asset: undefined }
+  /** @type {PerformanceResourceTiming[]} The list of script resource timing entries */
+  const scripts = globalScope.performance?.getEntriesByType('resource').filter(entry => entry.initiatorType === 'script') || []
+  if (scripts.length < 1 || !stack) return timings
+
+  try {
+    const mfeScriptUrl = extractUrlsFromStack(stack).at(-1) // array of URLs from the stack of the register API caller, the MFE script should be at the bottom
```
metal-messiah, line 59:
```suggestion
    const mfeScriptUrl = extractUrlsFromStack(stack).at(-1) // array of URLs from the stack of the register API caller, the MFE script should be at the bottom
    if (!mfeScriptUrl) return timings
```

metal-messiah:
> Last question: is `mfeScriptUrl.endsWith(scriptUrl) || scriptUrl.endsWith(mfeScriptUrl)` necessary? Aren't both url going to be the full https://... format?

for posterity, these were added just as a low-weight precaution in case the browser handles the file names in an unpredicted way.  They probably don't get applied in a meaningful way most of the time. We can re-evaluate later when we have higher confidence that they are no longer needed

PR #1679, opened 2026-01-13
feat: Add WebDriver detection attribute to PageView & PageViewTiming

cwli24:
`PageView`, `PageViewTiming` and initialPageLoad-category `BrowserInteraction` events will now report a `webdriverDetected` property corresponding to the `navigator.webdriver` set by the browser.
---

### Overview

Disclaimer: this is not a smart or heuristic driven approach to detect automation or bot traffic. It simply passes forward a browser API property that _can_ be inaccurate or altered by automation tools.

A true / false `webdriverDetected` property is attached to all PageView & PageViewTiming events or nodes. For BrowserInteraction, only the initialPageLoad node will have this custom property (routeChange or custom interactions will not).

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-367761

### DIFF
--- a/tests/specs/api.e2e.js
+++ b/tests/specs/api.e2e.js
@@ -2,6 +2,7 @@ import { checkAjaxEvents, checkJsErrors, checkMetrics, checkGenericEvents, check
 import { testAjaxEventsRequest, testAjaxTimeSlicesRequest, testBlobTraceRequest, testCustomMetricsRequest, testErrorsRequest, testEventsRequest, testInsRequest, testInteractionEventsRequest, testLogsRequest, testMetricsRequest, testMFEErrorsRequest, testMFEInsRequest, testRumRequest, testTimingEventsRequest } from '../../tools/testing-server/utils/expect-tests'
 import { rumFlags } from '../../tools/testing-server/constants'
 import { LOGGING_MODE } from '../../src/features/logging/constants'
+import { lambdaTestWebdriverFalse } from '../../tools/browser-matcher/common-matchers.mjs'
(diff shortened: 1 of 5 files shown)


src/common/util/webdriver-detection.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+export const webdriverDetected = (() => {
+  try {
+    // Standard WebDriver flag
+    if (typeof navigator !== 'undefined' && navigator.webdriver === true) {
+      return true
+    }
+
+    // Check for various automation-related properties
+    if (typeof window !== 'undefined' && typeof document !== 'undefined') {
```
metal-messiah: nitpick that definitely does not require fixing, but you could have leveraged the DRY `isBrowserScope` constant here.

src/common/util/webdriver-detection.js
```diff
@@ -0,0 +1,55 @@
...
+
+      // PhantomJS detection
+      if (window.callPhantom || window._phantom) {
+        return true
+      }
+
+      // Nightmare.js detection
+      if (window.__nightmare) {
+        return true
+      }
```
metal-messiah, line 49: do these types of globals get set at the beginning of the page lifecycle or in the middle?  If in the middle it might be better to make this export a getter, if at the beginning I think we are good to go and you can resolve this comment and merge

metal-messiah: actually nvm, i can see this is only imported in the agg module anyway, so disregard this.  LGTM

PR #1683, opened 2026-01-15
feat: Improve MFE tag setting

metal-messiah:
Enforce tags to be set for micro frontends by supplying an object of key-value pairs to the `tags` argument of the `register(...)` API.  This allows for tags to be specified with an explicit value instead of only supporting boolean tags. 
---

### Testing
Tests have been modified to validate this change in behavior

### DIFF
--- a/tests/specs/api.e2e.js
+++ b/tests/specs/api.e2e.js
@@ -433,7 +433,8 @@ describe('newrelic api', () => {
       })
     })
 
-    it('should exclude protected "name" and "id" keys from tags', async () => {
+    it('should exclude protected keys from tags', async () => {
+      // excluded keys right now are [name, id, type]
       const [mfeErrorsCapture] = await browser.testHandle.createNetworkCaptures('bamServer', [
         { test: testMFEErrorsRequest }
       ])


tests/components/api.test.js
```diff
@@ -942,41 +949,111 @@ describe('API tests', () => {
...
 
           expect(myApi.metadata.customAttributes).toEqual({
-            'source.module1': true,
+            'source.module1': 'shopping-cart',
             foo: 'bar',
             'application.version': '1.0.0'
           })
         })
 
         test('should exclude protected "name" and "id" keys from tags', () => {
```
ellisong, line 965: test name should also include "type"

src/loaders/api/register.js
```diff
@@ -21,6 +21,8 @@ import { recordCustomEvent } from './recordCustomEvent'
  * @typedef {import('./register-api-types').RegisterAPI} RegisterAPI
  */
 
+const PROTECTED_KEYS = ['name', 'id', 'type']
+
 /**
  * @experimental
  * IMPORTANT: This feature is being developed for use internally and is not in a public-facing production-ready state.
```
ahfuckit, line 28: then why can anybody see this?

PR #1687, opened 2026-02-02
chore: Monkey patch nrWrapper test

ellisong:
Fixed an e2e test related to monkey patching that caused our test suite to fail due to adding the new multi-agent warning code.
---

### Overview

Fixes the `should NOT warn if the monkey-patched function is nrWrapper` test in `monkey-patched.e2e.js`. This test needed to be updated to account for the multi-agent warning. [See this PR](https://github.com/newrelic/newrelic-browser-agent/pull/1686) for context.

### Testing

Make sure tests pass.

### DIFF
--- a/tests/specs/monkey-patched.e2e.js
+++ b/tests/specs/monkey-patched.e2e.js
@@ -25,6 +25,9 @@ describe('monkey-patched globals', () => {
     expect(logs[0][0].includes('New Relic Warning') && logs[0][0].includes('64')).toEqual(true) // 64 is the warning code for monkey-patched globals
     expect(logs[0][1]).toEqual('debug')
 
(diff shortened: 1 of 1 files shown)


tests/specs/monkey-patched.e2e.js
```diff
@@ -13,18 +13,4 @@ describe('monkey-patched globals', () => {
     expect(logs[1][0].includes('New Relic Warning') && logs[1][0].includes('64')).toEqual(true) // 64 is the warning code for monkey-patched globals
     expect(logs[1][1]).toEqual('function (){ \n        return origPerformanceNow.apply(performance, arguments);}')
   })
-
-  it('should NOT warn if the monkey-patched function is nrWrapper', async () => {
```
metal-messiah, line 17: Instead of removing this useful test, can you explain why this is failing with the new changes?

PR #1689, opened 2026-02-04
feat: Removing old SPA feature & related wrappings entirely

cwli24:
This concludes the complete removal of the old SPA feature that is replaced by Soft Navigations feature. Since agent version 1.307.0, the default "SPA" was already switched over, so no breaking changes from this version are expected.
---

### Overview

Removed a bunch of dead code & test files related to the old SPA feature (not to be confused with 'spa' build and the 'spa' RUM flag). Browser API wrappings that were only used by old SPA at this point are also cleaned up.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-224099
https://new-relic.atlassian.net/browse/NR-250803


src/features/spa/index.js
metal-messiah, line 1: Do you think we should treat this module the same way we did with PageAction module when that was removed in favor of the generic_events module? -- e.g. a redirection import?

cwli24: Mm I don't want to if that's what you're asking. The PA was really just the same feature renamed and it wasn't nearly as publicized as a deprecation (and there's still an init setting for it) whereas old SPA is a complete different feature removal & we've been telling people it's going to depr for years, so that's my argument.
What do you think? And exactly where would we need to keep it? I definitely don't want people to continue to reference "/spa" if it's going to actually be using "/soft_navigations"

metal-messiah: i think that makes enough sense to me.  Since the deprecation warning has been as long as it has, its likely a non-issue

metal-messiah (approved):

PR #1694, opened 2026-02-12
fix: Update measure API types

### LINKED ISSUE #1693: Type of `measure` method is missing `PerformanceMark` support.
Type of `measure` method is missing `PerformanceMark` support.

[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

[NOTE]: # ( Provide a brief overview of what the new feature is all about. )

Currently, the `measure` method types shows that the `start` and `end` properties should be a number.

In the [documentation](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/measure/#using-performancemark-arguments), though, it says that `PerformanceMark` (the return of `performance.mark`) should also be supported.

If we try using it, though, Typescript will complain that it isn't a number.

## Desired Behavior

[NOTE]: # ( Tell us how the new feature should work. Be specific. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

When using `newrelic.measure` in Typescript with `performance.mark`, there should be no error.

## Possible Solution

[NOTE]: # ( Not required. Suggest how to implement the addition or change. )

Update the types of `measure` to support `PerformanceMark`

metal-messiah:
Corrects an issue where the typings associated with the measure API did not accept PerformanceMark as a valid input for start or end values.
---

### Related Issue(s)
Fixes #1693 
[NR-523184](https://new-relic.atlassian.net/browse/NR-523184)

### Testing
Updated the typings test to accurately reflect the code and public docs.

[NR-523184]: https://new-relic.atlassian.net/browse/NR-523184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


metal-messiah:
> A couple more files that would require an update:
> 
> * src/loaders/api/register-api-types.js
> * src/interfaces/registered-entity.js
> 
> ❔ Maybe consider refactoring to define the typedefs once and reference 'em everywhere else

There's some quirks here with the way the types are generated that the IDE doesn't end up picking up the commentary from the types when they are not directly on the method.  Lets split that off into a new ticket

PR #1707, opened 2026-02-24
chore: Update browser-agent-2-query account reporting

ellisong:
Updates our browser-agent-2-query app to report to our new A2Q entities instead of a testing entity.
---

### Overview

Changing our app reporting from our test app to two separate A2Q apps: latest agent in US staging and released agent in US production. These changes are not customer facing and only affect internal reporting.

After approval and right before merging this PR, we need to change the secrets for A2Q license keys.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-525777

### Testing

- Tests pass.
- Latest agent should be on app ID 77606036 in US staging beacon.
- Released agent should be on app ID 1431909786 in US production beacon.

### DIFF
--- a/.github/actions/build-agent2query/templates/latest-config.js
+++ b/.github/actions/build-agent2query/templates/latest-config.js
@@ -34,4 +34,4 @@ NREUM.init = {
   }
 }
 NREUM.feature_flags = ['soft_nav', 'websockets']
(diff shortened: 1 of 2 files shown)


metal-messiah (approved):
Approved, with the following implemented along side/before merging:
- [x] update the PR to a "chore" so that a public release note is not generated
- [ ] Ensure the license key secrets are updated to the new accounts
  - [ ] A2Q_RELEASED_LICENSE_KEY
  - [ ] A2Q_LATEST_LICENSE_KEY
- [x] Chill request approved in #freeze-exceptions channel, or wait to merge until chill period ends

PR #1708, opened 2026-02-26
Add failing tests for pageUrl misattribution in FCP, CLS, INP

OzTK:
## Problem

PageViewTiming events (FCP, CLS, INP) are being attributed to the **current page URL** at harvest time instead of the **original page URL** at page load time. When soft navigations occur before the 30-second harvest interval, timing events get mis-attributed to the wrong page.

## Impact

- **Sample size drops** on original pages (e.g., slots, favourites)
- **Inflated metrics** on soft-nav destination pages
- **Inaccurate per-page performance data** across the application

This issue was observed in production after upgrading from v1.277.0 to v1.309.0, causing a 35% sample size drop and ~25% FCP increase on certain pages.

## Root Cause

The pageUrl fix that was applied to LCP in commit [04eb09fb](https://github.com/newrelic/newrelic-browser-agent/commit/04eb09fb27fb9d49dcb4a37cc503619217e9fafd) (Sept 18, 2025) was not applied to other vital metrics:

### ✅ Fixed (v1.298.0)
- **Largest Contentful Paint (LCP)** - includes `pageUrl` from `attribution.navigationEntry.name`

### ❌ Still Broken
- **First Contentful Paint (FCP)**
- **Cumulative Layout Shift (CLS)**
- **Interaction to Next Paint (INP)**

## This PR

This PR adds **failing unit tests** that demonstrate the issue. These tests verify that FCP, CLS, and INP should extract `pageUrl` from `attribution.navigationEntry.name` (which preserves the original URL from the browser's PerformanceNavigationTiming API) and include it in the attributes object.

### Tests Added:
- `tests/unit/common/vitals/first-contentful-paint.test.js`
- `tests/unit/common/vitals/cumulative-layout-shift.test.js`
- `tests/unit/common/vitals/interaction-to-next-paint.test.js`

### Expected Behavior:
All three tests will **FAIL** until the same fix applied to LCP is applied to these metrics:

```javascript
if (attribution.navigationEntry) attrs.pageUrl = cleanURL(attribution.navigationEntry.name)
```

## The Fix

Apply the same logic that exists in `src/common/vitals/largest-contentful-paint.js` to:
- `src/common/vitals/first-contentful-paint.js`
- `src/common/vitals/cumulative-layout-shift.js`
- `src/common/vitals/interaction-to-next-paint.js`

## Documentation

Full technical investigation available in the branch:
- [PAGEURL_MISATTRIBUTION_BUG.md](https://github.com/OzTK/newrelic-browser-agent/blob/fix/pageurl-misattribution-soft-nav/PAGEURL_MISATTRIBUTION_BUG.md)

## Related

- Issue #1566
- Original LCP fix: commit 04eb09fb27fb9d49dcb4a37cc503619217e9fafd

### DIFF
--- a/tests/unit/common/harvest/harvester.test.js
+++ b/tests/unit/common/harvest/harvester.test.js
@@ -144,3 +144,112 @@ describe('triggerHarvestFor', () => {
     expect(harvester.triggerHarvestFor(fakeAggregate, { })).toEqual({ payload: 'fakePayload', ranSend: true, endpointVersion: 2 })
   })
 })
+
+describe('FAILING: baseQueryString URL capture behavior', () => {
+  let mockSubmit
+  let agentRefWithBeacon
+
+  beforeEach(() => {
+    mockSubmit = jest.fn(() => ({ addEventListener: jest.fn() }))
+    agentRefWithBeacon = {
+      init: {
+        harvest: { interval: 1 },
+        ssl: true,
+        proxy: {}
+      },
+      info: {
+        errorBeacon: 'bam.nr-data.net',
+        applicationID: 'test-app-id',
+        licenseKey: 'test-license-key'
+      },
+      runtime: {
+        obfuscator: {
+          obfuscateString: jest.fn(url => url) // Just return the URL as-is for testing
+        },
+        session: {
+          state: {
+            value: 'test-session-id'
+          }
+        },
+        maxBytes: 1000000
+      },
+      ee: {
+        on: jest.fn()
+      }
+    }
+
+    // Mock the globalScope.location
+    delete global.window.location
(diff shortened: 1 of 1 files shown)


cwli24:
Thanks for bringing this to attention. We understand it causes confusion and mistrust when timings change the URLs they are reported under for SPA such that it's hard to piece them together and they appear miscounted for. We've standardized all timings to report under the same (initial) page URL in #1716.

This is still means continuous metrics such as INP & CLS will "bleed" into the data of the initial page even if it happened on a subsequent soft navigation "page", as soft nav timings are not yet a feature, but this at least normalize the querying of each set of timings from per visit.

PR #1709, opened 2026-02-26
chore: add manual dispatch option

metal-messiah:
Add a manual dispatch interface for the change tracking action, for better manual control and easier debugging
---

### DIFF
--- a/.github/workflows/manual-change-tracking.yml
+++ b/.github/workflows/manual-change-tracking.yml
@@ -41,7 +41,7 @@ on:
         options:
           - 'Basic'
           - 'Rolling'
-          - 'Rollback'
+          - 'Other'
           - 'Blue Green'
           - 'Canary'
           - 'Shadow'


.github/workflows/manual-change-tracking.yml
```diff
@@ -0,0 +1,160 @@
...
+      type:
+        description: 'Deployment Type'
+        required: true
+        type: choice
+        default: 'Basic'
+        options:
+          - 'Basic'
+          - 'Rolling'
+          - 'Rollback'
+          - 'Blue Green'
```
ptang-nr: Minor:  'Rollback' is not a valid value.  Looks like the accepted values are: 

https://docs.newrelic.com/docs/change-tracking/change-tracking-cli/
```
--deploymentType string   type of deployment, one of BASIC, BLUE_GREEN, CANARY, OTHER, ROLLING or SHADOW
```

metal-messiah:
```suggestion
          - 'Other'
          - 'Blue Green'
```

PR #1711, opened 2026-02-26
fix: Update RegExp to prevent un-minimized webpack build issue

cwli24:
Replaces the `$` character with its hexadecimal equivalent to prevent it being special replaced by webpack during build and breaking the agent.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-528802


src/common/deny-list/deny-list.js
metal-messiah, line 1: do we need to have any regression tests created to ensure that future us does not impose something that breaks the webpack build when un-minified?

PR #1712, opened 2026-03-02
feat: agentIdentifier removal

metal-messiah:
Simplify the agent by removing scoped references by agent identifier.  This allows all references to be stored in the agent reference itself, instead of a blending of agent references and scoped objects by lookup keys.
---

### Overview
This has been something we have wanting to do in the agent src code for many years.  This change was tasked to copilot, and then reviewed/modified by human.

### Related Issue(s)
A follow up to https://new-relic.atlassian.net/browse/NR-340432?atlOrigin=eyJpIjoiNjU5ODJlMTI3OTNiNDEzNmJmOTA0OTJlZmE4YjVhMDgiLCJwIjoiaiJ9

### DIFF
--- a/src/loaders/configure/configure.js
+++ b/src/loaders/configure/configure.js
@@ -50,20 +50,16 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
   if (!agent.runtime.configured) {
     Object.defineProperty(agent, 'beacons', {
       get () {
-        const beacons = [agent.info.beacon, agent.info.errorBeacon]
-        if (agent.init.proxy.assets) beacons.push(agent.init.proxy.assets)
-        if (agent.init.proxy.beacon) beacons.push(agent.init.proxy.beacon)
-        return beacons
+        return [agent.info.beacon, agent.info.errorBeacon, agent.init.proxy.assets, agent.init.proxy.beacon].filter(Boolean)
       }
     })
 
     Object.defineProperty(agent.runtime, 'denyList', {
       get () {
         // Compute the internal traffic list fresh each time to ensure beacons array is current
-        const currentInternalList = [...agent.beacons]
         return [
           ...(agent.init.ajax.deny_list || []),
-          ...(agent.init.ajax.block_internal ? currentInternalList : [])
+          ...(agent.init.ajax.block_internal ? agent.beacons : [])
         ]
       }
     })
--- a/src/common/config/runtime.js
+++ b/src/common/config/runtime.js
@@ -37,7 +37,7 @@ const RuntimeModel = {
     hiddenState.consented = value
   },
   customTransaction: undefined,
-  denyList: undefined,
+  denyList: [],
   disabled: false,
   /** @type {Map<string, {staged: boolean, priority: number}>} */
   drainRegistry: new Map(),


src/features/jserrors/aggregate/index.js
```diff
@@ -187,9 +187,10 @@ export class Aggregate extends AggregateBase {
     // still send EE events for other features such as above, but stop this one from aggregating internal data
     if (this.blocked) return
 
-    if (err.__newrelic?.[this.agentIdentifier]) {
-      params._interactionId = err.__newrelic[this.agentIdentifier].interactionId
-      params._interactionNodeId = err.__newrelic[this.agentIdentifier].interactionNodeId
+    const errNode = err.__newrelic?.[this.agentRef.agentIdentifier]
```
ptang-nr: How does `err.__newrelic` work/is this still used?  I can't seem to create the scenario to get `errNode` to be defined. 

(not necessarily a blocker but may be up for removal if it turns out to be dead code)

src/common/drain/drain.js
```diff
@@ -4,81 +4,78 @@
...
- * @param {string} agentIdentifier - A 16 character string uniquely identifying an agent.
+ * Ensures the agent has a drain registry initialized.
+ * @param {Object} agentRef - The agent reference object.
  */
-function curateRegistry (agentIdentifier) {
-  if (!agentIdentifier) throw new Error('agentIdentifier required')
-  if (!registry[agentIdentifier]) registry[agentIdentifier] = new Map()
+function ensureDrainRegistry (agentRef) {
+  if (!agentRef) throw new Error('agentRef required')
+  if (!agentRef.utils) agentRef.utils = {}
```
ptang-nr: Minor - consider moving the initiation of `utils = {}` to the agent constructor.  That way, we can be sure utils is always defined in various places.

src/features/jserrors/aggregate/index.js
metal-messiah: That is a scoped attribute on the error node itself when a websocket is being used and triggers an error in its event handlers.  I think it is still required to be used as defined

ptang-nr: Ah, true.  I thought `interactionId` and `interactionNodeId` dealt with SPA/SN and so I was trying to create a JS error while using `loader=spa`.  So for clarity, this block code is dealing with web sockets and I need to use a WebSocket to observe the code block executing?

src/common/drain/drain.js
metal-messiah: moved logic to runtime to avoid the need for another separate metadata object a9af4e2

src/loaders/configure/configure.js
```diff
@@ -44,7 +41,9 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
   const updatedInit = agent.init
   const internalTrafficList = [info.beacon, info.errorBeacon]
 
-  if (!alreadySetOnce.has(agent.agentIdentifier)) {
+  runtime.configured ??= false
```
cwli24: Not needed; already false

src/common/drain/drain.js
```diff
@@ -4,81 +4,77 @@
...
- * Agents without an identifier (as in the case of some tests) will be excluded from the registry.
- * @param {string} agentIdentifier - A 16 character string uniquely identifying an agent.
+ * Ensures the agent has a drain registry initialized.
+ * @param {Object} agentRef - The agent reference object.
  */
-function curateRegistry (agentIdentifier) {
-  if (!agentIdentifier) throw new Error('agentIdentifier required')
-  if (!registry[agentIdentifier]) registry[agentIdentifier] = new Map()
+function ensureDrainRegistry (agentRef) {
+  if (!agentRef) throw new Error('agentRef required')
```
cwli24: Can get rid of L44 here, under strong assumption that agentRef is valid and these internal func are provided proper arguments

src/common/util/console.js
```diff
@@ -17,7 +17,6 @@ export function warn (code, secondary) {
   if (typeof console.debug !== 'function') return
   console.debug(`New Relic Warning: https://github.com/newrelic/newrelic-browser-agent/blob/main/docs/warning-codes.md#${code}`, secondary)
   dispatchGlobalEvent({
-    agentIdentifier: null,
```
cwli24, line 20: Will only 1 main agent be dispatching global events now? None from other agents or sub-agent types?

src/features/jserrors/aggregate/index.js
cwli24: Well this was initially used for old SPA I think, and I just re-used it for WS. This seem like a missed remnant (`err.__newrelic[this.agentIdentifier].interactionId` isn't set anywhere anymore). These lines should be removed actually.

src/common/util/console.js
metal-messiah: yeah i dont think we need to bother with multiple agent debugging

src/loaders/configure/configure.js
```diff
@@ -41,28 +41,25 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
...
-    setTopLevelCallers(agent) // no need to set global APIs on newrelic obj more than once
-  }
+    agent.runtime.denyList = [
+      ...(updatedInit.ajax.deny_list || []),
+      ...(updatedInit.ajax.block_internal ? internalTrafficList : [])
+    ]
+    agent.runtime.ptid = agent.agentIdentifier
+    agent.runtime.loaderType = loaderType
 
-  runtime.denyList = [
```
cwli24, line 59: I faintly recall this was intentionally split out from the configured logic because `configure` can be called more than once, e.g. late init on the page, and that affects this ajax setting.

metal-messiah: Lets chat on this one at rodeo, I'm not fully following

metal-messiah: actually, nvm i do remember now.  I'll address that

cwli24 (approved):

PR #1714, opened 2026-03-03
feat: improve MFE timing logic

metal-messiah:
Improves the agent's ability to determine the correct MFE script calling the register API, which will help create more accurate timing information.
---

### Overview
In cases where MFE scripts are loaded/initialized via a generator script, such as is done in NR1, the MFE timings would be linked to the generator file and not the MFE file.  This change introduces better logic which grabs the file _closest_ to the agent script as the MFE script instead of the root file, which can be inaccurate in this case.

#### TL;DR
The reason one-vbp shows up as the root is that the module code is being executed as a "callback" of the VBP loader's generator. If this were "native" async code in a modern browser without transpilation, the browser might hide these implementation details or provide "Async Stack Traces" that show the original call. But because this is being run through a compatibility layer (the generator runner), the runner is visible as the primary caller.
The manifestation of this stack trace is a direct result of how generators (specifically transpiled async/await) work.

#### Description of the issue
In the stack trace mentioned in the slack thread, `one-vbp-543fd763.js` is acting as an Async Runner that is driving the execution of the nerdpack module.

1. Because many browsers/build environments don't support "native" async/await tools like Babel or TypeScript convert async functions into generators.
  - The Generator: The import() -- the logic in browser-application-preview nerdpack.
  - The Runner: A helper function (aka __awaiter or __generator) that calls .next() on the iterator.

2. Why the "wrong file" is the "root" of the stack
Stack traces only capture the current synchronous execution context.

When the platform (One VBP js) starts loading a nerdpack, it reaches a point where it has to wait (e.g., waiting for the network to download the chunk `browser-entity-preview--browser-application-preview~b21d8a93.js`). At that moment, the generator yields, the function pauses, and the stack is cleared.

Once the file downloads and is ready, the runner inside one-vbp is triggered by the Promise resolution.

The runner calls Generator.next(). This call re-enters the generator function. Because the runner itself is the function that initiats the specific iteration of the event loop, it sits at the bottom of the stack for everything that happens until the next yield or return.

### Related Issue(s)
See this thread for the context https://newrelic.slack.com/archives/C07GZKGJUJD/p1772557117212569

### Testing
A 5th case was added which replicates the generator stack pattern described above and linked in the slack thread.

### DIFF
--- a/src/common/util/script-tracker.js
+++ b/src/common/util/script-tracker.js
@@ -12,7 +12,8 @@ import { chrome, chromeEval, gecko } from './browser-stack-matchers'
  * @typedef {import('./register-api-types').RegisterAPITimings} RegisterAPITimings
  */
 
-let thisFile
+/** export for testing purposes */
+export let thisFile
 try {
   thisFile = extractUrlsFromStack(getDeepStackTrace()).at(0)
 } catch (err) {
@@ -66,7 +67,7 @@ function extractUrlsFromStack (stack) {
       urls.add(cleanURL(parts[2]))
     } else {
       // Fallback: match URLs using a generic .js pattern (non-greedy to handle ports and query params)
-      const fallbackMatch = line.match(/\(?((?:(?:https?|file):\/\/)?.*?\.js):\d+:\d+\)?/)
+      const fallbackMatch = line.match(/\(([^)]+\.js):\d+:\d+\)/) || line.match(/^\s+at\s+([^\s(]+\.js):\d+:\d+/)
       if (fallbackMatch && fallbackMatch[1]) {
         urls.add(cleanURL(fallbackMatch[1]))
       }
--- a/tests/unit/common/util/script-tracker.test.js
+++ b/tests/unit/common/util/script-tracker.test.js
@@ -139,8 +139,8 @@ describe('script-tracker', () => {
 
       // Stack trace references the external script
       mockStack = `Error
-    at findScriptTimings (internal:1:1)
-    at Object.register (internal:5:10)
+    at findScriptTimings (${scriptTrackerModule.thisFile}:1:1)
+    at Object.register (${scriptTrackerModule.thisFile}:5:10)
     at main (https://cdn.example.com/mfe-app.js:15:20)`
 
       const result = scriptTrackerModule.findScriptTimings()
@@ -175,8 +175,8 @@ describe('script-tracker', () => {
 
       // Stack references script tracked by observer
       mockStack = `Error
-    at findScriptTimings (internal:1:1)
-    at register (internal:10:5)
+    at findScriptTimings (${scriptTrackerModule.thisFile}:1:1)
+    at register (${scriptTrackerModule.thisFile}:10:5)
     at init (https://cdn.example.com/tracked-app.js:25:10)`
 
       // Import module to activate PerformanceObserver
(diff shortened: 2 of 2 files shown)


src/common/util/script-tracker.js
```diff
@@ -54,9 +61,15 @@ function extractUrlsFromStack (stack) {
 
   for (const line of lines) {
     // Try gecko format first, then chrome
-    const parts = line.match(gecko) || line.match(chrome)
+    const parts = line.match(gecko) || line.match(chrome) || line.match(chromeEval)
     if (parts && parts[2]) {
       urls.add(cleanURL(parts[2]))
+    } else {
+      // Fallback: match URLs using a generic .js pattern
+      const fallbackMatch = line.match(/\(?((?:https?:\/\/)?[^):\s]+\.js):\d+:\d+\)?/)
```
NandaKishorJeripothula: We might lose support for `http` as per this regex. Is it acceptable?

metal-messiah: Thanks for catching this.  The regex has been updated to [support https, http, and file protocols](https://github.com/newrelic/newrelic-browser-agent/pull/1714/commits/cddf0166d7edc5e59301c34e5e1b18a0c8696a65)

src/common/util/script-tracker.js
```diff
@@ -110,8 +123,11 @@ export function findScriptTimings () {
   const navUrl = globalScope.performance?.getEntriesByType('navigation')?.find(entry => entry.initiatorType === 'navigation')?.name || ''
 
   try {
-    const mfeScriptUrl = extractUrlsFromStack(stack).at(-1) // array of URLs from the stack of the register API caller, the MFE script should be at the bottom
+    const urls = extractUrlsFromStack(stack)
+    /** if there is exactly one url, this means the MFE script is running in the same file as the agent.  Otherwise, lets strip away the known agent file from any other file lines */
+    const mfeScriptUrl = (urls.length > 1 ? urls.filter(line => (!thisFile.endsWith(line) && !line.endsWith(thisFile))) : urls).at(0)
```
cwli24, line 129: I'm unclear how this alone addresses the problem from the description. I.e., if given a stack with multiple scripts (say, scriptA, scriptB, scriptC, agentScript), how this would return scriptC (the closest to the agent) as described.

cwli24: Wait, I see extractUrlsFromStack returns array with the top of the stack being first, and bottom being last. So this makes sense. Nvm.

tests/unit/common/util/script-tracker.test.js
```diff
@@ -742,6 +742,54 @@ init@https://cdn.example.com/gecko-app.js:20:10`
...
+      mockStack = `Error
+    at Object.register (internal:18:84024)
+    at register (internal:18:84952)
+    at mfe5-module (https://staging-one.nr-assets.net/nerdpacks/browser-entity-preview~b21d8a93.js:1:3275)
+    at Generator.<anonymous> (https://staging-one.nr-assets.net/platform/one-vbp-543fd763.js:1:32670)
+    at Generator.next (https://staging-one.nr-assets.net/platform/one-vbp-543fd763.js:1:33653)`
+
+      const result = scriptTrackerModule.findScriptTimings()
+
+      // Should select MFE chunk (middle), not generator runtime (bottom)
```
cwli24, line 786: I don't think `internal:_:___` result in anything on the url array, so you might need to insert another URL above mfe5-module to prove your point, and ensure it's not just picking off the top of the constructed URL stack now.

metal-messiah: [5f28900](https://github.com/newrelic/newrelic-browser-agent/commit/5f28900894ed38d7a22777a7f3f837b82fd4f7e3)

PR #1716, opened 2026-03-06
feat: Attribute original page URL to all PageViewTiming nodes

cwli24:
Re-attribute `pageUrl` for all timings, expanding upon previous similar work on LCP. `PageViewTiming` events will report the original hard navigation URL for all node types, even if the current URL at record or harvest time have changed (for SPA).
---

### Overview

Timings audited:

-    FP - not yet reported; once per page view, so should attribute to original page
-    FCP - not yet reported; once per page view, so should attribute to original page
-    LCP - already reported; once per page view, so should attribute to original page
-    FI/FID - not yet reported; once per page view, so should attribute to original page
-    INP - not yet reported; continuous over page life, but soft nav reporting not on, so attribute to original page instead of (SPA) url at capture
-    CLS - not yet reported; continuous over page life, but soft nav reporting not on, so attribute to original page instead of (SPA) url at capture
-    load - not yet reported; once per page view, so should attribute to original page
-    pagehide - not yet reported; every vischange->hidden over page life, but lacks guidance, so attribute to original page
-    unload - not yet reported; once per page view (at end) and should “pair” w/ load, so should attribute to original page as well

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-528815

### Testing

Revised + added

### DIFF
--- a/src/common/util/script-tracker.js
+++ b/src/common/util/script-tracker.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { globalScope, getNavigationEntry } from '../constants/runtime'
+import { globalScope } from '../constants/runtime'
 import { now } from '../timing/now'
 import { cleanURL } from '../url/clean-url'
 import { chrome, gecko } from './browser-stack-matchers'
@@ -107,7 +107,7 @@ export function findScriptTimings () {
   const timings = { registeredAt: now(), reportedAt: undefined, fetchStart: 0, fetchEnd: 0, asset: undefined, type: 'unknown' }
   const stack = getDeepStackTrace()
   if (!stack) return timings
-  const navUrl = getNavigationEntry()?.name || ''
+  const navUrl = globalScope.performance?.getEntriesByType('navigation')?.[0]?.name || ''
 
   try {
     const mfeScriptUrl = extractUrlsFromStack(stack).at(-1) // array of URLs from the stack of the register API caller, the MFE script should be at the bottom
--- a/tests/unit/common/util/script-tracker.test.js
+++ b/tests/unit/common/util/script-tracker.test.js
@@ -108,6 +108,9 @@ describe('script-tracker', () => {
       mockNavigationEntry = { name: 'https://example.com/page.html' }
 
       global.performance.getEntriesByType = jest.fn((type) => {
+        if (type === 'navigation') {
+          return [mockNavigationEntry]
+        }
         if (type === 'resource') {
           return []
         }


src/common/util/script-tracker.js
```diff
@@ -107,7 +107,7 @@ export function findScriptTimings () {
   const timings = { registeredAt: now(), reportedAt: undefined, fetchStart: 0, fetchEnd: 0, asset: undefined, type: 'unknown' }
   const stack = getDeepStackTrace()
   if (!stack) return timings
-  const navUrl = globalScope.performance?.getEntriesByType('navigation')?.find(entry => entry.initiatorType === 'navigation')?.name || ''
```
ptang-nr, line 124: Originally, this was looking for any matching `navigation` entries, and had no ties to the value of `responseStart`.  Would there be any scenario where it's valid to have responseStart === 0 and we'd still want the url?

cwli24: Good question. The `PerformanceResourceTiming` (of which the `PerformanceNavigationTiming` is based on) `responseStart` property notes the following 3 cases wherein the value can be `0`:

1. resource is a cancelled request
2. resource is cross-origin without Timing-Allow-Origin response header
3. resource was instantaneous retrieved from a cache

Addressing these cases:

1. If the hard navigation network request to our page was cancelled, we would not be executing this code as the page would've failed to load; this case is moot.
2. This page cannot be a cross-origin request against itself. E.g., if I load example.com, the navigation entry I get will be example.com. It can never be a third party like anotherexample.com, because that makes no sense.
3. This is the most possible, if at all. Service worker or browser cache can serve the initial document pretty fast. However, even on our blank vanilla test page (`instrumented.html`) with caching on, the lowest `responseStart` value I can get after many reloads is ~11ms, which is still `> 0`.

---So based on the above points, the practical answer is that there's no documented scenario wherein it's possible. However, if you want to be absolutely certain in theory, we could change this back to `globalScope.performance?.getEntriesByType('navigation')?.[0]?.name`

cwli24: Do you want that change?

src/features/page_view_timing/aggregate/index.js
```diff
@@ -71,6 +72,8 @@ export class Aggregate extends AggregateBase {
 
   addTiming (name, value, attrs) {
     attrs = attrs || {}
+    attrs.pageUrl = cleanURL(getNavigationEntry()?.name || initialLocation)
```
ptang-nr, line 75: I was trying to compare the [LCP fix](https://github.com/newrelic/newrelic-browser-agent/blob/04eb09fb/src/common/vitals/largest-contentful-paint.js#L30) and this one.

- does this fix in PVT make the LCP one redundant?
- it [sounds](https://github.com/GoogleChrome/web-vitals/issues/275#issuecomment-1310836105) like(?) web-vitals special-cased bfcache restores.  Similar to the other Q earlier - are we tying `responseStart` logic to URLs for every timing because timings + URLs from cases like bfcache restores are not valid?

cwli24: Oops, I was supposed to remove the pageUrl from LCP, but looks like I missed that.

There's no timing being reported when the page is re-served from the BFC right now. And even if there was, this would still correctly retrieve the original navigation entry.

cwli24: Pushed removal

src/common/util/script-tracker.js
ptang-nr: @cwli24 - hey, I started taking closer look at `responseStart > 0` logic because of https://new-relic.atlassian.net/browse/NR-531816, which seems to indicate that we could very well be seeing `responseStart = 0` happening in the wild (mostly in Safari, but not all.  I'm not entirely sure why).  I think it may be prudent to consider where we're applying the `responseStart > 0` check.

ptang-nr: tl;dr - yes, I don't think we should check for `responseStart > 0` when dealing with URLs.  If we think we need to toss out invalid entries, that may be something to look into (?) but that's not* what we're doing here.

src/features/page_view_timing/aggregate/index.js
ptang-nr: Overall, I don't follow why we need to check for `responseStart > 0` when dealing with URLs.  My reasoning is - there is some evidence we're seeing `responseStart = 0` in the wild that is causing a diff issue for ST.  If `responseStart = 0` can happen, then this check will push us to return the wrong URL for ~~LCP~~ INP(?) and CLS.  

Additionally, CWV lib should already be filtering out invalid entries, so in theory, we should just grab the URL from the entry and return it.

src/common/util/script-tracker.js
cwli24: That issue does not allude to anything regarding responseStart specifically.

I updated this to match closer to previous.

src/features/page_view_timing/aggregate/index.js
cwli24: This just keeps it DRY, and as you said, is a non issue if CWV has already vetted the timings being processed. The function is called getNavigationEntry so it is just getting the entry; it sounds like your concern is actually with the internals of the function itself. Also ST captures all resource entries, not just navigation entry, where some of the values can be 0s.

ptang-nr: Ugh, I'd been looking at the "standard build" logic in CWV, and that's why I wasn't following the "extra" `responseStart > 0` check 🤦 🫠

PR #1720, opened 2026-03-12
feat: Automatically Detect MFE Errors

metal-messiah:
Add support for automatic Micro Frontend (MFE) attribution of `JavaScriptError` events. Previously, `JavaScriptError` events were only ever captured globally by the container browser application. With this change, the agent cross-compares the stack trace of the thrown Error instance to the call stack signature of the MFE module that registered with the `register` API.  If the file that is responsible for registering with the browser agent is found in the stack trace of the thrown Error instance, that event will be attributed to the MFE instead of the container.

---

### Overview
Split from #1706 

Line by line in the trace, If any file path in the stack trace of the thrown Error instance contains a match to the file path captured during use of the .register API, the event is captured as belonging to that MFE.

### Testing


src/features/jserrors/aggregate/index.js
```diff
@@ -154,7 +251,7 @@ export class Aggregate extends AggregateBase {
      * the canonical stack trace excludes items like the column number increasing the hit-rate of different errors potentially
      * bucketing and ultimately resulting in the loss of data in NR1.
      */
-    var bucketHash = stringHashCode(`${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}_${params.hasReplay ? 1 : 0}_${target?.id || 'container'}`)
+    var bucketHash = stringHashCode(`${stackInfo.name}_${stackInfo.message}_${stackInfo.stackString}_${params.hasReplay ? 1 : 0}_${target?.id || 'container'}_${target?.instance || ''}`)
```
ellisong, line 254: how do we stringify `target?.instance` ? what variable type is `instance`?

metal-messiah: `target.instance` is a 8-digit hex string that uniquely identifies the register instance -- https://github.com/newrelic/newrelic-browser-agent/blob/main/src/loaders/api/register.js#L62

ellisong (approved):
LGTM if tests pass

PR #1721, opened 2026-03-12
feat: Automatically Detect MFE Logs

metal-messiah:
Add support for automatic Micro Frontend (MFE) attribution of `Log` events. Previously, `Log` events were only ever  _automatically_ captured globally by the container browser application. With this change, the agent cross-compares the call stack signature of the given log method (such as console.log, console.info, etc.) to the call stack signature of the MFE module that registered with the `register` API.  If the file that is responsible for registering with the browser agent is found in the call stack signature of the log method being triggered, that event will be attributed to the MFE instead of the container.

---

### Overview
Split from #1706 

Line by line in the trace, If any file path in the call stack of the log method executor contains a match to the file path captured during use of the .register API, the event is captured as belonging to that MFE.

### Testing


src/common/wrap/wrap-function.js
```diff
@@ -78,11 +80,17 @@ export function createWrapperWithEmitter (emitter, always) {
...
       let thrownError
+      let targets
 
       try {
         originalThis = this
         args = [...arguments]
 
+        // certain wrappers can inform the function wrapper to evaluate the stack of the executed wrapped function to find targets of the execution
+        // (e.g. wrap-logger can inform this method to find try to find the MFE source of a console.log)
+        targets = !!evaluateStack && !!agentRef?.init.api.allow_registered_children ? findTargetsFromStackTrace(agentRef) : [undefined] // undefined target always maps to the container agent
```
ptang-nr: Consider moving the fallback logic into `findTargetsFromStackTrace`.  It seems current usages align on this behavior (see also https://github.com/newrelic/newrelic-browser-agent/blob/main/src/common/wrap/wrap-xhr.js#L58-L60)

src/features/logging/aggregate/index.js
```diff
@@ -104,14 +98,19 @@ export class Aggregate extends AggregateBase {
...
+        attributes,
+        level
+      )
+
+      if (this.events.add(log)) this.reportSupportabilityMetric(LOGGING_EVENT + (autoCaptured ? 'Auto' : 'API') + '/Added')
+    }
 
-    if (this.events.add(log)) this.reportSupportabilityMetric(LOGGING_EVENT + (autoCaptured ? 'Auto' : 'API') + '/Added')
+    addEvent({ ...attributes, ...getVersion2Attributes(target, this) })
+    if (shouldDuplicate(target, this.agentRef)) addEvent({ ...attributes, ...getVersion2DuplicationAttributes(target, this) })
```
ptang-nr: What happens when `shouldDuplicate` evaluates to true and harvest endpoint != 2?  

(I'm trying to evaluate whether it's possible to send duplicate events to the same entity if `getVersion2Attributes` and `getVersion2DuplicationAttributes` both return `{}`)

ptang-nr (dismissed):
LGTM (always welcome a 2nd set of eyes)

src/features/logging/aggregate/index.js
metal-messiah: Thats a good callout.  I think for safety we should probably make sure that should duplicate is checking the endpoint version too.  maybe there's some consolidation that can be made around this logic in v2.js

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1721/commits/80e6264b7b7410ef4b3b3a93b1165bef23cf4ed0

src/common/wrap/wrap-function.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1721/commits/80e6264b7b7410ef4b3b3a93b1165bef23cf4ed0

src/common/util/v2.js
```diff
@@ -63,22 +63,36 @@ export function getVersion2Attributes (target, aggregateInstance) {
...
 }
 
+/**
+ * Finds the registered targets from the stack trace for a given agent reference.
+ * @param {*} agentRef The agent reference to use for finding targets.
+ * @returns {Array} An array of targets found from the stack trace. If no targets are found or allowed, returns an array with undefined.
+ */
 export function findTargetsFromStackTrace (agentRef) {
-  if (!agentRef?.init.api.allow_registered_children) return []
+  if (!agentRef?.init.api.allow_registered_children) return [undefined]
```
ptang-nr, line 95: (discussed on Slack - `[undefined]` is unintuitive, should revisit to see if we can represent the container differently)

ptang-nr (approved):

src/common/util/v2.js
metal-messiah: https://new-relic.atlassian.net/browse/NR-539907

PR #1723, opened 2026-03-13
feat: Automatically Detect MFE User Actions

metal-messiah:
Add support for automatic Micro Frontend (MFE) attribution of UserAction events via a `data-nr-mfe-id` data attribute. Previously, UserAction events were only ever captured globally by the container browser application. With this change, the agent now performs a parental chain lookup for elements associated with the action, identified by the `data-nr-mfe-id` attribute.

To associate UserAction events with a registered MFE, add a `data-nr-mfe-id` at the root DOM node controlled by your MFE package that matches the ID supplied to the `register` API.

---

### Overview
Split from #1706 

 If any element in the DOM chain contains a data-nr-mfe-id that matches an ID provided via the .register API, the event is captured as belonging to that MFE.

### Testing


src/loaders/api/register.js
```diff
@@ -61,6 +61,34 @@ function register (agentRef, target) {
...
+      }
+    }
+  })
+
+  // Only define attributes getter if it doesn't already exist
+  if (!Object.prototype.hasOwnProperty.call(target, 'attributes')) {
+    Object.defineProperty(target, 'attributes', {
+      get () {
+        return {
+          ...attrs,
```
ellisong: is it intentional for `...attrs` to be here rather than underneath line 87?

metal-messiah: this was a merge conflict remnant, removed in https://github.com/newrelic/newrelic-browser-agent/pull/1723/commits/c43b1b0c811ee355a476c9eaba737d40e6f8e141

ellisong (approved):
LGTM

PR #1724, opened 2026-03-13
feat: Automatically Detect MFE Page View Timings

metal-messiah:
Add support for automatic Micro Frontend (MFE) attribution of certain PageViewTiming events (LCP, INP, CLS) via a `data-nr-mfe-id` data attribute. Previously, these core web vitals were only ever captured globally by the container browser application. With this change, the agent now performs a parental chain lookup for elements associated with certain timing events, identified by the `data-nr-mfe-id` attribute.

The following PageViewTiming events can be observed this way:
- Largest contentful paint (LCP)
- Interaction to next paint (INP)
- Cumulative layout shift (CLS)

To associate these timings with a registered MFE, add a `data-nr-mfe-id` at the root DOM node affected by your MFE package that matches the ID supplied to the `register` API.

---

### Overview
Split from #1706 

If a parent element contains a data-nr-mfe-id that matches an ID provided via the .register API, the event is reassigned to that MFE.

### Testing
The vite-react-mfe test build is a great example app that implements the register API and supplies the correct set ups to replicate the behaviors across the auto-MFE PRs.  See the new e2e and jest tests around this new behavior and play around with that vite build to see this in action.


src/common/vitals/cumulative-layout-shift.js
```diff
@@ -18,6 +18,6 @@ if (isBrowserScope) {
       largestShiftValue: attribution.largestShiftValue,
       loadState: attribution.loadState
     }
-    cumulativeLayoutShift.update({ value, attrs })
+    cumulativeLayoutShift.update({ value, attrs, element: attribution.largestShiftTarget })
```
ptang-nr: I am not familiar with `largestShiftTarget`, so took a peek at the CWV [docs](https://github.com/googlechrome/web-vitals#CLSAttribution).  Looks like it represents 

> the first element (in document order) that shifted when the single largest layout shift that contributed to the page's CLS score occurred 

I think there's a nuance there that's not obvious and I'm worried about documentation. Should we call this out in the PR description and/or docs?

src/common/vitals/interaction-to-next-paint.js
```diff
@@ -25,6 +25,6 @@ if (isBrowserScope) {
       presentationDelay: attribution.presentationDelay,
       loadState: attribution.loadState
     }
-    interactionToNextPaint.update({ value, attrs })
+    interactionToNextPaint.update({ value, attrs, element: attribution.interactionTarget })
```
ptang-nr: Looks like `interactionTarget` is a string ([docs](https://github.com/googlechrome/web-vitals#inpattribution)).  There's a type mismatch since element is expected to be an `HTMLElement` in `analyzeElemPath`.

Same situation for cls's largestShiftTarget.

src/loaders/api/register.js
```diff
@@ -61,6 +61,18 @@ function register (agentRef, target) {
...
+      return {
+        'source.id': target.id,
+        'source.name': target.name,
+        'source.type': target.type,
+        'parent.type': target.parent?.type || V2_TYPES.BA,
+        'parent.id': target.parent?.id,
+        ...attrs // this is abstracted out to a getter so that the other attributes cant be overridden
+      }
+    }
+  })
```
ptang-nr: Possible dupe of lines 78-91?

src/common/vitals/interaction-to-next-paint.js
ptang-nr: Depending on where we end up aligning the type (for example, if `VitalMetric`'s `update()` expects element to be of type `Element`, would be good to add an integration test to "lock it in".

src/loaders/api/register.js
metal-messiah: i think some funky stuff happened from a merge conflict that i missed

metal-messiah: (including using attributes in the vitals that are the string and not the node) -- i'll get it sorted and re-test the PR

metal-messiah: @ptang-nr these should hopefully be addressed in https://github.com/newrelic/newrelic-browser-agent/pull/1724/commits/683a4140a16c595aa4c1a1cf86cf0f7fba9f32de can you take a look there?  Thanks

src/common/vitals/interaction-to-next-paint.js
```diff
@@ -25,6 +25,6 @@ if (isBrowserScope) {
       presentationDelay: attribution.presentationDelay,
       loadState: attribution.loadState
     }
-    interactionToNextPaint.update({ value, attrs })
+    interactionToNextPaint.update({ value, attrs, element: attribution.interactionTargetElement })
```
ptang-nr, line 28: Just a note for future review if/when we bump CWV version - looks like `interactionTargetElement` was removed in v5 by default.  https://github.com/GoogleChrome/web-vitals/blob/8e3554212e323a89d7a348c219dc86d0a570edbb/docs/upgrading-to-v5.md?plain=1#L30

ptang-nr (approved):

PR #1734, opened 2026-03-27
fix: Gql operation type detection logic for Ajax request

kamleshsharma:

Since the regex above expects the operation (query|mutation|subscription) at the start of the query string. This is not true in every case, especially for a GET request, where the client usually sends the query hash rather than the full query string. So, adding a capability where the application can pass the operationType explicitly

---

### Overview

Consider the below fetch all to a GraphQL server

`
fetch("https://www.test.com/graphql", {
  "headers": {
    "accept": "*/*",
    "accept-language": "en-IN,en-GB;q=0.9,en-US;q=0.8,en;q=0.7",
    "content-type": "application/json"
  },
  "referrer": "https://www.test.com/",
  "body": "{\"operationName\":\"validateUserEntitlement\",\"operation\":\"query\",\"variables\":{\"entityId\":\"140981\"},\"query\":\"\nfragment TestWidget on Test {\n  status\n}\n\nquery TestData() {\n  testWidget() {\n    ...TestWidget\n  }\n}\n",
  "method": "POST"
});
`

Here, the value for `contents.query` is 
`
fragment TestWidget on Test {
  status
}

query TestData() {
  testWidget() {
    ...TestWidget
  }
}
`

Now, if we try to do a regex check on the above query string, it will return undefined

`
const matches = contents.query.trim().match(/^(query|mutation|subscription)\s?(\w*)/)
const operationType = matches?.[1] // this will be undefined
`

### Testing

Make a fetch call similar to the above one from the Newrelic browser client to a GraphQL server, and check if the AjaxRequest is capturing the operation name and operationType in the dashboard or NRQL

### DIFF
--- a/src/features/ajax/aggregate/gql.js
+++ b/src/features/ajax/aggregate/gql.js
@@ -39,7 +39,13 @@ function parseSingleGQL (contents) {
 
   /** parses gql query string and returns [fullmatch, type match, name match] */
   const matches = contents.query.trim().match(/^(query|mutation|subscription)\s?(\w*)/)
-  const operationType = matches?.[1]
+  /** 
+   * Since the regex above expects the operation (query|mutation|subscription) at the start of the query string.
+   * This is not true in every case, especially in the case of a GET request, 
+   * where the client usually sends the query hash, not the full query string.
+   * So, adding a capability where the application can pass the operationType explicitly
+   * **/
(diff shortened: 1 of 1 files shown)


ptang-nr:
Hi @kamleshsharma , thank you for your submission.  This has been added as a Feature Request https://new-relic.atlassian.net/browse/NR-589225

PR #1740, opened 2026-04-03
chore: apply permissions to workflows

metal-messiah:
Set permission in workflows to adhere to code scanning requirements
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-535213

### DIFF
--- a/.github/workflows/delete-experiment.yml
+++ b/.github/workflows/delete-experiment.yml
@@ -16,6 +16,10 @@ on:
           - staging
           - standalone
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
   delete-experiment-on-s3:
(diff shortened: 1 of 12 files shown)


cwli24 (approved):

.github/workflows/delete-experiment.yml
```diff
@@ -16,6 +16,10 @@ on:
           - staging
           - standalone
 
+permissions:
+  contents: read
+  id-token: write
```
ptang-nr, line 21: @metal-messiah - what does `id-token` do here?

metal-messiah: this allows us to transition to OIDC later for AWS if we want to

PR #1752, opened 2026-04-10
feat: MFE iframe support - Beta Access

metal-messiah:
Adds an interface for capturing MFE data encapsulated in an iFrame.  This is experimental, is subject to change or total removal and should not be used in a production environment until GA status has been reached
---

### Overview
[Please see this HLD diagram which visualizes the data flow ](https://whimsical.com/nrxd/iframe-3t7ZxDZgDo1oLsJ4dDGqHp)


metal-messiah:
@copilot review this PR

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+   * @param {boolean} [bypassRegistration=false] - Whether to bypass waiting for registration
+   * @param {boolean} [needsResponse=false] - Whether to wait for a response from the parent
+   * @returns {Promise<void>}
+   */
+  async #postMessageToParent (type, data, bypassRegistration = false, needsResponse = false) {
+    if (this.blocked) return
+
+    const timestamp = now()
+    try {
+      await (bypassRegistration ? Promise.resolve() : this.#registrationPromise)
```
cwli24, line 296: There's no guarantee here that `this.#registrationPromise` actually succeeded; it could've failed and this would continue to send (e.g. vitals) after waiting for that promise to finish.

Since there's already a try-catch here, you can simply make `#register` rethrow its error caught to make the promise reject:
```js
warn(73, err)
this.blocked = true
throw err
```
That rejection would then cause this await here to exit to the empty catch, which is what you'd want.

(Doing this, you'd also want to add 
```js
this.#registrationPromise.catch(() => {})  // prevent unhandled rejection warning
```
)

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+  async #register (opts) {
+    try {
+      const response = await this.#postMethodToAgent(REGISTER, [opts])
+      if (response.metadata) Object.assign(this.metadata, response.metadata)
+
+      const timings = findScriptTimings()
+      // Send initial timing values
+      for (const [key, value] of Object.entries(timings)) {
+        if (key !== 'correlation') {
+          this.#postTimingToAgent(key, value)
```
cwli24: This and the next #postTimingToAgent() below on L121 ends up awaiting `this.#registrationPromise`. But that promise is the same promise that is `return reponse` on L126. So in essence, these calls in registration is cyclically waiting for the registration itself.

Now it does work because they will just go on and become microtasks rather than block this promise from resolving, which in turn would trigger the continuation of those microtasks. But introduces a race between them sending initial/update timings to parent. They should probably have the `bypassRegistration` option so that they are sent right away.

src/loaders/configure/configure.js
```diff
@@ -81,6 +82,13 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
       data: agent.config
     })
 
+    // Set up iframe postMessage listener for registered entities
+    if (agent.init.api.register.allow_iframe_bridge) {
+      globalScope.addEventListener('message', async (event) => {
```
cwli24: unnecessary `async` -- should be removed

src/loaders/configure/iframe-message-handler.js
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  output.entity = entity
+  const methodFn = entity[method]
+  if (typeof methodFn !== 'function') {
+    warn(35, method)
+    return output
+  }
+
+  output.result = await methodFn.apply(entity, args || [])
```
cwli24, line 178: At minimum, you should ensure the method is ok to run via `isSafeProperty(entity, method)`. (An explicit whitelist of allowed methods would be even better but isn't quite necessary here.) Wouldn't want prototype funcs, or worse the constructor, to be hijacked.

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+      lcp: { value: null },
+      fcp: { value: null },
+      inp: { value: null }
+    }
+  }
+
+  /** @private Map to store pending promise resolvers keyed by message ID */
+  #pendingMessages = new Map()
+  /** @private Unique ID for this iframe interface instance to correlate messages */
+  #iframeInterfaceId = Math.random()
```
cwli24: there's a `generateUuid()` for this that would be better suited & reduces collision chance + possibly falsey 0 value.

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+   * @param {string} type - The message type to send
+   * @param {object} data - The message payload to send
+   * @param {boolean} [bypassRegistration=false] - Whether to bypass waiting for registration
+   * @param {boolean} [needsResponse=false] - Whether to wait for a response from the parent
+   * @returns {Promise<void>}
+   */
+  async #postMessageToParent (type, data, bypassRegistration = false, needsResponse = false) {
+    if (this.blocked) return
+
+    const timestamp = now()
```
cwli24, line 294: it seems like this timestamp is never used anywhere in iframe-message-handler; can be removed

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+      }
+
+      // Validate iframeInterfaceId to prevent spoofing (message must be for this instance)
+      if (event.data.iframeInterfaceId !== this.#iframeInterfaceId) {
+        warn(75)
+        return
+      }
+
+      // Always sync metadata if provided to keep instance up to date
+      // if (event.data.metadata) Object.assign(this.metadata, event.data.metadata)
```
cwli24: what's going on here?

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+   * {@link https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/noticeerror/}
+   * @param {Error|string} error Provide a meaningful error message that you can use when analyzing data on browser's JavaScript errors page.
+   * @param {object} [customAttributes] An object containing name/value pairs representing custom attributes.
+   */
+  noticeError (error, customAttributes) {
+    this.#postMethodToAgent('noticeError', [castError(error), customAttributes])
+  }
+
+  /**
+   * Adds a user-defined identifier string to subsequent events on the page for the registered taret.
```
cwli24:
```suggestion
   * Adds a user-defined identifier string to subsequent events on the page for the registered target.
```

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+   * @private
+   */
+  #setupResponseListener () {
+    globalScope.addEventListener('message', (event) => {
+      if (this.blocked) return
+      // Validate message structure
+      if (event.data?.type !== IFRAME_API_RESPONSE) return
+
+      // Validate origin if we have specific allowed origins
+      if (!this.#isAllowedOrigin(event.origin)) {
```
cwli24: It's worth calling out that in the case of `#parentOrigin = '*'`, this is very permissive. One question is if parentOrigin cannot be determined (would fallback to '*'), should registration even be allowed?

And the other thing to address is the promise is not explicitly rejected here in the case that this iframe gets an `IFRAME_API_RESPONSE` message directed at it from someone else who is _not_ the parentOrigin. Instead, it's relying on it to timeout in order to reject. I think you should reject the promise right away here.

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+  /**
+   *
+   * @param {RegisterAPIConstructor} opts The options for setting up the registered iframe entity.
+   */
+  constructor (opts) {
+    warn(54)
+    // Store original descriptor for postMessage (before any function merging)
+    this.metadata.target = this.#targetDescriptor = opts
+
+    if (!isIFrameWindow(window)) {
```
cwli24:
```suggestion
    if (!isBrowserScope || !isIFrameWindow(globalScope)) {
```
guards against execution in Node env (i.e. SSR) and workers.

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+    this.#resourceObserver.observe({ type: 'resource', buffered: true })
+  }
+
+  /**
+   * Reports a single resource timing entry as an AJAX event, if it looks like a network request.
+   * @private
+   * @param {PerformanceResourceTiming} resource
+   */
+  #processResourceEntry (resource) {
+    if (!(resource.initiatorType in AJAX_INITIATOR_TYPES) || resource.responseStatus === 0) return
```
cwli24: `resource.responseStatus === 0` filters out any cross-origin request (w/o `Timing-Allow-Origin` header), even if a customer is calling their own API at a different domain in an MFE. Are you sure?

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+  /**
+   * Registers this entity with the parent agent, then seeds initial timings and FCP.
+   * @private
+   * @param {RegisterAPIConstructor} opts
+   * @returns {Promise<void>}
+   */
+  async #register (opts) {
+    try {
+      const response = await this.#postMethodToAgent(REGISTER, [opts])
+      if (response.metadata) Object.assign(this.metadata, response.metadata)
```
cwli24, line 153: In the event that `register.js` yields `target.blocked = true` on the parent side, you do get `response.metadata.target.blocked` here. The problem though is that isn't the same as `this.blocked`.

You can add `if (this.metadata.target.blocked) this.blocked = true` after this line. But a) you end up with two `blocked` flags for this entity, and b) it's not future proof -- if you uncomment the `Object.assign` on L219, it's back to not inheriting the parent's blocked state for this entity.

It may be better to combine `this.blocked` and `this.metadata.target.blocked`somehow into one flag, if `this.metadata.target` is always defined (even if empty object) in the lifecycle.

src/loaders/configure/iframe-message-handler.js
```diff
@@ -0,0 +1,218 @@
...
+ * Handles timing property updates from iframes. Mutates the timings object on the entity's metadata with the new value for the specified property.
+ * @param {MessageEvent} event - The message event containing the timing update
+ * @param {Object} agent - The browser agent instance
+ * @returns
+ */
+function handleTimingUpdate (event, agent) {
+  const entity = getValidEntity(event, agent)
+  if (!entity) return
+
+  if (entity.metadata?.timings) {
```
cwli24:
```suggestion
  if (entity.metadata.timings) {
```
redundant optional chaining: if `entity` is returned defined, `entity.metadata` will always be defined via `getValidEntity`.

Same thing with L64 in `handleVitalsUpdate` below.

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+
+      const timings = findScriptTimings()
+      // Send initial timing values
+      for (const [key, value] of Object.entries(timings)) {
```
cwli24: Minor improvement: skip falsy `value` in this loop. `findScriptTimings` return a lot of `reportedAt: undefined`, `fetchStart: 0`, etc. initially, which just overwrites the `entity.metadata.timings` in _iframe-message-handler_ with the same undefined/0 values. Not really harmful but just unnecessary postMessage calls.

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+
+  /**
+   * Registers this entity with the parent agent, then seeds initial timings and FCP.
+   * @private
+   * @param {RegisterAPIConstructor} opts
+   * @returns {Promise<void>}
+   */
+  async #register (opts) {
+    try {
+      const response = await this.#postMethodToAgent(REGISTER, [opts])
```
cwli24, line 152: This is important and related to initial timings right below this. `registeredAt` is set to the `now()` _of the iframe_, which strictly has a later offset from the parent window's `timeOrigin`. This solution will generalize the fix to future-proof all timings rather than just `registeredAt`:

```suggestion
      const response = await this.#postMethodToAgent(REGISTER, [opts, now()])
```
then propagate this to the parent window's `handleMethodCall`:
```js
if (method === REGISTER) {
  ...
  const iframeNow = args.[1]
  output.entity = agent[REGISTER](freshTarget)
  output.entity.metadata.target.iframeInterfaceId = iframeInterfaceId
  output.entity.metadata.target.iframeOrigin = event.origin
  Object.defineProperty(output.entity.metadata, 'clockOffset', {
    value: now() - iframeNow, // approx. top window timeOrigin <---> iframe registerEntity creation, base to all iframe timings
    enumerable: false // prevent serializeMetadata from sending this back to iframe
  })
  return output
}
```
and finally use it in parent's `handleTimingUpdate`:
```js
if (entity.metadata.timings) {
  const { property, value } = event.data
  if (isSafeProperty(entity.metadata.timings, property)) {
    const offsetForThisIframeEntity = entity.metadata.clockOffset || 0
    entity.metadata.timings[property] = typeof value === 'number' ? offsetForThisIframeEntity + value : value // type check guards operating on `asset` and `type` strings
  }
}
```

src/loaders/configure/configure.js
```diff
@@ -81,6 +82,13 @@ export function configure (agent, opts = {}, loaderType, forceDrain) {
...
+        handle('iframe-message', [event], undefined, 'IFRAME', agent.ee)
```
cwli24: Performance issue: you can be buffering a bunch of unrelated MessageEvent for a while until this drains, and holding onto all those `event.source` and `event.data` can be expensive for memory. You should pre-filter out the pertinent events now instead of waiting for `setupIframeMFEMessageListener` to do all the work:
```suggestion
(_import the `prefix` var from iframe-constant.js_)
if (typeof event.data?.type === 'string' && event.data.type.startsWith(prefix)) { // only buffer msg from iframe registerEntities
        handle('iframe-message', [event], undefined, 'IFRAME', agent.ee)
}
```

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+   * Instruments ajax using buffered resource timing so pre-registration entries are included.
+   * @private
+   */
+  #setupAjaxObserver () {
+    if (!globalScope.PerformanceObserver?.supportedEntryTypes?.includes('resource')) return
+
+    this.#resourceObserver = new globalScope.PerformanceObserver(list => {
+      list.getEntries().forEach(resource => this.#processResourceEntry(resource))
+    })
+    this.#resourceObserver.observe({ type: 'resource', buffered: true })
```
cwli24, line 207: Are you sure?
There's no identification of the initiator, so this can include:
* Ajax made by the iframe doc or a different MFE but is irrelevant to this registered entity specifically
* duplicate entries from 2+ RegisteredIframeEntity in same iframe

If so, should leave a comment on this particular behavior.

src/interfaces/registered-iframe-entity.js
cwli24: Also, you should probably add a jest test for this scenario to ensure #register setting block also causes iframe entity to be blocked

cwli24: Lastly, the `deregister` when called on the iframe entity should also ensure the final appropriate `.blocked` is set and understood to be set _on both ends_ (parent's and iframe's)

src/loaders/api/register.js
cwli24, line 143: `deregister` never removes an entity from the `agentRef.runtime.registeredEntities` list, and so every caller that uses `getRegisteredTargetsFromFilename` and `getRegisteredTargetsFromId` loops through that list, even if there are blocked or inactive entities, thereby increasing computation time. Is that intended?

cwli24: likewise, `deregister` also does not clean up this subscription `subscribeToPageUnload(reportTimings)` for both parent MFEs and iframe entities.

src/loaders/configure/iframe-message-handler.js
cwli24: this should have a jest test to ensure you cannot run those functions from the iframe entity

cwli24 (changes requested):

docs/warning-codes.md
```diff
@@ -142,4 +142,16 @@
...
+### 73
+`RegisteredIframeEntity failed to register with window context`
+### 74
+`RegisteredIframeEntity rejected message from unauthorized origin`
+### 75
+`RegisteredIframeEntity rejected message with mismatched iframeInterfaceId`
+### 76
+`Agent rejected post message, could not match with existing entity`
+### 77
+`Agent rejected post message, could not validate origin`
```
ellisong: Minor nitpick, but we should revisit whether all warning codes should have a period `.` at the end or not. The messages are inconsistent throughout.

src/loaders/api/register.js
```diff
@@ -78,7 +78,7 @@ function register (agentRef, target) {
   target.blocked = false
   if (typeof target.tags !== 'object' || target.tags === null || Array.isArray(target.tags)) target.tags = {}
   target.parent ??= {
-    get id () { return agentRef.runtime.appMetadata.agents[0].entityGuid }, // getter because this is asyncronously set
+    get id () { return agentRef.runtime.appMetadata.agents?.[0].entityGuid }, // getter because this is asyncronously set
```
ellisong: Rename `asyncronously` to `asynchronously`

src/loaders/configure/iframe-message-handler.js
```diff
@@ -0,0 +1,218 @@
...
+}
+
+/**
+ * Validates that the origin of the message event matches the expected origin for the registered entity
+ * @param {MessageEvent} event - The message event containing the method call or registration request
+ * @param {Object} entity - The registered entity associated with the iframeInterfaceId from the event data
+ * @returns {boolean}
+ */
+function isValidOrigin (event, entity) {
+  if (!entity || !event.origin) return false
```
ellisong: `MessageEvent.origin` can never be null or undefined. `!event.origin` is likely unnecessary.

src/interfaces/registered-iframe-entity.js
metal-messiah: I think I need to tweak this if it isnt actively consuming the iframe's event timestamp, that could lead to some inaccuracies down the line for API data

src/interfaces/registered-iframe-entity.js
metal-messiah: this is a remnant, will remove.  metadata sync-ing is now handled once during the register method's resolver, as the other methods wont cause desyncs of the metadata between window and iframe.

src/interfaces/registered-iframe-entity.js
metal-messiah: ok yeah that makes more sense i think.  better to be restrictive on this.

src/interfaces/registered-iframe-entity.js
metal-messiah: i believe this condition was just ported over from the existing ajax instrument module.  That said, i will dive into this to make sure the best behavior is applied

src/interfaces/registered-iframe-entity.js
metal-messiah: interesting, i think i overlooked that the resource api would be shared across frames.  i'll deep dive on this one

src/loaders/api/register.js
metal-messiah: thats mostly been intentional up to this point, but we can revisit this

src/interfaces/registered-iframe-entity.js
```diff
@@ -0,0 +1,449 @@
...
+   * @private
+   * @param {PerformanceResourceTiming} resource
+   */
+  #processResourceEntry (resource) {
+    if (!(resource.initiatorType in AJAX_INITIATOR_TYPES) || resource.responseStatus === 0) return
+    const params = { status: resource.responseStatus }
+    const metrics = { rxSize: resource.transferSize, duration: Math.floor(resource.duration), cbTime: 0 }
+    addUrl(params, resource.name)
+
+    this.#postMessageToParent(IFRAME_AJAX, {
```
ptang-nr: Consider sending the AJAX entries as a batch (With `buffered: true`, this fires immediately on instantiation with all prior resource entries.  I don't know too much, but from reading/googling about it, sounds like `postMessage` uses structured clone, which is considered a "heavy CPU-bound operation")

src/interfaces/registered-iframe-entity.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1752/commits/e20f3eecb3f23964610848de873fb0ac60f60c7f

src/interfaces/registered-iframe-entity.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1752/commits/8a2ec9eb262ece0d3adeadc3077f1840f6ed274b

metal-messiah: didnt go with bypassRegistration because we dont want the case of a timing occurring _during_ registration to get across the wire before registration is done for sure.  I just ended up moving it outside the register function to guarantee it was done executing first

src/loaders/configure/configure.js
metal-messiah: 752129d5d

src/loaders/configure/iframe-message-handler.js
metal-messiah: b33054775

src/interfaces/registered-iframe-entity.js
metal-messiah: e789c44fa

src/interfaces/registered-iframe-entity.js
metal-messiah: 466f99d0c

src/interfaces/registered-iframe-entity.js
metal-messiah: 87be939d2

src/interfaces/registered-iframe-entity.js
metal-messiah: 52e35db47

src/interfaces/registered-iframe-entity.js
metal-messiah: 37b10a1c1

src/interfaces/registered-iframe-entity.js
metal-messiah: 6bad62b56

src/interfaces/registered-iframe-entity.js
metal-messiah: 13d44751b

src/interfaces/registered-iframe-entity.js
metal-messiah: 2f421d17f

src/interfaces/registered-iframe-entity.js
metal-messiah: 28638faac

src/loaders/configure/configure.js
metal-messiah: 94ef30460

src/interfaces/registered-iframe-entity.js
metal-messiah: For now this should be treated as a known behavior of this module.  Lets complicate things further if customer feedback is received, otherwise we observe globally at the document and report for any and all registered iframe entities

src/loaders/api/register.js
metal-messiah: `registeredEntities` is also used for passive attribution (matching async errors/resource timings back to a target by filename/id via getRegisteredTargetsFromFilename/findTargetsFromStackTrace), so doing something like splicing on deregister could potentially break a correlation for any in-flight/async event tied to a since-deregistered entity, even though blocked already correctly no-ops future explicit calls. Leaving it as-is — the array growth is a real but lesser cost than losing attribution correctness. Same reasoning applies to the `subscribeToPageUnload` cleanup point, IMO its low impact since `pagehide` fires once per page life and `reportTimings()` is already idempotent, so not worth the added complexity right now.

docs/warning-codes.md
metal-messiah: 8fbf07dbd

src/loaders/api/register.js
metal-messiah: d514ee98e

src/loaders/configure/iframe-message-handler.js
metal-messiah: c5a012aea

src/interfaces/registered-iframe-entity.js
metal-messiah: 4af8e7cef

src/interfaces/registered-iframe-entity.js
cwli24: I see you've updated the constructor elsewhere to do the `.catch(() => {})`, so you don't need the rethrow on the await here:
image
^this is extraneous and effectively a no-op

When #registrationPromise is rejected, the await will throw the error for you.

src/loaders/configure/iframe-message-handler.js
cwli24: minor touch up but to keep it dry you should combine the two blocks:
```js
  output.entity = entity
  const methodFn = entity[method]
  if (!isSafeProperty(entity, method) || typeof methodFn !== 'function') {
    warn(35, method)
    return output
  }
```

src/interfaces/registered-iframe-entity.js
cwli24: when the main agent is executing those methods instructed by the iframe, is the `timestamp` supposed to be relative to the iframe itself (2 timelines) .. or is it supposed to be relative to the main agent (single timeline)?

src/interfaces/registered-iframe-entity.js
cwli24: Since you already addressed the 1st condition of `#isAllowedOrigin` in the constructor, `#isAllowedOrigin` is redundant & can be removed. This effectively becomes
```js
if (!(event.origin === this.#parentOrigin)) {
```

src/loaders/configure/iframe-message-handler.js
```diff
@@ -0,0 +1,276 @@
...
+ * @param {MessageEvent} event  - The message event containing the method call or registration request
+ * @param {Object} agent - The browser agent instance
+ * @returns
+ */
+function handleAjax (event, agent) {
+  const entity = getValidEntity(event, agent)
+  if (!entity) return
+  const { entries } = event.data
+  entries?.forEach(({ params, metrics, start, end, initiatorType }) => {
+    handle('xhr', [params, metrics, start, end, initiatorType, entity.metadata.target], undefined, FEATURE_NAMES.ajax, agent.ee)
```
ptang-nr, line 87: This was flagged by Claude - essentially clock offsets still has to be applied to the following:

- handleMethodCall:209: `pendingOriginTimestamps.set(entity, timestamp)` stores the raw iframe clock value. Every downstream consumer `(timeKeeper.correctRelativeTimestamp)` expects a container-relative value. This affects every API call — noticeError, addPageAction, log, recordCustomEvent, measure.

- handleAjax:107: `start` and `end` from the iframe's `PerformanceResourceTiming` are iframe-relative and forwarded raw. This affects all AJAX event timestamps and session trace placement.

src/interfaces/registered-iframe-entity.js
metal-messiah: 219f15892

src/loaders/configure/iframe-message-handler.js
metal-messiah: 219f15892

src/interfaces/registered-iframe-entity.js
metal-messiah: 392d97946

src/loaders/configure/iframe-message-handler.js
metal-messiah: 8d02faaf5

src/interfaces/registered-iframe-entity.js
metal-messiah: c328312f2

cwli24 (approved):

PR #1760, opened 2026-04-21
feat: Improve agent startup

ptang-nr:
Provide better support for scenarios where agent startup can be interrupted before RUM call completes successfully and feature session modes are saved into local storage (where applicable).
---

### Overview

In this PR, 
- the agent startup avoids defaulting and locking a new session into OFF mode for Session Trace, Session Replay, and/or Browser Logging when the first RUM call in a session is incomplete (for example, due to redirects or early navigations).  
- additionally when the RUM call completes, the agent will now cache the RUM response in the session/local storage such that the same feature modes can be reused for the duration of the session.
    - RUM call sends with fsh = 1, when it is the first RUM call in the session and there are no cached RUM response
    - RUM call sends with fsh = 0, when a cached RUM response is detected
- if a cached RUM response is present for a given session, the agent will use the cached response to continue startup

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-509779

### Testing

Extended integration tests in affected parts of the agent.
Manually tested scenario to ensure RUM response is cached in the session.


src/common/harvest/harvester.js
```diff
@@ -3,19 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import { SUPPORTABILITY_METRIC_CHANNEL } from '../../features/metrics/constants'
-import { FEATURE_TO_ENDPOINT, JSERRORS, RUM, EVENTS, FEATURE_NAMES, BLOBS, LOGS } from '../../loaders/features/features'
```
ptang-nr, line 6: Moved `send` function out to its own module s.t. Jest can spy on the the send function.

tests/unit/common/util/script-tracker.test.js
```diff
@@ -54,16 +54,17 @@ beforeEach(() => {
...
-      performanceObserverCallback = callback
-      return {
-        observe: jest.fn(),
-        disconnect: jest.fn()
-      }
-    }),
+
     Error: MockError
   }
+  global.window.PerformanceObserver = jest.fn((callback) => {
```
ptang-nr, line 60: Was hitting undefined error on line 69, moving the declaration out fixed it 🤷‍♀️

tests/specs/logging/harvesting.e2e.js
```diff
@@ -75,7 +75,9 @@ describe('logging harvesting', () => {
     })
 
     ;['api', 'api-wrap-logger', 'console-logger'].forEach(type => {
-      Object.keys(LOGGING_MODE).filter(mode => mode !== 'OFF').forEach(mode => {
+      // use INFO as the sample logging mode (there's unique behavior for console.log);
+      // all other modes should behave the same otherwise in terms of harvesting behavior
+      ['INFO'].forEach(mode => {
```
ptang-nr, line 80: Try to speed up this test - 3+ min to < 2 min.

src/common/session/session-entity.js
metal-messiah, line 21: Should `cachedRumResponse` be modeled in here?

src/features/session_replay/aggregate/index.js
```diff
@@ -89,18 +89,32 @@ export class Aggregate extends AggregateBase {
...
+            this.abort(ABORT_REASONS.ENTITLEMENTS)
+            this.reportSupportabilityMetric('SessionReplay/EnabledNotEntitled/Detected')
+          } else {
+            // avoid locking the session into OFF mode if entitled is unconfirmed
+            this.mode = null
+            this.recorder?.stopRecording?.()
+            this.syncWithSessionManager({ sessionReplayMode: this.mode })
+            this.recorder?.clearTimestamps?.()
+            while (this.recorder?.getEvents().events.length) this.recorder?.clearBuffer?.()
+          }
```
metal-messiah: This part is confusing me a bit.  When would `entitled` ever be null/undefined?

src/features/session_trace/aggregate/index.js
```diff
@@ -38,9 +38,15 @@ export class Aggregate extends AggregateBase {
 
   /** Sets up event listeners, and initializes this module to run in the correct "mode".  Can be triggered from a few places, but makes an effort to only set up listeners once */
   initialize (stMode, stEntitled, ignoreSession) {
-    this.entitled ??= stEntitled
-    if (!this.entitled) this.blocked = true
-    if (this.blocked) return this.deregisterDrain()
+    if (!this.blocked) this.entitled ??= stEntitled
+    if (this.entitled === null || this.entitled === undefined) {
```
metal-messiah: Im also confused how entitled could be supplied to this module as null/undefined.

metal-messiah: arent initialize() in both ST and SR triggered by getting the response to `waitForFlags`?  Shouldnt we just make sure that `waitForFlags` always returns flags?

metal-messiah (commented):
Just some general WIP comments, looking good!

src/features/session_replay/aggregate/index.js
ptang-nr: I think I was confused by a pre-existing logging agg test that emitted`'rumresp', {}`, and then wanting the features to have matching tests.  

We can always assume if RUM response comes back, entitlement + sampling decisions are there.  Updated!

src/features/session_trace/aggregate/index.js
ptang-nr: Confirmed LS writes/reads should be all-or-nothing, so `waitForFlags` should always return flags 👍 

Corrected the tests to remove the unconfirmed entitlement checks.  The `entitled` check here reflects strictly no entitlements, and will abort the feature (sets `mode` to OFF + `blocked` to true).

src/common/session/session-entity.js
ptang-nr: Added

src/features/page_view_event/aggregate/index.js
```diff
@@ -199,6 +201,8 @@ export class Aggregate extends AggregateBase {
       // Adding retry logic for the rum call will be a separate change; this.blocked will need to be changed since that prevents another triggerHarvestFor()
       this.ee.abort()
       return
+    } else if (this.isSessionTrackingEnabled && !this.agentRef.runtime.session.state.cachedRumResponse) {
+      this.agentRef.runtime.session.write({ cachedRumResponse: flags })
```
metal-messiah: It doesnt look like `app` is getting cached, is that by design?

metal-messiah (dismissed):
This looks good and makes sense. I assume `app` metadata is unneeded for the cache, but just want to double check that above ⬆️ before merging.

src/features/page_view_event/aggregate/index.js
ptang-nr: Sort of.  The design evolved and in the current state we are not keeping individual cache entries for each app in favor of reducing the cache complexity + keeping with the agent's current behavior of first app "wins".

(But if we'd like individual cache entries, that should be doable as well.)

src/common/harvest/harvester.js
cwli24: Jest can spy on send since it's exported even within this single module, no?

ptang-nr: From my attempts, I was unable to.  Best I understood is it's most likely running into the issue of the usage within harvester is internal so Jest was unable to eavesdrop.

ptang-nr: I think there were some potential "hacks" around it(?), but I didn't look too deeply.  Splitting `send` out to a module seemed the cleanest/most straightforward.

src/features/logging/aggregate/index.js
```diff
@@ -48,10 +48,14 @@ export class Aggregate extends AggregateBase {
       this.loggingMode ??= { auto, api } // likewise, don't want to overwrite the mode if it was set already
       const session = this.agentRef.runtime.session
       if (canEnableSessionTracking(agentRef.init) && session) {
-        if (session.isNew) this.#syncWithSessionManager()
-        else updateLocalLoggingMode(session.state.loggingMode, session.state.logApiMode)
+        // if session mode is already set, use it
+        if (session.state.loggingMode !== null || session.state.logApiMode !== null) {
```
cwli24: I think they needs to be a `&&`. They both should not be null if we're taking the mode from session, right?

cwli24: Actually this is too confusing to think about. I'd replace the original logic with the new definition rather than changing it up. That way we can be sure we're not introducing some edge case logic here to this change.

"if the session is new" ==> loggingMode **is null** (_OR_ logApiMode **is null**), agreed?
(technically they're always set as a pair, so the second check is optional and redundant)

cwli24: like:
```suggestion
        if (session.state.loggingMode == null || session.state.logApiMode == null) this.#syncWithSessionManager()
        else updateLocalLoggingMode(session.state.loggingMode, session.state.logApiMode)
```

src/features/logging/aggregate/index.js
```diff
@@ -48,10 +48,14 @@ export class Aggregate extends AggregateBase {
...
+          updateLocalLoggingMode(session.state.loggingMode, session.state.logApiMode)
+        } else {
+          this.#syncWithSessionManager()
+        }
       }
-      if (this.loggingMode.auto === LOGGING_MODE.OFF && this.loggingMode.api === LOGGING_MODE.OFF) {
+      if (!this.loggingMode.auto && !this.loggingMode.api) {
```
cwli24: double checking the logic here. The previous code would block the feature IF the session is new AND the feature flag returns them both as **off** _OR_ if the session exists AND the existing modes are both **off**, right?

So you _should not_ block the feature if the modes are nullish/undefined. That said, baed on the previous block, it's impossible for them to be nullish/undefined. (if they're off, they'd be `0`.) I think we should keep this as `== LOGGING_MODE.OFF` for clarity.

src/features/page_view_event/aggregate/index.js
```diff
@@ -225,6 +229,6 @@ export class Aggregate extends AggregateBase {
 
     this.drain()
     this.agentRef.runtime.harvester.startTimer()
-    activateFeatures(flags, this.agentRef)
+    activateFeatures(this.agentRef.runtime.session?.state.cachedRumResponse || flags, this.agentRef)
```
cwli24: Image

cwli24: (GH wouldn't let me make comment on the whole file for some reason so I'm putting it here)

src/features/session_replay/aggregate/index.js
```diff
@@ -89,6 +89,11 @@ export class Aggregate extends AggregateBase {
     const { error_sampling_rate, sampling_rate, autoStart, block_selector, mask_text_selector, mask_all_inputs, inline_images, collect_fonts } = agentRef.init.session_replay
 
     this.waitForFlags(['srs', 'sr']).then(([srMode, entitled]) => {
+      // set session to OFF mode only if there's a RUM response
+      if (!this.blocked && entitled === 0) {
+        this.mode = MODE.OFF
+        this.syncWithSessionManager({ sessionReplayMode: this.mode })
+      }
```
cwli24: I think these 2 lines can just go in the `if (!this.entitled)` block below. (this.blocked doesn't matter)

src/features/session_replay/aggregate/index.js
```diff
@@ -165,14 +170,16 @@ export class Aggregate extends AggregateBase {
 
     if (this.recorder?.trigger === TRIGGERS.API && this.agentRef.runtime.isRecording) {
       this.mode = MODE.FULL
-    } else if (!session.isNew && !ignoreSession) { // inherit the mode of the existing session
+    } else if (session.state.sessionReplayMode !== null && !ignoreSession) { // inherit the mode of the existing session
       this.mode = session.state.sessionReplayMode
     } else {
       // The session is new... determine the mode the new session should start in
       this.mode = srMode
+      this.syncWithSessionManager({ sessionReplayMode: this.mode })
```
cwli24: There's another sync down at L202, and it seems the code in between here and there can also change the mode again. I'm concern there may be unintended behavior from writing the mode potentially more than once and potentially with different mode values, if the return statements don't apply.

Should this be worked into the early return statement blocks instead?

src/features/session_replay/aggregate/index.js
```diff
@@ -165,14 +170,16 @@ export class Aggregate extends AggregateBase {
...
     }
-    // If off, then don't record (early return)
-    if (this.mode === MODE.OFF) return
+
+    // If off or undetermined, then don't record (early return)
+    if (!this.mode) return
```
cwli24: did you find a case wherein this.mode would still be null here?

src/features/utils/aggregate-base.js
```diff
@@ -119,7 +119,7 @@ export class AggregateBase extends FeatureBase {
       }
       function buildOutput (ref) {
         return flagNames.map(flag => {
-          if (!ref[flag]) return 0
+          if (ref[flag] === undefined) return null
```
cwli24: this is the key-value pairs from the RUM response, which should never have undefined values. `ref[flag]` should simply be cast to numbers

src/features/session_trace/aggregate/index.js
```diff
@@ -39,8 +39,10 @@ export class Aggregate extends AggregateBase {
   /** Sets up event listeners, and initializes this module to run in the correct "mode".  Can be triggered from a few places, but makes an effort to only set up listeners once */
   initialize (stMode, stEntitled, ignoreSession) {
     this.entitled ??= stEntitled
-    if (!this.entitled) this.blocked = true
-    if (this.blocked) return this.deregisterDrain()
+    if (!this.entitled) {
+      this.abort()
```
cwli24: `abort` calls `warn` which can be provided a "code", but I think we should pass a message here that ST is not entitled/etc to better assist troubleshooting for end users.

src/features/session_trace/aggregate/index.js
```diff
@@ -72,12 +74,15 @@ export class Aggregate extends AggregateBase {
...
     /** ST/SR sampling flow in BCS - https://drive.google.com/file/d/19hwt2oft-8Hh4RrjpLqEXfpP_9wYBLcq/view?usp=sharing */
     /** ST will run in the mode provided by BCS if the session IS NEW.  If not... it will use the state of the session entity to determine what mode to run in */
-    if (!this.agentRef.runtime.session.isNew && !ignoreSession) this.mode = this.agentRef.runtime.session.state.sessionTraceMode
+    if (this.agentRef.runtime.session.state.sessionTraceMode !== null && !ignoreSession) this.mode = this.agentRef.runtime.session.state.sessionTraceMode
     else this.mode = stMode
 
     /** If the mode is off, we do not want to hold up draining for other features, so we deregister the feature for now.
      * If it drains later (due to a mode change), data and handlers will instantly drain instead of waiting for the registry. */
-    if (this.mode === MODE.OFF) return this.deregisterDrain()
+    if (!this.mode) {
```
cwli24: If mode is OFF, it makes sense to write that to let other pages know ST is off for the session. But if it's null (again, is that even possible here?), we shouldn't ping other pages for a null/"unknown" mode.

src/features/utils/aggregate-base.js
ptang-nr: I think this was more about enforcing the `null` value to avoid defaulting to 0/OFF.

src/features/page_view_event/aggregate/index.js
ptang-nr: Updated to cache `app` to avoid collisions after our team discussion.

src/features/utils/aggregate-base.js
cwli24: Right but this is the RUM response, which is the single source and determinant of the feature's mode. Everything should either be on or off after this point. Having undecided/undefined/null flag or signal is not an option--that's indicating something went haywire or is faulty. We don't want some features to be running & reported while others to be in a limbo state, if the RUM response is corrupt.

ptang-nr: If the RUM response is corrupted, defaulting to 0/OFF masks the symptom, and the agent would run in a degraded state without any visibility (though we can now technically check the cached RUM response).

IMO, we should keep the agent's expected behavior fixed and internally consistent, independent of the correctness of other parts of the system.  Defaulting to 0/OFF is inconsistent with the ticket's goal of avoiding the OFF lock-in, and keeps us tightly coupled to the backend (think deployment).

cwli24: In that case, rather than returning `null` and relying on the features to handle the `null` properly (something that isn't audited for every feature + future features), it'd be better to execute some kind of failure mechanism here. Either that's aborting the feature and/or printing out a warning message, or something else.

SR still imports & start recorder even if its `this.mode == null` but it doesn't send anything. ST still have handlers registered and drains for `null` mode. The agent will persist in a dirty state, but some of these issues that likely impacts performance can still be masked.

ptang-nr: I think I'm going to flip this back to 0, but kind of under protest.

Agreed it'd be better to abort and not have the failure be masked.  I think defaulting to 0 is "clean" in one way but is still masking a problem.  After a session is over, there'd be no way to tell whether it was truly a 0 or the agent received an impossible value. (I think I did have the features to treat all falsy modes to shutdown initially but were asked to change back to explicit OFF checks 🤷‍♀️)

cwli24: If it's undefined, I think you can throw a new `warn` then return 0 for it. That leaves at least a trail for it. Do you want to do that?

ptang-nr: Ok, I added a warning for it.  Lmk if there's any change needed there.

cwli24 (approved):

PR #1762, opened 2026-04-29
feat: Enhance SPA API to support targetPageLoad option

### LINKED ISSUE #1756: soft_nav - Set name of initialPageLoad BrowserInteraction event
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
It does not seem possible to set the name of the initialPageLoad BrowserInteraction event.
window.newrelic.setPageViewName does work for PageView event.
window.newrelic.setCurrentRouteName does not work for the initialPageLoad BrowserInteraction event.
- (It is not clear why, but a simple reproduction is provided where it does not impact the captured data)
window.newrelic.interaction().setName(browserInteractionName); does not act on the initialPageLoad event.
- (As window.newrelic.interaction() will not return the initialPageLoad interaction).

## Steps to Reproduce

- JS snippet from the browser-application-settings tab on newrelic.com
- Call:
    - setPageViewName
    - setCurrentRouteName
    - interaction().setName
- Check data by querying NR

## Expected Behavior

- Some combination of `browserInteractionName, previousRouteName, targetRouteName,  targetGroupedUrl, previousGroupedUrl` is updated to use the custom route name.
    - Same / Similar to how setPageViewName sets a custom name for PageView

```html
<html lang="en">
    <head>
        <script type="text/javascript">

            /* SETUP - Add the JS Snippet from Application Settings here */
            
            const someCustomRouteName = '/some-example-path/*'
            const setPageViewNamesPrefixForBrowserInteraction = window.location.hostname + ':' + (window.location.port || (window.location.protocol === 'https:' ? '443' : '80'));
            const browserInteractionName = setPageViewNamesPrefixForBrowserInteraction + someCustomRouteName;
            console.log({browserInteractionName});
            window.newrelic.setPageViewName(someCustomRouteName, window.location.origin);
            window.newrelic.setCurrentRouteName(browserInteractionName);

            // Does not work - Does not get the initialPageLoad interaction
            window.newrelic.interaction().setName(browserInteractionName);
            
            // Does work, name is set to the custom name
            window.addInteraction = () => window.newrelic.interaction().setName(browserInteractionName).save().end();
        </script>
    </head>
    <body><div>Placeholder</div><button onClick=window.addInteraction()>Add interaction</button></body>
</html>
```

## Your Environment

- Current Chrome
- Was testing with `/*! For license information please see nr-loader-spa-1.313.0.min.js.LICENSE.txt */`

## Additional context

- See https://github.com/newrelic/newrelic-browser-agent/commit/762492896a7b96564269aab1aadeb6e44a4da242#diff-8e79f8cdc78d2ee0a0abe1f0428fcf9cb1c0bcf92d5cb49485ea14c3ff31ad54R208; change where the `interaction` call can not return the IPL.
- It is not clear why setCurrentRouteName does not impact the IPL; it may be a race condition or it may be an artifact of how the code is structured.
- You can notice that in the code above the code has to generate the setPageViewNamesPrefixForBrowserInteraction to match the format from PageView.
    - Is this intended? Can there be some method to ensure they match without having to replicate the logic?
- If it does not reproduce upon testing I can provide the first few network requests

cwli24:
Allow the SPA API to target and work on the `initialPageLoad` type `BrowserInteraction` event. A new option is added to `.interaction` that now binds the returned handle to the initial page load interaction at any point in time.
---

### Overview

Example:
```js
newrelic.interaction({ targetPageLoad: true }).setName('My Home Page')
```

Resolves #1756 

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-555406

Docs PR: https://github.com/newrelic/docs-website/pull/23890

### DIFF
--- a/src/features/soft_navigations/aggregate/index.js
+++ b/src/features/soft_navigations/aggregate/index.js
@@ -30,7 +30,6 @@ export class Aggregate extends AggregateBase {
       this.initialPageLoadInteraction.forceSave = true // unless forcibly ignored, iPL always finish by default
       const ixn = this.initialPageLoadInteraction
       this.events.add(ixn) // add the iPL ixn to the buffer for harvest
-      this.initialPageLoadInteraction = null
     })
 
     loadTime.subscribe(({ value: loadEventTime }) => {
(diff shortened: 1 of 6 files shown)


metal-messiah (approved):
I think this all looks good and makes sense.  A few tasks to check off before merging on this:
- [ ] Do we have adequate testing in place that ensures that API methods invoked on "finished" interactions successfully do "nothing"?
- [ ] Do we need to run this past API review?
- [ ] Do we have a ticket for updating the API public docs?

PR #1764, opened 2026-05-01
feat: Improve PageViewTiming load timing capture

cwli24:
The load timing is now more reliably captured for pages, inheriting the idea from web-vital's `onTTFB`. Before, it occasionally missed `loadEventEnd` from nav entry from a browser race condition, particularly for larger/longer-loaded pages, which caused load to be `0`.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-558319

### DIFF
--- a/src/common/vitals/load-time.js
+++ b/src/common/vitals/load-time.js
@@ -12,12 +12,15 @@ export const loadTime = new VitalMetric(VITAL_NAMES.LOAD_TIME)
 if (isBrowserScope) {
   const perf = globalScope.performance
   const handler = () => {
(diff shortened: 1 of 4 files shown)


metal-messiah (approved):
All makes sense to me.  Should this be a fix: or a feat: ?

cwli24:
> All makes sense to me. Should this be a fix: or a feat: ?

I'm not sure lol. But because the ticket was created as a story, not a bug, I'm rolling with it as a feat. I suppose the original implementation wasn't entirely "wrong".

PR #1765, opened 2026-05-04
fix: Patch handling of inlining images for Session Replay

ptang-nr:
Fixes an issue where Session Replay may leave an image broken in the live DOM.  In this version, the underlying rrweb version will avoid manipulating the live image on the DOM.
---

### Overview

In this PR, the agent will use an updated version of the forked rrweb lib to avoid manipulating images on the live DOM when `inline_images` is set to `true`.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-519780
Depends on: https://github.com/newrelic-forks/rrweb/pull/13

### Testing

Updated relevant unit tests


tests/specs/soft_navigations/xhr.e2e.js
```diff
@@ -161,12 +167,15 @@ describe('XHR SPA Interaction Tracking', () => {
...
+    await Promise.all([
+      interactionsCapture.waitForResult({ totalCount: 1 }),
+      browser.url(
+        await browser.testHandle.assetURL('soft_navigations/ajax/xhr-no-send.html')
+      ).then(() => browser.waitForAgentLoad())
+    ])
 
     const [interactionHarvests] = await Promise.all([
-      interactionsCapture.waitForResult({ timeout: 10000 }),
+      interactionsCapture.waitForResult({ totalCount: 2 }),
```
ptang-nr: The click is expected to be captured as a route change, just that it shouldn't have an 'ajax' node in its children.  So standardized to wait for 2 interactions, and then check.

tests/unit/features/session_replay/shared/recorder.test.js
```diff
@@ -1,34 +1,324 @@
...
+        const recorder = new Recorder(makeSrInstrument({ fix_stylesheets: false }))
+        recorder.audit(metaEvent())
+        expect(recorder.getEvents().events).toHaveLength(1)
+      })
+
+      test('stores the event even when there are incomplete stylesheets', () => {
+        // evaluate() is not called when fix_stylesheets is false in config, incompletes always 0
+        const recorder = new Recorder(makeSrInstrument({ fix_stylesheets: false }))
+        recorder.audit(metaEvent())
+        expect(recorder.getEvents().events).toHaveLength(1)
```
ellisong: Why do `stores the event even when there are incomplete stylesheets` and `stores the event directly when there are no incomplete stylesheets` tests have the same exact code?

ptang-nr: Good catch - looks like it's a bad test (can't actually mock the condition).  Dropped!

metal-messiah (approved):

PR #1638, opened 2025-11-25
feat: Make soft navigations feature the default SPA

cwli24:
Soft navigations feature is now the enforced running feature for SPA. The older and deprecated SPA feature cannot be selected or used as of this update.
---

### Overview

- Removed references to the old SPA feature in interaction.js and configure.js.
- Updated tests to eliminate reliance on the soft navigation feature flag.
- Adjusted interaction handling in ajax and error tests to accommodate new soft navigation logic.
- Cleaned up various test files by removing unnecessary feature flag initializations.
- Updated end-to-end tests to reflect changes in soft navigation handling and removed deprecated configurations.
- Marked several SPA-related tests as skipped to focus on soft navigation functionality.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-489970

### Testing

e2e & jest tests are updated to skip old spa (since old spa feature is no longer selectable; will be removed at later date) and to assume soft nav as the default feature

PR #1769, opened 2026-05-08
bug: PageViewTiming events mis-attributed to wrong browserTransactionName after soft navigation (regression since v1.277.0 to v1.312.1)

OzTK:
> **Note:** This PR contains a **failing test only** — no fix is included. It is submitted to provide New Relic with reproducible, evidence-backed proof of a production regression so the team can prioritise a fix. We are happy to collaborate on the solution.

---

## Summary

Since upgrading from **v1.277.0 → v1.312.1**, `PageViewTiming` Web Vitals events (LCP, FCP, INP, CLS) are being attributed to the **wrong `browserTransactionName`** in New Relic for SPA customers. Events captured on route A are consistently appearing under route B's transaction name — the route the user navigated *to*, not *from*.

This is a silent data-quality regression: total `PageViewTiming` volume is unchanged, so no alerts fire, but per-route Web Vitals dashboards are reporting metrics for the wrong pages.

---

## The Test

```
FAIL tests/components/page_view_timing/soft-nav-route-attribution.test.js

  ● REGRESSION: LCP captured on /groceries/pdp is sent with ref=/groceries/homepage
      after soft navigation, causing wrong browserTransactionName in New Relic

    expect(received).toContain(expected)

    Expected substring: "/groceries/pdp"
    Received string:    "http://localhost/groceries/homepage"
```

The test intercepts `XMLHttpRequest.prototype.open` to capture the exact beacon URL sent to NR's collector, then asserts on the `ref` query param — the field NR's backend uses to derive `browserTransactionName`.

**Scenario:**
1. User is on `/groceries/pdp`. LCP fires (2400ms) and is buffered in `PageViewTiming`.
2. User navigates to `/groceries/homepage` via a soft navigation. `window.location` changes.
3. Harvest fires (simulating the 30-second periodic interval, by which time location = `/homepage`).
4. The beacon URL contains `ref=http://localhost/groceries/homepage`.

**Expected:** `ref` = `/groceries/pdp` → NR attributes LCP to `browserTransactionName = "groceries/pdp"` ✓  
**Actual:** `ref` = `/groceries/homepage` → NR attributes LCP to `browserTransactionName = "groceries/homepage"` ✗

A companion test passes, confirming the correct behaviour: when harvest fires *before* navigation, `ref` correctly identifies `/groceries/pdp`.

### How to run

```bash
git clone https://github.com/newrelic/newrelic-browser-agent
cd newrelic-browser-agent && npm install

./node_modules/.bin/jest 
  tests/components/page_view_timing/soft-nav-route-attribution.test.js 
  --no-coverage
```

Expected: **1 failing, 1 passing.**

---

## Real-World Impact (Production Evidence)

We observed this regression immediately after deploying v1.312.1. Comparing a 4-day window **before** vs. **after** the upgrade (same days of week, controlled for traffic patterns):

| Route | Before | After | Delta |
|---|---|---|---|
| `groceries/pdp` | 13,352,459 | 12,609,507 | **−742K (−5.6%)** |
| `groceries/search` | 7,812,122 | 7,282,645 | **−530K (−6.8%)** |
| `groceries/browse` | 4,831,286 | 4,362,516 | **−469K (−9.7%)** |
| `groceries/global-homepage` | 5,511,948 | 6,143,713 | **+632K (+11.5%)** |
| `groceries/homepage` | 5,319,723 | 5,804,828 | **+485K (+9.1%)** |
| `groceries/buylists` | 894,742 | 1,062,533 | **+168K (+18.8%)** |

The pattern is unambiguous: **pages users navigate *from*** lose `PageViewTiming` volume; **pages users navigate *to*** gain the exact same amount. Total volume is conserved — the events are not lost, they are **systematically misattributed**.

This maps directly to the test failure: LCP captured on `/pdp` (a page navigated *from*) is sent with `ref=/homepage` (the page navigated *to*).

---

## What the `soft_navigations` feature does incorrectly

When a soft navigation occurs, buffered `PageViewTiming` events are not flushed before `window.location` changes. By the time the next periodic harvest fires, `window.location` reflects the destination route — so the beacon is sent with the wrong `ref`, and NR attributes all buffered source-route Web Vitals to the destination's `browserTransactionName`.

This behaviour regressed in **`d93f3f38` — "feat: Make soft navigations feature the default SPA (#1638)"**, which removed the old `spa` feature that had always flushed `PageViewTiming` synchronously on interaction completion.

---

## Fix Options

Two approaches would resolve this:

1. **Flush before navigation** — flush the `PageViewTiming` buffer when a soft navigation is detected (i.e. on the `newURL` event), before `window.location` changes. This preserves the existing `ref`-based attribution without any structural changes.

2. **Persist `ref` on buffered events** — snapshot `window.location` at the time each event is buffered and carry it through to the harvest, so the beacon `ref` reflects where the event was captured regardless of when the harvest fires.

---

## Severity Assessment

| Factor | Assessment |
|---|---|
| **Scope** | All SPA customers using `soft_navigations` (default since v1.312.1) |
| **Data quality** | Web Vitals per-route dashboards are silently wrong |
| **Detectability** | Near-zero — total volume is unchanged, no volume-based alerts fire |
| **Business impact** | LCP/INP/CLS targets assessed against data attributed to the wrong pages |
| **Introduced by** | v1.312.1 / commit `d93f3f38` (merged 2026-01-06) |

We consider this **high priority** and would welcome a patch release.


JustinRush-NR:
Thanks for the thorough test and explanation.

The attribute `pageUrl` is the window.location when the PageViewTiming event is captured. That attribute will always reflect the name/location at the time the Core Web Vitals were recorded for the PVT event.

browserTransactionName is meant to reflect the custom value set by the customer for PageView events using [setPageViewName()](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/setpageviewname/). It's not evaluated any differently for PageViewTiming events.

The ingest pipeline assigns a fallback value in the case where no view name is supplied using `ref` at the time of harvest. The timing of that harvest did change in [Dec 2024](https://github.com/newrelic/newrelic-browser-agent/pull/1298).

metal-messiah:
We believe this to have been addressed by other means, namely the browser agent APIs -- please reopen if that is not the case

PR #1776, opened 2026-05-15
feat: Use app-namespaced keys for session

cwli24:
The browser agent now namespaces the session info to per browser app, using the combination of `licenseKey` and `applicationID` provided at startup. This means different apps sharing the same origin and browser `localStorage` will no longer share the same session, which was an accepted but not ideal collision, causing feature modes to bleed from one over to another for the duration of a session previously.
---

### Overview

Each app will manage & map to its own session state object. This includes multiple agents of differing apps on the same page (though multi-agents on same page is _not_ supported by New Relic).

Expect some changes to session-related features: Replay, Trace, Logging. For example, originA:appB will now rely on its own RUM response flags and calculate its own feature mode rather than inherit from originA:appA.
 
### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-565063

### Testing

Tests modified + new tests added. Most session tests shifting over to namedspaced read/write covers the changes.

### DIFF
--- a/src/common/session/session-key.js
+++ b/src/common/session/session-key.js
@@ -1,0 +1,17 @@
+/**
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+// Deterministic 32-bit FNV-1a hash, rendered as a fixed 8-char hex string.
+function hashTo8Hex (value) {
+  let hash = 0x811c9dc5
+  for (let i = 0; i < value.length; i++) {
+    hash ^= value.charCodeAt(i)
+    hash = Math.imul(hash, 0x01000193)
+  }
+  return (hash >>> 0).toString(16).padStart(8, '0')
+}
+
+export function getAppSessionHash (licenseKey, applicationID) {
+  return hashTo8Hex(`${String(licenseKey)}:${String(applicationID)}`)
+}
--- a/src/features/session_replay/instrument/index.js
+++ b/src/features/session_replay/instrument/index.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 /**
@@ -30,8 +30,9 @@ export class Instrument extends InstrumentBase {
     setupPauseReplayAPI(agentRef)
 
     let session
+    const fullNamespacedKey = `${PREFIX}_${DEFAULT_KEY}::${agentRef.info.licenseKey}:${agentRef.info.applicationID}`
     try {
-      session = JSON.parse(localStorage.getItem(`${PREFIX}_${DEFAULT_KEY}`))
+      session = JSON.parse(localStorage.getItem(fullNamespacedKey))
     } catch (err) { }
 
     if (hasReplayPrerequisite(agentRef.init)) {
--- a/src/features/session_replay/instrument/index.js
+++ b/src/features/session_replay/instrument/index.js
@@ -7,7 +7,8 @@
  */
 
 import { handle } from '../../../common/event-emitter/handle'
-import { DEFAULT_KEY, MODE, PREFIX } from '../../../common/session/constants'
+import { MODE, SESSION_STORAGE_KEY_PREFIX } from '../../../common/session/constants'
+import { getAppSessionHash } from '../../../common/session/session-key'
 import { InstrumentBase } from '../../utils/instrument-base'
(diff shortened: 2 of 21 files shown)


src/features/session_replay/instrument/index.js
```diff
@@ -30,8 +30,9 @@ export class Instrument extends InstrumentBase {
     setupPauseReplayAPI(agentRef)
 
     let session
+    const fullNamespacedKey = `${PREFIX}_${DEFAULT_KEY}::${agentRef.info.licenseKey}:${agentRef.info.applicationID}`
```
metal-messiah: @cwli24 one thing we may have missed during CDD review is that storing the licensekey in persistent enduser storage may be seen as undesirable from a customer/internal perspective.  We should consider if this is an issue, and if so, we may need to at least hash this value instead of storing directly as human readable

tests/specs/npm/micro-agent.e2e.js
```diff
@@ -4,11 +4,34 @@ import { testErrorsRequest, testInsRequest, testLogsRequest, testRumRequest } fr
...
+        return modeReady && log1.drained === true && log2.drained === true
+      }),
+      {
+        timeout: 10000,
+        timeoutMsg: 'Both micro-agent logging aggregates never fully initialized'
+      }
+    )
+  }
+
+  it('shuts down logging for both micro-agents when logging mode is OFF', async () => {
```
ptang-nr: This is changing the meaning of the test.  It was meant to capture that the agent would've fallen back to OFF mode, because of shared session.

So with NS'd session, the expected behavior should be 1st agent is turned OFF but 2nd agent stays on DEBUG.  (🤔 I thought I was able to test for this by keeping the agents initializations staggered https://github.com/newrelic/newrelic-browser-agent/pull/1760/changes#diff-8a78c9a53f5aa4838d80d1874eecf31b6e96bc58a165d91ae53310ac38047b28)

cwli24: This test was problematic because it's very flaky. There's no mechanism right now to enforce a) which agent request RUM first, b) which request the test server replies to first/in order, and finally c) which agent actually processes the response first.

cwli24: As discussed, I'm just going to revert the changes to this file.

ptang-nr: @cwli24 thanks for giving that a go.  If the test is causing too much noise/pain in the meantime, I'd like to propose disabling it.  (I'd still like a chance to take a closer look when we merge this into https://github.com/newrelic/newrelic-browser-agent/pull/1760. Just haven't reasoned out why the staggered init appears to work in that branch but not in this branch 🤔)

cwli24: Actually I was able to readjust the tests properly.

src/common/session/session-key.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+// Deterministic 32-bit FNV-1a hash, rendered as a fixed 8-char hex string.
```
ptang-nr, line 5: I was unfamiliar with FNV-1a hash and did some googling (cool to learn about this!). 

However I do have some concerns.  Google noted that it's not cryptographically secure.  What do you think about using something that'll be crypto-secure? (for example, SHA-256). Looks like it should still be performant while covering our bases.

(I believe license keys are being treated as a secret.  See bullet 1e + its discussion thread in [OYC](https://newrelic.atlassian.net/wiki/spaces/TDP/blog/2023/11/28/3092121138/Own+your+Component#Work-to-be-prioritized%3A))

src/features/session_replay/instrument/index.js
```diff
@@ -30,8 +31,9 @@ export class Instrument extends InstrumentBase {
     setupPauseReplayAPI(agentRef)
 
     let session
+    const fullNamespacedKey = `${SESSION_STORAGE_KEY_PREFIX}${getAppSessionHash(agentRef.info.licenseKey, agentRef.info.applicationID)}`
     try {
-      session = JSON.parse(localStorage.getItem(`${PREFIX}_${DEFAULT_KEY}`))
+      session = JSON.parse(localStorage.getItem(fullNamespacedKey))
```
ptang-nr, line 36: I didn't entirely follow why SR had its own way of reading the session.  Couldn't this call the SessionEntity's `read()` function?  It'd be more centralized/consistent.

metal-messiah: session entity wont exist yet in instrument when SR is evaluating to be preloaded

ptang-nr (approved):

PR #1782, opened 2026-05-29
feat: Enhance obfuscation to support event filter

metal-messiah:
Obfuscation behavior can now be configured to apply only to a subset of event types, using the `eventFilter` property, which supports an array of event type strings per obfuscation rule.  When supplied, the obfuscation rule will only be applied when the event type matches one of the items in the eventFilter array. For backwards compatibility, not supplying an eventFilter property on configuration rules will apply the rule to all data.
---

### Overview
This PR introduces an event filter to the obfuscation mechanism.  This is a precursor to AJAX payload support. 

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-531298

### Testing
New tests have been added that validate that the rules can be applied to individual event types.


src/common/util/obfuscate.js
```diff
@@ -24,15 +25,34 @@ import { warn } from './console'
...
-    return this.agentRef.init.obfuscate || []
+    const allRules = this.agentRef.init.obfuscate || []
+
+    // If this instance has no specific event type, return all rules
+    if (!this.eventType) return allRules
+
+    // Filter rules to only those that apply to this instance's event type
+    return allRules.filter(rule => {
+      // If rule has no eventFilter, it applies to all events
+      if (!rule.eventFilter || !Array.isArray(rule.eventFilter) || rule.eventFilter.length === 0) {
```
ellisong: if rule.eventFilter has to be an array, then you should be able to leave out `!rule.eventFilter` check

src/common/util/obfuscate.js
```diff
@@ -60,6 +80,133 @@ export class Obfuscator {
...
+    }
+
+    // For generic obfuscators (no specific event types), check individual eventType properties
+    // This path is for features like generic_events that handle multiple event types
+    const eventTypesToObfuscate = new Set()
+    const globalRules = []
+    const eventSpecificRules = []
+
+    this.obfuscateConfigRules.forEach(rule => {
+      if (rule.eventFilter && Array.isArray(rule.eventFilter) && rule.eventFilter.length > 0) {
```
ellisong: can also leave out `rule.eventFilter` here

src/common/util/obfuscate.js
```diff
@@ -60,6 +80,133 @@ export class Obfuscator {
...
+   */
+  #applyFnWithEventTypeFilter (obj, fn, eventTypes, globalRules, shouldObfuscate) {
+    if (!obj || typeof obj !== 'object') return
+
+    // Determine the obfuscation state for this object
+    let currentShouldObfuscate = shouldObfuscate
+
+    // Check if this object has an eventType property
+    if ('eventType' in obj && typeof obj.eventType === 'string') {
+      currentShouldObfuscate = !!eventTypes.includes(obj.eventType)
```
ellisong: The `Array.prototype.includes()` function returns a boolean so the `!!` shouldn't be necessary

ellisong (commented):
Really nice work on the testing! See nitpicky comments

src/common/util/obfuscate.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1782/commits/1ed507b3b9e12ff6c215b46f61b9346a51f4dd77

src/common/util/obfuscate.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1782/commits/1ed507b3b9e12ff6c215b46f61b9346a51f4dd77

src/common/util/obfuscate.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1782/commits/1ed507b3b9e12ff6c215b46f61b9346a51f4dd77

ellisong (approved):
LGTM

PR #1789, opened 2026-06-09
fix: Calculate browser_stack_hash from truncated stack trace

ellisong:
Calculate the browser stack hash off of the truncated stack trace instead of the non-truncated stack trace. Before this change, first-occurrence harvests of JSE events can have a different browser stack hash from future harvests of the same stack trace if it was truncated. This will improve error grouping of JavaScriptError events.
---

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-558128

### Testing

Make sure unit tests pass. We can add more testing related to truncation logic if it helps.


src/features/jserrors/aggregate/index.js
```diff
@@ -267,7 +267,7 @@ export class Aggregate extends AggregateBase {
       params.stack_trace = truncateSize(stackInfo.stackString)
       this.observedAt[bucketHash] = Math.floor(this.agentRef.runtime.timeKeeper.correctRelativeTimestamp(time))
     } else {
-      params.browser_stack_hash = stringHashCode(stackInfo.stackString)
+      params.browser_stack_hash = canBeTruncated(stackInfo.stackString) ? stringHashCode(truncateSize(stackInfo.stackString)) : stringHashCode(stackInfo.stackString)
```
ptang-nr: Since `truncateSize` already checks input string length against the limit, there seems to be some redundancy.  It seems like we can call `stringHashCode(truncateSize(stackInfo.stackString))` directly here 🤔

ptang-nr: PS. What do you think about adding an agg unit test to ensure `browser_stack_hash` on second+ occurrence matches the expected hash of the truncated stack string?

ellisong: I agree with the redundancy and changed it to your suggestion. I also reverted `canBeTruncated` since that was extracted out of `truncateSize` due to using it in the new code.

This change should be tested, but we don't have any aggregate unit tests for JS errors that I'm aware of; they're mostly covering helper or utility methods. I found an example in `jserrors/with-softnav.test.js` at the component test level that I'll look into. I should have a component, if not an e2e, test ready by Thursday mid-day.

ptang-nr (approved):

PR #1768, opened 2026-05-05
feat: MFE core web-vitals detection

metal-messiah:
Add internal detection methods for FCP, INP, CLS, and LCP when spawned by a MFE
---

### Overview

This PR implements comprehensive Web Vitals tracking for Micro Frontend (MFE) components, adding detection and metadata collection for four Core Web Vitals metrics: First Contentful Paint (FCP), Largest Contentful Paint (LCP), Cumulative Layout Shift (CLS), and Interaction to Next Paint (INP).

#### Key Changes

**Unified Vitals Object Structure**
Each vital metric is tracked as a unified object containing both the measurement value and rich contextual metadata. These objects are reported directly as JSON in the `MicroFrontEndTiming` event:

```javascript
{
  'nr.vitals.fcp': {
    value: 1234.56,      // Timestamp when FCP occurred
    loadState: 'loading'  // Document readyState at FCP
  },
  'nr.vitals.lcp': {
    value: 2345.67,      // Timestamp when LCP occurred
    size: 480000,        // Element size in pixels (width × height)
    elTag: 'IMG',        // HTML tag name of LCP element
    eid: 'hero-image',   // Element ID (if present)
    elUrl: 'https://...' // Element URL (images/videos/background-images)
  },
  'nr.vitals.cls': {
    value: 0.123,                    // Cumulative layout shift score
    largestShiftValue: 0.05,         // Value of largest individual shift
    largestShiftTime: 1234,          // Timestamp of largest shift
    largestShiftTarget: 'div#hero',  // CSS selector of element that shifted
    loadState: 'complete'            // Document readyState when measured
  },
  'nr.vitals.inp': {
    value: 250,                      // Total interaction latency (ms)
    interactionTarget: 'button#submit', // CSS selector of interacted element
    interactionTime: 1234,           // Timestamp of interaction
    interactionType: 'click',        // Interaction type (click, keydown, etc.)
    inputDelay: 10,                  // Input delay (ms)
    processingDuration: 190,         // Event processing time (ms)
    presentationDelay: 50,           // Presentation delay (ms)
    nextPaintTime: 1484,             // Next paint timestamp
    loadState: 'interactive'         // Document readyState during interaction
  }
}
```

**MicroFrontEndTiming Event Attributes**
When an MFE deregisters (via `deregister()` API), a `MicroFrontEndTiming` event is generated with the following JSON object attributes:

**Important: these attributes are prefixed with nr. to ensure they are hidden while being evaluated**

[...]

PR #1801, opened 2026-06-25
chore: FAST-FOLLOW -- simplify and improve MFE vitals

metal-messiah:
Simplify and improve the implementation of MFE vitals detection.  values need to be relative to the MFE script, not the page origin.  Removed extraneous metadata until actually needed for build size and simplicity reasons.  Improved consistency, and testability.  Filing as `CHORE` since this is extending the already merged #1768 work 
---

### Overview
See [this dashboard](https://staging.onenr.io/0gR7nEVxkwo) for examples of the data in dev-one.


src/common/v2/mfe-vitals.js
```diff
@@ -247,8 +193,19 @@ export function trackMFEVitals (id) {
...
+    if (!isInMFE(event?.target, id)) return
+
+    disconnectLCP()
+    interactionEvents.forEach(type => {
+      globalScope.removeEventListener(type, handleInteraction, { passive: true })
+    })
+  }
+
+  interactionEvents.forEach(type => {
+    globalScope.addEventListener(type, handleInteraction, { passive: true })
```
ptang-nr, line 276: Should these listeners also be added to `vitals.disconnect`? (in case an interaction did not occur in the associated MFE or MFE is deregistered)

src/common/v2/mfe-vitals.js
```diff
@@ -247,8 +193,19 @@ export function trackMFEVitals (id) {
       // Observer may already be disconnected
     }
   }
-  ;['click', 'keydown', 'scroll'].forEach(type => {
-    globalScope.addEventListener(type, disconnectLCP, { once: true, passive: true })
+
+  const interactionEvents = ['click', 'keydown', 'scroll']
```
ptang-nr: Question for clarity - could `'pointerdown'` be another event or it's not as appropriate in this case because `'click'` is a stronger signal of intended user interaction?  (Claude mentioned there're some timing differences between when the two events fire in the same click event loop)

src/common/v2/mfe-vitals.js
```diff
@@ -5,58 +5,12 @@
 
 import { globalScope, isBrowserScope } from '../constants/runtime'
 import { now } from '../timing/now'
-import { cleanURL } from '../url/clean-url'
+// import { cleanURL } from '../url/clean-url'
```
ptang-nr: Cleanup (?)

src/common/v2/mfe-vitals.js
```diff
@@ -174,55 +148,27 @@ export function trackMFEVitals (id) {
...
       if (isInMFE(source.node, id)) {
-        clsValue += entry.value
-        vitals.cls ??= { value: 0, largestShiftValue: null, largestShiftTime: null, largestShiftTarget: null, loadState: null }
-        vitals.cls.value = clsValue
-
-        // Track largest shift
-        if (entry.value > (vitals.cls.largestShiftValue || 0)) {
-          vitals.cls.largestShiftValue = entry.value
-          vitals.cls.largestShiftTime = entry.startTime
-          vitals.cls.largestShiftTarget = getElementSelector(source.node)
```
ptang-nr, line 191: Just curious why some of these metadata got dropped.  Are they going to come back?  (saw some tests were only commented out)

src/common/v2/mfe-vitals.js
metal-messiah: this is a fair point, pointerdown will fire off before click does across all browsers.  Even though this is just to disconnect listeners, it still may be pragmatic to hook into pointer down to avoid any races. -- ill work on that bit

src/common/v2/mfe-vitals.js
metal-messiah: No not in this case, because we dont want things like INP to stop when the user interacts with the page. FCP disconnects itself, and CLS already accounts for user input with its emitter

src/common/v2/mfe-vitals.js
metal-messiah: Yeah while we are testing this internally, much of that info was not useful.  We will revisit if these are worth adding after we get it fully validated

src/common/v2/mfe-vitals.js
metal-messiah: nevermind, i understand your question better after re-reading it.  Yes, they probably should, i'll add that.

src/common/v2/mfe-vitals.js
```diff
@@ -174,58 +147,37 @@ export function trackMFEVitals (id) {
...
-      }
+    if (!entry.interactionId || !isInMFE(entry.target, id)) return
+    if (vitals.inp.value === null || entry.duration > vitals.inp.value) {
+      // an observed "INP" means _something_ rendered for the MFE, so at minimum we can make sure all the baseline values are populated for the vitals
+      populateVitalMinimums()
+      vitals.inp.value = entry.duration
     }
   })
 
+  const interactionEvents = ['pointerdown', 'keydown', 'scroll']
```
ptang-nr: Sorry, one more thing (flagged by Claude this time) - `scroll` events do not bubble, so the target for scroll events on window is `document`.

ptang-nr: Also spotted usage of an "isTrusted" attribute in this LCP [spec](https://w3c.github.io/largest-contentful-paint/#sec-report-largest-contentful-paint) draft, which looked interesting.  Not sure if we'd consider adopting that as well.
> If target’s [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global) is a [Window](https://html.spec.whatwg.org/multipage/nav-history-apis.html#window) object, event’s [type](https://dom.spec.whatwg.org/#dom-event-type) is [scroll](https://drafts.csswg.org/cssom-view-1/#eventdef-document-scroll) and its [isTrusted](https://dom.spec.whatwg.org/#dom-event-istrusted) is true, set target’s [relevant global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)’s [has dispatched scroll event](https://w3c.github.io/largest-contentful-paint/#has-dispatched-scroll-event) to true.

Then, report LCP...
> If either of window’s [has dispatched scroll event](https://w3c.github.io/largest-contentful-paint/#has-dispatched-scroll-event) or [has dispatched input event](https://wicg.github.io/event-timing#has-dispatched-input-event) is true, return.

ptang-nr:
Some additional findings from Claude (I'm not too sure about some of these, but thought they might be "interesting")

**`touchstart` is missing.** On mobile browsers, `touchstart` fires before the pointer event sequence. Without it, a mobile tap leaves LCP running until the delayed `click` event.

**LCP has two subtle correctness problems:**

1. `getBoundingClientRect()` is called inside the MutationObserver callback — at the moment a node is *added* to the DOM. Layout hasn't necessarily been computed yet. On a freshly-inserted element (especially dynamically-rendered content), the browser may return `{ width: 0, height: 0 }` because it hasn't reflowed yet. `0 > largestSize` (which starts at 0) is false, so the element is silently skipped as a candidate. LCP falls back to the `populateVitalMinimums` baseline — which means LCP === FCP in that case. Whether that's acceptable depends on how often lazy/async renders hit this.

2. The timestamp is `now()` at mutation time, not paint time. For images, the element enters the DOM before the image loads. The real LCP timestamp should be when the image renders, not when the `` tag is inserted. This is inherent to the mutation-based approach and a known limitation.

**CLS has a windowing gap:**

The real CLS spec uses a "session window" algorithm — groups layout shifts within 1-second windows (max 5 seconds per window), takes the highest-scoring window. This implementation accumulates *all* shifts forever. For a long-lived MFE, CLS will keep growing past what the spec would report. Spec-compliant tools like Lighthouse cap it. Whether this matters depends on how `deregister()` is used in practice.

Also, `buffered: true` picks up shifts that happened before the observer was registered. A pre-registration layout shift that happened to originate from nodes now inside the MFE's DOM area would inflate CLS. This could happen if the page shifts during load before `register()` is called.

**INP's `durationThreshold: 16` is below the spec value:**

The Web Vitals spec and the `web-vitals` library use 40ms as the threshold for what counts as a measurable interaction. At 16ms you're picking up fast interactions (click responses under 40ms) that the real INP metric ignores. This would generally make INP values *lower* than the real metric, since you're averaging in fast interactions. If you want parity with what Chrome reports, 40ms is the right threshold.

**Performance concerns**

**`getBoundingClientRect()` causes layout flush.** Calling it in a MutationObserver callback — where styles may be dirty from the very mutation that fired — forces a synchronous reflow. In a hot rendering path (e.g., a list of 50 items being added), this runs once per added node.

A mitigation: debounce the LCP size check, or at minimum only call `getBoundingClientRect()` if `isInMFE` returned true first (it already does this, so that part is fine — the walk happens before the rect call). The real cost is the observer firing frequency, not the rect call per se.

**Two document-wide MutationObservers per MFE.** With `subtree: true`, every DOM mutation anywhere on the page triggers both FCP and LCP callbacks — even after FCP disconnects itself (only the LCP observer persists). For a SPA with heavy rendering, every React/Vue render cycle fires the LCP callback. Each callback walks the DOM upward to find a `data-nr-mfe-id` attribute. Multiple registered MFEs multiply this: 3 MFEs = 3 LCP observers all doing ancestor walks on every mutation.

src/common/v2/mfe-vitals.js
```diff
@@ -125,107 +98,138 @@ const observePerformance = (observers, config, onEntry) => {
...
+  observePerformance(observers, { type: 'event', buffered: true, durationThreshold: 40 }, (entry) => {
+    if (!entry.interactionId || !isInMFE(entry.target, id)) return
+    if (vitals.inp.value === null || entry.duration > vitals.inp.value) {
+      // an observed "INP" means _something_ rendered for the MFE, so at minimum we can make sure all the baseline values are populated for the vitals
+      populateVitalMinimums()
+      vitals.inp.value = entry.duration
     }
   })
 
+  const interactionEvents = ['pointerdown', 'keydown']
```
ptang-nr, line 237: Asking just to clarify - are we explicitly not including 'scroll' as a qualifying user interaction for MFE vitals?  (Asking only because 'scroll' is technically a qualifying user interaction for web vitals, as long as it's trusted)

src/common/v2/mfe-vitals.js
```diff
@@ -125,107 +98,138 @@ const observePerformance = (observers, config, onEntry) => {
...
-          elUrl: getElementURL(elem)
+    // an observed "LCP" means _something_ rendered, so at minimum we can make sure all the baseline values are populated for the vitals
+    populateVitalMinimums()
+
+    if (resizeObs) {
+      try {
+        const elem = node.nodeType === 1 ? node : node.parentElement
+        if (elem && !observedElements.has(elem)) {
+          observedElements.add(elem)
+          resizeObs.observe(elem)
```
ptang-nr: For every mutated node, they're submitted to be observed by RO for their size.  It looks like for LCP, we should track an element's first rendered size.  

Here we're essentially observing from when an MFE is registered until the lifetime of page.  Should we stop observing sooner?

src/common/v2/mfe-vitals.js
```diff
@@ -125,107 +98,138 @@ const observePerformance = (observers, config, onEntry) => {
...
+          if (size > largestSize) {
+            largestSize = size
+            const elem = entry.target
+            // For img/video elements, wait for load event
+            if (['IMG', 'VIDEO'].includes(elem.tagName)) {
+              if (elem.complete) {
+                lcpObservedAt = now()
+              } else {
+                elem.addEventListener('load', () => {
+                  lcpObservedAt = now()
```
ptang-nr: For async images, would we need to grab the size on load?

ptang-nr: Also, Claude alerted me that `load` and `complete` won't work for HTMLVideoElement.

src/common/v2/mfe-vitals.js
```diff
@@ -247,8 +252,16 @@ export function trackMFEVitals (id) {
       // Observer may already be disconnected
     }
   }
-  ;['click', 'keydown', 'scroll'].forEach(type => {
-    globalScope.addEventListener(type, disconnectLCP, { once: true, passive: true })
+
+  const handleInteraction = (event) => {
+    if (!isInMFE(event?.target, id)) return
+
+    disconnectLCP()
```
ptang-nr, line 271: `disconnectLCP` also needs to disconnect `resizeObs`

src/common/v2/mfe-vitals.js
metal-messiah: yeah i think it has potential to create too much noise and processing burden in the observer, so i think we should just omit it for now

src/common/v2/mfe-vitals.js
metal-messiah: I'm not sure the right answer here.  "real" LCP only stops when the user interacts with the page or the page's visibility is changed. good call out on the resizing bit, let me look into that.

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1801/commits/adfcd1e014d0e2ae1092e87e115ed166c1a7011b

src/common/v2/mfe-vitals.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1801/commits/67d4afef6c85ed1ce91989d94e9824b9cd9d985f

src/common/v2/mfe-vitals.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1801/commits/67d4afef6c85ed1ce91989d94e9824b9cd9d985f

src/common/v2/mfe-vitals.js
```diff
@@ -155,22 +155,8 @@ export function trackMFEVitals (id, timings) {
...
-                  lcpObservedAt = now()
-                  resizeObs.unobserve(elem)
-                }, { once: true })
-              }
-            } else {
-              lcpObservedAt = now()
-              resizeObs.unobserve(elem)
-            }
+            lcpObservedAt = now()
+            resizeObs.unobserve(entry.target)
```
ptang-nr: I think technically this un-observes only the largest elements.  Should be ok to un-observe all elements at this point.

ptang-nr (dismissed):

ptang-nr (approved):

PR #1803, opened 2026-06-29
feat: Add websocket init config

ellisong:
Adds `web_sockets` block to NREUM init configuration. Setting `web_sockets.enabled` to `true` turns on the web sockets feature. The web sockets feature is disabled by default.
---

### Overview

Adds `web_sockets` configuration in the init block to turn on the web sockets feature. It can also be turned on through the `websockets` feature flag.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-553178

### Testing

Make sure tests pass, especially the library tests.


metal-messiah:
Can you also update the flag/init for
- [x] `publish-experiment.yml`
- [x] `internal-promotion > released.js` <-- I think we should be ok to expose this to NR1 at this point

metal-messiah (approved):

PR #1809, opened 2026-07-09
fix: Keep Ajax payload out of unrelated events

ptang-nr:
Drops captured AJAX request + payload info from unrelated events (AJAX timeslice metrics, session trace) via a shared `params` object.

---

### Overview

In this PR, we remove assignments to the `params` object inside `storeXHR`.  It was possibly a legacy mechanism for passing info to the old SPA implementation. Verified these fields are not pulled from `params` in the consumer.  The consumer should be pulling these fields from the AjaxNode/events directly.

### Related Issue(s)

[NR-582727](https://new-relic.atlassian.net/browse/NR-582727)

### Testing

Added component tests for the shape of params in timeslice metrics + session trace, as well as the Ajax event.

Params:
```
['method', 'status', 'host', 'hostname', 'pathname']
```

Ajax:
```
['method', 'status', 'domain', 'path', 'requestSize', 'responseSize', 'type', 'startTime', 
'endTime', 'callbackDuration', 'ajaxRequest.id', 'gql', 'requestQuery', 'requestHeaders', 
'responseHeaders', 'requestBody', 'responseBody']
```

[NR-582727]: https://new-relic.atlassian.net/browse/NR-582727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/features/ajax/aggregate/index.js
+++ b/src/features/ajax/aggregate/index.js
@@ -102,22 +102,22 @@ export class Aggregate extends AggregateBase {
       [AJAX_ID]: generateUuid() // all AjaxRequest events should have a unique identifier to allow for easier grouping and analysis in the UI
     }
 
-    event.gql = params.gql = parseGQL({
+    event.gql = parseGQL({
       body: ctx.requestBody,
       query: ctx.parsedOrigin?.search
     })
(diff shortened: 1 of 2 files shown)


ellisong (approved):
LGTM

PR #1810, opened 2026-07-09
feat: Do not obfuscate attribute keys

ellisong:
Do not obfuscate attribute keys in our addCustomAttributes method for the BEL serializer. This change affects attributes in AjaxRequest, PageViewTiming, and BrowserInteraction (soft nav only) events.

---

### Overview

We decided to not obfuscate attribute keys in our BEL serializer; we still obfuscate the attribute values. It was too easy and destructive with our obfuscation rules to corrupt the event attribute keys for AjaxRequest, PageViewTiming, and BrowserInteraction events. This PR adds a safeguard against that, assuming that users won't put sensitive data in their attribute keys.

### Related Issue(s)

JIRA: https://new-relic.atlassian.net/browse/NR-582722

### Testing

Make sure all tests pass. It's possible that a few e2e tests might be indirectly impacted.

### DIFF
--- a/src/common/serialize/bel-serializer.js
+++ b/src/common/serialize/bel-serializer.js
@@ -27,9 +27,9 @@ export function getAddStringContext (obfuscator, truncator) {
 
   return addString
 
-  function addString (str) {
+  function addString (str, obfuscate = true) {
     if (typeof str === 'undefined' || str === '') return ''
-    str = obfuscator?.obfuscateString(String(str)) ?? String(str)
+    str = obfuscate ? (obfuscator?.obfuscateString(String(str)) ?? String(str)) : String(str)
     str = truncator?.(str) ?? str
     if (hasOwnProp.call(stringTable, str)) {
       return numeric(stringTable[str], true)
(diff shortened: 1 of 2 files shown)


src/common/serialize/bel-serializer.js
```diff
@@ -27,9 +27,9 @@ export function getAddStringContext (obfuscator, truncator) {
 
   return addString
 
-  function addString (str) {
+  function addString (str, obfuscate = true) {
     if (typeof str === 'undefined' || str === '') return ''
-    str = obfuscator?.obfuscateString(String(str)) ?? String(str)
+    str = obfuscate ? (obfuscator?.obfuscateString(String(str)) ?? String(str)) : String(str)
```
metal-messiah, line 32: This is just a suggestion, feel free to ignore at your discretion -- does something like this feel easier to read than calling String(str) 3 times in the same ternary?
```suggestion
    str = String(str)
    if (obfuscate && obfuscator) str = obfuscator.obfuscateString(str)
```

metal-messiah (approved):

src/common/serialize/bel-serializer.js
ellisong: Let's go with what we have and we can clean the logic surrounding it later.

PR #1812, opened 2026-07-14
fix: Avoid capturing agent AJAX payloads

ptang-nr:
Prevent the AJAX `capture_payloads` feature from capturing the agent's own telemetry requests (harvests to NR beacon endpoints). 
---

### Overview

In this PR, `canCapturePayload` was refactored to accept an object and a `beacons` array to help detect and avoid capturing the agent's own AJAX payloads. 

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-582726

### Testing

Updated related unit + component tests.


metal-messiah (changes requested):
I think the code changes make sense.  Can you add an E2E that validates that agent beacon activity is not captured to round out the test robustness?

metal-messiah (approved):

PR #1813, opened 2026-07-16
feat: MFE vitals improvements

metal-messiah:
Implement quality of life improvements on the collection of MFE vital timings. 
- If no FCP value is observed within 10 seconds, shut down observers for memory saving and outlier control. 
- Decorate observed DOM trees with the MFE instance ID to facilitate debugging and prevent double-instrumentation of the same DOM tree from multiple instances pointed at the same MFE target.
---

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-597354


ellisong (approved):
LGTM if tests pass

PR #1815, opened 2026-07-16
chore: build size fix + improvements

metal-messiah:
The existing build size comment CI is not working correctly.  This gives an opportunity to both fix and improve it.
---


ellisong (approved):
LGTM

PR #1817, opened 2026-07-21
fix: Avoid obfuscating keys and reserved system vals

ptang-nr:
Preserves all keys as well as reserved system values for AjaxRequest and Page View Timing events.

---

### Overview

System fields from events that use the BEL serializer were previously passed through the user-configured obfuscator. This is now avoided by
- introducing a separate `addStringRaw` serializer that bypasses obfuscation for all keys, and 
- refactoring the `addCustomAttributes` function to accept `{ addKey, addVal }` serializers.  This creates more flexibility for callers to control which attributes can be preserved as system fields.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-584189

### Testing

Extended component tests to ensure all affected event fields are tested.

### DIFF
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -148,7 +148,7 @@ export class Aggregate extends AggregateBase {
   // serialize array of timing data
   serializer (eventBuffer) {
     if (!eventBuffer?.length) return ''
-    const { addStringRaw } = createStringAdders(getAddStringContext, this.obfuscator)
+    const { addString, addStringRaw } = createStringAdders(getAddStringContext, this.obfuscator)
     const obfuscateOnly = (val) => typeof val === 'string' ? (this.obfuscator?.obfuscateString(val) ?? val) : val
 
     var payload = 'bel.6;'
@@ -162,9 +162,11 @@ export class Aggregate extends AggregateBase {
 
       this.appendGlobalCustomAttributes(timing, { addVal: obfuscateOnly })
 
-      var attrParts = addCustomAttributes(timing.attrs, {
+      const { pageUrl, ...systemReservedAttrs } = timing.attrs || {}
+      var attrParts = addCustomAttributes(systemReservedAttrs, {
         addKey: addStringRaw, addVal: addStringRaw
       })
+      if (pageUrl) attrParts.push(...addCustomAttributes({ pageUrl }, { addKey: addStringRaw, addVal: addString }))
       if (attrParts && attrParts.length > 0) {
         payload += numeric(attrParts.length) + ';' + attrParts.join(';')
       }
--- a/src/features/page_view_timing/aggregate/index.js
+++ b/src/features/page_view_timing/aggregate/index.js
@@ -23,6 +23,7 @@ import { loadTime } from '../../../common/vitals/load-time'
 import { webdriverDetected } from '../../../common/util/webdriver-detection'
 import { cleanURL } from '../../../common/url/clean-url'
 import { EVENT_TYPES } from '../../../common/constants/events'
+import { createStringAdders } from '../../../common/payloads/payloads'
 
 export class Aggregate extends AggregateBase {
   static featureName = FEATURE_NAME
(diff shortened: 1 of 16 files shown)


src/features/soft_navigations/aggregate/interaction.js
```diff
@@ -140,6 +140,7 @@ export class Interaction extends BelNode {
   serialize (firstStartTimeOfPayload, agentRef, interactionObfuscator, ajaxObfuscator) {
     const isFirstIxnOfPayload = firstStartTimeOfPayload === undefined
     const addString = getAddStringContext(interactionObfuscator)
+    const addStringRaw = (str) => addString(str, false)
```
cwli24: Is there a reason this isn't just
```suggestion
const { addString, addStringRaw } = createStringAdders(getAddStringContext, interactionObfuscator)
```
?

src/features/page_view_timing/aggregate/index.js
```diff
@@ -146,20 +148,25 @@ export class Aggregate extends AggregateBase {
   // serialize array of timing data
   serializer (eventBuffer) {
     if (!eventBuffer?.length) return ''
-    var addString = getAddStringContext(this.obfuscator)
+    const { addString, addStringRaw } = createStringAdders(getAddStringContext, this.obfuscator)
+    const obfuscateOnly = (val) => typeof val === 'string' ? (this.obfuscator?.obfuscateString(val) ?? val) : val
```
cwli24: There's a bug here in that obfuscation happens first then serializing into string, but it doesn't account for object-type values. For example, consider a config like this:
```javascript
init: {
  obfuscate: [{ regex: /john\.doe/g, replacement: '***' }]
},
info: {
  jsAttributes: { user: { name: 'john.doe', role: 'admin' } }
}
```
`appendGlobalCustomAttributes` below is going to run:
`timingAttributes['user'] = obfuscateOnly({ name: 'john.doe', role: 'admin' })` which is going to bypass the obfuscate rule,
and proceed to `addCustomAttributes` with `addVal: addStringRaw`.

cwli24: I think the complication here is merging the system attributes and the user custom attributes (2 streams) into 1 (`timing.attrs`) which applies an uniform `addVal: addStringRaw` to everything.

One approach is to separate them:
1. `appendGlobalCustomAttributes`, instead of modifying `timing.attrs` directly / adding jsAttributes to it, becomes `getGlobalCustomAttributes` that returns an object of user attributes instead:
```js
#getGlobalCustomAttributes () {
    var reservedAttributes = ['size', 'eid', 'cls', 'type', 'fid', 'elTag', 'elUrl', 'net-type',
      'net-etype', 'net-rtt', 'net-dlink']
    return Object.fromEntries(
      Object.entries(this.agentRef.info.jsAttributes || {}).filter(([key]) => !reservedAttributes.includes(key))
    )
  }
```
2. the awkward one-off of adding `webdriverDetected` in the original func can be simply moved to `addTiming()` instead:
```js
attrs.webdriverDetected = webdriverDetected
```
3. lastly the part below can just become:
```js
      const userAttrs = this.#getGlobalCustomAttributes()

      const { pageUrl, ...systemReservedAttrs } = timing.attrs || {}
      var attrParts = addCustomAttributes(systemReservedAttrs, {
        addKey: addStringRaw, addVal: addStringRaw
      })
      attrParts.push(...addCustomAttributes(userAttrs, { addKey: addStringRaw, addVal: addString }))
```

cwli24: this handles both group of attributes accordingly, solves the stringify-obfuscate sequence, and conforms with how ajax obfuscation happens.

ptang-nr: Thanks for the suggested fixes.  Looks a lot better 👍 

Initially I got confused about whether `jsAttributes` would have object vals.  I did check [setCustomAttribute](https://docs.newrelic.com/docs/browser/new-relic-browser/browser-apis/setcustomattribute/#parameters) docs.  It stated "Custom attribute values cannot be complex objects, only simple types such as Strings, Integers and Booleans."  I just assumed that applied to attributes preset by customers 🫠

metal-messiah:
This looks fine to me 👍 -- @cwli24 when your comments are resolved in a satisfactory way I am good with this PR as well.

metal-messiah (approved):

cwli24 (approved):

PR #1820, opened 2026-07-27
fix: Deduplicate auto-instrumented MFE events

metal-messiah:
Fixes duplicate reporting of auto-instrumented events (AJAX requests, JS errors, console logs, WebSocket activity etc.) when a micro-frontend is registered with the `register()` API multiple times. Previously, a single auto-instrumented event originating from a repeatedly-registered MFE script would be reported once per registration — Auto-detected events are now collapsed to a single report per real occurrence when they resolve to the same underlying script and the same registered MFE `id`. Explicit API-driven calls (`.addPageAction()`, `.log()`, `.noticeError()`, etc.) and `MicroFrontEndTiming` events remain per-registration, as intended, and distinct MFEs that happen to share a script (e.g. two different MFEs registered from the same inline `<script>` block) continue to be reported independently.

---

### Overview

Customer teams building on top of the `register()` API (micro-frontend support) were calling `newrelic.register()` once per rendered component instance instead of once per MFE bundle (e.g. one nerdpack rendering 30 visualizations registered 30 times). Because auto-instrumentation attributes browser events to MFEs by matching JS call-stack URLs against each registered entity's resolved script asset (`getRegisteredTargetsFromFilename` / `findTargetsFromStackTrace` in `src/common/v2/utils.js`), every one of those 30 duplicate registrations independently matched the same real AJAX call, error, console log, or WebSocket event — producing 30 duplicate events instead of 1. This showed up as `ERR_INSUFFICIENT_RESOURCES` network flooding for at least one customer.

This PR adds a dedup step at the point auto-detection resolves matching registered targets:
- `dedupeRegisteredEntitiesByAsset` (new, in `src/common/v2/utils.js`) collapses registered entities down to one canonical entity per unique `(resolved script asset, target.id)` pair, preferring a non-deregistered target as canonical. Entities with an unresolved asset (e.g. inline scripts that couldn't be correlated) are never collapsed, since there's no signal they represent the same script. Entities that share a script but have *different* `id`s are also never collapsed, since that means the customer registered genuinely distinct MFEs (e.g. two different MFEs both registered from the same inline `<script>` block) — each should still receive its own copy of matched events.
- `dedupeTargetsByInstance` (new, same file) is a defensive second pass in `findTargetsFromStackTrace` that collapses duplicate targets that could re-enter the array via multiple matched stack frames.

Because both fan-out paths (`findTargetsFromStackTrace` for AJAX/WebSockets/logs, and jserrors' `#deriveTargets`) bottom out in `getRegisteredTargetsFromFilename`, this is a single choke-point fix — no changes were needed to any `wrap-*`, feature aggregate, or `register.js` code. Explicit register-API calls (`.addPageAction()`, `.log()`, `.noticeError()`, `.measure()`, `.recordCustomEvent()`) and `MicroFrontEndTiming` events bypass this matching entirely (they always carry their own explicit target) and are unaffected — they remain per-registration, as required.

### Related Issue(s)

- https://new-relic.atlassian.net/browse/NR-586577
- https://new-relic.atlassian.net/browse/NR-586579

### Testing

**Unit tests** (`tests/unit/common/util/v2.test.js`):
- `npm run test:unit -- v2.test.js`
- New coverage for `dedupeRegisteredEntitiesByAsset` and `dedupeTargetsByInstance`: same-asset+id collapse, blocked/non-blocked canonical selection, unresolved-asset entities never collapsed, and — critically — entities that share an asset but have *different* ids are never collapsed (regression guard for the "two distinct MFEs registered from the same inline script" scenario).
- New case on `getRegisteredTargetsFromFilename` simulating 30 duplicate registrations of the same MFE collapsing to 1 target, and a companion case mirroring `register-api.html` (two different MFE ids sharing an inline script) confirming both remain distinct.

**E2E test** (`tests/specs/api/register/general.e2e.js`, new fixture `tests/assets/js/mfe/mfe-dedup.js`):
- `npm run wdio -- --spec tests/specs/api/register/general.e2e.js`
- The fixture registers the same MFE `id` 30 times from one script, then triggers one auto-instrumented AJAX call, one auto-captured console log, one uncaught error, and 30 explicit `.log()` calls (one per registration).
- Test asserts each auto-detected event type is reported exactly once, while the 30 explicit log calls and 30 `MicroFrontEndTiming` events remain per-registration.

**Regression coverage:**
- Full unit + component suite: `npm run test:unit && npm run test:component` (1414 tests passing).
- Existing `tests/specs/api/register/flags.e2e.js` (`Base.with register`) continues to pass — this test specifically encodes the "two different MFEs sharing a script both get their own copy of matched events" behavior, which this fix preserves via the `id`-aware dedup key.
- Existing `tests/specs/api/register/auto-detection-*.e2e.js` suite (AJAX, errors, logs, WebSockets, user actions) continues to pass, confirming multi-script MFE attribution is unaffected.

### DIFF
--- a/src/common/v2/utils.js
+++ b/src/common/v2/utils.js
@@ -30,6 +30,13 @@ export function getRegisteredTargetsFromId (id, agentRef) {
 
 /**
  * Returns the registered target(s) associated with a given filename if found in the resource timing API during registration. Returns an empty array if no target is found.
+ * Multiple registrations that resolve to the same underlying script asset AND represent the same logical MFE (i.e.
+ * share the same customer-supplied `target.id`) are collapsed to a single target via
+ * {@link dedupeRegisteredEntitiesByAsset}, since auto-instrumented events (AJAX, JS errors, logs, WebSockets) should
+ * only be reported once per real occurrence when the same MFE was registered many times over -- not once per
+ * duplicate registration. Distinct MFEs (different `target.id`) that happen to share a script (e.g. two different
+ * MFEs both registered from the same inline `<script>` block) are intentionally NOT collapsed, since each is a
+ * genuinely different entity that should still receive its own copy of the matched event.
  * @param {string} filename
  * @param {*} agentRef
  * @returns {import("../../interfaces/registered-entity").RegisterAPIMetadataTarget[]}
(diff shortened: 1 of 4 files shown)


src/common/v2/utils.js
```diff
@@ -104,8 +156,30 @@ export function findTargetsFromStackTrace (agentRef) {
...
+
+/**
+ * Removes duplicate targets from an array, keyed by `target.instance`. This guards against the same canonical
+ * target re-entering the array via multiple matched stack-frame URLs (e.g. a recursive call whose stack contains the
+ * same file at multiple depths). Entries with no `instance` (i.e. `undefined`, meaning "the container agent")
+ * naturally collapse to a single entry too, which is the desired behavior.
+ * @param {Array} targets
+ * @returns {Array} deduped list of targets, preserving relative order of first occurrence
+ */
+export function dedupeTargetsByInstance (targets) {
```
ellisong, line 173: This can be rewritten as the following if it's easier to read/parse:
```
export function dedupeTargetsByInstance (targets) {
  const seen = new Set()
  return targets.filter(target => {
    const key = target?.instance
    if (seen.has(key)) return false
    seen.add(key)
    return true
  })
}
```

I'm fine either way.

ellisong (approved):
LGTM, see optional comment

PR #1823, opened 2026-07-29
feat: Improve ability to apply reasonable caps to MFE vitals

metal-messiah:
Fixed a bug in the Register API's micro-frontend (MFE) vitals tracking where a script's timing correlation could be reused after it went stale, causing First Contentful Paint (and other vitals) to be misreported as up to several seconds slower than they actually were for MFEs that remount without their underlying script reloading (e.g. in single-page applications).
---

### Overview
ScriptCorrelation objects created by the script tracker are keyed only by script URL and are never cleared. If the same script URL registers an MFE more than once in a session (e.g. an SPA remounting the same MFE without the script actually reloading, or a shared bundle registering multiple MFEs), findScriptTimings() would reuse the original correlation's DOM/performance timings for the new registration. Since scriptStart anchors the FCP/LCP vitals calculation, this made a fast remount report FCP as the time elapsed since the original page load — often several seconds — instead of the near-instant paint that actually occurred.

This PR fixes the issue by tracking which MFE ids have already claimed a given correlation (ScriptCorrelation.claimedBy). When the same MFE id claims a correlation again and the elapsed time since the correlation's recorded start exceeds a new CORRELATION_STALE_THRESHOLD_MS (10s) constant, the correlation is treated as stale and scriptStart/scriptEnd fall back to the registration's own registeredAt timestamp instead. mfe-vitals.js's FCP give-up timeout was also updated to reuse this same constant instead of a separate hardcoded 10000.

### Related Issue(s)
This is a followup to #1813 in that it didnt fully solve the issue.

### Testing
Added a unit test suite (tests/unit/common/util/script-tracker-correlations.test.js) covering:

A correlation claimed by the same MFE id past the staleness threshold falls back to registeredAt for scriptStart/scriptEnd.
A correlation claimed for the first time by a different MFE id is not treated as stale, even past the threshold (shared-bundle scenario).
Added an e2e test (tests/specs/api/register/script-correlation-remount.e2e.js) using a new asset (tests/assets/js/mfe/mfe-remount.js) that:

Loads a real external MFE script that registers an MFE (remount-mfe) on load.
Captures and validates the first mount's FCP.
Waits past the 10s staleness window without the script reloading.
Triggers a second register() call for the same MFE id from the same script (simulating an SPA remount).
Asserts the second mount's FCP is reported as the near-instant paint that just happened (< 1000ms), not the ~10s+ gap since the original load.


tests/assets/js/mfe/mfe-remount.js
```diff
@@ -0,0 +1,33 @@
+// Simulates an SPA that mounts an MFE, then remounts the SAME MFE (same id) much later
```
ellisong: This is a very minor nitpick throughout the code, but is it normal to use `an SPA` and `an MFE` instead of `a SPA` and `a MFE`?

tests/unit/common/util/script-tracker-correlations.test.js
```diff
@@ -529,6 +529,107 @@ describe('script-tracker correlations', () => {
...
+      test('discards a correlation reused by the same MFE past the staleness threshold', () => {
+        const scriptUrl = 'https://cdn.example.com/remount.js'
+        mockNavigationEntry = { name: 'https://example.com/' }
+        mockStack = `Error
+    at findScriptTimings (${scriptTrackerModule.thisFile}:1:1)
+    at init (${scriptUrl}:10:5)`
+        const mfeTarget = { id: 'mfe-a', instance: 'instance-1' }
+
+        const perfEntry = { name: scriptUrl, initiatorType: 'script', startTime: 10, responseEnd: 90 }
+        globalThis.performance.getEntriesByType = jest.fn((type) => {
```
ellisong, line 541: is the `getEntriesByType` function reusable between tests?

tests/unit/common/util/script-tracker-correlations.test.js
```diff
@@ -529,6 +529,107 @@ describe('script-tracker correlations', () => {
...
+        // The stale correlation must be discarded in favor of this registration's own timestamp
+        expect(secondTimings.scriptStart).toBe(secondTimings.registeredAt)
+        expect(secondTimings.scriptEnd).toBe(secondTimings.registeredAt)
+
+        // FCP observed shortly after the remount must be reported relative to *this*
+        // registration, not the original page load ~10s+ ago.
+        const fcpObservedAt = secondTimings.registeredAt + 50
+        const reportedFcp = fcpObservedAt - (secondTimings.scriptStart || secondTimings.registeredAt)
+        expect(reportedFcp).toBe(50)
+        expect(reportedFcp).toBeLessThan(10000)
```
ellisong: This line `expect(reportedFcp).toBeLessThan(10000)` doesn't make sense since it will fail beforehand if the value isn't `50`.

ellisong (commented):
see comments

src/common/v2/script-tracker.js
```diff
@@ -246,12 +250,24 @@ export function findScriptTimings () {
...
+    // timings still describe the *original* load, not this one. Detect that case so scriptStart/scriptEnd
+    // below can ignore the stale data instead of reporting it as if it were fresh.
+    const correlation = timings.correlation
+    const alreadyClaimedByThisMFE = !!mfeId && !!correlation?.claimedBy.has(mfeId)
+    if (correlation && mfeId) correlation.claimedBy.add(mfeId)
+    const isCorrelationStale = () => {
+      const correlationStart = correlation?.script.start
+      if (!alreadyClaimedByThisMFE || !correlationStart) return false
+      const staleness = timings.registeredAt - correlationStart
+      return staleness > CORRELATION_STALE_THRESHOLD_MS
```
ptang-nr, line 279: What happens for re-mounts that happen under 10s?  Would this mean MFE vitals can still be inflated by up to 10s?

tests/assets/js/mfe/mfe-remount.js
metal-messiah: `an` or `a` is ok for `MFE`, `an` is a typo for SPA (imo)

tests/unit/common/util/script-tracker-correlations.test.js
metal-messiah: Theres a global beforeEach mock at the top of the file that mocks the whole performance object for sharing in each test.  These are specific mock overrides just for that test

tests/assets/js/mfe/mfe-remount.js
metal-messiah:
```suggestion
// Simulates a SPA that mounts a MFE, then remounts the SAME MFE (same id) much later
```

tests/unit/common/util/script-tracker-correlations.test.js
metal-messiah: 0b149a0b0

src/common/v2/script-tracker.js
metal-messiah: Thats correct, there will still be variance, but the variance will be hard capped at 10s, to alleviate wild spikes while still ensuring the data can be reported well outside the recommended thresholds for vital data.

ptang-nr: Gotcha.  Initially, I was wondering if we can get away with only checking already claimed (not sure why 10s, or if that could fluctuate between customers) but sounds like it's used to weed out extreme values 🤔

ptang-nr (dismissed):
LGTM 👍 if Grady's ok too

ellisong (approved):
LGTM after tests pass

PR #53, opened 2021-05-20
Changelog and version update for release 1209

aubreymasten:
### Overview
Updated changelog and bumped version number for agent version 1209.

PR #1824, opened 2026-08-05
feat: Introduce v2 PageView & decouple connect service call

cwli24:
The old RUM request responsible for both retrieving app configurations and `PageView` generation is now split (temporarily under a feature flag). This makes the agent more resilient on problematic connect cycles or unsuccessful config retrieval, and the `PageView` event would no longer be the dependency blocking other events from harvesting.
---

### Overview

**This new behavior is gated by the `rum_v2` feature flag in this PR. Both versions of the code exists for now in the build.**

### Bootstrap / lifecycle

- Introduced a shared, one-time-per-agent `ensureRuntimeBootstrap` in _instrument-base.js_ that sets up session, Connector, and Harvester. AggregateBase#doOnceForAllAggregate was removed.
- Harvester's EOL (visibilitychange) listener relocated from constructor to `startTimer()`, called only after every feature has finished attempting its aggregate init. (This way, it always registers after feature-level visibilitychange listeners, e.g. web-vitals.)
- A new Connector unit handles the connect call with its own retry (up to 3 attempts, exponential backoff) and abort.
- Failure to construct Connector and Harvester now aborts the whole agent. Session-setup failure alone remains non-fatal and only warns.

### PVE v2

- Under `rum_v2`, PVE instrument loads a new aggregate_v2 module instead of the old aggregate. v2 shares the same harvest & retry logic as other features rather than v1's special-cased RUM call. A failed POST will not affect other features.
- v2 drops query params tt/us/ac/pr/xx/ua and fsh. It adds igp, sourced from the connect response.
- `activateFeatures`'s drained lifecycle event flag is now conditional on rum_v2 (false under v2, since drain would not have happened yet when all features await connect response).

### Misc/cleanup

- Removed the deprecated `ck` query param from all harvests.
- _send.js_: omits the trailing ? when a harvest has no query string; submitMethod now defaults via `getSubmitMethod()`; request headers can now be extended per-call via `localOpts.headers` (used by Connector's retry headers).
- Session Replay feature: added an abort handler to stop any in-progress preload recording if its aggregate never successfully loads; Recorder now guards against triggering an early harvest before its aggregate exists.
- _consent.js_: `pveAgg.sendRum?.()` — guarded since only the v1 aggregate has `sendRum`.

https://newrelic.atlassian.net/wiki/spaces/TDP/pages/5853020595/Request+Response+Contract+CDD+Connect+RUM+Harvest+v2+endpoints

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-539835
https://new-relic.atlassian.net/browse/NR-539865
https://new-relic.atlassian.net/browse/NR-539872
https://new-relic.atlassian.net/browse/NR-585094

### Testing

Bunch modified & added for jests & e2e

### DIFF
--- a/docs/warning-codes.md
+++ b/docs/warning-codes.md
@@ -143,3 +143,5 @@
 `A session replay payload failed to send and is being retried. Recording is paused during the retry period, and will resume when a successful harvest is made. Some replay activity may be missed during retry phases.`
 ### 71
 `An invalid feature mode was detected and set to "off".`
+### 72
+`Unable to initialized Connector and/or Harvester.`
--- a/docs/warning-codes.md
+++ b/docs/warning-codes.md
@@ -106,7 +106,7 @@
 ### 52
 `Unexpected problem encountered. There should be at least one app for harvest!`
 ### 53
-`Did not receive a valid entityGuid from connection response`
+`Failed to parse connect response.`
 ### 54
 `An experimental feature is being used. Support can not be offered for issues`
 ### 55
--- a/docs/warning-codes.md
+++ b/docs/warning-codes.md
@@ -144,4 +144,4 @@
 ### 71
 `An invalid feature mode was detected and set to "off".`
 ### 72
-`Unable to initialized Connector and/or Harvester.`
+`Unable to initialize Connector and/or Harvester.`
--- a/src/common/harvest/connector.js
+++ b/src/common/harvest/connector.js
@@ -1,0 +1,159 @@
+/**
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { isBrowserScope } from '../constants/runtime'
+import { send } from './send'
+import { TimeKeeper } from '../timing/time-keeper'
+import { activateFeatures } from '../util/feature-flags'
+import { now } from '../timing/now'
+import { warn } from '../util/console'
+import { VERSION } from '../constants/env'
+import { FEATURE_NAMES, FEATURE_TO_ENDPOINT } from '../../loaders/features/features'
+import { handle } from '../event-emitter/handle'
+import { SUPPORTABILITY_METRIC_CHANNEL } from '../../features/metrics/constants'
+
+const MAX_CONNECT_ATTEMPTS = 3
+const CONNECT_RETRY_BASE_MS = 3000 // affects each retry window proportionally (attempt 2: 0-3s, 3: 0-6s, 4: 0-12s, etc)
+
+export class Connector {
+  #agentRef
+  #connectStartTime
+  #numOfAttempts = 0
+
+  constructor (agentRef) {
+    if (!agentRef.init.feature_flags.includes('rum_v2')) return // the Connector does not run in v1; legacy PVE has its own BCS handling
+    this.#agentRef = agentRef
+    agentRef.runtime.timeKeeper = new TimeKeeper(agentRef.runtime.session)
+
+    if (isBrowserScope) {
+      const cached = agentRef.runtime.session && agentRef.runtime.session.state.cachedRumResponse
+      if (cached) {
+        if (cached.config) { // the old v1 RUM response does not have a config field, so this guards against breaking change on existing sessions at release time
+          this.#applyConnectResponse(cached)
+          return
+        } else this.#agentRef.runtime.session.reset() // this can be removed on the next version after rls, once the old format and sessions have expired or reset
+      }
+    }
+    this.makeConnectRequest()
+  }
+
+  makeConnectRequest () {
+    this.#connectStartTime = now()
+    return send(this.#agentRef, {
+      endpoint: `connect/2/${this.#agentRef.info.licenseKey}`,
+      payload: {
+        body: {},
+        qs: {
+          a: this.#agentRef.info.applicationID,
+          v: VERSION
+        }
+      },
+      localOpts: { sendEmptyBody: true },
+      raw: true,
+      featureName: 'connect',
+      cbFinished: this.#processConnectResponse.bind(this)
+    })
+  }
+
+  #applyConnectResponse (response) {
+    if (!Object.keys(this.#agentRef.runtime.appMetadata).length) this.#agentRef.runtime.appMetadata = response.app
+    activateFeatures(response.config, this.#agentRef)
+  }
+
+  #processConnectResponse ({ sent, status, retry, xhr, responseText }) {
+    const connectEndTime = now()
+    const session = this.#agentRef.runtime.session
+
+    if (session) {
+      // Double check that there isn't a response saved (e.g. from another agent/browser tab) btwn the async time of the request and now in a race-condition.
+      const cachedResp = session.state.cachedRumResponse
+      if (cachedResp) {
+        this.#applyConnectResponse(cachedResp)
+      }
+    }
+
+    const shouldRetry = sent && retry // important to note: no retry if HTTP status is 0
+    if (shouldRetry && ++this.#numOfAttempts < MAX_CONNECT_ATTEMPTS) {
+      // We use an exponential backoff calculation for the retry to avoid thundering herd, skipping any further processing.
+      const delay = Math.floor(Math.random() * CONNECT_RETRY_BASE_MS * (2 ** (this.#numOfAttempts - 1)))
+      setTimeout(this.makeConnectRequest.bind(this), delay)
+      return
+    }
+
+    if (status >= 400 || status === 0) {
+      warn(18, status)
+
+      // Get estimated payload size of our backlog.
+      const textEncoder = new TextEncoder()
+      const payloadSize = Object.values(newrelic.ee.backlog).reduce((acc, value) => {
+        if (!value) return acc
+        const encoded = textEncoder.encode(value)
+        return acc + encoded.byteLength
+      }, 0)
+
+      // Send SMs about the (last) failed connect request, then abort agent.
+      const BCSError = 'BCS/Error/'
+      const sm = [{
+        params: {
+          name: BCSError + status
+        },
+        stats: {
+          c: 1
+        }
+      }, {
+        params: {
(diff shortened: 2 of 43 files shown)


src/common/harvest/connector.js
```diff
@@ -0,0 +1,164 @@
...
+
+  #processConnectResponse ({ sent, status, retry, xhr, responseText }) {
+    const connectEndTime = now()
+    const session = this.#agentRef.runtime.session
+
+    if (session) {
+      // Double check that there isn't a response saved (e.g. from another agent/browser tab) btwn the async time of the request and now in a race-condition.
+      const cachedResp = session.state.cachedRumResponse
+      if (cachedResp) {
+        this.#applyConnectResponse(cachedResp)
```
metal-messiah: This seems like it will fall through and keep processing resp (TimeKeeper setup + #applyConnectResponse again below) even after applying cachedResp? If a cached response shows up mid-flight, should we `return` right after this line to ensure no other logic is applied if we already have the cached response during the race?

src/common/harvest/connector.js
```diff
@@ -0,0 +1,164 @@
...
+import { activateFeatures } from '../util/feature-flags'
+import { now } from '../timing/now'
+import { warn } from '../util/console'
+import { VERSION } from '../constants/env'
+import { FEATURE_NAMES, FEATURE_TO_ENDPOINT } from '../../loaders/features/features'
+import { handle } from '../event-emitter/handle'
+import { SUPPORTABILITY_METRIC_CHANNEL } from '../../features/metrics/constants'
+
+const MAX_CONNECT_ATTEMPTS = 3
+const CONNECT_RETRY_BASE_MS = 3000 // affects each retry window proportionally (attempt 2: 0-3s, 3: 0-6s, 4: 0-12s, etc)
```
metal-messiah, line 18: was this specific pattern reviewed with DEM, given their idempotency/rate-limiting counter concerns in the CDD, or is this just an arbitrary internal choice?

src/common/harvest/connector.js
```diff
@@ -0,0 +1,164 @@
...
+
+      this.#agentRef.ee.abort()
+      return
+    }
+
+    let resp
+    try {
+      resp = JSON.parse(responseText)
+    } catch (error) {
+      warn(53, error)
```
metal-messiah, line 144: This may need a more explicit warning code that more accurately describes the catch case, 53 seems to only warning about entityGuid

src/features/utils/instrument-base.js
```diff
@@ -103,40 +104,38 @@ export class InstrumentBase extends FeatureBase {
...
-          session = setupAgentSession(agentRef)
-        }
+      try { // in the interest of keeping loader file size small, some modules are lazy-loaded as part of the larger async chunk
+        await ensureRuntimeBootstrap(agentRef, this.ee, this.featureName)
       } catch (e) {
-        warn(20, e)
-        this.ee.emit('internal-error', [e])
-        handle(SESSION_ERROR, [e], undefined, this.featureName, this.ee)
+        warn(72, e)
+        this.ee.abort() // failed Connector or Harvester will cause entire agent to shutdown
```
metal-messiah, line 111: Do we know what the current failure cases would be here besides any unforeseen coding bugs?  Since this is a full-on abort() case, this would be helpful to understand any risks associated with full shut-down

src/loaders/agent.js
```diff
@@ -93,10 +93,11 @@ export class Agent extends AgentBase {
     try {
       const enabledFeatures = getEnabledFeatures(this.init)
       const featuresToStart = [...this.desiredFeatures]
+      const isRumV2Enabled = this.init.feature_flags.includes('rum_v2')
 
       featuresToStart.sort((a, b) => featurePriority[a.featureName] - featurePriority[b.featureName])
       featuresToStart.forEach(InstrumentCtor => {
-        if (!enabledFeatures[InstrumentCtor.featureName] && InstrumentCtor.featureName !== FEATURE_NAMES.pageViewEvent) return // PVE is required to run even if it's marked disabled
+        if (!enabledFeatures[InstrumentCtor.featureName] && (isRumV2Enabled || InstrumentCtor.featureName !== FEATURE_NAMES.pageViewEvent)) return // PVE is required to run even if it's marked disabled unless rum_v2 is enabled
```
metal-messiah, line 100: It worth double checking, will we see any issues from allowing disablement of PVE?  There are some UI experiences that assume a PVE will always exist, so this may be a risk

docs/warning-codes.md
```diff
@@ -142,4 +142,6 @@
 ### 70
 `A session replay payload failed to send and is being retried. Recording is paused during the retry period, and will resume when a successful harvest is made. Some replay activity may be missed during retry phases.`
 ### 71
-`An invalid feature mode was detected and set to "off".`
\ No newline at end of file
+`An invalid feature mode was detected and set to "off".`
+### 72
+`Unable to initialized Connector and/or Harvester.`
```
metal-messiah:
```suggestion
`Unable to initialize Connector and/or Harvester.`
```

src/common/harvest/connector.js
```diff
@@ -0,0 +1,164 @@
...
+      this.#agentRef.ee.abort()
+      return
+    }
+    if (session) session.write({ cachedRumResponse: resp }) // store the response for other pages later in same session or in parallel tabs
```
metal-messiah: The implementation of the new session shape (addition of `config`) only has 1-directional protection.  We may need to dig into this more.

For example, v1 flags -> v2 cache read will guard safely, but v2 flags -> v1 cache read could potentially throw errors/break agent.

We may need to make a push for DEM to return the same shape of flags, unless you have a safer way to handle this in both directions.

src/common/harvest/connector.js
cwli24: oop, yeah the `return` here got forgotten or stripped out somehow. Thanks. Gap in the test as well.

src/common/harvest/connector.js
cwli24: Same answer in gist as here https://newrelic.atlassian.net/wiki/spaces/APM/pages/5098700832/Splitting+initial+RUM+into+separate+connect+harvest+requests?focusedCommentId=5137596454

The base here specifically is arbitrary. The old always did it after 5s, so I landed somewhere in that ballpark. The max time it'd take for both retries to send, excluding network/backend time, would be 3s + 6s = 9s then.

src/common/harvest/connector.js
cwli24: This is taken directly from old PVE:
image

src/features/utils/instrument-base.js
cwli24: Aside from faulty constructor code, just failed imports similar to session manager and adblock.

src/loaders/agent.js
cwli24: Their app disappearing if it doesn't get the PageView keep-alive signal, and I think some logic in the blob consumer looks at PageViews maybe? I don't know every place. But this is also 100% on users, like it's pseudo mandatory. It'd be nice to trim these special branches or overly restrictive guardrails later. I'd say wait until someone actually hit this "duh" scenario then we can update docs to explicitly call this out.

src/common/harvest/connector.js
cwli24: Well, first I'd call out the impact to limited to existing sessions up to expiry, so the transient period would last no more than a few hours. v2 RUM resp being read by old PV can only occur if we're going backwards in version. In general, we shouldn't address "forward compatibility", like ensuring v1.300 works with the current agent version. 

I think the key here is this whole thing is rolled out behind a feature flag so that there's a controlled one directional phase out of the old to the new on deployment. I really want to get away from having to add a bunch of guardrails and being very careful/considerate of the shape of things every time we edit the session object and now its children object too. I think we need to tell people simply to not mix-and-match agent versions within the same origin, and to update the older ones.

src/common/harvest/connector.js
```diff
@@ -0,0 +1,165 @@
...
+        body: {},
+        qs: {
+          a: this.#agentRef.info.applicationID,
+          v: VERSION,
+          s: this.#agentRef.runtime.session?.state.value || '0'
+        }
+      },
+      localOpts: { sendEmptyBody: true, headers: retryHeaders },
+      raw: true,
+      featureName: 'connect',
```
ptang-nr: In send(), this leads to [harvest metadata tracking](https://github.com/newrelic/newrelic-browser-agent/blob/main/src/common/harvest/send.js#L101-L106), but it's a bit odd as there's no metadata to track for this "feature".

src/common/harvest/connector.js
metal-messiah: on pages that an old agent (one that supports the "cached" config pattern) is getting the new config, will it throw errors when it tries to extract the rum flags?  If yes, thats not really an acceptable answer imo, but if it just fails gracefully and re-connects to the BCS, then that is sufficient.  Can you confirm that here for brevity?

src/common/harvest/connector.js
metal-messiah: my opinion is to add a new warning code that is more explicit as to what is actually happening (something like "failed to parse connect response" or something like that).

cwli24: That's a change to the warning #53 itself but sure

src/common/harvest/connector.js
cwli24: There's **no error** for sure when an older agent retrieves a newer cachedRumResponse.

Feature behavior varies. The featuers like ajax and PVT that `waitForFlags(([]))` will still run. Features like jserror and trace that requires an entitlement/sampling flag will just be blocked.

src/common/harvest/connector.js
cwli24: Thank you; I'll add an early-quit check for connect as well in `trackHarvestMetadata`

src/common/harvest/connector.js
cwli24: --reverted to old format

metal-messiah (approved):

PR #1826, opened 2026-08-08
feat: improve v2 target standardization

metal-messiah:
Improves agent internals to provide an explicit reference for container agent targets while processing v2 MFE pathway data, rather than relying on side-effect driven undefined target assumptions.
---

### Overview
The goal here is to provide an actual target representing the container agent, that can be referenced and processed identically to how a v2 MFE target would be handled, ideally reducing branching cases and instead providing 1 handling path and a more concrete reference to trace for container agent data.

### Related Issue(s)
https://new-relic.atlassian.net/browse/NR-539907


ptang-nr (approved):
LGTM - enhanced readability w/o complicating payload ❤️

PR #1827, opened 2026-08-10
feat: MFE Manifest

metal-messiah:
Adds an optional `manifest` to the `register()` API so a Micro Frontend can declare its own known assets (scripts, stylesheets, images, fonts), letting the agent accurately attribute errors, logs, AJAX calls, WebSockets, and resource-timing events to it — even for assets loaded by a script other than the one that called `register()`. An opt-in `timingMethod` also lets a manifest widen `MicroFrontEndTiming` fetch/execution windows across its declared assets.

---

### Overview

`register()` used to only attribute activity back to the single script that called it, which doesn't fit setups like split bundles or a platform-level script registering many MFEs on their behalf. This adds a `manifest` option so an MFE can list its own files up front:

- `manifest: { assets: [{ matcher, type }] }` — matcher can be a string or RegExp; script-ness is inferred from `.js` or set explicitly via `type: 'script'`.
- Once a manifest is supplied, it's the *only* source of attribution for that MFE — we no longer also credit whatever script happened to call `register()`. This matters for registrar-style setups where one shared script registers many MFEs; its own errors shouldn't leak into any of them.
- `timingMethod: 'entry' | 'scripts' | 'all'` controls how much the manifest widens `MicroFrontEndTiming` (default leaves it untouched).
- New `totalWeight` and `renderBlocking` fields on `MicroFrontEndTiming`, accumulated from every matched asset regardless of `timingMethod`.
- Resource-timing (`BrowserPerformance`) attribution now checks the manifest too, including a retry for self-registering scripts whose resource entry resolves before they've called `register()`.
- Fixed a RegExp-matcher edge case found while testing on Firefox: a regex built in a different JS realm (iframe/Worker/WebDriver sandbox) wasn't being recognized — swapped `instanceof RegExp` for a realm-safe check.

### Related Issue(s)

- API Review: https://newrelic.atlassian.net/wiki/x/CwA-ZAE
- https://new-relic.atlassian.net/browse/NR-578225
- https://new-relic.atlassian.net/browse/NR-578233

### Testing

- Unit: `npx jest tests/unit/common/util/manifest.test.js tests/unit/common/util/v2.test.js tests/unit/common/util/script-tracker.test.js tests/unit/features/generic_events/aggregate/index.test.js`
- E2E: `tests/specs/api/register/manifest.e2e.js` and `tests/specs/api/register/auto-detection-resources.e2e.js`
- Local run: `npm run wdio:smoke -- tests/specs/api/register/manifest.e2e.js -b chrome@latest` (or `firefox@latest`)


src/common/v2/script-tracker.js
```diff
@@ -219,14 +250,134 @@ function subscribeToLatePerformanceEntry (timings, mfeScriptUrl) {
...
+  }
+
+  // Non-script assets (css/images/fonts) never execute, so only script assets widen the execution window or anchor asset/type.
+  if (asset.isScript) {
+    const correlation = findCorrelation(cleanURL(entry.name))
+    if (correlation) {
+      const widenScriptWindow = () => {
+        const { start: scriptStart, end: scriptEnd } = correlation.script
+        if (scriptStart) timings.scriptStart = timings.scriptStart > 0 ? Math.min(timings.scriptStart, scriptStart) : scriptStart
+        if (scriptEnd) timings.scriptEnd = timings.scriptEnd > 0 ? Math.max(timings.scriptEnd, scriptEnd) : scriptEnd
```
cwli24: There's an issue with the way this works when `timingMethod` is `script`/`all`. These properties are `timingFactory` get/set objects. Once the `override` is set once, the live lookup (`getLazyValue()`) is never called again, and the `override` is always used instead. That means any single manifest asset that's already resolved can freeze the `timings.scriptEnd` value under its first execution of this func, and no later completion with a higher `correlation.dom.end` is ever factored in.

Doesn't seem like any tests caught this because they're not going through the `timingFactory` code path.

src/common/v2/script-tracker.js
```diff
@@ -197,6 +207,27 @@ function applyPerformanceEntry (timings, entry) {
...
+ * manifest asset) into a timings object. Shared by both the entry-script path (`applyPerformanceEntry`) and the
+ * manifest path (`applyManifestEntry`) so `totalWeight`/`renderBlocking` reflect every asset actually detected,
+ * regardless of which path found it.
+ * @param {RegisterAPITimings} timings
+ * @param {PerformanceResourceTiming} entry
+ */
+function applyResourceWeight (timings, entry) {
+  // transferSize is 0 for cross-origin responses without a Timing-Allow-Origin header (a browser privacy
+  // restriction, not evidence of a zero-byte asset) -- adding 0 is the correct behavior either way.
+  timings.totalWeight = (timings.totalWeight || 0) + (entry.transferSize || 0)
```
cwli24, line 259: This double counts in the case that the `.register` calling script lists itself as one of the `manifest.assets`. It's added the first time from `findScriptTimings` ancestor caller, then added again from `applyManifestTimings` ancestor caller, without any de-dupe in between.

metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1827/commits/8a46bed6dbd0e67b8d9f2a232c17364043ed0c2d

src/common/v2/script-tracker.js
metal-messiah: https://github.com/newrelic/newrelic-browser-agent/pull/1827/commits/8a46bed6dbd0e67b8d9f2a232c17364043ed0c2d

src/common/v2/manifest.js
```diff
@@ -0,0 +1,124 @@
...
+      isScript = isScriptType(type)
+    } else {
+      // Invalid `type` never overrides -- fall back to inference-only behavior rather than forcing `isScript: false`,
+      // so a typo'd `type` can't silently break stack-trace attribution for an otherwise-valid `.js` matcher.
+      warnInvalidManifestEntry(type)
+    }
+  }
+
+  if (isRegExp(matcher)) {
+    return { pattern: matcher, test: (url) => matcher.test(cleanURL(url)), isScript }
```
ptang-nr: Not necessarily blocking, just a note - this may fail to match silently if customer happens to use 'g' or 'y' flags, due to matcher keeping internal state via [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex).

cwli24 (dismissed):

src/loaders/api/register.js
```diff
@@ -49,6 +52,7 @@ export const warnings = {
   disabled: single(() => warn(55)),
   invalidTarget: single((target) => warn(48, target)),
   deregistered: single(() => warn(68)),
+  invalidTimingMethod: single((value) => warn(80, value))
```
metal-messiah:
```suggestion
  invalidTimingMethod: single((value) => warn(80, value)),
```

ptang-nr (approved):

PR #1828, opened 2026-08-10
fix: Apply hard cap to MFE FCP

metal-messiah:
Fixes an issue where MFE vitals could occasionally report out of bounds FCP/LCP values (e.g. after a browser tab freeze or suspend) instead of being dropped like other stale data.
---

### Overview
Some MFEs were still occasionally reporting FCP values with large durations (2+ hours) despite the existing 10-second give-up timeout added in #1823. Assumed root cause: setTimeout timers can pause while a tab is frozen/suspended (bfcache, mobile background) even though performance.now() keeps advancing through real wall-clock time — so on resume, a mutation callback observing the first paint can win the race against the disconnect callback and set an our of bounds FCP before the disconnect cycle can veto it.

This PR adds a stale-clock cap directly to trackMFEVitals in mfe-vitals.js:

- The FCP value getter now drops (undefined) any relative timestamp exceeding the existing 10s threshold, instead of reporting it.
- The LCP value getter cascades: if FCP is invalid, LCP is dropped too, since LCP can never occur before FCP and would be derived from the same untrustworthy clock.
- populateVitalMinimums now checks staleness before seeding fcpObservedAt, and disconnects immediately if the very first observed render is already stale, rather than seeding bad data and waiting for the timeout.
- A disconnected flag now guards every vitals-writing path (setters, resize/CLS/INP observer callbacks) so a callback already in flight when disconnect() runs can't sneak a late write in afterward.
- Fixed an ordering bug uncovered while making this change: the "MFE already rendered before we started observing" fast path called populateVitalMinimums() before observers was populated and before disconnect had its real implementation assigned, so an early stale-detection there would have silently no-op'd. Moved it to run after full setup.

### Related Issue(s)
Follow up to https://new-relic.atlassian.net/browse/NR-597354 as that work was not completely successful.

### Testing
Unit: extended tests/unit/common/v2/mfe-vitals.test.js with cases for the FCP cap, the FCP→LCP cascade, and rejecting writes after disconnect() (41/41 passing).
E2E (wdio, run manually): extended tests/specs/api/register/mfe-vitals.e2e.js — the shared validateVitals helper now universally enforces the 10s FCP cap and the FCP→LCP cascade invariant across all existing scenarios, and the existing "give up after 10s" test now explicitly documents that it also exercises the cascade with a qualifying LCP-sized element added after the timeout.
To test manually: register an MFE, then simulate a stalled clock/backgrounded tab before first paint and confirm no FCP/LCP is reported instead of an inflated value; run npx jest tests/unit/common/v2/mfe-vitals.test.js and the mfe-vitals.e2e.js wdio spec.

### DIFF
--- a/src/common/v2/mfe-vitals.js
+++ b/src/common/v2/mfe-vitals.js
@@ -124,6 +124,9 @@ const observePerformance = (observers, config, onEntry) => {
 export function trackMFEVitals (target, timings) {
   let fcpObservedAt
   let lcpObservedAt
+  // Once disconnect() runs, no further vitals data should be written - even from an observer callback that was
+  // already in flight when disconnect() was called.
+  let disconnected = false
 
   const getTimeRelativeToScriptStart = (capturedAt) => {
     if (capturedAt == null) return capturedAt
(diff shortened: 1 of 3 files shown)


ellisong (approved):
LGTM

PR #1830, opened 2026-08-10
chore: fix broken slack message

metal-messiah:
This PR fixes a broken github action where slack messages were failing to publish. See https://github.com/newrelic/newrelic-browser-agent/actions/runs/31226217091 for an example
---


ellisong (approved):
LGTM

For simplified context, the top-level "text" field is used for the notification and the "blocks" is used for the message itself.

PR #1838, opened 2026-08-27
feat: Add instrumentation to track races between tabs

ellisong:
Adds a supportability metric to track a race condition that occurs when another page runs at the same time with an updated session state.
---

### Overview

We're adding a supportability metric named `Session/RaceCondition/Seen`. This metric reports when you have the same agent (same app ID and license key) running in multiple tabs and they attempt to update the shared session state at the same time.

Angler: https://source.datanerd.us/agents/angler/pull/859

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-546642

### Testing

Make sure existing tests pass. This new supportability metric is troublesome to automatically test.


metal-messiah (approved):

metal-messiah:
Approved and merged Angler PR as well

Issue #1832, opened 2026-08-13
Soft Navigation API


tunetheweb:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

[NOTE]: # ( Provide a brief overview of what the new feature is all about. )

Chrome has now released it's soft navigation measurement support from Chrome 151:
https://developer.chrome.com/docs/web-platform/soft-navigations

We have also updated our `web-vitals` reference library in v6 to illustrate how to measure Web Vitals using these new APIs: https://github.com/GoogleChrome/web-vitals#report-metrics-for-soft-navigations

It's also documented on MDN:

- https://developer.mozilla.org/en-US/docs/Web/API/PerformanceSoftNavigation
- https://developer.mozilla.org/en-US/docs/Web/API/InteractionContentfulPaint
- Compatibility data is WIP: https://github.com/mdn/browser-compat-data/pull/30195

I also presented this API to the RUM CG meeting yesterday ([slides](https://docs.google.com/presentation/d/1ux41hS-AmfgihZlEhA3jilkGvnxTsECO0h_b1k_Nemc/edit?slide=id.g3f6d623949c_0_0#slide=id.g3f6d623949c_0_0)) where feedback from other RUM providers was positive, with many already using these new APIs and getting better results than their previous heuristics when attempting to measure API sites.

If this API is of any interest to New Relic, then I'm happy to answer any questions on it or discuss it.

## Desired Behavior

[NOTE]: # ( Tell us how the new feature should work. Be specific. )
[TIP]:  # ( Do NOT give us access or passwords to your New Relic account or API keys! )

Consider using the new PerformanceSoftNavigation and InteractionContentfulPaint APIs to measure soft navigations.

## Possible Solution

[NOTE]: # ( Not required. Suggest how to implement the addition or change. )

Consider using the new PerformanceSoftNavigation and InteractionContentfulPaint APIs to measure soft navigations.

## Additional context

[TIP]:  # ( Why does this feature matter to you? What unique circumstances do you have? )

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-605664

metal-messiah:
Thanks Barry! Good to see you again, it's been a while.  I will escalate this to our leadership

PR #540, opened 2023-05-24
chore(main): release 1.233.0

newrelic-browser-agent-team:
When this PR is merged, a new tagged release will be created with the notes below.
---

### Features

* Capture metrics for usage of MooTools and certain polyfills ([#539](https://github.com/newrelic/newrelic-browser-agent/issues/539)) ([903a7e1](https://github.com/newrelic/newrelic-browser-agent/commit/903a7e1e93a09f4e94ae76e95659a203ecb2896b))
* Update agent internals in early preparation for new features ([#532](https://github.com/newrelic/newrelic-browser-agent/issues/532)) ([1ee675d](https://github.com/newrelic/newrelic-browser-agent/commit/1ee675d232fb9a233c36ad0a2b998739fbaa1b8b))

### Bug Fixes

* Address "configurable" warnings arising from user-agent module ([#546](https://github.com/newrelic/newrelic-browser-agent/issues/546)) ([7a7dace](https://github.com/newrelic/newrelic-browser-agent/commit/7a7daceeaed603396805d81c9afd4cac4364cb40))
* Ensure runtime is preserved for late-configuration cases ([#538](https://github.com/newrelic/newrelic-browser-agent/issues/538)) ([229b8ed](https://github.com/newrelic/newrelic-browser-agent/commit/229b8ed0f4bc558cae62dd34ab6d32c302084d0b))
* Refactor usage of Array.from to address MooTools conflict ([#544](https://github.com/newrelic/newrelic-browser-agent/issues/544)) ([f1e6336](https://github.com/newrelic/newrelic-browser-agent/commit/f1e63367f653a957b0a401fbd21d24a1ca898bbd))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #1846, opened 2026-09-02
fix: Upgrade web-vitals from 4.2.4 to 6.2.1 to plug a listener leak

### LINKED ISSUE #1845: Bundled web-vitals 4.2.4 leaks a document `visibilitychange` listener per INP report cycle
## Description

`@newrelic/browser-agent` pins `web-vitals@4.2.4`. That version's `whenIdle()` helper registers a `document` `visibilitychange` listener through `onHidden()` on **every invocation**, and never removes it — no `once`, no `removeEventListener`, no signal:

```js
// web-vitals@4.2.4 src/lib/onHidden.ts
export const onHidden = (cb: () => void) => {
  document.addEventListener('visibilitychange', () => {
    if (document.visibilityState === 'hidden') {
      cb();
    }
  });
};
```

```js
// web-vitals@4.2.4 src/lib/whenIdle.ts
} else {
  handle = rIC(cb);
  onHidden(cb); // <-- new permanent listener on each call
}
```

`whenIdle()` is called from `onINP.ts:95` on every event-entry observer batch, from `attribution/onINP.ts:163` on every attribution cleanup cycle, and from `onLCP.ts:107` on LCP finalization. Since the agent's page_view_timing feature subscribes to `onINP` (and `onLCP`), every instrumented page accumulates roughly two permanent listeners per interaction batch for the life of the page. Each listener retains its closure, and every tab switch dispatches all of them — precisely the moment users expect an idle tab to be cheap.

This is a known upstream bug: GoogleChrome/web-vitals#534 ("whenIdle add never removed visibilitychange event listener"), fixed by GoogleChrome/web-vitals#538 and first shipped in web-vitals 5.0.0 (tightened further in 5.0.3 via #627). The current `whenIdleOrHidden()` (6.2.1) registers with `{ once: true, capture: true }`, removes the listener when the idle callback wins the race, and cancels the stale idle callback in the hidden path. No fixed 4.x release exists — 4.2.4 is the last 4.x.

We measured this in a long-lived SPA (multi-hour sessions are our norm): after ~45 minutes of ordinary interaction, `document` had accumulated **112 `visibilitychange` listeners**, all registered by the agent's bundled web-vitals copy. A per-type listener census showed the entire residual was this one type. Growth is interaction-driven, not time-driven: flat across a 120 s idle period, and climbing with clicks/keystrokes.

One measurement note: because the agent wraps `addEventListener`, DevTools' listener views attribute all of these registrations to the agent's wrapper, which makes the growth easy to miss or misattribute. We had to patch `Document.prototype.addEventListener` ahead of the agent and capture registration stacks to attribute them to `web-vitals`' `onHidden`.

## Steps to Reproduce

1. Before the agent loads, install a counter:

```html
<script>
  const orig = Document.prototype.addEventListener;
  window.__vcCount = 0;
  Document.prototype.addEventListener = function (type, ...args) {
    if (type === 'visibilitychange') window.__vcCount++;
    return orig.call(this, type, ...args);
  };
</script>
```

2. Load any page instrumented with the browser agent (any recent version up to and including 1.321.0) with the page_view_timing feature enabled (default).
3. Interact with the page — clicks, key presses — for a few minutes.
4. Log `window.__vcCount` periodically.

## Expected Behavior

The `visibilitychange` listener count on `document` stays bounded (single digits) regardless of session length and interaction volume — which is the behavior with web-vitals >= 5.0.0.

Per-type census from a ~45-minute session (Chrome, staging environment of our app, listener count on `document`):

```
document total: 155
  visibilitychange     112     <-- all from bundled web-vitals onHidden
  keydown                9
  pointerdown            4
  ... everything else 1-2 each
```

## Your Environment

* Chrome 140 (reproduces in any browser; the leak is unconditional in the library code path)
* Browser agent versions ~1.270 through 1.321.0 (all pin a web-vitals 4.x with the leak; 4.2.4 pinned since v1.284.0)
* Observed via the npm package (`@newrelic/browser-agent`), applies equally to CDN loaders since web-vitals is compiled in

## Additional context

I understand a major web-vitals bump has been deliberately deferred before (#1474 / #1477 — "too many consequences that would need to be planned around"), so here is what our analysis of the 4.2.4 → 6.2.1 boundary says those consequences actually are for this agent's usage surface (five `web-vitals/attribution` imports + the core-build import in `registered-iframe-entity.js`):

* **One hard API change:** `LCPAttribution.element` was renamed to `target` in v5 (GoogleChrome/web-vitals#585). Without a code change the `element` attribute on `lcp` PageViewTiming events would silently disappear. Trivial to absorb while keeping the existing NR attribute name.
* **INP attribution fields became optional:** v6 emits "dummy" INP reports (`entries: []`, value 8, no `interactionTarget`/`interactionTime`/`interactionType`/`nextPaintTime`) after bfcache restores when only sub-threshold interactions occurred (GoogleChrome/web-vitals#724). Needs a small guard in the INP wrapper, plus a product decision on whether to keep or filter those reports.
* **`onFID` removal (v5's headline break) is moot** — the agent already removed FID usage (#1319, #1395) and computes `fi` from native `first-input` entries.
* **Test environments:** web-vitals >= 5 calls `performance.getEntriesByType()` unguarded at metric registration ("Baseline Widely available" policy), which jsdom does not implement — this repo's own jest suites need a one-line polyfill in the shared setup, and npm consumers who run their instrumented apps under jest/jsdom will see the vitals-dependent features disable themselves (warn code 34) instead of silently no-op'ing as with v4.
* **Dist syntax level:** v5+ dropped ES5 output; 6.2.1 ships ~ES2022. Within the agent's `.browserslistrc` (last 10 versions of the majors) this is fine, but since node_modules bundle untranspiled into the CDN loaders, browsers far outside the support matrix would move from degraded behavior to a parse error in the chunk that carries web-vitals.
* **Value shifts to note in release notes, not code:** attribution selector strings sort their class lists from v5 (#518); INP breakdowns are clamped to the INP duration (#528, #540); prerendered pages now report FCP/LCP (#621); the idle callback is capped at 1 s so `reportAllChanges` subscribers hear more often on busy pages (#755); the v6 attribution build keeps a bounded (50-entry) module-scope resource-timing observer for LCP attribution.
* **Everything else the agent reads is shape-stable** between 4.2.4 and 6.2.1: both entry points (`web-vitals`, `web-vitals/attribution`) still exist with `onCLS`/`onFCP`/`onINP`/`onLCP`/`onTTFB`, and the CLS/FCP/TTFB attribution fields the agent consumes are unchanged.

There is also a strategic upside: web-vitals v6 mainlines soft-navigation support (`reportSoftNavs`, `navigationId` on every metric), which #1832 asks about — a v6 bump is a prerequisite for adopting that.

Precedent-wise, this is the same shape as #1379, where a web-vitals patch bump was taken specifically because it fixed a listener-registration leak.

I have the migration prepared (dependency pin, the LCP field rename, a guard for the v6 synthetic INP report, a small safe-registration helper so a missing Performance Timeline API degrades instead of throwing at import, and matching unit/component test updates — full jest/types/build green against 6.2.1) and I'm happy to open a PR if you're open to it.

gesposito:
Upgrades the bundled web-vitals library from 4.2.4 to 6.2.1, fixing an unbounded `document` `visibilitychange` listener leak: web-vitals 4.x registers a permanent listener on every INP report cycle, so long-lived pages accumulate hundreds of listeners. Also absorbs the v5/v6 attribution API changes the agent is exposed to (the `LCPAttribution.element` → `target` rename and newly-optional INP/CLS attribution fields) with no change to the attribute names the agent emits.
---

### Overview

web-vitals 4.2.4's `whenIdle()` calls `onHidden()` on every invocation, and `onHidden()` does a bare `document.addEventListener('visibilitychange', ...)` with no `once` and no removal. `onINP` invokes `whenIdle()` on every event-entry observer batch and every attribution cleanup cycle, so pages instrumented with page_view_timing leak roughly two permanent listeners per interaction batch for the life of the page (we measured 112 after ~45 minutes of ordinary use in a long-lived SPA). Upstream fixed this in web-vitals 5.0.0 (GoogleChrome/web-vitals#534 → #538) and there is no fixed 4.x release, so this bumps to the current 6.2.1.

Changes, mapped to the 4 → 6 breaking-change surface this agent actually touches:

* **`package.json` / `package-lock.json`**: `web-vitals` exact pin `4.2.4` → `6.2.1` (kept exact per #1133). Both entry points the agent imports (`web-vitals`, `web-vitals/attribution`) and all five consumed metric functions are unchanged in v6.
* **`src/common/vitals/largest-contentful-paint.js`**: `LCPAttribution.element` was renamed to `target` in v5 (GoogleChrome/web-vitals#585). The wrapper now reads `attribution.target` but keeps emitting the existing `element` attribute on `lcp` PageViewTiming events, so nothing changes for NRQL consumers. Without this, the attribute would silently disappear.
* **`src/common/vitals/interaction-to-next-paint.js`**: v6 emits a synthetic INP report (`value: 8`, `entries: []`, no interaction attribution) after a bfcache restore when every interaction stayed below the duration threshold (GoogleChrome/web-vitals#724). v4 reported nothing in that situation, so the wrapper skips reports without entries to keep only measured interactions — otherwise a constant 8 ms sample would be harvested as an `inp` PageViewTiming node indistinguishable from a real one.
* **`src/common/vitals/vital-metric.js`**: several attribution fields are optional as of v6 (`INPAttribution.interactionTarget` when the interacted element was removed before reporting, `CLSAttribution.largestShiftTarget` likewise, and every CLS field on a zero-shift report) where v4 coerced them to `''`. `VitalMetric.update()` now drops undefined-valued attrs once, centrally, so they are omitted from the payload instead of being serialized as null attributes — the same outcome the LCP wrapper already produced for its optional fields, without per-field guards in every wrapper.
* **`src/common/vitals/register-vital.js`** (new) and the five wrappers plus `src/interfaces/registered-iframe-entity.js`: web-vitals >= 5 calls `performance.getEntriesByType()` unguarded, synchronously, while a metric is registered (its "Baseline Widely available" policy dropped v4's feature check). The agent registers metrics at module top level, so wherever that API is missing — a consumer running an instrumented app under Jest/jsdom, a webview shimming `performance` — v6 would throw at import time, the aggregate chunk import would reject with warn code 34, and because page_view_event and soft_navigations import the FCP wrapper too, the PageView event itself would be lost. The `RegisteredIframeEntity` constructor would throw into customer code the same way. `registerVital()` wraps each registration in the try/catch idiom `first-paint.js` already uses for `PerformanceObserver`, so a missing API degrades to "that metric is not reported", exactly as with v4.
* **`tools/jest/globals.mjs`**: jest-environment-jsdom does not implement the Performance Timeline, so the same unguarded call meant 8 suites / 71 tests that transitively load the real library (page_view_timing aggregate unit tests, page_view_event and soft_navigations component tests) failed at module init with `TypeError: performance.getEntriesByType is not a function` — surfacing only as opaque 5 s hook timeouts because the import failure is caught and logged through the muted console. With `registerVital()` those suites no longer fail, but the polyfill (`() => []`, which the agent's own `getNavigationEntry()` already treats as "no entry") keeps the callback paths exercised under jsdom; it is defined writable/configurable because some tests redefine it.
* **`onFID`** (v5's headline removal) is not used by the agent — no change needed.

Behavioral notes for reviewers (library-side changes that can shift reported values; no agent code involved):

* The synthetic post-bfcache INP report described above is skipped for v4 parity. If you would rather adopt web-vitals' intent (so INP p75 isn't biased toward pages with only fast interactions), dropping the `entries` check in the INP wrapper passes them through; they arrive with only `metricId`/`inputDelay`/`processingDuration`/`presentationDelay`/`loadState`.
* Attribution selector strings sort their class lists as of v5 (#518) — `largestShiftTarget` / `interactionTarget` / LCP target strings can change form for multi-class elements.
* INP breakdowns are clamped to the INP duration and `nextPaintTime` to `processingStart` as of v5 (#528, #540, #546) — sub-metric values become more accurate.
* Prerendered pages now report FCP/LCP (#621); previously 4.2.4 could silently never report there.
* The idle callback is capped at 1 s (#755), so `reportAllChanges` subscribers (CLS here, plus cls/inp in `registered-iframe-entity`) hear more often on busy pages; both consumers are idempotent per-report handlers.
* The v6 attribution build registers a module-scope resource-timing PerformanceObserver retaining at most 50 entries (LCP subpart attribution) — new, bounded runtime footprint; it is feature-detected inside try/catch, so non-browser scopes are unaffected.
* web-vitals 5+ dist output is no longer ES5 (6.2.1 is ~ES2022, including `Array.prototype.at`). This is well within `.browserslistrc` (last 10 versions of the majors) and webpack/terser parse it fine, but since node_modules bundle untranspiled into the CDN loaders, browsers far outside the support matrix would now hit a parse error in the chunk carrying web-vitals rather than degrading. npm consumers whose bundlers don't transpile node_modules inherit the same floor — worth a release-note mention.
### Related Issue(s)

* Fixes #1845
* GoogleChrome/web-vitals#534 / GoogleChrome/web-vitals#538 (the upstream bug and fix)
* #1474 / #1477 (previously-deferred v5 bumps)
* #1832 (Soft Navigation API — web-vitals v6 mainlines `reportSoftNavs`, so this bump is a prerequisite)
* #1379 (precedent: leak-motivated web-vitals bump)

### Testing

* `tests/unit/common/vitals/largest-contentful-paint.test.js`: fixture updated to the v5+ `target` field; asserts the outgoing attribute is still named `element`.
* `tests/unit/common/vitals/interaction-to-next-paint.test.js`: new tests feeding a v6-shaped synthetic report (`entries: []`, `inputDelay: 0`, `processingDuration: 0`, `presentationDelay: value`, no interaction fields) and asserting it is not reported, and a report with an unresolved `interactionTarget` asserting the attr is omitted; existing full-shape test unchanged.
* `tests/unit/common/vitals/cumulative-layout-shift.test.js`: new test asserting `largestShiftTarget` is omitted when undefined.
* `tests/unit/common/vitals/vital-metric.test.js`: new test for the undefined-attr omission in `update()`.
* `tests/unit/common/vitals/register-vital.test.js` (new), plus a "does NOT throw if web-vitals registration throws" test in each of the five wrapper suites and in `registered-iframe-entity.test.js` (a throwing `onCLS` no longer aborts the constructor and the other vitals still register).
* `tests/components/page_view_timing/aggregate.test.js` and `index.test.js`: the web-vitals mock fixtures now carry the complete attribution shape (and `id`/`entries`) real web-vitals emits — they previously passed empty or partial attribution objects, which only satisfied the key-presence assertions because undefined attrs serialized as null attributes; the LCP/FCP/INP assertions now check values too.
* Full local jest runs against the installed 6.2.1 pass: 923 unit tests (93 suites), 568 component tests (33 suites). `tsc -b` + `tsd` clean; the CDN webpack build compiles and every loader carries the v6 `whenIdleOrHidden` registration (`{once: true, capture: true}`).
* Coverage limits: the jest suites mock web-vitals for the wrapper-level tests, so real-library behavior (including the leak fix itself) is exercised by the WDIO e2e specs (`tests/specs/pvt/timings.e2e.js` etc.), which need the LambdaTest matrix a maintainer must dispatch; none of those specs assert on the renamed/now-optional attribution fields. Locally, loading a built loader and counting `visibilitychange` registrations on `document` (patch `Document.prototype.addEventListener` before the agent) shows the count stays bounded with this change and grows monotonically with interaction without it.

### DIFF
--- a/src/interfaces/registered-iframe-entity.js
+++ b/src/interfaces/registered-iframe-entity.js
@@ -7,6 +7,7 @@
 import { onCLS, onFCP, onINP, onLCP } from 'web-vitals'
 // internal
 import { globalScope, isBrowserScope } from '../common/constants/runtime'
+import { registerVital } from '../common/vitals/register-vital'
 import { isIFrameWindow } from '../common/dom/iframe'
 import { now } from '../common/timing/now'
 import { warn } from '../common/util/console'
@@ -179,12 +180,12 @@ export class RegisteredIframeEntity {
    */
   #setupVitalsListeners () {
     VITALS.forEach(([vitalFn, property]) => {
-      vitalFn(({ value }) => {
+      registerVital(() => vitalFn(({ value }) => {
         this.metadata.vitals[property].value = value
         this.#postMessageToParent(IFRAME_VITALS_UPDATE, {
           entries: [{ property, value }]
         })
-      }, { reportAllChanges: property === 'cls' || property === 'inp' })
+      }, { reportAllChanges: property === 'cls' || property === 'inp' }))
     })
   }
 
--- a/src/interfaces/registered-iframe-entity.js
+++ b/src/interfaces/registered-iframe-entity.js
@@ -180,7 +180,11 @@ export class RegisteredIframeEntity {
    */
   #setupVitalsListeners () {
     VITALS.forEach(([vitalFn, property]) => {
-      registerVital(() => vitalFn(({ value }) => {
+      registerVital(() => vitalFn(({ value, entries: vitalEntries }) => {
+        /* web-vitals v6 reports a synthetic INP (value 8, no entries) after bfcache restores when
+          every interaction stayed below the duration threshold; skip those so only measured
+          interactions are reported, matching the guard in src/common/vitals/interaction-to-next-paint.js */
+        if (property === 'inp' && !vitalEntries?.length) return
         this.metadata.vitals[property].value = value
         this.#postMessageToParent(IFRAME_VITALS_UPDATE, {
           entries: [{ property, value }]
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@newrelic/rrweb": "^1.1.2",
         "fflate": "0.8.2",
-        "web-vitals": "4.2.4"
+        "web-vitals": "6.2.1"
       },
       "devDependencies": {
         "@babel/cli": "^7.23.4",
@@ -25736,9 +25736,9 @@
       }
     },
     "node_modules/web-vitals": {
-      "version": "4.2.4",
-      "resolved": "https://registry.npmjs.org/web-vitals/-/web-vitals-4.2.4.tgz",
-      "integrity": "sha512-r4DIlprAGwJ7YM11VZp4R884m0Vmgr6EAKe3P+kO0PPj3Unqyvv59rczf6UiGcb9Z8QxZVcqKNwv/g0WNdWwsw==",
+      "version": "6.2.1",
+      "resolved": "https://registry.npmjs.org/web-vitals/-/web-vitals-6.2.1.tgz",
+      "integrity": "sha512-rLcLXA2sx6+9dE88NHFubwTtGxpK4yYBLj6qHPdFoCaLr0cXGb4efOqtKLlm4loGA4OEKHIQKMKzZkKyOh5ctw==",
       "license": "Apache-2.0"
     },
     "node_modules/webdriver": {
--- a/package.json
+++ b/package.json
@@ -225,7 +225,7 @@
   "dependencies": {
     "@newrelic/rrweb": "^1.1.2",
     "fflate": "0.8.2",
-    "web-vitals": "4.2.4"
+    "web-vitals": "6.2.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.23.4",
--- a/src/common/vitals/cumulative-layout-shift.js
+++ b/src/common/vitals/cumulative-layout-shift.js
@@ -1,16 +1,17 @@
 /**
- * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 import { onCLS } from 'web-vitals/attribution'
 import { VITAL_NAMES } from './constants'
 import { VitalMetric } from './vital-metric'
+import { registerVital } from './register-vital'
 import { isBrowserScope } from '../constants/runtime'
 
 export const cumulativeLayoutShift = new VitalMetric(VITAL_NAMES.CUMULATIVE_LAYOUT_SHIFT, (x) => x)
 
 if (isBrowserScope) {
-  onCLS(({ value, attribution, id }) => {
+  const handleCLS = ({ value, attribution, id }) => {
     const attrs = {
       metricId: id,
       largestShiftTarget: attribution.largestShiftTarget,
@@ -19,5 +20,6 @@ if (isBrowserScope) {
       loadState: attribution.loadState
     }
     cumulativeLayoutShift.update({ value, attrs })
-  }, { reportAllChanges: true })
+  }
+  registerVital(() => onCLS(handleCLS, { reportAllChanges: true }))
 }
--- a/src/common/vitals/first-contentful-paint.js
+++ b/src/common/vitals/first-contentful-paint.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 import { onFCP } from 'web-vitals/attribution'
@@ -7,6 +7,7 @@ import { onFCP } from 'web-vitals/attribution'
 import { iOSBelow16, initiallyHidden, isBrowserScope } from '../constants/runtime'
 import { VITAL_NAMES } from './constants'
 import { VitalMetric } from './vital-metric'
+import { registerVital } from './register-vital'
 
 export const firstContentfulPaint = new VitalMetric(VITAL_NAMES.FIRST_CONTENTFUL_PAINT)
 
@@ -27,7 +28,7 @@ if (isBrowserScope) {
     // ignore
     }
   } else {
-    onFCP(({ value, attribution }) => {
+    const handleFCP = ({ value, attribution }) => {
       if (initiallyHidden || firstContentfulPaint.isValid) return
       const attrs = {
         timeToFirstByte: attribution.timeToFirstByte,
@@ -35,6 +36,7 @@ if (isBrowserScope) {
         loadState: attribution.loadState
       }
       firstContentfulPaint.update({ value, attrs })
-    })
+    }
+    registerVital(() => onFCP(handleFCP))
   }
 }
--- a/src/common/vitals/interaction-to-next-paint.js
+++ b/src/common/vitals/interaction-to-next-paint.js
@@ -1,17 +1,21 @@
 /**
- * Copyright 2020-2025 New Relic, Inc. All rights reserved.
+ * Copyright 2020-2026 New Relic, Inc. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 import { onINP } from 'web-vitals/attribution'
 import { VitalMetric } from './vital-metric'
 import { VITAL_NAMES } from './constants'
+import { registerVital } from './register-vital'
 import { isBrowserScope } from '../constants/runtime'
 
 export const interactionToNextPaint = new VitalMetric(VITAL_NAMES.INTERACTION_TO_NEXT_PAINT)
 
 if (isBrowserScope) {
   /* Interaction-to-Next-Paint */
-  onINP(({ value, attribution, id }) => {
+  const handleINP = ({ value, attribution, id, entries }) => {
+    /* web-vitals v6 reports a synthetic INP (value 8, no entries, no interaction attribution) after bfcache restores when
+      every interaction stayed below the duration threshold; skip those so only measured interactions are reported, as in v4 */
+    if (!entries?.length) return
     const attrs = {
       metricId: id,
       eventTarget: attribution.interactionTarget, // event* attrs deprecated in v4, kept for NR backwards compatibility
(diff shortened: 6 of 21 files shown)


metal-messiah:
Thanks for reporting this and submitting the PR @gesposito ! We will be looking into validating compatibility with the dependency version bump this week and we will be back with any revisions needed, but we greatly appreciate it!  Stay tuned

src/interfaces/registered-iframe-entity.js
```diff
@@ -179,12 +180,12 @@ export class RegisteredIframeEntity {
...
-      vitalFn(({ value }) => {
+      registerVital(() => vitalFn(({ value }) => {
         this.metadata.vitals[property].value = value
         this.#postMessageToParent(IFRAME_VITALS_UPDATE, {
           entries: [{ property, value }]
         })
-      }, { reportAllChanges: property === 'cls' || property === 'inp' })
+      }, { reportAllChanges: property === 'cls' || property === 'inp' }))
     })
   }
```
metal-messiah, line 194:
```suggestion
#setupVitalsListeners () {
    VITALS.forEach(([vitalFn, property]) => {
      registerVital(() => vitalFn(({ value, entries: vitalEntries }) => {
        /* web-vitals v6 reports a synthetic INP (value 8, no entries) after bfcache restores when
          every interaction stayed below the duration threshold; skip those so only measured
          interactions are reported, matching the guard in src/common/vitals/interaction-to-next-paint.js */
        if (property === 'inp' && !vitalEntries?.length) return
        this.metadata.vitals[property].value = value
        this.#postMessageToParent(IFRAME_VITALS_UPDATE, {
          entries: [{ property, value }]
        })
      }, { reportAllChanges: property === 'cls' || property === 'inp' }))
    })
  }
```

metal-messiah: @gesposito This seems to be subject to the same synthetic-INP issue that [you guarded against in interaction-to-next-paint.js](https://github.com/newrelic/newrelic-browser-agent/pull/1846/changes#diff-6ae18ff2a3885b638c6002942208a03405c26f39d18e2c868d315339d5d62de8R18) and seems like it should apply here too — after a bfcache restore with no qualifying interactions, onINP fires with synthetic value and an empty entries array. Since this listener is shared across all four vitals, the guard would need to be scoped to property === 'inp' specifically (CLS/LCP/FCP don't appear to have this synthetic-report behavior). Otherwise a phantom value would get forwarded to the parent as if it were a real measurement.

metal-messiah (changes requested):
@gesposito This looks really good, well done. Ran full E2E suite locally -- all passing.  One small change request if you don't mind :)

metal-messiah:
Relates to https://new-relic.atlassian.net/browse/NR-605664

ptang-nr (approved):

metal-messiah (approved):

metal-messiah:
Thanks @gesposito !

Issue #1836, opened 2026-08-25
Use ReportingObserver API to report Content Security Policy Violations since page load


welkup:
## Summary

The current approach to reporting on Content Security Policy violations is done through the [securitypolicyviolation event](https://developer.mozilla.org/en-US/docs/Web/API/Element/securitypolicyviolation_event).

CSP Violations that occur prior to the browser agent initialization are missed.

The [ReportingObserver API](https://developer.mozilla.org/en-US/docs/Web/API/ReportingObserver) is another way to detect CSP violations. It has a `buffered` configuration flag that allows reporting on violations that occur before the browser agent is initialized.

## Desired Behavior

SecurityPolicyViolations would be reported in the same way, only now they will encompass the entire page load (not only those violations that occur after the event handler is attached).

## Additional context

We are loading the browser agent asynchronously, so a number of resources are loaded prior to the browser agent.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-610205

metal-messiah:
Thanks for the suggestion @welkup! I have added a ticket to our backlog to look into this --- stay tuned

metal-messiah:
Will be included in release 1.322.0 #1844 Thanks for the suggestion!

welkup:
Awesome! Thank you so much!

PR #1848, opened 2026-09-04
feat: Include observer API for SPV events

cwli24:
Use the `ReportingObserver` API to catch `securitypolicyviolation` events that occurred prior to the agent loader execution. In practice, this mainly affects chromium browsers, as existing configs and bugs affect Firefox and Safari collection.
---

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-610205

### Testing

Updated

### DIFF
--- a/src/features/generic_events/instrument/index.js
+++ b/src/features/generic_events/instrument/index.js
@@ -61,12 +61,26 @@ export class Instrument extends InstrumentBase {
         handle('ws-complete', [nrData, targets], undefined, this.featureName, this.ee)
       })
     }
-    if (securityPolicyViolationEnabled && typeof globalScope.ReportingObserver === 'function') {
-      const cspObserver = new globalScope.ReportingObserver((reports) => { // csp-violation reports do not have timestamps of their own
-        reports.forEach(report => handle('spv', [report, now()], undefined, FEATURE_NAMES.genericEvents, this.ee))
-      }, { types: ['csp-violation'], buffered: true })
-      cspObserver.observe()
-      this.removeOnAbort.signal.addEventListener('abort', () => cspObserver.disconnect(), { once: true })
+    if (securityPolicyViolationEnabled) {
+      /** ReportingObserver's `csp-violation` type isn't reliably enabled everywhere out of the box:
+       *  - Safari only delivers it when the page sets a report-to CSP directive -- https://bugs.webkit.org/show_bug.cgi?id=320750;
+       *  - Firefox ships it behind the `dom.reporting.enabled` pref, which defaults to false;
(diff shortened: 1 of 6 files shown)


src/features/generic_events/instrument/index.js
```diff
@@ -61,8 +62,24 @@ export class Instrument extends InstrumentBase {
...
+            handle('spv', [normalizeCspViolation(report), now()], undefined, FEATURE_NAMES.genericEvents, this.ee)
+          })
+          cspObserver.disconnect()
+        }, { types: ['csp-violation'], buffered: true })
+        cspObserver.observe()
+      }
+
       globalScope.addEventListener('securitypolicyviolation', (evt) => {
-        handle('spv', [evt], undefined, FEATURE_NAMES.genericEvents, this.ee)
+        handle('spv', [normalizeCspViolation(evt), evt.timeStamp], undefined, FEATURE_NAMES.genericEvents, this.ee)
```
metal-messiah, line 82: since this handles buffered + live, is there a reason we should keep this in the instrument vs just doing it all in the agg?

src/features/generic_events/instrument/index.js
```diff
@@ -132,5 +149,26 @@ export class Instrument extends InstrumentBase {
     else this.deregisterDrain()
   }
 }
+/** `Report.body` (ReportingObserver) and `SecurityPolicyViolationEvent` (securitypolicyviolation) describe the same
+ * violation with differently-named/shaped fields -- collapse both into one shape so downstream code doesn't care
+ * which mechanism produced it. `source` is either of those: a Report (fields under `.body`, URLs as blockedURL/
+ * documentURL) or the event (fields at the top level, URLs as blockedURI/documentURI). Also normalizes disposition's
+ * "reporting" (Report body) to "report" (event) so the value is consistent regardless of source. */
+function normalizeCspViolation (source) {
```
metal-messiah, line 157: same here, couldnt we normalize this in the agg

metal-messiah:
#1836

src/features/generic_events/instrument/index.js
cwli24: Not particularly. The pros here are that the styling is cleaner with it being obvious that the different subscription methods generates different "event" shapes and that the buffer doesn't hold onto references to full entries/reports, only the relevant information.

metal-messiah: Ok, i'm fine either way on this, the build size implication isnt significant --- I'll leave the decision to you.  Otherwise this PR LGTM -- i'll make this as approved.  If you end up wanting to shift it to the agg LMK and I can re-review.

metal-messiah (approved):

PR #1851, opened 2026-09-08
chore: Daily dispatch improvements

metal-messiah:
Extends the Daily Dispatch Slack report with clearer PR triage, build-size visibility for the upcoming release, and basic repo/npm health stats.
---

### Overview
See the new format [here](https://newrelic.slack.com/archives/C04E6RHTBCK/p1788886487261439).

Reworks `.github/actions/daily-dispatch/index.js` to give the daily dispatch more useful, better-organized triage information:

- **PR triage is now split into five sections** instead of one flat "PRs Needing Review" list:
  - `🔖 Release PR` — the release-please PR, called out on its own since it's release automation, not a PR that needs the normal review triage.
  - `🌍 External Contributor PRs` — PRs from non-org-member authors.
  - `🆕 Never Reviewed` — no reviewer activity yet.
  - `🟠 Needs Revisions` — reviewer feedback given, author hasn't pushed since.
  - `🔁 Needs Re-Review` — author pushed since the last feedback, reviewer hasn't looked again.
  - External-contributor classification uses GitHub's `authorAssociation`, but also treats anyone already tracked in the dispatch's reviewer map as internal regardless of what `authorAssociation` reports — that field silently under-reports org membership (falls back to `CONTRIBUTOR`/`NONE`) when the querying token can't see the author's org membership, which is the case for the default `GITHUB_TOKEN` and any maintainer whose org membership is private.
- **Version Status now includes build size vs. the latest release.** The `size-compare` job in `pull-request-checks.yml` already comments an asset size report (tagged `<!-- browser_agent asset size report -->`) on every PR, including the release-please PR since it's a normal PR against `main`. The dispatch parses that comment and shows each asset's current size and percent change vs. the latest release tag, colored green/yellow/red to match the report's own thresholds. The comparison against `main` is dropped since the release PR is synced with `main` and that diff is always ~0%. If the comment isn't there yet (checks still running), the dispatch links out to the PR's checks instead of silently omitting the section.
- **New "📊 Repository Stats" section** (moved to the top of the message): stars, forks, watchers, and weekly npm downloads (from the public `api.npmjs.org` downloads endpoint, degrading to "unavailable" instead of failing the whole dispatch if that call errors).
- **New "🚀 Deployment Status" section**: the last successful deployment per environment (`nr1-dev`, `nr1-staging`, `nr1-us-prod`, `nr1-eu-prod`, `nr1-jp-prod`, `public-release`). This required scanning each deployment's full status history for a `SUCCESS` entry rather than trusting `latestStatus`, since GitHub flips a deployment's `latestStatus` to `INACTIVE` once a newer deployment supersedes it in the same environment — that's normal lifecycle, not a failure indicator. Also caught and fixed a `last`/`first` connection-pagination bug during development: `deployments(last: 10, orderBy: CREATED_AT DESC)` returns the *tail* of that already-sorted list (the oldest deployments in the sampled range), not the most recent ones — it needs `first: 10`.
- The per-PR sections now render one Slack block per PR instead of two, and cap at 6 PRs each, to keep the overall message comfortably under Slack's 50-block-per-message limit now that PRs are split across five sections instead of one.
- Added matching metrics fields (`externalPrCount`, `neverReviewedCount`, `needsRevisionsCount`, `needsReReviewCount`) to the existing `Br0ws3rMetrics` custom event payload.

No workflow or permissions changes were needed — everything is derivable from data the existing `GITHUB_TOKEN` scopes and public APIs already allow.

### Related Issue(s)

N/A

### Testing

Ran the action script directly against the live repository (`GITHUB_TOKEN=$(gh auth token) GITHUB_REPOSITORY=newrelic/newrelic-browser-agent node .github/actions/daily-dispatch/index.js`) and inspected the generated Slack Block Kit payload and metrics JSON after each change to confirm:
- PR bucketing matches the actual review state of real open PRs (spot-checked against `gh pr list` and individual PR timelines).
- The release-please PR appears only in its own section, never in the four review-state/external buckets.
- The asset size report parses correctly against the real comment on the open release-please PR, including the header/separator-row bleed-through bug found and fixed along the way.
- Deployment dates match what's shown on GitHub's own `/deployments/<environment>` pages for each tracked environment.
- npm weekly downloads match `https://api.npmjs.org/downloads/point/last-week/@newrelic/browser-agent` and degrade to "unavailable" without crashing when that call fails.
- Total Slack block count for the current repo state (34 blocks) stays well under Slack's 50-block limit.

### DIFF
--- a/.github/actions/daily-dispatch/index.js
+++ b/.github/actions/daily-dispatch/index.js
@@ -50,6 +50,61 @@ const reviewerCommentEvents = (pr) => pr.timelineItems.nodes.flatMap((item) => {
   return []
 })
 
+// Renders one PR-list section (title, PR links, assignee/age/status context lines).
+// `statusSuffix(pr)` returns the bucket-specific trailing note for the context line, or ''.
+const renderPrSection = (blocks, textLines, { emoji, title, prList, emptyText, mttmHours, statusSuffix }) => {
+  if (prList.length === 0) {
+    blocks.push(sectionBlock(`*${emoji} ${title}*\n${emptyText}`))
+    textLines.push(`${title}: none`)
+    blocks.push(dividerBlock())
+    return
+  }
+
+  blocks.push(sectionBlock(`*${emoji} ${title}*\n${prList.length} PR${prList.length === 1 ? '' : 's'}, oldest first:`))
+  textLines.push(`${title}: ${prList.length}`)
+
+  // Capped low, and one Slack block per PR (not two), to keep the overall payload
+  // comfortably under Slack's 50-block-per-message limit now that PRs are split
+  // across four sections instead of one.
+  const maxDetailed = 6
+  for (const pr of prList.slice(0, maxDetailed)) {
+    const assignees = pr.assignees.nodes.map((assignee) => assignee.login)
+    const authorLogin = pr.author?.login
+
+    const prLink = `<${pr.url}|#${pr.number} ${escapeSlack(pr.title)}>`
+    const assigneeMentions = assignees.filter((login) => login !== authorLogin).map(mentionFor)
+
+    let prText
+    if (assignees.length > 0) {
+      prText = assigneeMentions.length > 0
+        ? `${prLink}\n*Assigned to:* ${assigneeMentions.join(' ')}`
+        : prLink
+    } else {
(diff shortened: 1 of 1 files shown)


ellisong (approved):
LGTM

PR #1854, opened 2026-09-10
chore: make jest more resilient to flakes

metal-messiah:
---

### DIFF
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -46,16 +46,21 @@ jobs:
       - name: Install project dependencies
         run: npm ci
       - name: Run unit tests
-        run: npm run test -- --coverage
+        uses: ./.github/actions/retry-command
+        with:
+          command: npm run test -- --coverage
+          max_attempts: 2
       - name: Find pull request
         id: pull-request-target
         if: ${{ inputs.coverage }}
         uses: ./.github/actions/find-pull-request
+        continue-on-error: true
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
       - name: Upload pr code coverage
         if: ${{ inputs.coverage && steps.pull-request-target.outputs.results }}
         uses: codecov/codecov-action@v4
+        continue-on-error: true
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           override_commit: ${{ fromJSON(steps.pull-request-target.outputs.results).head_sha }}
@@ -66,6 +71,7 @@ jobs:
       - name: Upload branch code coverage
         if: ${{ inputs.coverage && !steps.pull-request-target.outputs.results }}
         uses: codecov/codecov-action@v4
+        continue-on-error: true
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: unit-tests
@@ -73,6 +79,7 @@ jobs:
       - name: Archive code coverage results
         if: ${{ inputs.coverage }}
         uses: actions/upload-artifact@v4
+        continue-on-error: true
         with:
           name: jest-code-coverage-report
           path: coverage/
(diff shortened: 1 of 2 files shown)


cwli24 (approved):
I don't think I've ever had jest tests fail wherein there wasn't something wrong or it's a race/timing issue that isn't fixable. Just saying. Wouldn't want to hide jest tests that should be fixed in actuality.

.github/workflows/jest.yml
```diff
@@ -46,16 +46,21 @@ jobs:
       - name: Install project dependencies
         run: npm ci
       - name: Run unit tests
-        run: npm run test -- --coverage
+        uses: ./.github/actions/retry-command
+        with:
+          command: npm run test -- --coverage
+          max_attempts: 2
```
ptang-nr, line 52: Are we seeing unexpected flakiness from a unit test? I think taking a look at the test itself to see if they can be adjusted first would be preferable.

ptang-nr: Welp, ok, I can see marching on for network hiccups 👍 
https://github.com/newrelic/newrelic-browser-agent/actions/runs/34497346330/job/102939068621
```
Summary of all failing tests
FAIL tests/unit/features/session_replay/shared/recorder.test.js
  ● Recorder › audit() › when fix_stylesheets is disabled › does not take new snapshot

    write EPIPE
```
But probably would prefer to revisit the test for things like this 🤷‍♀️ 
https://github.com/newrelic/newrelic-browser-agent/actions/runs/34254112083/job/102155457915

```
FAIL tests/components/api.test.js (8.512 s)
  ● API tests › APIs › finished › should execute as expected

    expect(received).toBeLessThanOrEqual(expected)

    Expected: <= 1
    Received:    7

      148 |         expectHandled('storeEventMetrics', ['finished', { time: expect.any(Number) }])
      149 |         const storeEventMetricsCall = handleModule.handle.mock.calls.find(callArr => callArr[0] === 'storeEventMetrics')
    > 150 |         expect(Math.abs(storeEventMetricsCall[1][1].time - n)).toBeLessThanOrEqual(1) // should be unix timestamp
          |                                                                ^
      151 |
      152 |         expectHandled('storeSupportabilityMetrics', ['API/addToTrace/called'])
      153 |

      at Object.toBeLessThanOrEqual (tests/components/api.test.js:150:64)
```

ptang-nr: One more (for future reminder)
https://github.com/newrelic/newrelic-browser-agent/actions/runs/28981593141/job/86001391117

```
FAIL tests/components/api.test.js (9.73 s)
  ● API tests › APIs › finished › should allow argument as expected

    expect(received).toBeLessThan(expected)

    Expected: < 10000
    Received:   10520

      182 |         expectHandled('api-addPageAction', [expect.any(Number), 'finished']) // unix timestamp
      183 |         const addPageActionCall = handleModule.handle.mock.calls.find(callArr => callArr[0] === 'api-addPageAction')
    > 184 |         expect(addPageActionCall[1][0]).toBeLessThan(10000) // should be relative timestamp
          |                                         ^
      185 |         expect(Math.abs(addPageActionCall[1][0] - (n + 1000))).toBeLessThanOrEqual(1) // should be relative timestamp, account for rounding errors
      186 |       })
      187 |

      at Object.toBeLessThan (tests/components/api.test.js:184:41)
```

metal-messiah:
> I don't think I've ever had jest tests fail wherein there wasn't something wrong or it's a race/timing issue that isn't fixable. Just saying. Wouldn't want to hide jest tests that should be fixed in actuality.

i have to re-run the jest suite semi-regularly.  if a test fails twice in a row its a real issue that needs to be fixed, but if a test fails by 1ms or something, thats a flake we shouldnt be concerned with

PR #1864, opened 2026-09-19
feat: Harvest SR on switch from ERROR to FULL mode

ptang-nr:
Update agent to handle session replays the same way as session traces, and trigger a replay harvest when SR is switching from `error` to `full` mode.
---

### Overview

Today when running in SR `error` mode, the agent keeps a window of 15-30s worth of session replay prior to a JavaScript error.  Then when a JS error is encountered, it will trigger SR to switch to `full` mode for the rest of that session.  The captured SR data would fallback to the ongoing harvest cycle, which may be up to 30s later.  

Following previous changes to improve reliability of agent harvests, we trigger a harvest when SR switches from `error` to `full` mode.

### Related Issue(s)

https://new-relic.atlassian.net/browse/NR-620516

### Testing

Extended SR aggregate/error mode integration testing.


metal-messiah (commented):
Is there a stable way to test this change in a wdio spec?  something like `tests/specs/session-replay/harvest.e2e.js`? 

Maybe mirroring the `should harvest early if exceeds preferred size` test where the interval is purposefully set to `60` seconds, but we expect a harvest to occur before that (due to errors triggering an immediate harvest). What do you think there?

metal-messiah (approved):
LGTM with passing test suite

PR #1865, opened 2026-09-21
chore: improve jest consistency

metal-messiah:
Fixes the CI retry wrapper so a flaky command actually gets retried instead of failing the job on the first attempt, and reduces jest worker contention in CI to cut down on the underlying flake.

---
### Overview
The `retry-command` composite action's bash step runs with `set -e`. Inside the retry `until` loop, `bash -c "$RETRY_COMMAND"` (or the `timeout`-wrapped variant) sat as a bare command, not guarded by an `if`/`&&`/`||`. Under `set -e`, a non-zero exit from that line aborted the whole script immediately, so the code that checks the exit status, prints the retry warning, sleeps, and increments `attempt` never ran — every wrapped step failed outright on its first flake, regardless of `max_attempts`. This affected the jest workflow's CI run (`.github/workflows/jest.yml`), where an intermittent `write EPIPE` crash in the coverage-instrumented jest run was hitting exactly this dead retry path.

Fixed by capturing the exit code via `&& exit_code=0 || exit_code=$?` for both branches, so the retry loop's logic actually executes on failure. Also capped the CI jest run to `--maxWorkers=2` to reduce worker/stdout contention, which is the likely trigger for the `write EPIPE` crash itself.

### Related Issue(s)
https://github.com/newrelic/newrelic-browser-agent/actions/runs/35624090043/job/106417960735

### Testing
Reproduced the `set -e` bug locally with a minimal bash repro (a failing command inside an `until` loop body aborts the script under `-e` before the retry logic runs), and verified the fix restores the intended behavior: attempt 1 fails, prints the retry warning, attempt 2 succeeds, script exits 0. No jest/wdio coverage applicable — this is a CI workflow/composite-action change, not `src` behavior.


ellisong (approved):
LGTM

PR #1866, opened 2026-09-21
chore: skip experiment publish for chore PRs

metal-messiah:
Skips the automatic nr1-experiments publish workflow for pull requests titled as chores, avoiding unnecessary experiment builds/deploys for non-behavioral changes.

---
### Overview
Adds an `if` condition to the `publish-experiment-to-s3` job in `publish-experiment.yml` so PR-triggered runs (`opened`/`synchronize`) are skipped when the PR title starts with `chore`. Manual `workflow_dispatch` runs are unaffected, since the check only applies to `pull_request` events.

### Related Issue(s)
N/A

### Testing
Workflow-only change (`.github/workflows/publish-experiment.yml`); no jest/wdio coverage applies. Verified the `if` expression logic manually against both trigger types (`pull_request` with a `chore:`-titled PR, and `workflow_dispatch`).


ellisong (approved):
LGTM

